PR #6, opened 2022-05-23
feat: expose more envs in udf and udsink

### LINKED ISSUE #5: expose Namespace, Pipeline and Vertex name as environment variables
As a UDF or UDSink developer, I would like to emit custom metrics. WIth the custom metrics, I would like to attach namespace, vertex, and pipeline names as tags so I can filter those metrics easily. Today there is no way to get these names.

whynowy:

Closes #5.

### DIFF
--- a/pkg/isb/jetstream/writer_test.go
+++ b/pkg/isb/jetstream/writer_test.go
@@ -164,9 +164,9 @@ func TestJetStreamBufferWriterBufferFull(t *testing.T) {
 	_, errs := jw.Write(ctx, messages)
 	assert.Equal(t, len(errs), 2)
 	for _, errMsg := range errs {
-		assert.Nil(t, errMsg) // Buffer not full, expected no error
+		assert.Nil(t, errMsg) // Buffer not full, expect no error
 	}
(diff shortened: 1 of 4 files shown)


pkg/isb/jetstream/writer_test.go
```diff
@@ -165,6 +165,9 @@ func TestJetStreamBufferWriterBufferFull(t *testing.T) {
 	_, errs := jw.Write(ctx, messages)
 	assert.Equal(t, len(errs), 10)
 
+	// wait 2 seconds so that the buffer full check is done.
+	time.Sleep(2 * time.Second)
```
vigith, line 169: adding this wait won't have any benefit. the assertion has already happened.

whynowy: Adding this is to make sure the isFull check has been run.

whynowy: Which is for line 172.

whynowy: Updated test case.

vigith (approved):

PR #15, opened 2022-05-26
feat: expose min/max replicas

whynowy:
Only honors the min replicas, auto-scaling is not there yet.


controllers/pipeline/controller.go
```diff
@@ -419,13 +415,25 @@ func buildVertices(pl *dfv1.Pipeline) map[string]dfv1.Vertex {
...
 		copyLimits(pl, vCopy)
+		replicas := int32(1)
+		if pl.Status.Phase == dfv1.PipelinePhasePaused {
+			replicas = int32(0)
+		} else {
+			x := vCopy.Scale
+			if x.Min != nil && *x.Min > 1 && replicas < *x.Min {
+				replicas = *x.Min
+			}
+			if x.Max != nil && *x.Max > 1 && replicas > *x.Max {
```
vigith, line 426: so only `Min` is honored as of today since `Max` does not make sense without auto-scaling. right?

what will happen is someone runs `k scale vtx ....` ?

whynowy: Today, `scale vtx` will still be honored, but with auto-scaling, it will not - unless we allow some kind of rules, e.g. scale to a number which is out of min <-> max.

vigith (approved):

Issue #2, opened 2022-05-21
Kafka e2e test broken


whynowy:
Kafka e2e test is broken with following error.

2022/05/21 00:54:13 create Kafka topic "e2e-topic-jtl8v"
2022/05/21 00:54:13 GET http://127.0.0.1:8378/kafka/create-topic?topic=e2e-topic-jtl8v
2022/05/21 00:54:13 > 500 Internal Server Error
2022/05/21 00:54:13 > kafka: controller is not available
    suite.go:63: test panicked: 500 Internal Server Error
        goroutine 102 [running]:
        runtime/debug.Stack()
        	/usr/local/Cellar/go/1.18.1/libexec/src/runtime/debug/stack.go:24 +0x65
        github.com/stretchr/testify/suite.failOnPanic(0xc000682340)
        	/Users/jwang21/go/pkg/mod/github.com/stretchr/testify@v1.7.0/suite/suite.go:63 +0x3e
        panic({0x2189a20, 0xc000614a80})
        	/usr/local/Cellar/go/1.18.1/libexec/src/runtime/panic.go:838 +0x207
        github.com/numaproj/numaflow/test/fixtures.InvokeE2EAPI({0x23acbc9?, 0xf?}, {0xc0006ad508?, 0x1?, 0x1?})
        	/Users/jwang21/workspace/numaproj/numaflow/test/fixtures/e2eapi.go:31 +0x37e
        github.com/numaproj/numaflow/test/fixtures.CreateKafkaTopic()
        	/Users/jwang21/workspace/numaproj/numaflow/test/fixtures/kafka.go:15 +0xe9
        github.com/numaproj/numaflow/test/kafka-e2e.(*KafkaSuite).TestKafkaSourceSink(0xc0002569a0)
        	/Users/jwang21/workspace/numaproj/numaflow/test/kafka-e2e/kafka_test.go:77 +0x4a
        reflect.Value.call({0xc0003fb800?, 0xc000114050?, 0x13?}, {0x23953b6, 0x4}, {0xc0000f9e70, 0x1, 0x1?})
        	/usr/local/Cellar/go/1.18.1/libexec/src/reflect/value.go:556 +0x845
        reflect.Value.Call({0xc0003fb800?, 0xc000114050?, 0xc0002569a0?}, {0xc000082e70, 0x1, 0x1})
        	/usr/local/Cellar/go/1.18.1/libexec/src/reflect/value.go:339 +0xbf
[...]

PR #16, opened 2022-05-27
ci: fix kafka e2e test

whynowy:

Fixes: https://github.com/numaproj/numaflow/issues/2


test/kafka-e2e/kafka_test.go
```diff
@@ -13,7 +13,8 @@ import (
 )
 
 //go:generate kubectl -n numaflow-system delete statefulset zookeeper kafka-broker --ignore-not-found=true
-//go:generate kubectl apply  -k ../../config/apps/kafka -n numaflow-system
+//go:generate kubectl apply -k ../../config/apps/kafka -n numaflow-system
+// Wait for zookeeper to come up
```
vigith, line 17: is there way we can verify that zookeeper is up?

vigith (approved):

test/kafka-e2e/kafka_test.go
whynowy: It's difficult. Technically, the order should be zookeeper comes up, and then broker comes up. Needs more effort to do that, simply wait for a while to make it work.

PR #21, opened 2022-05-27
docs: Update README and DEVELOPMENT

edlee2121:
Signed-off-by: Edward Lee <edward_lee@intuit.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,19 +6,24 @@
 
 ## Summary
 
-NumaFlow is a Kubernetes-native platform to run massive parallel data processing or streaming jobs.
+NumaFlow is a Kubernetes-native platform for running massive parallel data processing and streaming jobs.
 
-Each pipeline is specified as a Kubernetes custom resource which consists of one or more source vertices, data processing vertices and sink vertices. Each vertex runs zero or more pods with auto scaling.
+A NumaFlow Pipeline is implemented as a Kubernetes custom resource, and consists of one or more source, data processing, and sink vertices.
 
-NumaFlow targets to achieve `Exactly-Once` semantics, which means from the data from the source vertex to the sink vertex will be processed `Exactly-Once`.
+NumaFlow installs in less than a minute and is easier and cheaper for simple data processing applications than full-featured stream processing platforms.
 
-## Core Principles
+## Key Features
 
-- Easy to use for an engineer in any language
-- Install and up and running in < 1 min (minimal setup)
-- Cheaper than Flink, Samza, etc. when TPS is < 10K TPS
+* Kubernetes-native: If know Kubermetes, you already know 90% of what you need to use NumaFlow.
+* Language agnostic: Use your favority programming language.
+* Exactly-Once semantics: No input element is duplicated or lost even as pods are rescheduled or restarted.
+
+## Roadmap
+
+* Auto-scaling with back-pressure: Each vertex automatically scales from zero to whatever is needed.
+* Data aggregation (e.g. group-by)
 
 ## Resources
 - Check out [QUICK START](docs/QUICK_START.md) to try it out.
 - Take a look at [DEVELOPMENT](docs/DEVELOPMENT.md) to set up development environment.
-- Refer to [CONTRIBUTING](CONTRIBUTING.md) to contribute to the project.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -6,19 +6,24 @@
...
 ## Summary
 
-NumaFlow is a Kubernetes-native platform to run massive parallel data processing or streaming jobs.
+NumaFlow is a Kubernetes-native platform for running massive parallel data processing and streaming jobs.
 
-Each pipeline is specified as a Kubernetes custom resource which consists of one or more source vertices, data processing vertices and sink vertices. Each vertex runs zero or more pods with auto scaling.
+A NumaFlow Pipeline is implemented as a Kubernetes custom resource, and consists of one or more source, data processing, and sink vertices.
 
-NumaFlow targets to achieve `Exactly-Once` semantics, which means from the data from the source vertex to the sink vertex will be processed `Exactly-Once`.
+NumaFlow installs in less than a minute and is eaasier and chaper to run than full-featured stream processing platforms like Flink and Samza.
```
vigith, line 13: we do not support features like reduce which is in Flink or Samza yet. To be politically correct, should we remove `like Flink and Samza`?

README.md
```diff
@@ -6,19 +6,24 @@
...
-## Core Principles
+## Key Features
 
-- Easy to use for an engineer in any language
-- Install and up and running in < 1 min (minimal setup)
-- Cheaper than Flink, Samza, etc. when TPS is < 10K TPS
+* Kubernetes-native: If know Kubermetes, you already know 90% of what you need to use NumaFlow.
+* Language agnostic: Use your favority programming language.
+* Exactly-Once semantics: No input element is duplicated or lost even as pods are rescheduled or restarted.
+* Auto-scaling with back-pressure: Each vertex automatically scales from zero to whatever is needed.
```
vigith, line 20: this is work in progress

README.md
whynowy: type `eaasier`

docs/DEVELOPMENT.md
```diff
@@ -1,46 +1,38 @@
...
-## Useful Commands
+### Useful Commands
 
 - `make build`
-  Build binaries into `./dist` directory.
+  Binaries are placed in `./dist`.
 
 - `make codegen`
-  If there's any model changes (`./pkg/api/`), this command need to be run to regenerate code.
+  Run after making changes to `./pkg/api/` Can we add a Makefile rule to do this automatically?
```
whynowy, line 29: There's a CI job to check if this is done or not.

README.md
edlee2121: Good idea. FYI, that was already there later in the doc.

README.md
edlee2121: Will move it to roadmap section

docs/DEVELOPMENT.md
edlee2121: What about for users running it manually on their laptop though?

edlee2121:
Pushed new changes

docs/DEVELOPMENT.md
whynowy: `codegen` is already a rule, developers need to run it manually, do you mean adding a rule to check if `make codegen` is run?

edlee2121: The comment seems to say that if anything in pkg/api changes, users need to make codegen before they do make build? Can't make build automatically run codegen if anything changes?

whynowy: Do not want to make `make build` run codegen automatically because:

1. `codegen` does not create any new files, but regenerate the content of several files, so we can not use makefile rule prerequisites to check file existence, but have to run it every time if we do that.
2. It takes too long to run codegen (around 1 min), and there's no need to run it every time - only needed when changes in pkg/api happen.

vigith (approved):

docs/DEVELOPMENT.md
edlee2121: If it always regenerates a particular file, we can use a makefile rule for that file.
Otherwise, we could touch a dummy file each time codegen runs and create a rule for the dummy file.
Definitely don't want to run codegen every time.

PR #26, opened 2022-05-28
docs: Added examples/README

edlee2121:
Please explain what this PR does


examples/README.md
```diff
@@ -0,0 +1,104 @@
...
+      source:
+        generator:
+          rpu: 5
+          duration: 1s
+    - name: filter
+      udf:
+        builtin:
+          name: filter
+          kwargs:
+            expression: int(json(payload).id) < 100
```
vigith, line 94: we do not have `.id` field in the `generator` so this will not have any effect.

vigith (approved):

examples/README.md
edlee2121: Yes, planning to replace this with a better example once we have http pull source.

PR #50, opened 2022-06-07
Changed all errors logs in redis isb to debug logs

Krithika3:


vigith:
there are still many log entries in the log, we might have to tackle this differently

vigith (commented):
i think we shouldn't mask `error` as `debug`.. we might have to change the test log level perhaps?

Krithika3:
> i think we shouldn't mask `error` as `debug`.. we might have to change the test log level perhaps?

I think if we have them as error logs, even the lowest log level will display them if I am not mistaken?

Krithika3:
I think another reason for so many logs in the test is also the refreshInterval that we set. Let me take a look at that.

vigith:
> > i think we shouldn't mask `error` as `debug`.. we might have to change the test log level perhaps?
> 
> I think if we have them as error logs, even the lowest log level will display them if I am not mistaken?

Setting log level to fatal is one option, so only fatal and above will be printed.

whynowy:
> > > i think we shouldn't mask `error` as `debug`.. we might have to change the test log level perhaps?
> > 
> > 
> > I think if we have them as error logs, even the lowest log level will display them if I am not mistaken?
> 
> Setting log level to fatal is one option, so only fatal and above will be printed.

No.. no, `Fatal` exits the application.

vigith:
> > > > i think we shouldn't mask `error` as `debug`.. we might have to change the test log level perhaps?
> > > 
> > > 
> > > I think if we have them as error logs, even the lowest log level will display them if I am not mistaken?
> > 
> > 
> > Setting log level to fatal is one option, so only fatal and above will be printed.
> 
> No.. no, `Fatal` exits the application.

this is just for test log.level

whynowy:
Why logrus is needed?

Krithika3:
> Why logrus is needed?

I think we can use zap

Krithika3:
the watermark test failed again. What is the way around this?

vigith:
> the watermark test failed again. What is the way around this or how do we fix it?

I don't think it is the watermark test that failed, but rather

```
# github.com/numaproj/numaflow/pkg/isb/redis [github.com/numaproj/numaflow/pkg/isb/redis.test]
Error: pkg/isb/redis/read_test.go:36:2: undefined: log
Error: pkg/isb/redis/read_test.go:68:2: undefined: log
Error: pkg/isb/redis/read_test.go:166:2: undefined: log
Error: pkg/isb/redis/read_test.go:289:2: undefined: log
```

Krithika3:
> > the watermark test failed again. What is the way around this or how do we fix it?
> 
> I don't think it is the watermark test that failed, but rather
> 
> ```
> # github.com/numaproj/numaflow/pkg/isb/redis [github.com/numaproj/numaflow/pkg/isb/redis.test]
> Error: pkg/isb/redis/read_test.go:36:2: undefined: log
> Error: pkg/isb/redis/read_test.go:68:2: undefined: log
> Error: pkg/isb/redis/read_test.go:166:2: undefined: log
> Error: pkg/isb/redis/read_test.go:289:2: undefined: log
> ```

Ah ok,thanks

Krithika3:
I closed this as this is not the right way to do this. I will open a new PR with the changes

PR #51, opened 2022-06-08
feat: async publishing for kafka sink

whynowy:


pkg/sinks/kafka/kafka.go
```diff
@@ -102,37 +96,50 @@ func (tk *ToKafka) GetName() string {
...
+			select {
+			case err := <-tk.producer.Errors():
+				idx := err.Msg.Metadata.(int)
+				errs[idx] = err.Err
+				sent++
+			case m := <-tk.producer.Successes():
+				idx := m.Metadata.(int)
+				errs[idx] = nil
+				sent++
+			case <-timeout:
```
whynowy, line 120: There will be problems here, need more stuff...

vigith: can you send a counter metric and also a log line so we can track these timeouts?

whynowy: done.

vigith (approved):

PR #53, opened 2022-06-10
feat: introduce source buffer and sink buffer

whynowy:

Introduced source buffer and sink buffer concepts, which will create buckets but no streams.
Original buffer is internally edge buffer.


whynowy:
Here you go! @vigith

vigith (approved):

PR #54, opened 2022-06-10
Integration of generic fetch interface 

kohlisid:
Integration of the fetch/publisher KV interface for watermarks

### DIFF
--- a/pkg/watermark/publish/publisher.go
+++ b/pkg/watermark/publish/publisher.go
@@ -28,16 +28,13 @@ type Publisher interface {
 
 // Publish publishes the watermark for a processor entity.
 type Publish struct {
-	ctx             context.Context
-	entity          processor.ProcessorEntitier
-	keyspace        string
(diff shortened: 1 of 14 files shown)


pkg/watermark/publish/publisher.go
```diff
@@ -28,16 +28,13 @@ type Publisher interface {
...
-	js              nats.JetStreamContext
-	log             *zap.SugaredLogger
-	otBucket        nats.KeyValue
-	headWatermark   processor.Watermark
+	ctx            context.Context
+	entity         processor.ProcessorEntitier
+	heartbeatStore store.WatermarkKVStorer
+	otStore        store.WatermarkKVStorer
+	log            *zap.SugaredLogger
+	otBucket       nats.KeyValue
```
whynowy, line 36: Still has `nats`?

kohlisid: Will check and remove!

PR #56, opened 2022-06-13
chore: switch JetStream sync writing to async

whynowy:

### DIFF
--- a/pkg/isb/jetstream/writer.go
+++ b/pkg/isb/jetstream/writer.go
@@ -164,6 +164,7 @@ func (jw *jetStreamWriter) Write(ctx context.Context, messages []isb.Message) ([
 			futures[index] = future
 		}
 	}
+	futureCheckDone := make(chan struct{})
 	cctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 	wg := new(sync.WaitGroup)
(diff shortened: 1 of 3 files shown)


pkg/isb/jetstream/writer.go
```diff
@@ -137,37 +137,70 @@ func (jw *jetStreamWriter) Close() error {
...
-			} else {
+			select {
+			case pubAck := <-fu.Ok():
 				writeOffsets[idx] = &writeOffset{seq: pubAck.Sequence}
+				errs[idx] = nil
 				jw.log.Debugw("Succeeded to publish a message", zap.String("stream", pubAck.Stream), zap.Any("seq", pubAck.Sequence), zap.Bool("duplicate", pubAck.Duplicate), zap.String("domain", pubAck.Domain))
+			case err := <-fu.Err():
+				errs[idx] = err
+				isbWriteErrors.With(labels).Inc()
+			case <-ctx.Done():
```
vigith, line 186: i think we might need to add a timeout here too, what if we lost `fu.Ok()` and/or `fu.Err()`? we will end up waiting indefinitely.. perhaps you can use context with timeout?

whynowy: That is not needed. 

There's a timeout at line 198, which will issue a `cancel()` to end this.

vigith (approved):

pkg/isb/jetstream/writer.go
vigith: you are right.

PR #58, opened 2022-06-15
Added name to service spec

Krithika3:
I was working on a document for metrics and there was an ask to figure out how to scrape the metrics using the prometheus operator. As part of the exercise I had created some k8s serviceMonitors to monitor services(based on labels) and scrape the corresponding metric. While doing so I figured that the serviceMonitor expected a `name` field on the service spec Port object. Added the same.

This has been tested by running a sample pipeline in my local k3d cluster

### DIFF
--- a/pkg/apis/numaflow/v1alpha1/vertex_types.go
+++ b/pkg/apis/numaflow/v1alpha1/vertex_types.go
@@ -73,22 +73,14 @@ func (v Vertex) GetHeadlessServiceName() string {
 }
 
 func (v Vertex) GetServiceObjs() []*corev1.Service {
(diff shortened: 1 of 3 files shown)


pkg/apis/numaflow/v1alpha1/vertex_types.go
```diff
@@ -81,6 +81,14 @@ func (v Vertex) GetServiceObjs() []*corev1.Service {
 }
 
 func (v Vertex) getServiceObj(name string, headless bool, port int) *corev1.Service {
+	var servicePortName string
```
whynowy, line 84: Pass the port name as a parameter.

Krithika3: Done.

PR #68, opened 2022-06-22
test: e2e test for filtering builtin function

whynowy:


test/fixtures/util.go
```diff
@@ -285,3 +355,83 @@ func podLogContains(ctx context.Context, client kubernetes.Interface, namespace,
...
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				if !s.Scan() {
+					return
+				}
+				data := s.Bytes()
+				fmt.Println(string(data))
```
vigith, line 385: do we need this?

vigith (approved):

test/fixtures/util.go
whynowy: This is to print out the watched log, useful for debugging.

PR #70, opened 2022-06-23
feat: define buffer limits on edges

whynowy:

`bufferMaxLength` and `bufferUsageLimit` are the properties for a buffer instead of a vertex, which is currently incorrectly defined in Vertex level.

This PR removes these 2 properties from Vertex Limits and add them into Edge Limits. In Pipeline level Limits definition, they remain in the same place.


pkg/daemon/server/service/isbs_query_service.go
```diff
@@ -29,28 +29,22 @@ func (is *isbSvcQueryService) ListBuffers(ctx context.Context, req *daemon.ListB
...
+		bufferLength, bufferUsageLimit := getBufferLimits(is.pipeline, edge)
 		usage := float64(bufferInfo.TotalMessages) / float64(bufferLength)
 		if x := (float64(bufferInfo.PendingCount) + float64(bufferInfo.AckPendingCount)) / float64(bufferLength); x < usage {
 			usage = x
 		}
 		b := &daemon.BufferInfo{
 			Pipeline:         &is.pipeline.Name,
-			FromVertex:       &vFrom.Name,
-			ToVertex:         &vTo.Name,
+			FromVertex:       &edge.From,
```
vigith, line 46: `FromVertex` pointing to `edge.From` data could be confusing (same for `To`)? Rename `FromVertex` to `FromEdge` ?

vigith (approved):

pkg/daemon/server/service/isbs_query_service.go
whynowy: Not edgeFrom and edgeTo?

If we want to rename, do it later together with UI change - it's only used in the UI.

vigith: @Krithika3 fyi

PR #73, opened 2022-06-24
chore: open source UI

Krithika3:
Signed-off-by: Krithika Vijayakumar <krithika_vijayakumar@intuit.com>


Krithika3:
This PR just moves the UI directory over

ui/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Prerequisites
+
+Follow the instructions below to set up your development environment.
+
+### Install requirements
+
+These tools are required for development.
+
+1. [`NumaFlow`](https://github.com/numaproj/numaflow/blob/master/docs/DEVELOPMENT.md): Follow the document to prepare `Dataflow` development environment.
```
whynowy, line 25: `Dataflow`?

ui/README.md
```diff
@@ -0,0 +1,90 @@
...
+1. [`node`](https://nodejs.org/en/download/): NodeJS - `brew install node`.
+1. [`yarn`](https://yarnpkg.com/): Package manager - `brew install yarn`
+
+### Create a dev cluster
```
whynowy, line 29: Maybe we don't need all the stuffs below, they should be covered in the root development doc.

Krithika3: Removed

ui/README.md
```diff
@@ -1,6 +1,6 @@
 # NumaFlow UI
 
-![Numaflow Image](../docs/images/Numa.svg)
+![Numaflow Image](../docs/architecture/assets/numa.svg)
```
whynowy, line 3: Not under `architecture` but directly under `docs`. The ones under `architecture` should also move over later.

Krithika3: Done. should we move the others too?

vigith: let's not move the architecture yet. We can keep the "Improvement Proposals" and the related architecture separate.

vigith (approved):

vigith:
I have a question about the UX code structure. Why not keep wrappers along with the main code and avoid utils dir altogether?

Krithika3:
> I have a question about the UX code structure. Why not keep wrappers along with the main code and avoid utils dir altogether?

I would like to discuss this with our UI team before we do it. Merging this for now

PR #79, opened 2022-06-28
feat: expose pending messages and processing rate

whynowy:

Processing rate and pending message are exposed in each vertex's prometheus, which will be used later for auto scaling calculation.


pkg/metrics/metrics.go
```diff
@@ -7,17 +7,144 @@ import (
...
+						for i := len(items) - 2; i >= 0; i-- {
+							if now-items[i].timestamp < ms.pendingLookbackSeconds {
+								total += items[i].pending
+								num++
+							} else {
+								break
+							}
+						}
+						pendingMessages.With(labels).Set(float64(total / num))
+					}
```
vigith, line 136: should we emit these metrics to prometheus too?

vigith (approved):

pkg/metrics/metrics.go
whynowy: I assume you meant shall we display these metrics (they are already exposed/emmited) - This is the tricky part.

The `pending_messages` and `processing_rate` represent the vertex level metrics, not a pod, but we are exposing them from every individual pod, this will be a problem when displaying in the dashboard, that we should only display the metrics from one pod (or avg).

whynowy: This the reason of the discussion about should we only expose these metrics from 1 pod.

vigith: we could expose from all pods, the numbers should be more or less the same.. we can let the user configure how they want to display avg/max/min etc?

Krithika3: I agree with Vigith that we should expose from all the pods and let the user decide when they are displaying these metrics on the type of stat to be used. We just need to call this out specifically in our doc that these numbers are vertex level and should roughly be the same and the aggregation we need to use is just `avg/max/min`?

PR #80, opened 2022-06-29
feat: kafka source pending messages

whynowy:

Expose Kafka source pending messages.


pkg/sources/kafka/reader.go
```diff
@@ -136,6 +140,17 @@ func (r *KafkaSource) Ack(_ context.Context, offsets []isb.Offset) []error {
 }
 
 func (r *KafkaSource) Start() <-chan struct{} {
+	adminClient, err := sarama.NewClusterAdmin(r.brokers, r.config)
```
vigith, line 143: should we give an option not to create a cluster-admin client? I don't think it will always be available for every deployment and we won't be able to even start the kafka source.

whynowy: Don't be scaled by the word `admin`, it's just a particular word used in sarama lib.

It's true there's a requirement to support these `admin` APIs, which is the minimal broker version has to be `0.10.0.0`, I think this shouldn't be an issue for most of the kafka installations. Other than that, I didn't find there's anything special to enable it from Kafka server side.

whynowy: Changed to best effort.

vigith (approved):

PR #81, opened 2022-06-29
chore: added yarn build to docker file

Krithika3:
In the below PR we have added the following:
1. Dockerfile: Added yarn build to the docker image and copied the UI in the main numaflow binary. (I did not research if there was a way to compress this yet)
2. The image size prior to adding UI director `ui/build` is:
Screen Shot 2022-06-28 at 8 48 37 PM
3. The image size after adding UI director `ui/build` is:
Screen Shot 2022-06-28 at 9 14 51 PM
4. The make start (which runs make image in turn runs docker-build) command with cached package.json changes and yarn build takes the following time:
Screen Shot 2022-06-28 at 10 41 58 PM
5.  I think this is reasonable and we could take a run at the release github action to see how it behaves.
6. running `kubectl apply -f ./config/install.yaml` now installs the numaflow-server too. (I have not updated the README.md yet, we could take a stab at it if this approach is ok)


config/base/numaflow-server-sa.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
Krithika3, line 1: Do we need a separate service account for numaflow-server? Currently I have re-used the existing one

Krithika3: What is the best practice?

config/cluster-install/rbac/numaflow-server-binding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
```
Krithika3, line 2: I have added the clusterRoleBinding and the clusterRole, but re-used the existing ones. Let me know if it is good practice to add new cluster roles for server

Dockerfile
```diff
@@ -1,4 +1,16 @@
 ARG ARCH=$TARGETARCH
+####################################################################################################
+FROM docker.io/node:18 as numaflow-ui
```
Krithika3, line 3: We can do the `yarn build` in the `make image` which would expect the user to have yarn installed, we could circumvent it by adding a flag requesting ui vs not requesting it.

Krithika3: I think it is not warranted to do this if the docker build is not taking very long

config/cluster-install/rbac/numaflow-cluster-role.yaml
```diff
@@ -42,6 +43,14 @@ rules:
       - update
       - patch
       - delete
+  - apiGroups:
+    - metrics.k8s.io
```
Krithika3, line 47: Is it ok to re-use the existing cluster role?

Dockerfile
whynowy: 👍

config/base/numaflow-server/numaflow-server-deployment.yaml
```diff
@@ -0,0 +1,44 @@
...
+        app.kubernetes.io/component: numaflow-ux
+    spec:
+      serviceAccountName: numaflow-sa
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 9737
+      containers:
+        - name: numaflow-server
+          image: quay.io/numaproj/numaflow:latest
+          args: [ "server" ]
```
whynowy, line 24: Use yaml approach.

config/base/numaflow-server/numaflow-server-deployment.yaml
```diff
@@ -0,0 +1,44 @@
...
+      labels:
+        app.kubernetes.io/part-of: numaflow
+        app.kubernetes.io/component: numaflow-ux
+    spec:
+      serviceAccountName: numaflow-sa
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 9737
+      containers:
+        - name: numaflow-server
```
whynowy, line 22: Use a name like `main` maybe?

config/cluster-install/rbac/numaflow-cluster-role.yaml
whynowy: No.

config/cluster-install/rbac/numaflow-server-binding.yaml
whynowy: Please use a separate set of sa, role and binding.

config/base/numaflow-server/numaflow-server-deployment.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: apps/v1
```
whynowy, line 1: Also add a ClusterIP Service.

config/cluster-install/rbac/numaflow-cluster-role.yaml
Krithika3: Done

whynowy:
I can see the time increase is about 2 mins for each image build.

config/base/numaflow-server/numaflow-server-service.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Service
+metadata:
+  name: numaflow-server-svc
+spec:
+  type: ClusterIP
+  ports:
+  - targetPort: 8443
+    port: 8443
+  selector:
+    app.kubernetes.io/component: numaflow-ux
```
whynowy, line 11: Use all the labels:

```
        app.kubernetes.io/part-of: numaflow
        app.kubernetes.io/component: numaflow-ux
```

config/cluster-install/rbac/numaflow-server-cluster-role.yaml
```diff
@@ -0,0 +1,57 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: numaflow-server-role
+rules:
+  - apiGroups:
+      - numaflow.numaproj.io
+    verbs:
+      - create
```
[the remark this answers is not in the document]
Krithika3: Done

Krithika3:
> I can see the time increase is about 2 mins for each image build.

whynowy (approved):
LGTM

PR #93, opened 2022-07-06
chore: refactor source watermark

vigith:
refactor source watermark

### DIFF
--- a/pkg/sources/source.go
+++ b/pkg/sources/source.go
@@ -37,10 +37,9 @@ func (sp *SourceProcessor) Start(ctx context.Context) error {
 	// watermark variables no-op initialization
 	var fetchWatermark fetch.Fetcher = generic.NewNoOpWMProgressor()
 	// publishWatermark is a map representing a progressor per edge, we are initializing them to a no-op progressor
-	var publishWatermark map[string]publish.Publisher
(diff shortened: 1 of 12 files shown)


pkg/sources/source.go
```diff
@@ -24,29 +28,43 @@ type SourceProcessor struct {
...
-func (u *SourceProcessor) Start(ctx context.Context) error {
+func (sp *SourceProcessor) Start(ctx context.Context) error {
 	log := logging.FromContext(ctx)
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 	var writers []isb.BufferWriter
-	switch u.ISBSvcType {
+
+	// watermark variables no-op initialization
+	var fetchWatermark fetch.Fetcher = generic.NewNoOpWMProgressor()
```
whynowy, line 38: Sources need to do it twice (source bucket, and the another normal bucket) right? Where is that implemented?

whynowy (approved):

pkg/sources/source.go
vigith: https://github.com/numaproj/numaflow/blob/source-watermark-refactor/pkg/watermark/generic/generic.go#L154

PR #94, opened 2022-07-07
feat: Add icon and other minor changes

Krithika3:
Explain what this PR does.


ui/.eslintrc
```diff
@@ -10,6 +10,7 @@
     "plugin:@typescript-eslint/recommended"
   ],
   "rules": {
-    "no-debugger": 0
+    "no-debugger": 0,
+    "@typescript-eslint/no-explicit-any": ["off"]
```
Krithika3, line 14: I had added this to turn off eslint on any types, which are not avoidable when we are making fetch calls. I will consult with Shakira on this to see if there is a better way to make the types not be `any`

ui/src/components/pipeline/Pipeline.css
```diff
@@ -1,2 +1,7 @@
-.Pipeline {
+.react-flow__edge-textwrapper {
```
Krithika3, line 1: Thanks to Shakira for this fix :)

ui/src/components/pipeline/graph/Graph.css
```diff
@@ -2,6 +2,7 @@
     flex-grow: 1;
     position: relative;
     height: 100%;
+    cursor: move;
```
Krithika3, line 5: Thanks to Shakira here too :)

vigith (approved):

PR #105, opened 2022-07-08
chore: controller deployment rename and more docs.

whynowy:
1. Controller deployment renamed to `numaflow-controller`;
2. Doc names changed to lower cases - it's too hard to always use upper cases.
3. More docs.


docs/debugging.md
```diff
@@ -22,14 +22,14 @@ spec:
         env:
           - name: NUMAFLOW_DEBUG
             value: "true" # DO NOT forget the double quotes!!!
-    - name: output
+    - name: out
       sink:
         log: {}
   edges:
-    - from: input
+    - from: in
```
vigith, line 29: shouldn't this be `inp`?

docs/environment-variables.md
```diff
@@ -0,0 +1,44 @@
+# Environment Variables
+
+For the `main` container of vertex pods, environment variable `NUMAFLOW_DEBUG` can be set to `"true"` for [debugging](./debugging.md).
```
vigith, line 3: should we callout explicitly that the `true` should be in `"` ?

vigith (approved):

docs/debugging.md
whynowy: oops.. updated.

docs/environment-variables.md
whynowy: Removed the ", since it's been called out in the next linked page.

PR #106, opened 2022-07-08
chore: Remove publishKeyspace

kohlisid:
Remove publishKeyspace from processor entity and publisher signature.


pkg/watermark/processor/entity.go
```diff
@@ -65,7 +62,7 @@ const _defaultKeySeparator = "_"
 
 // NewProcessorEntity returns a new `ProcessorEntity`.
 // TODO: remove publishKeyspace
```
vigith, line 64: remove TODO

vigith (approved):

PR #119, opened 2022-07-21
fix: daemon service rest api not working

whynowy:

Daemon service should be accessible through either gRPC or REST.


pkg/apis/proto/daemon/daemon.proto
```diff
@@ -71,6 +71,6 @@ service DaemonService {
   };
 
   rpc GetVertexMetrics (GetVertexMetricsRequest) returns (GetVertexMetricsResponse) {
-    option (google.api.http).get = "/api/v1/namespace/{namespace}/pipelines/{pipeline}/vertices/{vertex}/metrics";
```
whynowy, line 74: We don't need namespace in the url, as the daemon service is namespaced, which already have the namespace in the host url.

Krithika3: Thanks

vigith (approved):

PR #122, opened 2022-07-24
feat: add watermark to UI

Krithika3:
Let us decide if we like the placement here:
Screen Shot 2022-07-24 at 2 45 43 PM

And will not be displayed if isWatermark enabled is not true


server/routes/routes.go
```diff
@@ -33,6 +33,6 @@ func v1Routes(r gin.IRouter) {
 	r.GET("/metrics/namespaces/:namespace/pods/:pod", handler.GetPodMetrics)
 	r.GET("/namespaces/:namespace/pipelines/:pipeline/edges", handler.ListPipelineEdges)
 	r.GET("/namespaces/:namespace/pipelines/:pipeline/edges/:edge", handler.GetPipelineEdge)
-	r.GET("/namespaces/:namespace/pipelines/:pipeline/vertices/:vertex/metrics", handler.GetVertexMetrics)
+	r.GET("/namespaces/:namespace/pipelines/:pipeline/vertices/:vertex/metrics", handler.GetVertexMetricsAndWatermark)
```
Krithika3, line 36: maybe we should not call this handle something else

vigith:
Can we show the full date-timestamp rather than just time?

examples/1-simple-pipeline.yaml
```diff
@@ -3,6 +3,8 @@ kind: Pipeline
 metadata:
   name: simple-pipeline
 spec:
+  watermark:
```
whynowy, line 6: I would be better to create a new file particularly with watermark turned on, and leave this one for simple use case demonstration.

Krithika3: Sorry, I had merged this by mistake. Will remove it :) and add a separate one for watermark

Krithika3: Added an example for watermark

Krithika3:
> The placement is perfect. Can we show the full date-timestamp rather than just time?
Screen Shot 2022-07-25 at 8 46 36 PM

examples/1-simple-pipeline.yaml
vigith: let's not add watermark example yet. we need to iron few more details

ui/src/components/pipeline/Pipeline.tsx
```diff
@@ -83,23 +85,72 @@ export function Pipeline() {
...
                 json["processingRates"]["5m"].toFixed(2);
-              processinRates.ratePerFifteenMin =
+              vertexMetrics.ratePerFifteenMin =
                 json["processingRates"]["15m"].toFixed(2);
-              vertexToRateMap.set(vertex.name, processinRates);
+              vertexToMetricsMap.set(vertex.name, vertexMetrics);
             });
         })
-      ).then(() => setVertexRate(vertexToRateMap));
+      ).then(() => setVertexMetrics(vertexToMetricsMap));
```
bbehnke, line 98: Add catch here for potential error scenario.

ui/src/components/pipeline/Pipeline.tsx
```diff
@@ -71,9 +74,8 @@ export function Pipeline() {
     }
   }, [pipeline]);
 
-  // This useEffect is used to obtain all the rates for a given vertex in a pipeline.
-  useEffect(() => {
-    const vertexToRateMap = new Map();
+  const getMetrics = () => {
```
bbehnke, line 77: Convert to useCallback

examples/1-simple-pipeline.yaml
Krithika3: ok, removed it

ui/src/components/pipeline/Pipeline.tsx
```diff
@@ -83,23 +85,72 @@ export function Pipeline() {
...
+
+  // This useEffect is used to obtain metrics for a given vertex in a pipeline and refreshes every 5 minutes
+  useEffect(() => {
+    getMetrics();
+    const interval = setInterval(() => {
+      getMetrics();
+    }, 300000);
+
+    return () => clearInterval(interval);
+  }, [pipeline]);
```
bbehnke, line 110: Add getMetrics as dependency. Don't want to use stale instance once its update to useCallback.

ui/src/components/pipeline/Pipeline.tsx
```diff
@@ -83,23 +85,72 @@ export function Pipeline() {
...
+              vertexToWatermarkMap.set(vertex.name, vertexWatermark);
+            });
+        })
+      ).then(() => setVertexWatermark(vertexToWatermarkMap));
+    }
+  };
+
+  // This useEffect is used to obtain watermark for a given vertex in a pipeline and refreshes every 5 seconds
+  useEffect(() => {
+    getWatermark();
```
bbehnke, line 138: Same for getWatermaerk, useCallback and add as dependency.

ui/src/components/pipeline/Pipeline.tsx
```diff
@@ -83,23 +85,72 @@ export function Pipeline() {
...
+              const vertexWatermark = {} as VertexWatermark;
+              vertexWatermark.isWaterMarkEnabled = json["isWatermarkEnabled"];
+              vertexWatermark.watermark = json["watermark"];
+              vertexWatermark.watermarkLocalTime = new Date(
+                json["watermark"] * 1000
+              ).toLocaleString();
+              vertexToWatermarkMap.set(vertex.name, vertexWatermark);
+            });
+        })
+      ).then(() => setVertexWatermark(vertexToWatermarkMap));
```
bbehnke, line 132: catch

ui/src/components/pipeline/graph/Node.css
```diff
@@ -17,4 +17,24 @@
 
 .node-tooltip {
     font-size: 1.2em;
+}
+
+.node-watermark {
+    width: 50px;
```
bbehnke, line 23: fit-content maybe better than fixed width. Not sure if a high TPS would make it overflow or not.

bbehnke (changes requested):

bbehnke (approved):
LGTM

whynowy:
The rate unit is `xx/sec` not `xx/min`.

Krithika3:
> The rate unit is `xx/sec` not `xx/min`.

Changed

whynowy (approved):

PR #124, opened 2022-07-29
feat: Add watermark for sink vertex

kohlisid:
Adding watermark progressors for the sink vertex

Signed-off-by: Sidhant Kohli <sidhant.kohli@gmail.com>


test/watermark-e2e/functional-wm-test.go
```diff
@@ -0,0 +1 @@
+package watermark_e2e
```
vigith, line 1: please remove this file

vigith (commented):
can you add Kafka too?

vigith:
please expose `<sink_vertex>_SINK` via daemon service.

whynowy:
My understanding is all the sinks should use `noop` right?

pkg/sinks/sink.go
```diff
@@ -32,6 +35,15 @@ func (u *SinkProcessor) Start(ctx context.Context) error {
 	var reader isb.BufferReader
 	var err error
 	fromBufferName := u.VertexInstance.Vertex.GetFromBuffers()[0].Name
+
+	// watermark variables no-op initialization
+	var fetchWatermark fetch.Fetcher = generic.NewNoOpWMProgressor()
+	// publishWatermark is a map representing a progressor per edge, we are initializing them to a no-op progressor
+	publishWatermark := make(map[string]publish.Publisher)
+	for _, buffer := range u.VertexInstance.Vertex.GetToBuffers() {
```
whynowy, line 43: Be aware `GetToBuffers()` returns a particular `sink buffer`, make sure this is what we want.

vigith:
> My understanding is all the sinks should use `noop` right?

sink could have a watermark based on when it wrote to the final sink. today sink's watermark is of the incoming edge's watermark. the whole watermark is right shifted.

pkg/daemon/server/service/pipeline_watermark_query.go
```diff
@@ -53,24 +52,39 @@ func newVertexWatermarkFetcher(pipeline *v1alpha1.Pipeline) *watermarkFetchers {
...
 	wmFetcher.fetchMap = vertexWmMap
 	return wmFetcher
 }
 
+func createWatermarkFetcher(ctx context.Context, pipelineName string, fromBufferName string, vertexName string) *generic.GenericFetch {
+	log := logging.FromContext(ctx)
+	hbBucket := isbsvc.JetStreamProcessorBucket(pipelineName, fromBufferName)
+	hbWatch, err := jetstream.NewKVJetStreamKVWatch(ctx, pipelineName, hbBucket, clients.NewInClusterJetStreamClient())
+	if err != nil {
+		log.Fatalw("JetStreamKVWatch failed", zap.String("HeartbeatBucket", hbBucket), zap.Error(err))
```
whynowy, line 75: Please return error and let the invoker to decide how to handle the error instead of letting it panic here.

pkg/daemon/server/service/pipeline_watermark_query.go
```diff
@@ -53,24 +52,39 @@ func newVertexWatermarkFetcher(pipeline *v1alpha1.Pipeline) *watermarkFetchers {
...
+	}
+	otBucket := isbsvc.JetStreamOTBucket(pipelineName, fromBufferName)
+	otWatch, err := jetstream.NewKVJetStreamKVWatch(ctx, pipelineName, otBucket, clients.NewInClusterJetStreamClient())
+	if err != nil {
+		log.Fatalw("JetStreamKVWatch failed", zap.String("OTBucket", otBucket), zap.Error(err))
```
whynowy, line 80: Same here.

pkg/sinks/sink.go
```diff
@@ -100,15 +115,15 @@ func (u *SinkProcessor) Start(ctx context.Context) error {
...
-func (u *SinkProcessor) getSinker(reader isb.BufferReader, logger *zap.SugaredLogger) (Sinker, error) {
+func (u *SinkProcessor) getSinker(reader isb.BufferReader, logger *zap.SugaredLogger, fetchWM fetch.Fetcher, publishWM map[string]publish.Publisher) (Sinker, error) {
 	sink := u.VertexInstance.Vertex.Spec.Sink
 	// TODO: add watermark
 	if x := sink.Log; x != nil {
-		return logsink.NewToLog(u.VertexInstance.Vertex, reader, logsink.WithLogger(logger))
+		return logsink.NewToLog(u.VertexInstance.Vertex, reader, fetchWM, publishWM, logsink.WithLogger(logger))
 	} else if x := sink.Kafka; x != nil {
-		return kafkasink.NewToKafka(u.VertexInstance.Vertex, reader, kafkasink.WithLogger(logger))
+		return kafkasink.NewToKafka(u.VertexInstance.Vertex, reader, nil, nil, kafkasink.WithLogger(logger))
```
whynowy, line 124: Why are they `nil, nil` here?

pkg/sinks/kafka/kafka.go
```diff
@@ -65,7 +67,7 @@ func NewToKafka(vertex *dfv1.Vertex, fromBuffer isb.BufferReader, opts ...Option
 			forwardOpts = append(forwardOpts, forward.WithReadBatchSize(int64(*x.ReadBatchSize)))
 		}
 	}
-	f, err := forward.NewInterStepDataForward(vertex, fromBuffer, map[string]isb.BufferWriter{vertex.Name: toKafka}, forward.All, applier.Terminal, nil, nil, forwardOpts...)
+	f, err := forward.NewInterStepDataForward(vertex, fromBuffer, map[string]isb.BufferWriter{generic.GetSinkOutboundEdge(vertex): toKafka}, forward.All, applier.Terminal, nil, nil, forwardOpts...)
```
whynowy, line 70: Can we get the buffer name from the Vertex object directly instead of adding an extra function `GetSinkOutboundEdge`?

pkg/sinks/logger/log.go
```diff
@@ -54,7 +57,7 @@ func NewToLog(vertex *dfv1.Vertex, fromBuffer isb.BufferReader, opts ...Option)
 			forwardOpts = append(forwardOpts, forward.WithReadBatchSize(int64(*x.ReadBatchSize)))
 		}
 	}
-	isdf, err := forward.NewInterStepDataForward(vertex, fromBuffer, map[string]isb.BufferWriter{name: toLog}, forward.All, applier.Terminal, nil, nil, forwardOpts...)
+	isdf, err := forward.NewInterStepDataForward(vertex, fromBuffer, map[string]isb.BufferWriter{generic.GetSinkOutboundEdge(vertex): toLog}, forward.All, applier.Terminal, fetchWatermark, publishWatermark, forwardOpts...)
```
whynowy, line 60: Same here.

pkg/sinks/kafka/kafka.go
vigith: I suggested this because tomorrow, we might have different logic for each sink's outbound buffer.

whynowy: If this is the case, should we change the logic in

```
func GenerateSinkBufferName(namespace, pipelineName, vertex string) string {
	return fmt.Sprintf("%s-%s-%s_SINK", namespace, pipelineName, vertex)
}
```

Today we have several different things for edge/bugger:

1. Buffer name (includes edge buffer, source bugger and sink buffer), which is open to external;
2. Stream name (generated from buffer name), it's for internal stream/bucket creation.

`GetSinkOutboundEdge` introduces another thing open to external, and it's generated from buffer name.

whynowy: BTW - in `publishWatermark`, the map key is the sink buffer name, which is not consistent. 

Also, it's `nil, nil` here in this line.

pkg/daemon/server/service/pipeline_watermark_query.go
```diff
@@ -56,33 +57,37 @@ func newVertexWatermarkFetcher(pipeline *v1alpha1.Pipeline) *watermarkFetchers {
 		// Can be checked by querying sinkName_SINK in the service
 		if vertex.Sink != nil {
 			toBufferName := v1alpha1.GenerateSinkBufferName(pipeline.Namespace, pipelineName, vertex.Name)
-			fetchWatermark := createWatermarkFetcher(ctx, pipelineName, toBufferName, vertex.Name)
+			fetchWatermark, err := createWatermarkFetcher(ctx, pipelineName, toBufferName, vertex.Name)
+			if err != nil {
+				log.Fatalw("failed to create watermark fetcher", zap.Error(err))
```
whynowy, line 62: Can we return the error all the way up?

pkg/sinks/sink.go
kohlisid: Have updated in the new commit

pkg/sinks/sink.go
kohlisid: Yes, we are looking for the sink buffer for checking this

pkg/sinks/kafka/kafka.go
kohlisid: @whynowy Which key are you referring to?
Rest is done

vigith: use this `vertex.GetToBuffers()[0].Name`

pkg/daemon/server/service/pipeline_watermark_query.go
```diff
@@ -53,22 +53,43 @@ func newVertexWatermarkFetcher(pipeline *v1alpha1.Pipeline) *watermarkFetchers {
...
-		if err != nil {
-			log.Fatalw("JetStreamKVWatch failed", zap.String("HeartbeatBucket", hbBucket), zap.Error(err))
+		// Adding an extra entry for the sink vertex to check the watermark value progressed out of the vertex.
+		// Can be checked by querying sinkName_SINK in the service
+		if vertex.Sink != nil {
+			toBufferName := v1alpha1.GenerateSinkBufferName(pipeline.Namespace, pipelineName, vertex.Name)
+			fetchWatermark, err := createWatermarkFetcher(ctx, pipelineName, toBufferName, vertex.Name)
+			if err != nil {
+				log.Errorw("failed to create watermark fetcher", zap.Error(err))
+				return nil, err
```
whynowy, line 63: return nil, fmt.Errorf("xxxxxx, %w", err)

whynowy (approved):

kohlisid:
> did you come across a case where `len(vertex.GetToBuffers())` is `0` ?

The 0 len check is added to ensure correctness for the test cases. It fail without this check

whynowy:
> > did you come across a case where `len(vertex.GetToBuffers())` is `0` ?
> 
> The 0 len check is added to ensure correctness for the test cases. It fail without this check

Should fix the test case then.

PR #126, opened 2022-08-03
test: add e2e test for watermark

kohlisid:
Add e2e functional test for checking watermark correctness.

Signed-off-by: Sidhant Kohli <sidhant.kohli@gmail.com>

### DIFF
--- a/test/e2e/functional_test.go
+++ b/test/e2e/functional_test.go
@@ -5,6 +5,7 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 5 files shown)


test/e2e/functional_test.go
```diff
@@ -183,6 +184,81 @@ func (s *FunctionalSuite) TestFlatmapUDF() {
...
+			// timeout for checking watermark progression
+			ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+			defer cancel()
+			go func() {
+				<-ctx.Done()
+				if ctx.Err() == context.DeadlineExceeded {
+					assert.Failf(s.T(), "isWatermarkProgressing timed out", "")
+				}
+			}()
+			assert.Truef(s.T(), isWatermarkProgressing(ctx, client, pipelineName, vertex, 3), "isWatermarkProgressing")
```
whynowy, line 236: If `isWatermarkProgressing` takes more than 1 minute, even though the spawned goroutine has asserted failure, but this guy will not end, which means the `wg` will not end, is that right?

Normally, we do the reverse way like below:

```golang
ctx, cancel := xxxx
defer cancel()
done := make(chan struct{})
go func() {
  isWatermarkProgressing...
  close(done)
}()
select {
  case <- ctx.Done():
    assert.Fail.....
  case <- done:
}
```

vigith:
Can you rebase with master? I am surprised that it is not showing "update branch"

test/e2e/functional_test.go
```diff
@@ -183,6 +184,81 @@ func (s *FunctionalSuite) TestFlatmapUDF() {
...
+		}(vertex)
+	}
+	wg.Wait()
+}
+
+// isWatermarkProgressing checks whether the watermark for a given vertex is progressing monotonically.
+// progressCount is the number of progressions the watermark value should undertake within the timeout deadline for it
+// to be considered as valid.
+func isWatermarkProgressing(ctx context.Context, client *daemonclient.DaemonClient, pipelineName string, vertexName string, progressCount int) bool {
```
vigith, line 238: how about we make this function return error too? this way if context has been cancelled, you can `return ctx.Error()`

test/e2e/functional_test.go
```diff
@@ -183,6 +184,81 @@ func (s *FunctionalSuite) TestFlatmapUDF() {
...
+	prevWatermark := int64(-1)
+	for i := 0; i < progressCount; i++ {
+		currentWatermark := prevWatermark
+		for currentWatermark <= prevWatermark {
+			// select ctx.done
+			wm, err := client.GetVertexWatermark(ctx, pipelineName, vertexName)
+			if err != nil {
+				return false
```
vigith, line 246:
```suggestion
				return false, err
```

test/e2e/functional_test.go
```diff
@@ -183,6 +184,81 @@ func (s *FunctionalSuite) TestFlatmapUDF() {
...
+			}
+			currentWatermark = *wm.Watermark
+			select {
+			case <-ctx.Done():
+				return false
```
vigith, line 251:
```suggestion
				return false, ctx.Error()
```

test/e2e/functional_test.go
```diff
@@ -183,6 +184,81 @@ func (s *FunctionalSuite) TestFlatmapUDF() {
...
+			default:
+				time.Sleep(10 * time.Millisecond)
+			}
+
+		}
+		prevWatermark = currentWatermark
+	}
+	return true
```
vigith, line 259:
```suggestion
	return true, nil
```

test/e2e/functional_test.go
```diff
@@ -183,6 +184,81 @@ func (s *FunctionalSuite) TestFlatmapUDF() {
...
+
+// isWatermarkProgressing checks whether the watermark for a given vertex is progressing monotonically.
+// progressCount is the number of progressions the watermark value should undertake within the timeout deadline for it
+// to be considered as valid.
+func isWatermarkProgressing(ctx context.Context, client *daemonclient.DaemonClient, pipelineName string, vertexName string, progressCount int) bool {
+	prevWatermark := int64(-1)
+	for i := 0; i < progressCount; i++ {
+		currentWatermark := prevWatermark
+		for currentWatermark <= prevWatermark {
+			// select ctx.done
```
vigith, line 243: delete

vigith (approved):

Issue #88, opened 2022-06-30
NoOp watermark support


vigith:
use NoOp watermark progressor and avoid nil checks in forwarder.

vigith:
this is not done, we have `nil` check in forwarder.

vigith:
we need to fix sources, they are still passing `nil`

kohlisid:
@vigith - I'll pick this up and check the what's the issue

whynowy:
> 

@kohlisid - I think we are good, we don't need to get rid of those `nil`s particularly, as we implementing WM for all the sources, they will be gone.

PR #141, opened 2022-08-18
bug: watermark needs nil check

vigith:
add back nil check in forwarder because sources are still passing `nil` for WM fetchers and publishers


chromevoid (approved):

vigith:
@kohlisid the sources are still passing in `nil`, eg. https://github.com/numaproj/numaflow/blob/main/pkg/sources/http/http.go#L165

vigith:
reopens #88

PR #151, opened 2022-08-30
introducing numaflow message

ashwinidulams:
Explain what this PR does.

Introducing a numamessage interface that exposes functions to set and retrieve necessary fields on any message processed by the numaflow platform. Intention is to hide away the internal details of envelope structs that are used today for platform processing. for example isb.Message struct can implement numaflow interface.

This will allow us to introduce additional structs with only the necessary state.


ashwinidulams:
closing this as there are unwanted commits.

PR #152, opened 2022-08-30
introducing numamessage abstractions

ashwinidulams:
Explain what this PR does.

Introducing a numamessage interface that exposes functions to set and retrieve necessary fields on any message processed by the numaflow platform. Intention is to hide away the internal details of envelope structs that are used today for platform processing. for example isb.Message struct can implement numaflow interface.

This will allow us to introduce additional structs with only the necessary state.


pkg/messages/interfaces.go
```diff
@@ -0,0 +1,41 @@
...
+	encoding.BinaryUnmarshaler
+	// SetID sets the ID for a message
+	SetID(id string)
+	// ID is the unique identifier for a message
+	ID() string
+	// SetEventTime represents the timestamp of the message depending on the time characteristics of the pipeline.
+	// please note that EventTime does not mean only the event time.
+	// This is intended to be set once and should be immutable. Therefore, an error could be thrown if
+	// an attempt is made to overwrite the time on a message.
+	SetEventTime(t time.Time) error
```
vigith, line 23: I don't think we need to worry about errors. we can assume that the error checks are handled by the framework itself (in other words, only good data is used to call setters).

pkg/messages/interfaces.go
```diff
@@ -0,0 +1,41 @@
...
+	// EventTime returns the timestamp of the message. Calling Time() on a message should give a meaningful error
+	// for the user to understand that a time has not been set yet.
+	EventTime() (time.Time, error)
```
vigith, line 26: no need of error

pkg/messages/interfaces.go
```diff
@@ -0,0 +1,41 @@
...
+	// Key returns the key of the message
+	Key() string
+	// Payload returns the payload received from the source it was read.
+	Payload() ([]byte, error)
```
vigith, line 30: no need of error

ashwinidulams:
There are a few reasons why i believe we need this abstraction

- adding/deleting struct fields should not require us to change unrelated functions that dont own the struct.
- adding a new message thats appropriate for the context (for example a Read message makes sense only in the Reader context). beyond that offsets or watermark is not required. 
- with an interface, introducing a new struct becomes fairly trivial and the struct implementation can be opinionated about the behavior of different methods in the interface. 
- interface aligns pretty closely with the primitives in a data pipeline (defined by the pipeline model) and the anticipation is almost no changes. 
- There is some transformation expected at the points where data is exchanged and i believe that is an added advantage instead of forcing every system to align with one representation that keeps on growing with the addition of new attributes and behaviors. 
- in other words, we can define structs that are appropriate to the context instead of trying to force all the components to align with one uber struct which makes the system that is modular but is joined at the hip by the contract enforced by this uber struct.

ashwinidulams:
i am not opinionated about how we name this struct. we can definitely rename it to something appropriate like a data record, or event or something similar.

pkg/messages/interfaces.go
ashwinidulams: done

pkg/messages/interfaces.go
ashwinidulams: done

vigith:
please add `Ack()` also to interface.

pkg/messages/interfaces.go
```diff
@@ -0,0 +1,43 @@
...
+	ReadMessage
+	encoding.BinaryMarshaler
+	encoding.BinaryUnmarshaler
+	// SetID sets the ID for a message
+	SetID(id string)
+	// ID is the unique identifier for a message
+	ID() string
+	// SetEventTime represents the timestamp of the message depending on the time characteristics of the pipeline.
+	// please note that EventTime does not mean only the event time.
+	// This is intended to be set once and should be immutable. Therefore, an error could be thrown if
```
whynowy, line 21: No error thrown any more.

pkg/messages/interfaces.go
```diff
@@ -0,0 +1,43 @@
...
+	// an attempt is made to overwrite the time on a message.
+	SetEventTime(t time.Time)
```
whynowy, line 23: Same here.

pkg/messages/interfaces.go
```diff
@@ -0,0 +1,43 @@
...
+	// Key returns the key of the message
+	Key() string
+	// Payload returns the payload received from the source it was read.
+	Payload() []byte
+}
+
+// ReadMessage exposes functions to set and get in-vertex stream characteristics of a NumaMessage.
+type ReadMessage interface {
+	// Offset returns the offset of the message from the source it was read.
+	Offset() isb.Offset
```
whynowy, line 36: Should we move `Offset` to the same package?

whynowy:
Can you refactor current code with a new implementation? I'd love to see how it looks like.

whynowy:
I. believed this one has been merged in another PR, can we close it? @ashwinidulams

PR #153, opened 2022-08-30
chore: use string as the type of key

whynowy:

Explain what this PR does.


pkg/udf/applier/uds_http.go
```diff
@@ -166,7 +166,8 @@ retry:
 			Header: isb.Header{
 				PaneInfo: parentPaneInfo,
 				ID:       fmt.Sprintf("%s-%d", offset.String(), i),
-				Key:      key,
+				// TODO: change to use string directly after SKD upgrade
```
vigith, line 169:
```suggestion
				// TODO: change to use string directly after SDK upgrade
```

vigith (approved):

PR #155, opened 2022-09-01
pbq implementation

yhl25:
Explain what this PR does.
persistent buffer queue implementation backed with memory store and no-op store


pkg/pbq/pbq.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/numaproj/numaflow/pkg/pbq/store"
+	"github.com/numaproj/numaflow/pkg/shared/logging"
+	"go.uber.org/zap"
+	"time"
+)
+
+var COBErr error = errors.New("error while writing to pbq, pbq is closed")
+var EOF error = errors.New("error while reading, EOF")
+
+type PBQ struct {
```
vigith, line 16: comment PQB

pkg/pbq/pbqmanager.go
```diff
@@ -0,0 +1,164 @@
...
+// Manager helps in managing the lifecycle of PBQ instances
+type Manager struct {
+	options *store.Options
+	pbqMap  map[string]*PBQ
+	log     *zap.SugaredLogger
+}
+
+// NewManager returns new instance of manager
+// We don't intend this to be called by multiple routines.
+func NewManager(ctx context.Context, opts ...store.SetOption) (*Manager, error) {
```
vigith, line 28: we could make NewManager take PBQ options like `opts ...Options` and one of the option for the PBQ could be `WithStoreOptions(opts ...Store.Options)`.

WDYT?

pkg/pbq/pbq.go
```diff
@@ -0,0 +1,157 @@
...
+	cob         bool // cob to avoid panic in case writes happen after close of book
+	partitionID string
+	options     *store.Options
+	isReplaying bool
+	manager     *Manager
+	log         *zap.SugaredLogger
+}
+
+// NewPBQ accepts size and store and returns new PBQ
+func NewPBQ(ctx context.Context, partitionID string, persistentStore store.Store, pbqManager *Manager, options *store.Options) (*PBQ, error) {
```
vigith, line 28: here we do not require `store.Options` since we are passing in `store.Store`, it requires PBQ options.

pkg/pbq/pbqmanager.go
```diff
@@ -0,0 +1,164 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"sync"
+	"time"
+)
+
+var NonExistentPBQErr error = errors.New("missing PBQ for the partition")
+
+// Manager helps in managing the lifecycle of PBQ instances
+type Manager struct {
+	options *store.Options
```
vigith, line 21:
```suggestion
	storeOpts *store.Options
	pbqOpts   *Options
```

pkg/pbq/pbq.go
```diff
@@ -0,0 +1,157 @@
...
+
+	return p, nil
+}
+
+// WriteFromISB writes message to pbq and persistent store
+// We don't need a context here as this is invoked for every message.
+func (p *PBQ) WriteFromISB(ctx context.Context, message *isb.Message) (writeErr error) {
+	// if we are replaying records from the store, writes should be blocked
+	for {
+		if !p.isReplaying {
```
vigith, line 49: `isReplaying` will have race condition. Both Writer and Reader is accessing it via go routines.

vigith (commented):
Let's also add `Read` via channel, else we might be busy fixing the code that will be deleted soon.

vigith:
let's also remove replaying logic from the current code since we do not support replaying. we will revisit once we have a concrete store implementation.

pkg/pbq/pbq.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/numaproj/numaflow/pkg/shared/logging"
+	"go.uber.org/zap"
+	"time"
+)
+
+var COBErr error = errors.New("error while writing to pbq, pbq is closed")
+var EOF error = errors.New("error while reading, EOF")
+
+type PBQ struct {
+	Store       store.Store
```
whynowy, line 17: Lowercase?

pkg/pbq/pbqmanager.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	return pbqManager, nil
+}
+
+// ListPartitions returns all the pbq instances
+func (m *Manager) ListPartitions() []*PBQ {
+
+	pbqList := make([]*PBQ, len(m.pbqMap))
+	i := 0
+	for _, val := range m.pbqMap {
```
whynowy, line 50: You need a lock to access the map.

pkg/pbq/pbqmanager.go
```diff
@@ -0,0 +1,164 @@
...
+		return nil
+	case dfv1.InMemoryType:
+		return nil
+	case dfv1.FileSystemType:
+		return errors.New("not implemented")
+	}
+
+	for _, value := range m.pbqMap {
+		// set replay flag so that it replays messages from the store during startup
+		value.SetIsReplaying(ctx, true)
```
whynowy, line 111: No need to use `replaying`, instead, a lock is good enough.

pkg/pbq/pbqmanager.go
```diff
@@ -0,0 +1,164 @@
...
+		go func(q *PBQ) {
+			defer wg.Done()
+			var closeErr error
+			var attempt int
+			closeErr = wait.ExponentialBackoffWithContext(ctx, PBQCloseBackOff, func() (done bool, err error) {
+				closeErr = q.CloseWriter()
+				if closeErr != nil {
+					attempt += 1
+					m.log.Errorw("Failed to close pbq, retrying", zap.Any("attempt", attempt), zap.Any("partitionID", q.partitionID), zap.Error(closeErr))
+					return false, err
```
whynowy, line 141: When you return error, the retry ends.

pkg/pbq/store/options.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func (o *Options) ReadTimeoutSecs() int {
+	return o.readTimeoutSecs
+}
+
+func DefaultOptions() *Options {
+	return &Options{
+		maxBatchSize:    dfv1.DefaultPBQMaxBufferSize,
+		syncDuration:    dfv1.DefaultPBQSyncDuration,
```
whynowy, line 45: use `time.Duration.`

pkg/pbq/store/memory/store.go
```diff
@@ -0,0 +1,87 @@
...
+	"context"
+	"github.com/numaproj/numaflow/pkg/isb"
+	"github.com/numaproj/numaflow/pkg/pbq/store"
+	"github.com/numaproj/numaflow/pkg/pbq/util"
+	"github.com/numaproj/numaflow/pkg/shared/logging"
+	"go.uber.org/zap"
+)
+
+// MemoryStore implements PBQStore which stores the data in memory
+type MemoryStore struct {
```
whynowy, line 13: `memoryStore`

pkg/pbq/store/memory/store.go
```diff
@@ -0,0 +1,87 @@
...
+	writePos    int64
+	readPos     int64
+	storage     []*isb.Message
+	options     *store.Options
+	log         *zap.SugaredLogger
+	partitionID string
+}
+
+// NewMemoryStore returns new memory store
+func NewMemoryStore(ctx context.Context, partitionID string, options *store.Options) (*MemoryStore, error) {
```
whynowy, line 24: return interface.

pkg/pbq/store/interfaces.go
```diff
@@ -0,0 +1,19 @@
+package store
+
+import (
+	"github.com/numaproj/numaflow/pkg/isb"
+)
+
+// Store provides methods to read, write and delete data from a durable store.
+type Store interface {
+	// ReadFromStore returns all the messages from the persisted store
+	ReadFromStore(size int64) ([]*isb.Message, bool, error)
```
whynowy, line 10: `Read()`?

`Store.ReadFromStore()` is weird.

pkg/pbq/store/interfaces.go
```diff
@@ -0,0 +1,19 @@
...
+	// WriteToStore writes message to persisted store
+	WriteToStore(msg *isb.Message) error
```
whynowy, line 12: `Write()`?

pkg/pbq/interfaces.go
```diff
@@ -0,0 +1,28 @@
...
+	"context"
+	"github.com/numaproj/numaflow/pkg/isb"
+)
+
+// Reader provides methods to read from PBQ.
+type Reader interface {
+	// ReadFromPBQ reads from PBQ
+	ReadFromPBQ(ctx context.Context, size int64) ([]*isb.Message, error)
+	// CloseReader closes the reader.
+	CloseReader() error
```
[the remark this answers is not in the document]
yhl25: since pbq implements both reader and writer interfaces, they cannot have the same method name

pkg/pbq/pbqmanager.go
yhl25: yes, I have added

pkg/pbq/pbqmanager.go
yhl25: yeah it is better that way, I have added separate options for pbq now

pkg/pbq/pbq.go
yhl25: done

pkg/pbq/pbq.go
yhl25: I have removed isReplaying flag

pkg/apis/numaflow/v1alpha1/const.go
```diff
@@ -110,6 +110,23 @@ const (
 	DefaultTargetProcessingSeconds = 20  // Default targeted time in seconds to finish processing all the pending messages for a source
 	DefaultTargetBufferUsage       = 50  // Default targeted percentage of balanced buffer usage
 	DefaultReplicasPerScale        = 2   // Default maximum replicas to be scaled up or down at once
+
+	// Default Persistent Buffer Queue Options
+	DefaultPBQMaxBufferSize   = 100000        // Default buffer size for pbq in bytes
+	DefaultPBQSyncDuration    = 2             // Default sync duration for pbq in seconds
+	DefaultPBQStoreType       = "file-system" // Default store type in string
```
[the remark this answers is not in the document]
yhl25: done

pkg/pbq/pbq.go
yhl25: done

pkg/pbq/pbqmanager.go
```diff
@@ -0,0 +1,164 @@
...
+		i++
+	}
+	return pbqList
+}
+
+// GetPBQ returns pbq for the given partitionID, if createIfMissing is set to true
+// it creates and return a new pbq instance
+func (m *Manager) GetPBQ(ctx context.Context, partitionID string, createIfMissing bool) (*PBQ, bool, error) {
+
+	pbqInstance, ok := m.pbqMap[partitionID]
```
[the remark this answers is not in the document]
yhl25: done

pkg/pbq/pbqmanager.go
yhl25: yeah realised it now, I am returning nil now

pkg/pbq/store/options.go
yhl25: done

pkg/pbq/store/memory/store.go
yhl25: done

pkg/pbq/store/memory/store.go
yhl25: done

pkg/pbq/pbq.go
```diff
@@ -0,0 +1,101 @@
...
+	output      chan *isb.Message
+	cob         bool // cob to avoid panic in case writes happen after close of book
+	partitionID string
+	options     *Options
+	manager     *Manager
+	log         *zap.SugaredLogger
+}
+
+// Write writes message to pbq and persistent store
+// We don't need a context here as this is invoked for every message.
```
vigith, line 24:
```suggestion
```

pkg/pbq/pbq.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	// we need context to get out of blocking write
+	select {
+	case p.output <- message:
+		writeErr = p.store.Write(message)
+		return
+	case <-ctx.Done():
+		// closing the output channel will not cause panic, since its inside select case
+		close(p.output)
+		writeErr = p.store.Close()
```
vigith, line 40: why are we closing it here? `CloseWriter` is doing this.

pkg/pbq/pbq.go
```diff
@@ -0,0 +1,101 @@
...
+	cob         bool // cob to avoid panic in case writes happen after close of book
+	partitionID string
+	options     *Options
+	manager     *Manager
+	log         *zap.SugaredLogger
+}
+
+// Write writes message to pbq and persistent store
+// We don't need a context here as this is invoked for every message.
+func (p *PBQ) Write(ctx context.Context, message *isb.Message) (writeErr error) {
```
vigith, line 25:
```suggestion
func (p *PBQ) Write(ctx context.Context, message *isb.Message) (error) {
```

usually this is done to manipulate the return result via `defer`

pkg/pbq/pbq.go
vigith, line 40:
```suggestion
		writeErr = ctx.Err()
```
`ctx.Done` implicitly means that the write hasn't succeeded.

pkg/pbq/pbq.go
```diff
@@ -0,0 +1,101 @@
...
+	select {
+	case p.output <- message:
+		writeErr = p.store.Write(message)
+		return
+	case <-ctx.Done():
+		// closing the output channel will not cause panic, since its inside select case
+		close(p.output)
+		writeErr = p.store.Close()
+	}
+	return
```
vigith, line 42:
```suggestion
	return writerErr
```

pkg/pbq/pbq.go
```diff
@@ -0,0 +1,101 @@
...
+	if p.cob {
+		p.log.Errorw("failed to write message to pbq, pbq is closed", zap.Any("partitionID", p.partitionID), zap.Any("header", message.Header))
+		writeErr = COBErr
+		return
+	}
+	// we need context to get out of blocking write
+	select {
+	case p.output <- message:
+		writeErr = p.store.Write(message)
+		return
```
vigith, line 36:
```suggestion
```

pkg/pbq/pbq.go
```diff
@@ -0,0 +1,101 @@
...
+	// if cob we should return
+	if p.cob {
+		p.log.Errorw("failed to write message to pbq, pbq is closed", zap.Any("partitionID", p.partitionID), zap.Any("header", message.Header))
+		writeErr = COBErr
+		return
+	}
+	// we need context to get out of blocking write
+	select {
+	case p.output <- message:
+		writeErr = p.store.Write(message)
```
vigith, line 35:
```suggestion
		// this store.Write is an `inSync` flush (if need be). The performance will be very bad but the system is correct.
		// TODO: shortly in the near future we will move to async writes.
		writeErr = p.store.Write(message)
```

pkg/pbq/pbq.go
```diff
@@ -0,0 +1,101 @@
...
+		// closing the output channel will not cause panic, since its inside select case
+		close(p.output)
+		writeErr = p.store.Close()
+	}
+	return
+}
+
+//CloseOfBook closes output channel
+func (p *PBQ) CloseOfBook() {
+	close(p.output)
```
vigith, line 47: does the order matter?

vigith: ideally it shouldn't matter because there is only a single writer that is orchestrating both write and close.

pkg/pbq/pbqmanager.go
vigith: why do we need this lock? how many managers will be active at any point? is manager supposed to be accessed concurrently? 

One could argue that if `GetPQB` returns NonExistentPBQ, the entire PBQ could be managed through a go-routine, and hence a lock is required. 

Please document this thought in the `Manager` struct.

pkg/pbq/pbqmanager.go
```diff
@@ -0,0 +1,199 @@
...
+		go func(q *PBQ) {
+			defer wg.Done()
+			var closeErr error
+			var attempt int
+			closeErr = wait.ExponentialBackoffWithContext(ctx, PBQCloseBackOff, func() (done bool, err error) {
+				closeErr = q.CloseWriter()
+				if closeErr != nil {
+					attempt += 1
+					m.log.Errorw("Failed to close pbq, retrying", zap.Any("attempt", attempt), zap.Any("partitionID", q.partitionID), zap.Error(closeErr))
+					return false, nil
```
vigith, line 154: a 1 liner doc here

pkg/pbq/pbqmanager.go
```diff
@@ -0,0 +1,199 @@
...
+		wg.Add(1)
+		go func(q *PBQ) {
+			defer wg.Done()
+			var closeErr error
+			var attempt int
+			closeErr = wait.ExponentialBackoffWithContext(ctx, PBQCloseBackOff, func() (done bool, err error) {
+				closeErr = q.CloseWriter()
+				if closeErr != nil {
+					attempt += 1
+					m.log.Errorw("Failed to close pbq, retrying", zap.Any("attempt", attempt), zap.Any("partitionID", q.partitionID), zap.Error(closeErr))
```
vigith, line 153:
```suggestion
					m.log.Warnw("Failed to close pbq, retrying", zap.Any("attempt", attempt), zap.Any("partitionID", q.partitionID), zap.Error(closeErr))
```

pkg/pbq/store/interfaces.go
```diff
@@ -0,0 +1,19 @@
+package store
+
+import (
+	"github.com/numaproj/numaflow/pkg/isb"
+)
+
+// Store provides methods to read, write and delete data from a durable store.
```
vigith, line 7:
```suggestion
// Store provides methods to read, write and delete data from the store.
```

pkg/pbq/store/memory/store.go
```diff
@@ -0,0 +1,87 @@
...
+
+// ReadFromStore will return upto N messages persisted in store
+// this function will be invoked during bootstrap if there is a restart
+func (m *memoryStore) Read(size int64) ([]*isb.Message, bool, error) {
+	if m.IsEmpty() || m.readPos >= m.writePos {
+		m.log.Errorw(store.ReadStoreEmptyErr.Error())
+		return []*isb.Message{}, true, nil
+	}
+
+	size = util.Min(size, m.writePos-m.readPos)
```
vigith, line 47: please document the thought

pkg/pbq/store/memory/store.go
```diff
@@ -0,0 +1,87 @@
...
+// ready for GC
+func (m *memoryStore) GC() error {
+	m.storage = nil
+	m.writePos = -1
+	return nil
+}
+
+// IsEmpty check if there are any records persisted in store
+func (m *memoryStore) IsEmpty() bool {
+	return m.writePos == 0
```
vigith, line 86:
```suggestion
	return m.closed == true || m.writePos == 0
```
does it make sense to do this?

pkg/pbq/store/noop/store.go
```diff
@@ -0,0 +1,36 @@
...
+func (p *PBQNoOpStore) Close() error {
+	return nil
+}
+
+func (p *PBQNoOpStore) GC() error {
+	return nil
+}
+
+func (p *PBQNoOpStore) IsEmpty() bool {
+	return false
```
vigith, line 35: shouldn't this be `true`?

perhaps we do not require `IsEmpty` to be in the interface. if yes, unexport it.

pkg/pbq/pbq.go
yhl25: yeah, agreed

pkg/pbq/pbq.go
yhl25: it shouldn't matter

pkg/pbq/pbqmanager.go
yhl25: I have documented this

pkg/pbq/pbqmanager.go
yhl25: done

pkg/pbq/store/memory/store.go
yhl25: done

pkg/pbq/store/memory/store.go
yhl25: no, closed means no more writes are allowed
isEmpty should return true only when the store is created and there are no writes

pkg/pbq/store/noop/store.go
yhl25: done, removed it from the interface

vigith (approved):

pkg/pbq/interfaces.go
```diff
@@ -0,0 +1,28 @@
+package pbq
+
+import (
+	"context"
+	"github.com/numaproj/numaflow/pkg/isb"
+)
+
+// Reader provides methods to read from PBQ.
+type Reader interface {
```
whynowy, line 9: Based on my understanding, the the PBQ is a channel concept, the read and write happen in same application, and read and write even share the same instance. Then it seems to make no sense to have 2 interfaces defined.

Also, based on the code below, there's only one implementation, it's not related to any type of `store` (store is an interface), and your `NewPBQ` returns an `*PBQ` instead of an interface, then what's the point of having interfaces defined here?

whynowy: Maybe merging these 2 interfaces makes more sense?

vigith: yes, i think we should merge both `Reader` and `Writer` interface into `ReaderWriter` interface. It will look more cleaner. We really do not have to worry about the separation of concern because we are not leaking this to SDK.

ashwinidulams: As of now, its only one implementation, but depending on how  windowing shapes up, we may have to have composite readers vs single readers etc. 

one of the reasons why reader and writer are two interfaces is to make sure that we give access to behavior that is expected instead of throwing open all the functions. also the intent becomes clear.

can we merge these two later? if needed, we can introduce a new interface which composes both reader and writer something like `PBQReadWriter`

ashwinidulams: To the comment about *PBQ, i agree. returning a pointer when you have interfaces defined probably doesn't make a lot of sense. PBQManager could instead expose methods for reader/writer

vigith: @ashwinidulams what do you mean by composite readers?

vigith: @yhl25 some refactoring has been made.. please merge https://github.com/yhl25/numaflow/pull/1

vigith: the interface now allows you to do 
```golang
pbq := NewPBQ() ReaderWriter
// for left side
ReadLoop(pbq.(PBQWriter))
// for right loop
ProcessAndForward(pbq.(PBQReader))
```

ashwinidulams: in case of session windowing, depending on the order in which elements arrive, one could have isolated windows to begin with which may need to be merged (once the elements in the gaps arrive). To handle such a situation its not easy to move the data around and so a partitioner will be required to orchestrate elements from multiple partitions to be read by a single reader.  a Pbq reader will be required to support reading from multiple partitions thereby creating a fan-in scenario. this is what i mean by a composite reader.

pkg/pbq/pbq_test.go
```diff
@@ -0,0 +1,245 @@
...
+
+	for _, msg := range writeMessages {
+		err = pq.Write(ctx, &msg)
+	}
+	pq.CloseOfBook()
+
+	assert.Error(t, err, store.WriteStoreFullErr)
+	// check if the messages are persisted in store
+	pbqCh := pq.ReadCh()
+	var storeMessages = make([]*isb.Message, 0)
```
ashwinidulams, line 240: lets add separate test cases for store in a different PR. we should use PBQ to verify the store contents.

vigith (approved):
LGTM

pkg/pbq/pbqmanager.go
```diff
@@ -0,0 +1,204 @@
...
+	for _, val := range m.pbqMap {
+		pbqList[i] = val
+		i++
+	}
+
+	return pbqList
+}
+
+// GetPBQ returns pbq for the given partitionID
+func (m *Manager) GetPBQ(partitionID string) (ReadWriteCloser, error) {
```
whynowy, line 106: Minor - think this could simply return `ReadWriteCloser` without using an error, which would be easier to use - there's no other error could be generated in the implementation, so using `nil` instead of a particular error to tell if there's a PBQ existing is more convenient.

pkg/pbq/pbq.go
```diff
@@ -0,0 +1,100 @@
...
+		return COBErr
+	}
+	var writeErr error
+	// we need context to get out of blocking write
+	select {
+	case p.output <- message:
+		// this store.Write is an `inSync` flush (if need be). The performance will be very bad but the system is correct.
+		// TODO: shortly in the near future we will move to async writes.
+		writeErr = p.store.Write(message)
+	case <-ctx.Done():
```
whynowy, line 42: I think this `case` is not needed, also there's no need to use `select`.

Should not `close(p.output)` there as well.

vigith: we cannot close it here because it will be signaling a close-of-book. ctx.Done was added there if avoid a holdup during the shutdown (in case there is a back-pressure and we are not able to write to the forwarding channel).

whynowy: When using `select`, if it's already in a `case` execution (let's say, here in `p.store.Write()`), it will not respond to another `case` (`ctx.Done()`). If `p.store.Write()` is done with an error return, it breaks out the `select` directly, so there's no need to have a `case <- ctx.Done()` here.

whynowy: If we want to address the store writing back-pressure issue, we should introduce a `context` into `p.store.Write()`.

yhl25: select is to avoid infinite blocking on the output channel

vigith: the `ctx` in only for the channel, ATM we are not worried about the write to the store in a blocked state. we will have to revisit when we have a proper disk store. there is already a TODO for this.

whynowy: > select is to avoid infinite blocking on the output channel

See my comments above, that will not work.

yhl25: yeah, I know it won't work on the store, but as mentioned by @vigith we are not worried about the writing to the store getting blocked, context and select are to avoid blocked write on the output channel

pkg/pbq/pbqmanager.go
yhl25: Okay

pkg/pbq/pbq.go
whynowy: Makes sense!

One last question - do we really want to `close(p.output)` here? Once it's closed, how to continue writing to it?

yhl25: we are closing the channel on ctx.Done(), so there will not be any writes after that right?

whynowy: This is to assume the `write()` either succeed or block until exit (and no way to continue). Imagine the invoker wants to call `Write()` with a timeout in the context, in this case there's no way to let him/her retry after timeout.

yhl25: Yeah, `Write()` with timeout is not supported, I don't think we should. @vigith @ashwinidulams WDYT?

ashwinidulams: we shouldn't support writes with timeout. actually there is no difference between a blocking write and a write with timeout in this context since if the write doesn't go through or is taking long time, then it means there is some back pressure. with timeout, its no different from a blocking write on the left side of PBQ.

we will also offer option for pipeline developers to pick async flush option which will eliminate the need for blocking.

whynowy: Ok.

whynowy (approved):
LGTM

PR #157, opened 2022-09-04
feat: windowing operations

vigith:
support the following windowing strategies
* fixed windows
* sliding windows

### DIFF
--- a/pkg/window/strategy/sliding/sliding.go
+++ b/pkg/window/strategy/sliding/sliding.go
@@ -1,0 +1,53 @@
+// Package sliding implements Sliding windows.  Sliding windows are defined by a window size and slide
+// period, e.g. hourly windows starting every minute. Fixed windows are really a special case of sliding windows where
+// size equals period.
+package sliding
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/numaproj/numaflow/pkg/window"
+)
+
+// Sliding implements Sliding window.
+type Sliding struct {
+	Length          time.Duration
+	PeriodInSeconds int
+	sliceCount      int
+}
+
+var _ window.Windower = (*Sliding)(nil)
+
+// NewSliding returns a Sliding window.
+func NewSliding(length time.Duration, periodInSeconds int) *Sliding {
+	sliceCount := int(length.Seconds() / float64(periodInSeconds))
+
+	return &Sliding{
+		Length:          length,
+		PeriodInSeconds: periodInSeconds,
+		sliceCount:      sliceCount,
+	}
+}
+
+// AssignWindow assigns a window for the given eventTime.
+func (s *Sliding) AssignWindow(eventTime time.Time) []*window.IntervalWindow {
+	slidingWindows := make([]*window.IntervalWindow, s.sliceCount)
+
+	tmp := eventTime.Truncate(s.Length)
+	for i := 0; i < s.sliceCount; i++ {
+		start := tmp
+		slidingWindows[i] = &window.IntervalWindow{
+			Start: start,
+			End:   start.Add(time.Second * time.Duration(s.PeriodInSeconds)),
+		}
+		fmt.Println(slidingWindows[i].Start, slidingWindows[i].End)
+		tmp = tmp.Add(-1 * time.Second * time.Duration(s.PeriodInSeconds))
+	}
+	return slidingWindows
+}
+
+// TrackWindowByKey tracks Sliding window.
+func (s *Sliding) TrackWindowByKey(_ string, _ *window.IntervalWindow) {
+	return
+}
--- a/pkg/window/strategy/sliding/sliding_test.go
+++ b/pkg/window/strategy/sliding/sliding_test.go
@@ -1,0 +1,128 @@
+package sliding
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/numaproj/numaflow/pkg/window"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSliding_AssignWindow(t *testing.T) {
+	loc, _ := time.LoadLocation("UTC")
+	baseTime := time.Unix(1651129201, 0).In(loc)
+
+	type fields struct {
+		Length          time.Duration
+		PeriodInSeconds int
+	}
+
+	tests := []struct {
+		name      string
+		fields    fields
+		eventTime time.Time
+		want      []*window.IntervalWindow
+	}{
+		{
+			name: "2in_1min",
+			fields: fields{
+				Length:          120 * time.Second,
+				PeriodInSeconds: 60,
+			},
+			eventTime: baseTime,
+			want: []*window.IntervalWindow{
+				{
+					Start: time.Unix(1651129200, 0).In(loc),
+					End:   time.Unix(1651129260, 0).In(loc),
+				},
+				{
+					Start: time.Unix(1651129200-60, 0).In(loc),
+					End:   time.Unix(1651129260-60, 0).In(loc),
+				},
+			},
+		},
+		{
+			name: "2min-as-sec_1min",
+			fields: fields{
+				Length:          2 * time.Minute,
+				PeriodInSeconds: 60,
+			},
+			eventTime: baseTime,
+			want: []*window.IntervalWindow{
+				{
+					Start: time.Unix(1651129200, 0).In(loc),
+					End:   time.Unix(1651129200+60, 0).In(loc),
+				},
+				{
+					Start: time.Unix(1651129200-60, 0).In(loc),
+					End:   time.Unix(1651129200+60-60, 0).In(loc),
+				},
+			},
+		},
+		{
+			name: "4min_2min",
+			fields: fields{
+				Length:          4 * time.Minute,
+				PeriodInSeconds: 120,
+			},
+			eventTime: baseTime,
+			want: []*window.IntervalWindow{
+				{
+					Start: time.Unix(1651129200, 0).In(loc),
+					End:   time.Unix(1651129200+120, 0).In(loc),
+				},
+				{
+					Start: time.Unix(1651129200-120, 0).In(loc),
+					End:   time.Unix(1651129200+120-120, 0).In(loc),
+				},
+			},
+		},
+		{
+			name: "4min_4min",
+			fields: fields{
+				Length:          4 * time.Minute,
+				PeriodInSeconds: 240,
+			},
+			eventTime: baseTime,
+			want: []*window.IntervalWindow{
+				{
+					Start: time.Unix(1651129200, 0).In(loc),
+					End:   time.Unix(1651129200+240, 0).In(loc),
+				},
+			},
+		},
+		{
+			name: "4min_5min",
+			fields: fields{
+				Length:          4 * time.Minute,
+				PeriodInSeconds: 300,
+			},
+			eventTime: baseTime,
+			want:      []*window.IntervalWindow{},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			s := NewSliding(tt.fields.Length, tt.fields.PeriodInSeconds)
+			got := s.AssignWindow(tt.eventTime)
+			equal, err := windowIntervalEqual(got, tt.want)
+			assert.NoError(t, err)
+			assert.True(t, equal)
+		})
+	}
+}
+
+func windowIntervalEqual(got, want []*window.IntervalWindow) (bool, error) {
+	if len(got) != len(want) {
+		return false, fmt.Errorf("len(got)=%d NOT EQUAL TO len(want)=%d", len(got), len(want))
+	}
+
+	for idx, intervalWindow := range want {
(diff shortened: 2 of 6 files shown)


pkg/window/aligned/active.go
```diff
@@ -0,0 +1,98 @@
+// Package aligned maintains the state of active keyed windows in a vertex.
+// Keyed Window maintains the association between set of keys and an interval window.
```
vigith, line 2:
```suggestion
// KeyedWindow maintains the association between a set of keys and an interval window. KeyedWindow is a superset of non-keyed windows.
```

pkg/window/aligned/active.go
```diff
@@ -0,0 +1,98 @@
...
+// aligned provides functionality Add, Get window for a given interval, remove windows whose End time is before the
```
vigith, line 3:
```suggestion
// aligned also provides the lifecycle management of an interval window. Watermark is used to trigger the expiration of windows.
```

pkg/window/aligned/active.go
```diff
@@ -0,0 +1,98 @@
...
+// current watermark.
```
vigith, line 4:
```suggestion
```

pkg/window/aligned/active.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/numaproj/numaflow/pkg/window"
+	"sync"
+	"time"
+)
+
+// ActiveWindows maintains the state of active windows
+// All the operations in ActiveWindows order the entries in the ascending order of start time.
+// So the earliest window is at the front and the oldest window is at the end.
+type ActiveWindows struct {
+	entries *list.List
```
vigith, line 18: can you document entries? let's explicitly say whether Head() grows v/s Tail().

pkg/window/aligned/active.go
```diff
@@ -0,0 +1,98 @@
...
+// CreateKeyedWindow adds a new keyed window for a given interval window
+func (aw *ActiveWindows) CreateKeyedWindow(iw *window.IntervalWindow) *KeyedWindow {
+	aw.lock.Lock()
+	defer aw.lock.Unlock()
+
+	kw := NewKeyedWindow(iw)
+
+	// this could be the first window
+	if aw.entries.Len() == 0 {
+		aw.entries.PushFront(kw)
```
vigith, line 39: do we need to run the rest of the code if this is the first entry?

pkg/window/aligned/active.go
```diff
@@ -0,0 +1,98 @@
...
+
+	// late arrival
+	if earliestWindow.Start == kw.End || earliestWindow.Start.After(kw.End) {
+		aw.entries.PushFront(kw)
+	} else if recentWindow.End == kw.Start || recentWindow.End.Before(kw.Start) {
+		// early arrival
+		aw.entries.PushBack(kw)
+	} else {
+		// a window in the middle
+		for e := aw.entries.Front(); e != nil; e = e.Next() {
```
vigith, line 53: will it be amortized better if we come from `Back()` ?

pkg/window/aligned/active.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	return kw
+}
+
+// GetKeyedWindow returns an existing window for the given interval
+func (aw *ActiveWindows) GetKeyedWindow(iw *window.IntervalWindow) *KeyedWindow {
+	aw.lock.RLock()
+	defer aw.lock.RUnlock()
+	// check if we already have a window
+	for e := aw.entries.Front(); e != nil; e = e.Next() {
```
vigith, line 69: Is Back better?

vigith (commented):
partial review

pkg/window/aligned/active.go
ashwinidulams: addressed.

pkg/window/aligned/active.go
ashwinidulams: fair. changed.

pkg/window/aligned/active.go
ashwinidulams: true. changed.

pkg/window/aligned/active.go
ashwinidulams: addressed

pkg/window/aligned/active.go
```diff
@@ -0,0 +1,100 @@
...
+	if aw.entries.Len() == 0 {
+		aw.entries.PushFront(kw)
+		return kw
+	}
+
+	earliestWindow := aw.entries.Front().Value.(*KeyedWindow)
+	recentWindow := aw.entries.Back().Value.(*KeyedWindow)
+
+	// late arrival
+	if earliestWindow.Start == kw.End || earliestWindow.Start.After(kw.End) {
```
vigith, line 48: let's use `time.Equal()` instead of `==`

pkg/window/aligned/active.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	return kw
+}
+
+// GetKeyedWindow returns an existing window for the given interval
+func (aw *ActiveWindows) GetKeyedWindow(iw *window.IntervalWindow) *KeyedWindow {
+	aw.lock.RLock()
+	defer aw.lock.RUnlock()
+	// check if we already have a window
+	for e := aw.entries.Back(); e != nil; e = e.Prev() {
```
vigith, line 71: you can return early if an element is not present, because the list is stored. we do not have to loop through.

ashwinidulams: done.

pkg/window/aligned/active.go
ashwinidulams: done.

pkg/window/strategy/sliding/sliding_test.go
```diff
@@ -0,0 +1,154 @@
...
+			eventTime: baseTime,
+			want: []*window.IntervalWindow{
+				{
+					Start: time.Unix(1651129200, 0).In(loc),
+					End:   time.Unix(1651129200+240, 0).In(loc),
+				},
+			},
+		},
+		{
+			name: "4min_5min",
```
whynowy, line 96: What will happen with this kind of cases? Shall we treat it as invalid requirements?

Similarly, `4min_3min`.

vigith: `4min_3min` is a valid input. if the period is greater than the length, then we can treat it as invalid for the time being.

pkg/window/windower.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+// Windower assigns the element to zero or more windows.
+type Windower interface {
+	// AssignWindow assigns the event to the window based on give window configuration.
+	AssignWindow(eventTime time.Time) []*IntervalWindow
+}
+
+// IntervalWindow has the window boundary details.
+type IntervalWindow struct {
```
whynowy, line 14: The name `IntervalWindow` is confusing when it is together with `KeyedWindow`, see the code below.

```
func (aw *ActiveWindows) CreateKeyedWindow(iw *window.IntervalWindow) *keyed.KeyedWindow {
```

pkg/window/keyed/keyed.go
```diff
@@ -0,0 +1,54 @@
...
+	partitions := make([]PartitionId, len(kw.Keys))
+	for i, key := range kw.Keys {
+		partitions[i] = Partition(kw.IntervalWindow, key)
+	}
+
+	return partitions
+}
+
+// Partition returns the partitionId for a given window and a key
+func Partition(window *window.IntervalWindow, key string) PartitionId {
```
whynowy, line 52: Make it private?

pkg/window/strategy/fixed/lifecycle.go
```diff
@@ -0,0 +1,130 @@
...
+	if aw.entries.Len() == 0 {
+		aw.entries.PushFront(kw)
+		return kw
+	}
+
+	earliestWindow := aw.entries.Front().Value.(*keyed.KeyedWindow)
+	recentWindow := aw.entries.Back().Value.(*keyed.KeyedWindow)
+
+	// late arrival
+	if earliestWindow.Start.Equal(kw.End) || earliestWindow.Start.After(kw.End) {
```
whynowy, line 55: `if kw.End.Before(earliestWindow.Start) {`

pkg/window/strategy/fixed/lifecycle.go
```diff
@@ -0,0 +1,130 @@
...
+		aw.entries.PushFront(kw)
+	} else if recentWindow.End.Equal(kw.Start) || recentWindow.End.Before(kw.Start) {
```
whynowy, line 57: User `after`?

pkg/window/strategy/sliding/sliding.go
```diff
@@ -0,0 +1,55 @@
...
+		PeriodInSeconds: periodInSeconds,
+		sliceCount:      sliceCount,
+	}
+}
+
+// AssignWindow assigns a window for the given eventTime.
+func (s *Sliding) AssignWindow(eventTime time.Time) []*window.IntervalWindow {
+	slidingWindows := make([]*window.IntervalWindow, s.sliceCount)
+
+	tmp := eventTime.Truncate(s.Length)
```
whynowy, line 45: I don't think this is right, I'll give more information later.

Since we are not doing sliding window at this time, we can revisit in the future.

whynowy:
IMG_4019 copy

The reason why said the sliding implementation is wrong, check the `4min_3min` example attached.

whynowy:
So for sliding window, it should be:

t = time.Truncate(interval)

If length >= interval:
- From `t` to `t + interval - (length - interval)` -> 1 window;
- Otherwise -> 2 windows.

Else (length < interval):
- From `t` to `t + length` -> 1 window;
- Otherwise -> 0 window.

whynowy:
So for sliding window, it should be:

t = time.Truncate(interval)

N = int(length/interval)

If length >= interval:
- From `t` to `t + interval - (length - interval)` -> N window;
- Otherwise -> N + 1 windows.

Else (length < interval):
- From `t` to `t + length` -> 1 window;
- Otherwise -> 0 window.

pkg/window/keyed/keyed.go
ashwinidulams: we need access to this from the readloop which needs to compute partition id to identify the right pbq to write to.

ashwinidulams: this may change as readloop is developed

pkg/window/strategy/fixed/lifecycle.go
ashwinidulams: `kw.Start.After(recentWindow.End)`?  if so, its the same. are you suggesting that writing like this may be more readable?

pkg/window/strategy/fixed/lifecycle.go
whynowy: I meant, `Equal || After` is `!Before`

pkg/window/strategy/fixed/lifecycle.go
whynowy: Similarly here.

pkg/window/strategy/sliding/sliding.go
ashwinidulams: yes sliding window may need some changes even from lifecycle standpoint. we can revisit again after finishing the entire loop with tumbling initially.

pkg/window/keyed/keyed.go
```diff
@@ -0,0 +1,54 @@
...
+	return PartitionId(fmt.Sprintf("%s-%v-%v", key, window.Start.Unix(), window.End.Unix()))
```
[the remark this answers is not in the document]
ashwinidulams: fair.

pkg/window/strategy/fixed/lifecycle.go
ashwinidulams: makes sense.

pkg/window/strategy/fixed/lifecycle.go
ashwinidulams: right

ashwinidulams: made the change.

pkg/window/strategy/fixed/lifecycle.go
ashwinidulams: made the change.

pkg/window/keyed/keyed.go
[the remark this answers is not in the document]
ashwinidulams: done.

pkg/window/windower.go
ashwinidulams: i am going to leave this as is for now. Will tackle this when we go about implementing sliding windows. 
may be bounded and unbounded windows make more sense.

whynowy (approved):
LGTM

vigith:
since the sliding window is wrong, let's delete that part of the code and add it later the right way. we are not interested in sliding window as of now.

ashwinidulams:
> since the sliding window is wrong, let's delete that part of the code and add it later the right way. we are not interested in sliding window as of now.

removed.

PR #166, opened 2022-09-08
chore: add mkdocs check

whynowy:
1. Add mkdocs check;
2. Upgrade go version
3. Deprecate `ioutil`


pkg/sources/generator/tickgen.go
```diff
@@ -114,7 +114,9 @@ func WithReadTimeOut(timeout time.Duration) Option {
 
 // NewMemGen fuction creates an instance of generator.
 // ctx  - context passed by the cmd/start.go a new context with cancel
-//        is created for use by this vertex.
+//
```
vigith, line 117: is this intended? i think that comment was part of prev line

pkg/watermark/processor/entity.go
```diff
@@ -57,7 +57,8 @@ var _ ProcessorEntitier = (*ProcessorEntity)(nil)
 
 // _defaultKeySeparator is the key separate when we have shared OT buckets.
 // NOTE: we can only use `_` as the separator, Jetstream will not let any other special character.
-//       Perhaps we can encode the key using base64, but it will have a performance hit.
+//
```
vigith, line 60: I think this is also unintended. It does look like a block code comment.

vigith (approved):

pkg/sources/generator/tickgen.go
whynowy: It's go 1.19 format.

pkg/watermark/processor/entity.go
whynowy: Same here, 1.19 format.

PR #169, opened 2022-09-13
feat: Shuffle implementation

yhl25:
hash-based shuffle implementation with fnv


pkg/shuffle/shuffle.go
```diff
@@ -0,0 +1,43 @@
...
+func (s *Shuffle) ShuffleMessages(messages []*isb.Message) map[string][]*isb.Message {
+	buffersCount := uint64(len(s.bufferIdentifiers))
+
+	// hash of the message key returns a unique hashValue
+	// mod of hashValue will decide which isb it will belong
+	hashMap := make(map[string][]*isb.Message)
+	for _, message := range messages {
+		hashValue := s.generateHash(message.Key)
+		hashValue = hashValue % buffersCount
+		hashMap[s.bufferIdentifiers[hashValue]] = append(hashMap[s.bufferIdentifiers[hashValue]], message)
```
ashwinidulams, line 34: if you already know the size of the array (equal to the number of messages here), you could avoid append.

pkg/shuffle/shuffle.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+	return hashMap
+}
+
+func (s *Shuffle) generateHash(key string) uint64 {
+	s.hash.Reset()
```
vigith, line 40: why do we need to Reset every time we generate a hash?

pkg/shuffle/shuffle.go
```diff
@@ -0,0 +1,43 @@
...
+// ShuffleMessages accepts list of isb messages and returns the mapping of isb to messages
+func (s *Shuffle) ShuffleMessages(messages []*isb.Message) map[string][]*isb.Message {
+	buffersCount := uint64(len(s.bufferIdentifiers))
+
+	// hash of the message key returns a unique hashValue
+	// mod of hashValue will decide which isb it will belong
+	hashMap := make(map[string][]*isb.Message)
+	for _, message := range messages {
+		hashValue := s.generateHash(message.Key)
+		hashValue = hashValue % buffersCount
```
ashwinidulams, line 33: should we consider generating hash ranges (some pre-fixed multiple of no of isbs) and distributing the hash range? wondering if that will ensure more even distribution.

pkg/shuffle/shuffle.go
yhl25: size is not known, it's not equal to the number of messages here..

pkg/shuffle/shuffle.go
yhl25: probability would still remain the same, that is 1 / (number of isbs), I don't think it will make any difference

pkg/shuffle/shuffle.go
yhl25: hash stores the state, since we need to discard the state everytime we need to reset the hash

pkg/shuffle/shuffle.go
ashwinidulams: yes. missed that.

pkg/shuffle/shuffle.go
```diff
@@ -0,0 +1,43 @@
...
+func NewShuffle(bufferIdentifiers []string) *Shuffle {
+	return &Shuffle{
+		bufferIdentifiers: bufferIdentifiers,
+		hash:              fnv.New64(),
+	}
+}
+
+// ShuffleMessages accepts list of isb messages and returns the mapping of isb to messages
+func (s *Shuffle) ShuffleMessages(messages []*isb.Message) map[string][]*isb.Message {
+	buffersCount := uint64(len(s.bufferIdentifiers))
```
vigith, line 26: can't we store that into the `Shuffle` struct during `NewXXX`, so you don't have to call `len` for every `ShuffleMessages` call?

yhl25: true, changed

pkg/shuffle/shuffle.go
```diff
@@ -0,0 +1,44 @@
...
+func NewShuffle(bufferIdentifiers []string) *Shuffle {
+	return &Shuffle{
+		bufferIdentifiers: bufferIdentifiers,
+		buffersCount:      len(bufferIdentifiers),
+		hash:              fnv.New64(),
+	}
+}
+
+// ShuffleMessages accepts list of isb messages and returns the mapping of isb to messages
+func (s *Shuffle) ShuffleMessages(messages []*isb.Message) map[string][]*isb.Message {
```
yhl25, line 27: reason I am using  `map[bufferId][]*isb.Message ` as a return type for shuffle is, writer can do parallel writes to isbs

ashwinidulams (approved):
LGTM

vigith (approved):

whynowy (approved):

PR #177, opened 2022-09-19
fix: numaflow-go udf example & docs

chromevoid:
**Fix numaflow golang udf example and the `Golang` link**
image

**Fix docs**
Before:
image
image
Now:
image

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -5,15 +5,13 @@ Install Numaflow and run a couple of example pipelines.
 ## Prerequisites
 
 A Kubernetes cluster is needed to try out Numaflow. A simple way to create a local cluster is using Docker Desktop.
-<ul>
-  <li><a href="https://www.docker.com/">Docker</a></li>
-  <li><a href="https://www.docker.com/blog/how-kubernetes-works-under-the-hood-with-docker-desktop/">Docker Desktop</a></li>
-</ul>
+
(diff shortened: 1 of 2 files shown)


docs/quick-start.md
```diff
@@ -5,11 +5,15 @@ Install Numaflow and run a couple of example pipelines.
 ## Prerequisites
 
 A Kubernetes cluster is needed to try out Numaflow. A simple way to create a local cluster is using Docker Desktop.
-* https://www.docker.com/
-* https://www.docker.com/blog/how-kubernetes-works-under-the-hood-with-docker-desktop/
+<ul>
+  <li><a href="https://www.docker.com/">Docker</a></li>
```
vigith, line 9: won't `[text](url)` style work?, something like
* [foo](foo.example.com)
* [bar](bar.example.com)

chromevoid: unfortunately no..

image
image

whynowy: Should not use HTML tags in the markdown.

chromevoid: Should we abandon the list style?

chromevoid: discussed offline..
thank you @whynowy for the tips

need one more blank line to render the markdown style list

whynowy (approved):

PR #193, opened 2022-09-29
fix: update udf fetchWatermark and publishWatermark initial values

chromevoid:
fix: update udf fetchWatermark and publishWatermark initial values

tested using redis isbsvc and simple pipeline with
```yaml
  watermark:
    disabled: true
```
image
image


whynowy (approved):
LGTM

PR #194, opened 2022-09-30
chore: update quickstart

KeranYang:
Update quickstart instructions.

* Move numaflow UI creation from "A Simple Pipeline" section to "Installation" section, since the UI is shared by both simple and advanced pipeline.
* Add a "Clean up" section to close the port forwarding.

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -54,7 +54,6 @@ kubectl logs -f simple-pipeline-out-0-xxxx main
 2022/08/25 23:59:39 (out) {"Data":"jk4nN/a7Dhc=","Createdts":1661471978707963534}
 
 ```
-
 View the UI for the simple pipeline at https://localhost:8443/
 ![Numaflow UI](assets/numaflow-ui-simple-pipeline.png)
 
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -22,6 +22,13 @@ kubectl apply -n numaflow-system -f https://raw.githubusercontent.com/numaproj/n
 kubectl apply -f https://raw.githubusercontent.com/numaproj/numaflow/stable/examples/0-isbsvc-jetstream.yaml
 ```
 
+Port forward the UI to https://localhost:8443/
+```shell
+kubectl -n numaflow-system port-forward deployment/numaflow-server 8443:8443
+```
+
+Keep the current terminal running, open a new terminal tab to run the following example pipelines.
+
 ## A Simple Pipeline
 
 Create a `simple pipeline`, which contains a source vertex to generate messages, a processing vertex that echos the messages, and a sink vertex that logs the messages.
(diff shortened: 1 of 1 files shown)


docs/quick-start.md
```diff
@@ -22,6 +22,13 @@ kubectl apply -n numaflow-system -f https://raw.githubusercontent.com/numaproj/n
 kubectl apply -f https://raw.githubusercontent.com/numaproj/numaflow/stable/examples/0-isbsvc-jetstream.yaml
 ```
 
+Port forward the UI to https://localhost:8443/
```
whynowy, line 25: This is not part of installation, should not be here.

docs/quick-start.md
```diff
@@ -106,7 +113,7 @@ kubectl logs -f even-odd-odd-sink-0-a6p0n main
 2022/09/07 22:30:19 (odd-sink) 103
 ```
 
-View the UI for the advanced pipleline at https://localhost:8443/
+View the UI for the advanced pipeline at https://localhost:8443/
```
whynowy, line 116: 👍

docs/quick-start.md
```diff
@@ -116,6 +123,13 @@ The pipeline can be deleted by
 kubectl delete -f https://raw.githubusercontent.com/numaproj/numaflow/stable/test/e2e/testdata/even-odd.yaml
 ```
 
+## Clean up
```
whynowy, line 126: Don't think this section needs to be mentioned particularly. BTW - regular `Ctrl+c` can terminate it.

docs/quick-start.md
```diff
@@ -89,7 +96,7 @@ isbsvc-default-js-2                3/3     Running   0          10m
 # Port-forward the HTTP endpoint so we can post data from the laptop
 kubectl port-forward even-odd-in-0-xxxx 8444:8443
 
-# Post data to the HTTP endpoint
+# Open a new terminal tab to post data to the HTTP endpoint
```
whynowy, line 99: Feel like a simple `Post xxxx` makes more sense, more detail like `open a terminal` might confuse ppl, as lots of ppl do not use it, and even have no idea about it.

KeranYang: Make sense, will address.

docs/quick-start.md
KeranYang: How about adding a ## Port Forwarding section?

docs/quick-start.md
KeranYang: Agree that a regular `Ctrl+c` is better. I was using `Ctrl+z`, which didn't close the port hence made it occupied, causing "Address in use" error the very next time I played with this tutorial. I guess it should be obvious to use `Ctrl+c`. Will remove.

docs/quick-start.md
whynowy: The expected user of Numaflow is someone who knows Kubernetes, and this is a very common operation in k8s world, don't think this needs to be in a section.

KeranYang: Makes sense. I moved `kubectl -n numaflow-system port-forward deployment/numaflow-server 8443:8443` command out of `A Simple Pipeline` section because it's shared by both pipelines so it could be executed before any pipeline creation. Do you think I should keep it in the `A Simple Pipeline` then?

vigith: The doc currently does not have sections for each command, it is a single blob with commands and their outputs. Should we split it out to make it more readable?

KeranYang: I made these changes assuming readers are new to k8s like myself. Since numaflow is targeting users who knows k8s, I agree that we can keep this quickstart doc as it is. I will update this PR to just fix typo.

KeranYang:
I am cancelling this PR as most of the changes I proposed are not needed as per our discussion. I will post a separate PR to fix the minor typo.

PR #202, opened 2022-10-07
chore: add read and written bytes counter

whynowy:

Explain what this PR does.

### DIFF
--- a/pkg/isb/forward/forward.go
+++ b/pkg/isb/forward/forward.go
@@ -179,6 +179,7 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
 	// let's track only the last element's watermark
 	processorWM := isdf.fetchWatermark.GetWatermark(readMessages[len(readMessages)-1].ReadOffset)
 	for _, m := range readMessages {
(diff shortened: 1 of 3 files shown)


pkg/isb/forward/forward.go
```diff
@@ -318,6 +319,10 @@ func (isdf *InterStepDataForward) writeToBuffers(ctx context.Context, messageToS
 
 // writeToBuffer forwards an array of messages to a single buffer and is a blocking call or until shutdown has been initiated.
 func (isdf *InterStepDataForward) writeToBuffer(ctx context.Context, toBuffer isb.BufferWriter, messages []isb.Message) (writeOffsets []isb.Offset, err error) {
+	var totalBytes float64
+	for _, m := range messages {
```
vigith, line 323: is there a way we can avoid this extra loop of O(N)? Can we collect this info while writing to buffer? We can do summation while querying from prometheus.

whynowy: Based on our implementation, it's going to be very tricky, because we do batch writing and need to consider failure case.

vigith (approved):

PR #213, opened 2022-10-12
chore: reference the Payload directly when writing to Redis ISB to avoid extra marshaling

### LINKED ISSUE #204: replace JSON format of isb.message with a binary format
Today we have implemented `MarshalBinary` and `UnmarshalBinary` to encode and decode using JSON as the intermediary format. We should try out using binary formats. 

The code change itself is minimal, we should try `encoding/gob` first.

- [ ] Run the performance test before the change (measure CPU usage, TPS, etc.)
- [ ] Make the change
- [ ] Compare the performance

xdevxy:
Closes: #204 

This PR reference the `Payload` directly in `isb.message.body` to avoid extra marshaling in Redis ISB implementation. Note that we need to revisit it when `Body` structure changes for [this](https://github.com/numaproj/numaflow/pull/213#issuecomment-1278016520) reason. 

`Testing`: Unit tests.

### DIFF
--- a/pkg/isb/stores/redis/read.go
+++ b/pkg/isb/stores/redis/read.go
@@ -325,11 +325,7 @@ func getHeaderAndBody(field string, value interface{}) (msg isb.Message, err err
 		return msg, fmt.Errorf("header unmarshal error %w", err)
 	}
 
-	err = msg.Body.UnmarshalBinary([]byte(value.(string)))
-	if err != nil {
-		return msg, fmt.Errorf("body unmarshal error (%v) %w", []byte(value.(string)), err)
-	}
-
+	msg.Body.Payload = []byte(value.(string))
 	return msg, nil
 }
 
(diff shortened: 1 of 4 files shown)


vigith:
I was thinking we will keep the structure as is, but just reference the Payload directly when writing to Redis ISB.

xdevxy:
> I was thinking we will keep the structure as is, but just reference the Payload directly when writing to Redis ISB.

Talked to @vigith offline, he suggests we may add more fields to `Body` structure later so better to keep it. Theoretically, keeping the `Body` structure and reference the Payload directly when writing to Redis ISB can open the window to miss properly processing new fields add to `Body` later if we don't manually track what has been used in the `Body` structure in each implementation of ISB. However, this is trivial to track ATM, so following his suggestion in this PR.

vigith (approved):

PR #215, opened 2022-10-13
perf: avoid busy loop

vigith:


pkg/reduce/readloop/ordered.go
```diff
@@ -67,52 +69,76 @@ func (op *orderedProcessor) process(ctx context.Context,
...
+					} else {
+						break
+					}
 				}
+				op.Lock()
+				rm := currElement
+				currElement = currElement.Next()
+				op.taskQueue.Remove(rm)
+				op.Unlock()
+				break
```
vigith, line 138: do we need this `break`?

ashwinidulams (approved):

PR #216, opened 2022-10-13
feat: in-memory watermark store for better testing

chromevoid:
watermark unit tests with in mem store


pkg/watermark/store/inmem/kv_watch.go
```diff
@@ -0,0 +1,100 @@
...
+	return updates
+}
+
+// GetKVName returns the KV store (bucket) name.
+func (k *inMemWatch) GetKVName() string {
+	return k.bucketName
+}
+
+// Close closes the connection.
+func (k *inMemWatch) Close() {
```
vigith, line 99: Can you inspect how `Close` is invoked? We do have `Close` in the interface. Can you make sure we are using `Close`? Without correctly closing, we might run into connection leaks and memory leaks associated with it.

chromevoid: 👌

chromevoid: opened a new issue https://github.com/numaproj/numaflow/issues/224

vigith (approved):

pkg/watermark/store/inmem/kv_watch.go
```diff
@@ -0,0 +1,100 @@
+package inmem
+
+import (
+	"context"
+	"math/rand"
```
whynowy, line 5: No `math/rand`

pkg/watermark/store/inmem/kv_watch.go
```diff
@@ -0,0 +1,100 @@
...
+		kvHistory:    []store.WatermarkKVEntry{},
+		updatesChMap: make(map[string]chan store.WatermarkKVEntry),
+		log:          logging.FromContext(ctx).With("pipeline", pipelineName).With("bucketName", bucketName),
+	}
+	return k, nil
+}
+
+var letters = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
+
+func randSeq(n int) string {
```
whynowy, line 42: use https://github.com/numaproj/numaflow/blob/main/pkg/shared/util/string.go#L10

PR #218, opened 2022-10-14
fix: update default watermark to -1. Fixes #133

### LINKED ISSUE #133: Watermark shows -62135596800 when a fresh pipeline started without any traffic
**Describe the bug**

Screen Shot 2022-08-08 at 5 53 25 PM

KeranYang:
Fixes: #133

This change is to fix #133 where watermark shows -62135596800 when a fresh pipeline started without any traffic. The daemon service shows -62135596800 because when the OT store is empty, the watermark fetcher returns time.Time{} as default watermark, which transforms to epoch as -62135596800. In this change I am proposing -1 as default watermark to indicate there is no valid watermark yet.

Before:
image

After:
Screen Shot 2022-10-14 at 2 09 16 PM


pkg/watermark/fetch/processor_to_fetch.go
```diff
@@ -58,9 +58,8 @@ func NewProcessorToFetch(ctx context.Context, processor processor.ProcessorEntit
 		otWatcher:      watcher,
 		log:            logging.FromContext(ctx),
 	}
-	if watcher != nil {
```
vigith, line 61: any particular reason why you did this? (ie, moved to `startTimeLineWatcher`)?

pkg/watermark/fetch/source_fetcher.go
```diff
@@ -32,6 +32,7 @@ func NewSourceFetcher(ctx context.Context, sourceBufferName string, processorMan
 
 // GetHeadWatermark returns the latest watermark of all the processors.
 func (e *sourceFetcher) GetHeadWatermark() processor.Watermark {
+	e.log.Infof("Fetching head watermark from source buffer %s...", e.sourceBufferName)
```
vigith, line 35: log.debug?

pkg/watermark/fetch/source_fetcher.go
```diff
@@ -42,8 +43,10 @@ func (e *sourceFetcher) GetHeadWatermark() processor.Watermark {
 		}
 	}
 	if epoch == math.MinInt64 {
-		return processor.Watermark(time.Time{})
+		e.log.Infof("Didn't find any existing head watermarks across all processors, using default watermark %d", int64(-1))
```
vigith, line 46: log.debug?

pkg/watermark/fetch/source_fetcher.go
```diff
@@ -42,8 +43,10 @@ func (e *sourceFetcher) GetHeadWatermark() processor.Watermark {
...
+		return processor.Watermark(time.Unix(-1, 0))
 	}
+	e.log.Infof("Found head watermark epoch %d", epoch)
```
vigith, line 49: log.debug?

pkg/watermark/generic/progressor.go
```diff
@@ -1,13 +0,0 @@
-package generic
-
-import (
-	"github.com/numaproj/numaflow/pkg/watermark/fetch"
-	"github.com/numaproj/numaflow/pkg/watermark/publish"
-)
-
-// Progressor interface defines how the watermark can be progressed.
-// FIXME: delete this
```
vigith, line 9: 👍

pkg/watermark/processor/entity.go
```diff
@@ -21,6 +21,12 @@ func (w Watermark) String() string {
 	return t.Format(time.RFC3339)
 }
 
+func (w Watermark) Unix() int64 {
+	var location, _ = time.LoadLocation("UTC")
```
vigith, line 25: should we move this as a global var? else we will have to keep loading all the time. perhaps do via `func init() {....}()` and `panic` if we cannot load the location

vigith: Also, what will happen if we do not do LoadLocation? Are we seeing failures? All of our timestamps should be in UTC. If this is mostly for tests, then let's do LoadLocation only in tests.

pkg/watermark/fetch/source_fetcher.go
whynowy, line 49: Use `Debugf`

pkg/watermark/fetch/source_fetcher.go
whynowy, line 35: Use `Debugf`

pkg/watermark/fetch/source_fetcher.go
whynowy, line 46: `Debugf`

pkg/watermark/fetch/processor_to_fetch.go
```diff
@@ -98,6 +97,9 @@ func (p *ProcessorToFetch) IsDeleted() bool {
 }
 
 func (p *ProcessorToFetch) startTimeLineWatcher() {
+	if p.otWatcher == nil {
```
whynowy, line 100: Don't add the check, let it panic instead in case this really happens (even though it won't happen), otherwise it will help hide the issue.

pkg/watermark/fetch/edge_fetcher.go
```diff
@@ -33,10 +33,11 @@ func NewEdgeFetcher(ctx context.Context, edgeName string, processorManager *Proc
 	}
 }
 
-// GetHeadWatermark returns the watermark using the HeadOffset (latest offset). This
+// GetHeadWatermark returns the watermark using the HeadOffset (the latest offset among all processors). This
 // can be used in showing the watermark progression for a vertex when not consuming the messages
 // directly (eg. UX, tests,)
 func (e *edgeFetcher) GetHeadWatermark() processor.Watermark {
+	e.log.Infof("Fetching head watermark for edge %s...", e.edgeName)
```
whynowy, line 40: `Debugf`

pkg/watermark/fetch/edge_fetcher.go
```diff
@@ -56,9 +57,11 @@ func (e *edgeFetcher) GetHeadWatermark() processor.Watermark {
 	}
 	e.log.Debugf("GetHeadWatermark: %s", debugString.String())
 	if epoch == math.MaxInt64 {
-		return processor.Watermark(time.Time{})
+		e.log.Infof("Didn't find any existing head watermarks across all processors, using default watermark %d...", int64(-1))
```
whynowy, line 60: `Debugf`

pkg/watermark/fetch/edge_fetcher.go
```diff
@@ -56,9 +57,11 @@ func (e *edgeFetcher) GetHeadWatermark() processor.Watermark {
...
+		return processor.Watermark(time.Unix(-1, 0))
 	}
 
+	e.log.Infof("Found head watermark epoch %d.", epoch)
```
whynowy, line 64: `Debugf`

pkg/daemon/server/service/pipeline_watermark_query.go
```diff
@@ -118,7 +118,10 @@ func (ps *pipelineMetadataQuery) GetVertexWatermark(ctx context.Context, request
 		log.Errorf("watermark fetcher not available for vertex %s in the fetcher map", vertexName)
 		return nil, fmt.Errorf("watermark not available for given vertex, %s", vertexName)
 	}
-	vertexWatermark := time.Time(vertexFetcher.GetHeadWatermark()).Unix()
+
+	vertexWatermark := vertexFetcher.GetHeadWatermark().Unix()
+	log.Infof("Head watermark retrieved for vertex %s is %d.", vertexName, vertexWatermark)
```
whynowy, line 123: `Debugf`

pkg/watermark/processor/entity.go
whynowy, line 25: Do we need the location here? `time.Unix()` does not depend on the location associated with it.

pkg/watermark/fetch/processor_to_fetch.go
KeranYang: My thought behind this change is that, since `watcher` is a member of ProcessorToFetch p, so it should be p's responsibility to do NULL check on watcher. By moving this check to ProcessorToFetch, we ensure ProcessorToFetch never start the watcher if watcher is NULL and on the other hand, all future callers on startTimeLineWatcher don't need to worry about watcher being NULL.

pkg/daemon/server/service/pipeline_watermark_query.go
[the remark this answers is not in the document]
KeranYang: Thanks! I was using level INFO during local development because otherwise I couldn't see the logs. Do you know how to tell my local env to show DEGUB level logs? (set EnvVar NUMAFLOW_DEBUG?)

I set it to INFO level as I think it could be useful informations to show in PROD, which help us debugging at PROD stage. The tradeoff is that we might flood the PROD memory with a lot of logs. Do we have a standard within the team on when to use what log level?

pkg/watermark/fetch/processor_to_fetch.go
KeranYang: Makes sense. I am thinking I can keep the check but instead of `return`, explicitly panic. Please see https://github.com/numaproj/numaflow/pull/218#discussion_r996082512 on why I moved the null check from outside this method to inside.

pkg/daemon/server/service/pipeline_watermark_query.go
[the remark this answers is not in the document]
vigith: We do not log repeated events over and over as Info as it spoils the log file. Makes debugging even tougher. Usually, `Info` is once in a while, mostly while bootup and shutdown. `Error` can be logged often, but in some cases, we group and summarize if the errors that occur are the same (e.g., buffer-full errors).

while debugging you can set `os.SetEnv("NUMAFLOW_DEBUG", "true")` in your test function.

pkg/watermark/processor/entity.go
KeranYang: Thanks, I will remove.
> The Time.Unix() function in Go language is used to yield “t” as a Unix time that is the number of seconds passed from January 1, 1970, in UTC and the output here doesn’t rely upon the location connected with t.

KeranYang: Agree. If we don't load location, the default location will be UTC and the behaviour will be the same(no failures). I will remove loading location here.

pkg/daemon/server/service/pipeline_watermark_query.go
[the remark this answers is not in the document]
KeranYang: Cool got it! Thanks!

pkg/watermark/fetch/processor_to_fetch.go
KeranYang: I decide to change it back. It seems we do have a case where the watcher can be nil and it's OK (heartbeat watcher for a source vertex).
```
        // we do not care about heartbeat watcher if this is a source vertex
	if v.hbWatcher != nil {
		go v.startHeatBeatWatcher()
	}
```
For such case, if I move null check from outside `startHeatBeatWatcher` to inside and panic, the thread will panic instead of skipping `startHeatBeatWatcher`. This could be a backward-incompatible change.

vigith: that is a bug in our code, can you open an issue or PR for it? there should never be a nil watcher. We wrote no-op watcher and storer for that purpose.

vigith (commented):
I felt we are logging too much, even though it is. log-level debug. What do you think?

pkg/watermark/fetch/edge_fetcher.go
```diff
@@ -33,10 +33,11 @@ func NewEdgeFetcher(ctx context.Context, edgeName string, processorManager *Proc
 	}
 }
 
-// GetHeadWatermark returns the watermark using the HeadOffset (latest offset). This
+// GetHeadWatermark returns the watermark using the HeadOffset (the latest offset among all processors). This
 // can be used in showing the watermark progression for a vertex when not consuming the messages
 // directly (eg. UX, tests,)
 func (e *edgeFetcher) GetHeadWatermark() processor.Watermark {
+	e.log.Debugf("Fetching head watermark for edge %s...", e.edgeName)
```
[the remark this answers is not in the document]
KeranYang: Personally I would like to add as much meaningful logs as possible, making it easier to troubleshoot issues. Not sure if having a lot of logs is a bad thing in k8s's world or not...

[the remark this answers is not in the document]
vigith: I agree; unfortunately, we do not have multiple levels of logging levels available (e.g., we do not have a verbose level). Since this is a high throughput system, we will be seeing a ton lot of logs just by moving the log-level to the next lowest level (i.e, Info to Debug). Also, the logs you have added are in the hot-code path, meaning they will be invoked for every message where the watermark operation is called. 

I am not proposing to lose important information but to use some tricks to make the logs more informational while reducing the number of log lines. E.g., there is a log entry to print the name of the edge, you could use [With](https://pkg.go.dev/go.uber.org/zap#Logger.With) to print out the edge for every log line.

pkg/watermark/fetch/processor_to_fetch.go
vigith: could you please create a new issue for this?

KeranYang: New issue created: #230

pkg/watermark/fetch/edge_fetcher.go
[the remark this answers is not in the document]
KeranYang: That's a good point. I didn't realize I was putting logs on a very hot code path. Will remove. Thanks!

vigith (approved):

whynowy:
@vigith @KeranYang - how about using `0`?

KeranYang:
> @vigith @KeranYang - how about using `0`?

I don't have a preference as long as we keep it consistent across all vertices.

Krithika3:
I think -1 is better than 0.

whynowy:
Be aware `0` means `1969-12-31 16:00:00 -0800 PST`, `-1` means `1969-12-31 15:59:59 -0800 PST`. @KeranYang - do you see the time when you hover on the number?

whynowy:
Anyways, we can get this in, and let's see how it looks like, and revise if it's needed.

PR #222, opened 2022-10-14
fix: Show human-readable timestamps in Daemon Service log

akhilchoudhary2k:
Signed-off-by: Akhil <akhil.choudhary2000@gmail.com>

Will show Daemon Service log shows timestamp in human-readable timestamps instead of epoch values.

### DIFF
--- a/pkg/shared/logging/log.go
+++ b/pkg/shared/logging/log.go
@@ -5,6 +5,7 @@ import (
 	"os"
 
 	zap "go.uber.org/zap"
+	zapcore "go.uber.org/zap/zapcore"
 )
 
 // NewLogger returns a new zap.SugaredLogger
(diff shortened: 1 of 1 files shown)


pkg/shared/logging/log.go
```diff
@@ -18,6 +19,7 @@ func NewLogger() *zap.SugaredLogger {
 	}
 	// Config customization goes here if any
 	config.OutputPaths = []string{"stdout"}
+	config.EncoderConfig.EncodeTime = zapcore.TimeEncoderOfLayout(time.RFC3339)
```
whynowy, line 22: As I commented in the issue, we have no plan to change this.

whynowy:
@akhilchoudhary2k - Thanks for having interest in working on this project, if you'd like, I can pick another task for you, let me know.

PR #228, opened 2022-10-16
feat: support disabling TLS and changing port for UX server

### LINKED ISSUE #227: Support http for numaflow UX server
# Summary

It looks like the server currently only supports https through a generated key pair.  It would be great to support http as well.  If supporting http, it would also be good to support configuring a different port number, since 8443 suggests https.

# Use Cases

I'm currently working with a setup that is using envoy sidecars to route traffic in/out of containers (specifically istio mtls service mesh), so allowing the server to accept unencrypted traffic would be sufficient/preferred for this use-case.

Alternatively I can look at configuring the envoy sidecar to use an https connection but skip cert validation when routing traffic into 8443 similar to what k8s is doing for liveness probes.

---

dseapy:
Closes #227

* Adds support for disabling TLS with `--insecure` (server will listen by default at 8080)
* Adds support for overriding the default port with `--port=8081` or `-p 8081`.

### DIFF
--- a/cmd/commands/server.go
+++ b/cmd/commands/server.go
@@ -6,12 +6,15 @@ import (
 )
 
 func NewServerCommand() *cobra.Command {
+	var insecure bool
+
 	command := &cobra.Command{
 		Use:   "server",
 		Short: "Start a Numaflow server",
 		Run: func(cmd *cobra.Command, args []string) {
-			svrcmd.Start()
+			svrcmd.Start(insecure)
 		},
 	}
+	command.Flags().BoolVar(&insecure, "insecure", false, "Whether to disable on TLS, defaults to false.")
 	return command
 }
--- a/cmd/commands/server.go
+++ b/cmd/commands/server.go
@@ -7,14 +7,19 @@ import (
 
 func NewServerCommand() *cobra.Command {
 	var insecure bool
+	var port int
 
 	command := &cobra.Command{
 		Use:   "server",
 		Short: "Start a Numaflow server",
 		Run: func(cmd *cobra.Command, args []string) {
-			svrcmd.Start(insecure)
+			if !cmd.Flags().Changed("port") && insecure {
+				port = 8080
+			}
+			svrcmd.Start(insecure, port)
 		},
 	}
 	command.Flags().BoolVar(&insecure, "insecure", false, "Whether to disable on TLS, defaults to false.")
+	command.Flags().IntVarP(&port, "port", "p", 8443, "Port to listen on, defaults to 8443 or 8080 if insecure is set")
 	return command
 }
--- a/cmd/commands/server.go
+++ b/cmd/commands/server.go
@@ -21,7 +21,7 @@ func NewServerCommand() *cobra.Command {
 			svrcmd.Start(insecure, port)
 		},
 	}
-	command.Flags().BoolVar(&insecure, "insecure", false, "Whether to disable on TLS, defaults to false.")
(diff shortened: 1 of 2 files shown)


dseapy:
One thing that might not be ideal is that if someone is disabling tls they would generally not want to be on `8443` as one might assume it's encrypted, which requires setting both `--port` and `--secure` flags.  Seems like changing `8443` to a port that is not known to be related to the protocol might be ideal, but didn't know if that change would be desired or not as it would be a breaking change.
Another possible alternative is defaulting to different ports based on protocol (i.e. `8080` for http and `8443` for https) outside of the cobra definition, not sure if that's a good idea.

cmd/commands/server.go
```diff
@@ -6,12 +6,19 @@ import (
 )
 
 func NewServerCommand() *cobra.Command {
+	var (
+		port   int
+		secure bool
```
whynowy, line 11: How about use `insecure` and default to `false`? This is more common.

whynowy:
> One thing that might not be ideal is that if someone is disabling tls they would generally not want to be on `8443` as one might assume it's encrypted, which requires setting both `--port` and `--secure` flags. Seems like changing `8443` to a port that is not known to be related to the protocol might be ideal, but didn't know if that change would be desired or not as it would be a breaking change. Another possible alternative is defaulting to different ports based on protocol (i.e. `8080` for http and `8443` for https) outside of the cobra definition, not sure if that's a good idea.

Let's do `8080` if TLS is disabled? @vigith

cmd/commands/server.go
dseapy: will do, yeah that seems more natural.  saw argo-workflows had `secure` so initially had gone with that approach.

vigith:
> Let's do `8080` if TLS is disabled? @vigith

Yes, let's do that. We still can support the port override, but let it be 8080 by default when `--insecure` is provided.

dseapy:
> > Let's do `8080` if TLS is disabled? @vigith
> 
> Yes, let's do that. We still can support the port override, but let it be 8080 by default when `--insecure` is provided.

pushed a commit to default to `8080` for http.  did remove the port override before seeing comment.  sounds like from discussion in slack it might be ok without it, if ya'll decide it's preferred to add port override into the PR just let me know & can implement.

whynowy:
> > > Let's do `8080` if TLS is disabled? @vigith
> > 
> > 
> > Yes, let's do that. We still can support the port override, but let it be 8080 by default when `--insecure` is provided.
> 
> pushed a commit to default to `8080` for http. did remove the port override before seeing comment. sounds like from discussion in slack it might be ok without it, if ya'll decide it's preferred to add port override into the PR just let me know & can implement.

Port override is still needed.

cmd/commands/server.go
```diff
@@ -6,12 +6,22 @@ import (
...
 		Short: "Start a Numaflow server",
 		Run: func(cmd *cobra.Command, args []string) {
-			svrcmd.Start()
+			if !cmd.Flags().Changed("port") && insecure {
+				port = 8080
+			}
+			svrcmd.Start(insecure, port)
 		},
 	}
+	command.Flags().BoolVar(&insecure, "insecure", false, "Whether to disable on TLS, defaults to false.")
```
vigith, line 24:
```suggestion
	command.Flags().BoolVar(&insecure, "insecure", false, "Whether to disable TLS, defaults to false.")
```

vigith (commented):
a minor comment

Issue #230, opened 2022-10-17
Fix heartbeat and offset timeline watermark store watcher NULL check


KeranYang:
**Describe the bug**
When creating a new ProcessorManager, right after the creation, we start a heartbeat watcher. When newing a ProcessorToFetch, we a timeline watcher.

With in the code, the implementation does a NULL check before invoking watcher as follow.
```
if watcher != nil {
		go p.startTimeLineWatcher()
	}
```

```
if v.hbWatcher != nil {
		go v.startHeatBeatWatcher()
	}
```
Vigith mentioned in [here](https://github.com/numaproj/numaflow/pull/218#discussion_r996174518) that there should NEVER be a nil watcher. Having such NULL check could potentially hide the failure scenario where we do have a nil watcher during runtime. We should remove the NULL check and let it panic if nil, to surface the unexpected failure scenario.

**Expected behavior**
The system panics when having a nil watermark store watcher object.

**Screenshots**

**Environment (please complete the following information):**

- Kubernetes: [e.g. v1.18.6]
- Numaflow: [e.g. v0.5.1]

**Additional context**

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

Issue #224, opened 2022-10-15
Fix Watermark store and watcher Close() routine 


chromevoid:
Right now the Close() function in watermark store and in watermark watcher are not used anywhere, so, when we are using jetstream store for watermark, the nats connection keeps opening. We need to revisit how to properly shutdown the watermark store and watermark watcher.

PR #238, opened 2022-10-18
fix: Watermark close fix and removed the nil check

jy4096:
Fixes: #224
Fixes: #230

### DIFF
--- a/api/json-schema/schema.json
+++ b/api/json-schema/schema.json
@@ -1440,8 +1440,7 @@
         }
       },
       "required": [
-        "replicas",
-        "availableReplicas"
+        "replicas"
       ],
       "type": "object"
     },
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -1391,8 +1391,7 @@
       "description": "StatefulSetStatus represents the current state of a StatefulSet.",
       "type": "object",
       "required": [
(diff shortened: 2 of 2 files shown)


pkg/watermark/generic/jetstream/generic.go
```diff
@@ -50,6 +50,7 @@ func BuildWatermarkProgressors(ctx context.Context, vertexInstance *v1alpha1.Ver
 
 	var fetchWatermark fetch.Fetcher
 	if fromBuffer.Type == v1alpha1.SourceBuffer {
+		// TODO: does it use a non-nil hbWatch?
```
chromevoid, line 53: Meaning...
is this really executed for source vertex?
https://github.com/numaproj/numaflow/blob/e953699e321baae2c46af9e7b2bce56b5ba935ce/pkg/watermark/fetch/processor_manager.go#L55-L57

vigith: also #230

vigith: can you check the source's code and see whether we need this check?

chromevoid: 👀

whynowy (approved):
LGTM

vigith (approved):

Issue #134, opened 2022-08-09
Watermark displays incorrectly


whynowy:
**Describe the bug**

With a simple 3-vertex pipeline with `generator` source, start it with only the source with replica=1, and the other vertices with 0 replica. It is expected only the source vertex replica to show timestamp moving, however from the screenshot below, the 2nd vertex shows the same watermark, which is wrong.

<img width="1215" alt="Screen Shot 2022-08-08 at 5 58 59 PM" src="https://user-images.githubusercontent.com/2832028/183540855-ad388a59-3107-45fe-9018-8eb783288e91.png">

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**

- Kubernetes: [e.g. v1.18.6]
- Numaflow: [e.g. v0.5.1]

**Additional context**

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

vigith:
is this is caching issue on the UI? or is our backend returning number for a vertex that has been scaled down?

KeranYang:
As new data comes in, the **input** vertex populates the offset timeline store(otStore) to save most recent timestamps for each of the offsets. With the graph above, assuming we never had any active PODs for **p1** and **output** vertices, the system is in a state where otStore populated by **input** is growing and otStores to be populated by **p1** and **output** are empty.

The deamon service uses fetchers to grab vertex's level watermarks and shows them at front-end. In our current implementation, watermark fetcher for vertex Vn gets watermark from the otStore that is populated by vertex Vn-1. Therefore, the watermark showing on vertex Vn, is the most recent timestamp that ALL data before/at that timestamp has been **processed by Vn-1** and **ready for Vn to process**. That's why although the 2nd vertex has 0 pod and never had processed any data, it still has a moving watermark.

I had a discussion with @whynowy and we think that watermark for Vn should be the timestamp before which all data has already been **processed** by Vn, instead of being **ready to be processed** by Vn.

The change could be to update the daemon service to fetch watermarks using publishers instead of fetchers. Investigating...

PR #245, opened 2022-10-20
fix: update vertex watermark fetching logic. Fixes: #134

KeranYang:
Fixes: #134 

This change updates how we retrieve vertex level watermark, from fetching `from the input edge` to `from output edges`.

Before this change, we retrieve by getting the head watermark from the in-coming edge of the vertex. The problem of such approach is that when we have vertex N with 0 pods, if vertex N-1 is processing data, we end up showing moving watermarks for vertex N, which could be misleading.

With this change, we now show the vertex watermark as the timestamp before which all data **has already been processed** by Vn, instead of **being ready to be processed** by Vn.

Testing:

1. Tested with a simple pipeline.

Before change:
before

After change:
Screen Shot 2022-10-19 at 1 29 43 PM

Corner case:
When a vertex scales down to 0 pods, the watermark will show -1 since there is no active progressors. This has been communicated with Vigith and Derek and we agree it's ok to display -1.

corner-case

2. Tested with conditional forwarding and verified the vertex before branching out will display latest watermark among out edges.

Screen Shot 2022-10-21 at 4 33 35 PM

### DIFF
--- a/pkg/isb/forward/forward.go
+++ b/pkg/isb/forward/forward.go
@@ -251,13 +251,14 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
 
 	// forward the highest watermark to all the edges to avoid idle edge problem
 	// TODO: sort and get the highest value
-	for _, offsets := range writeOffsets {
-		for _, publisher := range isdf.publishWatermark {
+	for bufferName, offsets := range writeOffsets {
+		if publisher, ok := isdf.publishWatermark[bufferName]; ok {
 			if isdf.opts.vertexType == dfv1.VertexTypeSource || isdf.opts.vertexType == dfv1.VertexTypeUDF {
 				if len(offsets) > 0 {
 					publisher.PublishWatermark(processorWM, offsets[len(offsets)-1])
 				}
 				// This (len(offsets) == 0) happens at conditional forwarding, there's no data written to the buffer
+				// TODO: Should also publish to those edges without writing (fall out of conditional forwarding)
 			} else { // For Sink vertex, and it does not care about the offset during watermark publishing
 				publisher.PublishWatermark(processorWM, nil)
 			}
--- a/pkg/isb/forward/forward.go
+++ b/pkg/isb/forward/forward.go
@@ -251,14 +251,13 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
 
 	// forward the highest watermark to all the edges to avoid idle edge problem
 	// TODO: sort and get the highest value
-	for bufferName, offsets := range writeOffsets {
-		if publisher, ok := isdf.publishWatermark[bufferName]; ok {
+	for _, offsets := range writeOffsets {
+		for _, publisher := range isdf.publishWatermark {
 			if isdf.opts.vertexType == dfv1.VertexTypeSource || isdf.opts.vertexType == dfv1.VertexTypeUDF {
 				if len(offsets) > 0 {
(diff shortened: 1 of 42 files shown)


pkg/isb/forward/forward.go
```diff
@@ -251,14 +251,13 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
 
 	// forward the highest watermark to all the edges to avoid idle edge problem
 	// TODO: sort and get the highest value
-	for bufferName, offsets := range writeOffsets {
-		if publisher, ok := isdf.publishWatermark[bufferName]; ok {
+	for _, offsets := range writeOffsets {
+		for _, publisher := range isdf.publishWatermark {
```
whynowy, line 255: Is this right?

pkg/watermark/generic/generic.go
```diff
@@ -7,14 +7,14 @@ import (
...
-// NewGenericEdgeFetch returns a Fetcher, where bufferName is the from buffer of the vertex that is currently processing.
-// fetcher is an interface for retrieving both the heartbeat and the offset watermark timeline (Vn-1 vertex).
+// NewGenericEdgeFetch returns a Fetcher, which fetches watermarks from bufferName by consulting the corresponding storeWatcher
 func NewGenericEdgeFetch(ctx context.Context, bufferName string, storeWatcher store.WatermarkStoreWatcher) fetch.Fetcher {
 	processorManager := fetch.NewProcessorManager(ctx, storeWatcher)
 	return fetch.NewEdgeFetcher(ctx, bufferName, processorManager)
 }
 
 // NewGenericSourceFetch returns Fetcher, where sourceBufferName is the source buffer of the source vertex.
+// TODO remove this - we no longer need a source fetcher since we now look at the out edge to determine watermark for a vertex.
```
whynowy, line 17: Isn't it being used in source vertex?

pkg/isb/forward/forward.go
KeranYang: Based on my observation, the `writeOffsets` as an edge-to-offset mapping, contains ALL the out edges and their corresponding written offsets.

In the map, some of the out edges can point to an empty offset because of falling off conditional forwarding.

Before this code change, if we wrote msgs to offset 8 for edge A and didn't write anything to edge B, we only publish offset 8 to edge A. After this change, offset 8 also gets published to edge B.

With this code change, for any written offset, we invoke ALL the publishers to publish the watermark so that ALL out edges share the same watermark.

Let me do some manual tests to double confirm.

pkg/watermark/generic/generic.go
KeranYang: You are right. I will remove this comment.

pkg/isb/forward/forward.go
vigith: Any reason why we should be writing to all the edges? We should not be publishing watermarks to edges for which no data is written.

KeranYang: Was trying to address https://github.com/numaproj/numaflow/blob/6870d2a4558107cf824e5ba91ff8c65125a43eed/pkg/isb/forward/forward.go#L261 , which also simplifies fetching vertex level watermark as we just need to look into one of the out edges instead of looping through all.

I need to learn more about the impact of publishing to ALL out edges. Let's sync up offline on it.

whynowy (approved):
LGTM for now.

Pending items:
1. Do not hard code to use JetStream
2. Single jetstream connection, and need to handle `close`

vigith (approved):

PR #249, opened 2022-10-21
feat: Support running UX server with namespace scope. Fixes #248

### LINKED ISSUE #248: Support namespaced UX Server
# Summary

Currently if using Role/RoleBinding there are errors in the UX Server when trying to list namespaces.  To support running the server without cluster-wide permissions, it would be good to add --namespaced and --managed-namespace flags similar to how the controller currently does.

# Use Cases

User installing numaflow may not have sufficient permissions on their cluster to do a cluster-wide install.  Alternatively one may want to run multiple instances of the server restricted to their namespaces.

---

dseapy:

Add support for running UX server restricted to a namespace.

### DIFF
--- a/server/cmd/start.go
+++ b/server/cmd/start.go
@@ -20,12 +20,15 @@ var (
 	}
 )
 
-func Start(insecure bool, port int) {
+func Start(insecure bool, port int, namespaced bool, managedNamespace string) {
 	logger := logging.NewLogger().Named("server")
 	router := gin.New()
 	router.Use(gin.Logger())
 	router.RedirectTrailingSlash = true
 	router.Use(static.Serve("/", static.LocalFile("./ui/build", true)))
+	if namespaced {
+		router.Use(Namespace(managedNamespace))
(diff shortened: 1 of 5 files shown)


server/cmd/start.go
```diff
@@ -20,12 +20,15 @@ var (
...
 
-func Start(insecure bool, port int) {
+func Start(insecure bool, port int, namespaced bool, managedNamespace string) {
 	logger := logging.NewLogger().Named("server")
 	router := gin.New()
 	router.Use(gin.Logger())
 	router.RedirectTrailingSlash = true
 	router.Use(static.Serve("/", static.LocalFile("./ui/build", true)))
+	if namespaced {
+		router.Use(Namespace(managedNamespace))
```
whynowy, line 30: 👍

whynowy (approved):
LGTM!

I was thinking of raising an issue about this yesterday after making changes to the installation manifests, but you already submitted a PR for it!

PR #250, opened 2022-10-21
chore: introduce simple int offset

vigith:

Offset is a monotonically increasing integer. For most of the internal representations, int64 makes more sense rather than the string representation. We can avoid an unwanted conversion if we have the support of int64. 

This will help when saving the message in WAL (a hot code path) where we can encode/decode the offset as int64.

### DIFF
--- a/pkg/isb/interfaces.go
+++ b/pkg/isb/interfaces.go
@@ -88,7 +88,7 @@ func (so SimpleStringOffset) AckIt() error {
 	return nil
 }
 
-// SimpleIntOffset is an Offset convenient function for implementations without needing AckIt() when offset is a string.
+// SimpleIntOffset is an Offset convenient function for implementations without needing AckIt() when offset is a int64.
 type SimpleIntOffset func() int64
 
 func (si SimpleIntOffset) String() string {
--- a/pkg/isb/interfaces.go
+++ b/pkg/isb/interfaces.go
@@ -73,17 +73,32 @@ type Offset interface {
 	AckIt() error
 }
 
-// SimpleOffset is an Offset convenient function for implementations without needing AckIt()
-type SimpleOffset func() string
+// SimpleStringOffset is an Offset convenient function for implementations without needing AckIt() when offset is a string.
+type SimpleStringOffset func() string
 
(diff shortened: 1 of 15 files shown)


pkg/isb/stores/simplebuffer/buffer.go
```diff
@@ -112,7 +112,7 @@ func (b *InMemoryBuffer) Write(_ context.Context, messages []isb.Message) ([]isb
 			errs[idx] = nil
 			b.buffer[currentIdx].dirty = true
 			b.writeIdx = (currentIdx + 1) % b.size
-			writeOffsets = append(writeOffsets, isb.SimpleOffset(func() string {
+			writeOffsets = append(writeOffsets, isb.SimpleStringOffset(func() string {
```
whynowy, line 115: use the new one?

pkg/isb/interfaces.go
```diff
@@ -73,17 +73,32 @@ type Offset interface {
...
+func (so SimpleStringOffset) Sequence() (int64, error) {
 	return strconv.ParseInt(so(), 10, 64)
 }
 
-func (so SimpleOffset) AckIt() error {
+func (so SimpleStringOffset) AckIt() error {
+	return nil
+}
+
+// SimpleIntOffset is an Offset convenient function for implementations without needing AckIt() when offset is a string.
```
whynowy, line 91: `is an int64`

pkg/isb/stores/simplebuffer/buffer.go
```diff
@@ -176,7 +176,7 @@ func (b *InMemoryBuffer) Read(ctx context.Context, count int64) ([]*isb.ReadMess
 			}
 		}
 
-		readMessage := isb.ReadMessage{Message: msg, ReadOffset: isb.SimpleOffset(func() string { return strconv.Itoa(int(currentIdx)) })}
+		readMessage := isb.ReadMessage{Message: msg, ReadOffset: isb.SimpleStringOffset(func() string { return strconv.Itoa(int(currentIdx)) })}
```
whynowy, line 179: same here?

whynowy (approved):

PR #255, opened 2022-10-22
fix: retry only the failed offsets

yhl25:

retry only the failed offsets and use exponential backoff for Ack

### DIFF
--- a/pkg/isb/forward/forward.go
+++ b/pkg/isb/forward/forward.go
@@ -295,8 +295,9 @@ func (isdf *InterStepDataForward) ackFromBuffer(ctx context.Context, offsets []i
 		Steps:    math.MaxInt64,
 		Duration: time.Millisecond * 10,
 	}
+	var ackOffsets = offsets
 	ctxClosedErr := wait.ExponentialBackoffWithContext(ctx, ackRetryBackOff, func() (done bool, err error) {
-		errs := isdf.fromBuffer.Ack(ctx, offsets)
+		errs := isdf.fromBuffer.Ack(ctx, ackOffsets)
 		summarizedErr := errorArrayToMap(errs)
 		var failedOffsets []isb.Offset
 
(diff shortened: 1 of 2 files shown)


pkg/isb/forward/forward.go
```diff
@@ -285,24 +287,44 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
...
-			// TODO: implement retry with backoff etc.
-			time.Sleep(isdf.opts.retryInterval)
+
+			// retry only the failed offsets
+			for i, offset := range offsets {
+				if errs[i] != nil {
+					failedOffsets = append(failedOffsets, offset)
+				}
+			}
+			offsets = failedOffsets
```
vigith, line 312: usually you don't do this approach, because it is mutating the function argument. It is not wrong but usually frowned upon 😄 

does it make sense to do
```go
func (isdf *InterStepDataForward) ackFromBuffer(ctx context.Context, offsets []isb.Offset) error {
   var offsets := offsets
```
to give a feeling that this is a local copy?

vigith: @whynowy, any comments?

ashwinidulams: does it so happen with Jetstream that if Ack operation on one offset fails but the subsequent offset may suceed concurrently/immediately?
also if it need not be in a sequence, can we not parallelize this Ack operation?

ashwinidulams: ofcourse if the Ack is inherently async, then parallelism doesn't matter

vigith: Ack is done concurrently https://github.com/numaproj/numaflow/blob/main/pkg/isb/stores/jetstream/reader.go#L235-L246

whynowy: The values of Slice parameter go with pointers, needs to be very careful, as long as there's no `slice[index] = xxx`, it should be okay. Line 312 is already to define a local variable, should be safe. But it would be better to avoid ambiguity :)

vigith (approved):

vigith (approved):

PR #261, opened 2022-10-23
fix: nil pointer deref when running example with minimal CRD. Fixes #260

### LINKED ISSUE #260: nil pointer dereference when running 1-simple-pipeline with minimal crd
**Describe the bug**
When installing with a namespaced mode it is suggested that user should use minimal crd (ref: https://github.com/numaproj/numaflow/blob/29f15d57a125c54b6de8c4cd29d16e3d8473b655/docs/installation.md#namespace-scope).  When doing this the `simple-pipeline` example (ref: https://github.com/numaproj/numaflow/blob/27e6a8755bd1092b31d93fac891d1decd006a093/examples/1-simple-pipeline.yaml) errors with a panic:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1a66f29]

goroutine 1 [running]:
github.com/numaproj/numaflow/pkg/sources.(*SourceProcessor).getSourcer(0xc0009a7d78, {0xc0001abd80, 0x1, 0x1}, {0x23b7b80, 0xc0005ee690}, 0x23b2100?, {0x23b7bd0, 0xc0007f82e0}, 0xc000493b70)
	/home/dseapy/github/numaflow/pkg/sources/source.go:145 +0x4c9
github.com/numaproj/numaflow/pkg/sources.(*SourceProcessor).Start(0xc0009a7d78, {0x23cc038, 0xc0001e83f0})
	/home/dseapy/github/numaflow/pkg/sources/source.go:95 +0x465
github.com/numaproj/numaflow/cmd/commands.NewProcessorCommand.func1(0xc00059f680?, {0x2076944?, 0x2?, 0x2?})
	/home/dseapy/github/numaflow/cmd/commands/processor.go:67 +0x525
github.com/spf13/cobra.(*Command).execute(0xc00059f680, {0xc000130fe0, 0x2, 0x2})
	/home/dseapy/go/pkg/mod/github.com/spf13/cobra@v1.2.1/command.go:856 +0x67c
github.com/spf13/cobra.(*Command).ExecuteC(0x33a1440)
	/home/dseapy/go/pkg/mod/github.com/spf13/cobra@v1.2.1/command.go:974 +0x3bd
github.com/spf13/cobra.(*Command).Execute(...)
	/home/dseapy/go/pkg/mod/github.com/spf13/cobra@v1.2.1/command.go:902
github.com/numaproj/numaflow/cmd/commands.Execute()
	/home/dseapy/github/numaflow/cmd/commands/root.go:16 +0x25
main.main()
	/home/dseapy/github/numaflow/cmd/main.go:8 +0x17
```

This is caused due to `msgSize` missing in the example, which doesn't get defaulted when using minimal crd.

**To Reproduce**

1. Install minimal crd in cluster
2. kubectl apply the `1-simple-pipeline.yaml` example
3. kubectl logs the `simple-pipeline-in-0-...` pod

**Expected behavior**
* Error message provides indication as to what is missing.
* Warning in installation guide that full crd leverages defaults which will not be populated by the minimal crd.  Therefore some examples may be incomplete if using the minimal crd.

- Kubernetes: v1.25.3
- Numaflow: head of main

---

dseapy:

Fixes #260.

Avoids nil pointer dereference.  Adds note about using minimal CRD with examples that may rely on defaults present in full CRD.

### DIFF
--- a/pkg/sources/source.go
+++ b/pkg/sources/source.go
@@ -141,7 +141,7 @@ func (sp *SourceProcessor) getSourcer(writers []isb.BufferWriter, fetchWM fetch.
 			return nil, fmt.Errorf("generator rpu is not set")
 		}
 		if x.MsgSize == nil {
-			return nil, fmt.Errorf("generator msgsize is not set")
+			return nil, fmt.Errorf("generator msgSize is not set")
 		}
 		if x.Duration == nil {
 			return nil, fmt.Errorf("generator duration is not set")
--- a/pkg/sources/source.go
+++ b/pkg/sources/source.go
@@ -142,7 +142,7 @@ func (sp *SourceProcessor) getSourcer(writers []isb.BufferWriter, fetchWM fetch.
 		if l := sp.VertexInstance.Vertex.Spec.Limits; l != nil && l.ReadTimeout != nil {
 			readOptions = append(readOptions, generator.WithReadTimeOut(l.ReadTimeout.Duration))
 		}
-		return generator.NewMemGen(sp.VertexInstance, int(*x.RPU), *x.MsgSize, x.Duration.Duration, writers, fetchWM, publishWM, publishWMStores, readOptions...)
+		return generator.NewMemGen(sp.VertexInstance, writers, fetchWM, publishWM, publishWMStores, readOptions...)
 	} else if x := src.Kafka; x != nil {
 		readOptions := []kafka.Option{
 			kafka.WithGroupName(x.ConsumerGroupName),
--- a/pkg/sources/source.go
+++ b/pkg/sources/source.go
@@ -136,6 +136,17 @@ func (sp *SourceProcessor) getSourcer(writers []isb.BufferWriter, fetchWM fetch.
 	src := sp.VertexInstance.Vertex.Spec.Source
 
 	if x := src.Generator; x != nil {
+		// generator fields can be nil if using minimal CRD that doesn't provide defaults
+		if x.RPU == nil {
+			return nil, fmt.Errorf("generator rpu is not set")
+		}
+		if x.MsgSize == nil {
+			return nil, fmt.Errorf("generator msgsize is not set")
+		}
+		if x.Duration == nil {
+			return nil, fmt.Errorf("generator duration is not set")
+		}
+
 		readOptions := []generator.Option{
 			generator.WithLogger(logger),
 		}
--- a/pkg/sources/source.go
+++ b/pkg/sources/source.go
@@ -136,17 +136,6 @@ func (sp *SourceProcessor) getSourcer(writers []isb.BufferWriter, fetchWM fetch.
 	src := sp.VertexInstance.Vertex.Spec.Source
 
 	if x := src.Generator; x != nil {
-		// generator fields can be nil if using minimal CRD that doesn't provide defaults
-		if x.RPU == nil {
-			return nil, fmt.Errorf("generator rpu is not set")
-		}
-		if x.MsgSize == nil {
-			return nil, fmt.Errorf("generator msgSize is not set")
-		}
-		if x.Duration == nil {
-			return nil, fmt.Errorf("generator duration is not set")
-		}
-
 		readOptions := []generator.Option{
 			generator.WithLogger(logger),
 		}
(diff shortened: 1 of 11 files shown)


pkg/sources/source.go
```diff
@@ -136,6 +136,17 @@ func (sp *SourceProcessor) getSourcer(writers []isb.BufferWriter, fetchWM fetch.
 	src := sp.VertexInstance.Vertex.Spec.Source
 
 	if x := src.Generator; x != nil {
+		// generator fields can be nil if using minimal CRD that doesn't provide defaults
+		if x.RPU == nil {
+			return nil, fmt.Errorf("generator rpu is not set")
```
dseapy, line 141: instead of throwing an error we could set here to the defaults, but then we are duplicating defaults present in the kubebuilder annotations (which populate the full crd).  but seems like we are already doing that for replicas. ref: 

* https://github.com/numaproj/numaflow/blob/27e6a8755bd1092b31d93fac891d1decd006a093/pkg/apis/numaflow/v1alpha1/redis_buffer_service.go#L124
* https://github.com/numaproj/numaflow/blob/27e6a8755bd1092b31d93fac891d1decd006a093/pkg/apis/numaflow/v1alpha1/redis_buffer_service.go#L44

maybe best not use crd defaults at all & just default in code?

anyway, ya'll let me know what changes are desired.

whynowy: Add the validation to https://github.com/numaproj/numaflow/blob/main/pkg/reconciler/pipeline/validate.go

whynowy:
@dseapy - I remember I left a comments about adding validation, somehow it disappeared...

I would like to correct myself - I think it should be consistent for the behaviors of minimal and full CRDs, since full CRD has below default values, we should also have default values when minimal CRD is enforced. We should only validate those mandatory fields, WDYT?

```
type GeneratorSource struct {
	// +kubebuilder:default=5
	// +optional
	RPU *int64 `json:"rpu,omitempty" protobuf:"bytes,1,opt,name=rpu"`
	// +kubebuilder:default="1s"
	// +optional
	Duration *metav1.Duration `json:"duration,omitempty" protobuf:"bytes,2,opt,name=duration"`
	// Size of each generated message
	// +kubebuilder:default=8
	// +optional
	MsgSize *int32 `json:"msgSize,omitempty" protobuf:"bytes,3,opt,name=msgSize"`
}
```

dseapy:
> @dseapy - I remember I left a comments about adding validation, somehow it disappeared...
Sorry I think i resolved it which collapsed it (https://github.com/numaproj/numaflow/pull/261#discussion_r1002638962), or maybe there were other comments that disappeared.

> I think it should be consistent for the behaviors of minimal and full CRDs
Sounds good.  I can remove the warning in the installation.md.

> we should also have default values when minimal CRD is enforced.
To make sure I'm understanding, are you suggesting setting default values for gpu/duration/msgSize in the code (that's the only way to set them for minimal CRD afaik), and if so where would be the best place - inside `getSourcer` before they are used or somewhere else?

> We should only validate those mandatory fields
To make sure I'm understanding, by this you mean remove the validation checks for gpu/duration/msgSize as there will now exist defaults?  If so, sounds good to me.

---
If defaults end up getting set in the code, then seems like having the defaults in the full CRD is only resulting in duplication of default values.  Would it make sense to remove the kubebuilder annotations that set the defaults and solely rely on code defaults in this case, or is there value in keeping those defaults?

whynowy:
> > we should also have default values when minimal CRD is enforced.
> 
> To make sure I'm understanding, are you suggesting setting default values for gpu/duration/msgSize in the code (that's the only way to set them for minimal CRD afaik), and if so where would be the best place - inside `getSourcer` before they are used or somewhere else?

The `generator.NewMemGen` function has redundant parameters, we can remove `int(*x.RPU), *x.MsgSize, x.Duration.Duration`, since `vertexInstance` is already there, then we can make the default values in the generator implementation, similar to the Kafka one. 

> 
> > We should only validate those mandatory fields
> 
> To make sure I'm understanding, by this you mean remove the validation checks for gpu/duration/msgSize as there will now exist defaults? If so, sounds good to me.

Yes

> 
> If defaults end up getting set in the code, then seems like having the defaults in the full CRD is only resulting in duplication of default values. Would it make sense to remove the kubebuilder annotations that set the defaults and solely rely on code defaults in this case, or is there value in keeping those defaults?

I don't have an answer for it, maybe we can keep them for now, and revisit later...

whynowy (approved):
LGTM, thanks!

PR #265, opened 2022-10-25
fix: main branch make build failure: math.MaxInt64 for int type

chromevoid:
- A separate PR to fix main branch make build failure: math.MaxInt64 for int type https://github.com/numaproj/numaflow/actions/runs/3316672732/jobs/5478733505#step:4:2410
  ref: https://github.com/numaproj/numaflow/blob/47229e28fe8f13f977f8e5dafcf16c504295a223/pkg/isb/forward/forward.go#L296
- ~~Add make build in the ci~~ remove the build step for now

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -112,6 +112,20 @@ jobs:
       - name: Test
         run: make test-code
 
+  build-binaries:
+    name: Build Binaries
+    timeout-minutes: 10
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup Go
+        uses: actions/setup-go@v3.1.0
+        with:
+          go-version: 1.19
+      - name: Build binaries
+        run: make build
+
   lint:
     name: Lint
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -114,8 +114,8 @@ jobs:
 
   build-binaries:
     name: Build Binaries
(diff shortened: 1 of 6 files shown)


pkg/isb/forward/forward.go
```diff
@@ -292,7 +292,7 @@ func (isdf *InterStepDataForward) ackFromBuffer(ctx context.Context, offsets []i
 	var ackRetryBackOff = wait.Backoff{
 		Factor:   1,
 		Jitter:   0.1,
-		Steps:    math.MaxInt64,
+		Steps:    math.MaxInt,
```
whynowy, line 295: There are several other places having same issue, fix all of them.

.github/workflows/ci.yaml
```diff
@@ -112,6 +112,20 @@ jobs:
       - name: Test
         run: make test-code
 
+  build-binaries:
```
whynowy, line 115: Why image build does not fail?

chromevoid: You mean “There are several other places having same issue”, so the make build should have failed?…

vigith: He meant, there are many places you see int64, fix all those.

whynowy: And curious why image build didn't fail, it also has binary build included.

chromevoid: > And curious why image build didn't fail, it also has binary build included.

could be in release.yaml it's using `ubuntu-20.04` while here we are using `ubuntu-latest`?...

whynowy (approved):

PR #278, opened 2022-10-26
feat: Watermark millisecond. Fixes #201

### LINKED ISSUE #201: move watermark at millisecond granularity
Currently, the watermark is progressing at second granularity. This is too slow for near-realtime systems. Please make sure the progression at the millisecond level will not affect the overall load on the watermark store.

Tasks 
- [ ] write a proposal on what it takes to make watermark progress at the millisecond level
- [ ] implement millisecond progression

### LINKED ISSUE #266: Reduce or document watermark ignored pod logs in simple pipeline
# Summary

Currently when running simple-pipeline from https://github.com/numaproj/numaflow/tree/main/examples#numaflow-by-example it prints out lots of logs in all the vertex pods that look like the following:
```
{"level":"info","ts":1666688845.801714,"logger":"numaflow.udf-processor","caller":"publish/publisher.go:88","msg":"New watermark is ignored because it's older than the current watermark","vertex":"simple-pipeline-cat","head":"2022-10-25T09:07:24Z","new":"2022-10-25T09:07:24Z"}
```

It's not clear if it's expected to get all these messages or if it's an issue.  I think since they are logged at `info` rather than `debug` or `warn` it's not clear if it's a problem or something I can ignore.  I'm not sure the best thing to do, but maybe one of the following:

* Consider changing the log level if it makes sense
* If it's something new user is likely to be seeing a lot in the logs, add documentation either in the examples' README.md or in the docs' watermarks.md.  `1-simple-pipeline-wm-disabled.yaml` prevents the additional logs, but watermarks are enabled on the readme example.

# Use Cases

When getting started with examples it would be nice if the logs were cleaner to be sure install went smoothly.  It would also make seeing the other logging (i.e. `2022/10/25 09:30:51 (out) {"Data":"Cv52GfRFIRc=","Createdts":1666690251805359626}`) easier without relying on something like `| grep -v older` on kubectl logs.

---

chromevoid:
fixes #201 
fixes #266 

Updated the watermark time unit from second to millisecond.
image

chore #266 
Updated the publish watermark log messages and level.
https://github.com/numaproj/numaflow/blob/ac9c14b4b7f1e7109df11def1f914dc619b8e948/pkg/watermark/publish/publisher.go#L87-L96

Created a new issue for watermark timestamp tooltip display.
https://github.com/numaproj/numaflow/issues/280


pkg/watermark/fetch/processor_manager_inmem_test.go
```diff
@@ -42,7 +42,7 @@ func TestFetcherWithSameOTBucket_InMem(t *testing.T) {
 	err = ot.PutKV(ctx, fmt.Sprintf("%s%s%d", "p1", "_", epoch), b)
 	assert.NoError(t, err)
 
-	epoch += 60
+	epoch += 60000000
```
KeranYang, line 45: 60000?

pkg/watermark/publish/publisher.go
```diff
@@ -83,10 +85,13 @@ func (p *publish) PublishWatermark(wm processor.Watermark, offset isb.Offset) {
 	}
 	// update p.headWatermark only if wm > p.headWatermark
 	if wm.After(time.Time(p.headWatermark)) {
-		p.log.Debugw("New watermark updated for head water mark", zap.String("head", p.headWatermark.String()), zap.String("new", wm.String()))
+		p.log.Debugw("New watermark is updated for the head watermark", zap.String("head", p.headWatermark.String()), zap.String("new", wm.String()))
 		p.headWatermark = wm
+	} else if wm.Before(time.Time(p.headWatermark)) {
+		p.log.Warnw("New watermark is older than the current watermark", zap.String("head", p.headWatermark.String()), zap.String("new", wm.String()))
```
KeranYang, line 91: nit: `Skip publishing the new watermark because it's older than the current watermark` ? just to be explicit on the skip.

KeranYang (approved):

KeranYang (approved):

vigith (approved):

PR #282, opened 2022-10-27
feat: add pipeline node counts and age to printcolumn. Closes #267

### LINKED ISSUE #267: Add relevant information to printcolumn for pipeline resources
# Summary

`kubectl get vertex` prints:
```
NAME                  PHASE     REASON   MESSAGE   DESIRED   CURRENT
simple-pipeline-cat   Running                      1         1
simple-pipeline-in    Running                      1         1
simple-pipeline-out   Running                      1         1
```
but `kubectl get pipeline` only prints:
```
NAME              PHASE   MESSAGE
simple-pipeline 
```

Maybe `phase` should be populated or removed.  Additionally it would be nice to see more printcolumns with relevant info, not sure what makes sense, some ideas:
* number of vertices
* number of edges
* age
* maybe something similar to `ready` (i.e. 5/5) that you get with po/deploy/sts, but for vertices?

# Use Cases

When running multiple pipelines, having `kubectl get` provide additional info beyond the pipeline name would be nice as a quick way to get a good overview.

---

dseapy:

Closes #267.

Adds printer columns to Pipeline resources.  Specifically adds counts for vertices, sources, sinks, udfs, as well as the age of the Pipeline resource.

Only shows sources, sinks and udfs when `-owide` is used.

`kubectl get pipeline`:
```
NAME                    PHASE     MESSAGE   VERTICES   AGE
simple-pipeline         Running             3          20m
```
`kubectl get pipeline -owide`:
```
NAME                    PHASE     MESSAGE   VERTICES   SOURCES   SINKS   UDFS   AGE
simple-pipeline         Running             3          1         1       1      20m
```


pkg/reconciler/pipeline/controller.go
```diff
@@ -140,9 +140,11 @@ func (r *pipelineReconciler) reconcileNonLifecycleChanges(ctx context.Context, p
 	pl.Status.InitConditions()
 	if err := ValidatePipeline(pl); err != nil {
 		log.Errorw("Validation failed", zap.Error(err))
+		pl.Status.ResetTopologyCounts()
```
whynowy, line 143: This means if it's a failure caused by an update, we also reset all the numbers... There's nothing wrong with it, i'm just thinking, if keeping the numbers is better?

whynowy:
Showing the vertices count looks great, maybe we don't need to show the number of edges? WDYT @vigith

pkg/reconciler/pipeline/controller.go
dseapy: yeah that's what it's doing, i wasn't sure if we wanted to clear them or not.  I'll go ahead and remove the reset.

dseapy:
> Showing the vertices count looks great, maybe we don't need to show the number of edges? WDYT @vigith

good point, i'm not sure i know of a great use-case for showing them. can remove.

dseapy:
> Showing the vertices count looks great, maybe we don't need to show the number of edges? WDYT @vigith

went ahead and made the change for now.  if @vigith or you do want them back, i can git revert the commits.

PR #285, opened 2022-10-29
feat: support adding init containers to vertices. Closes #284

### LINKED ISSUE #284: Custom init containers for vertex pods
# Summary

Add the ability for users to add init containers to the udf pods.

# Use Cases

Currently the udf container is responsible for downloading the ML model.  Instead it would be preferred for my use-case if there was an init-container (maintained by devops, not data-scientists developing the udf containers) that was responsible for acquiring the model and sharing it with the udf container through a volumemount.  This provides a couple advantages:
* Ensures consistency for configuration/implementation of acquiring the model.
* Provides an easy way to migrate if/when storage implementation changes (i.e. custom s3 endpoint, move from s3 to another storage provider, etc).
* Simplifies debugging & any required fixes by devops in case there are network-related issues, since they own a single container.

---

dseapy:

Closes #284.

Allows setting a list of `initContainers` to run on vertices.


pkg/reconciler/pipeline/validate.go
```diff
@@ -133,6 +133,15 @@ func validateVertex(v dfv1.AbstractVertex) error {
 	if min > max {
 		return fmt.Errorf("vertex %q: max number of replicas should be greater than or equal to min", v.Name)
 	}
+	for _, ic := range v.InitContainers {
+		if ic.Name == dfv1.CtrInit ||
+			ic.Name == dfv1.CtrMain ||
+			ic.Name == "numa" ||
```
vigith, line 139: since we do not have `numa` container, we can remove `ic.Name == "numa"` check, we can fix that by changing the const value of `CtrMain` later.

In case someone creates a container by the name `numa` by then, it won't be backward compatible 😄

vigith (commented):
I think we should also add this feature to our documentation too? https://numaflow.numaproj.io/user-defined-functions/

pkg/reconciler/pipeline/validate.go
whynowy: Could you change the value of `CtrMain` to `numa` in this PR?

whynowy:
> I think we should also add this feature to our documentation too? https://numaflow.numaproj.io/user-defined-functions/

Add a doc `Init Containers` in https://numaflow.numaproj.io/pipeline/, next to `Autoscaling` is better.

dseapy:
> > I think we should also add this feature to our documentation too? https://numaflow.numaproj.io/user-defined-functions/
> 
> Adding a doc `Init Containers` in https://numaflow.numaproj.io/pipeline/, next to `Autoscaling` is better.

ah, sorry, just saw this after pushing.  will address in the morning.

docs/init-containers.md
```diff
@@ -0,0 +1,49 @@
+# Init Containers
```
whynowy, line 1: This is awesome!

`mkdocs.yml` change is missing.

docs/quick-start.md
```diff
@@ -41,7 +41,7 @@ simple-pipeline-cat-0-kqrhy                  2/2     Running     0          10s
 simple-pipeline-in-0-rhpjm                   1/1     Running     0          11s
 
 # Watch the log for the `output` vertex
-kubectl logs -f simple-pipeline-out-0-xxxx main
+kubectl logs -f simple-pipeline-out-0-xxxx numa
```
whynowy, line 44: Removing the container name from the command line is better and works - there's a period of time before this change gets released (however the doc would have been published), and someone follows this guide will fail.

docs/quick-start.md
```diff
@@ -96,12 +96,12 @@ curl -kq -X POST -d "103" https://localhost:8444/vertices/in
 curl -kq -X POST -d "104" https://localhost:8444/vertices/in 
 
 # Watch the log for the even vertex
-kubectl logs -f even-odd-even-sink-0-nf2ql main
+kubectl logs -f even-odd-even-sink-0-nf2ql numa
```
whynowy, line 99: Same here.

docs/quick-start.md
```diff
@@ -96,12 +96,12 @@ curl -kq -X POST -d "103" https://localhost:8444/vertices/in
...
 2022/09/07 22:29:40 (even-sink) 102
 2022/09/07 22:29:40 (even-sink) 104
 
 # Watch the log for the odd vertex
-kubectl logs -f even-odd-odd-sink-0-a6p0n main
+kubectl logs -f even-odd-odd-sink-0-a6p0n numa
```
whynowy, line 104: Same here.

vigith (approved):

whynowy (approved):

whynowy:
oops, conflicts...

Issue #237, opened 2022-10-18
Unary gRPC message size limit 4MB


whynowy:
# Summary

There's a 4MB message size limit for unary gRPC message size, this is going to impact:

1. UDSink, which always do batch requests;
2. UDF which has single message over 4MB.

We need to fix this in 2 steps:

- [x] #299
- [x] #335

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

shashank10456:
I would like to contribute to this.

PR #294, opened 2022-10-31
feat: add support to configure max message size

chromevoid:
feat: #237 task 1

Note: merge and release https://github.com/numaproj/numaflow-go/pull/21 before reviewing this PR.


chromevoid:
closed as discussed: it would be better to keep the default 4 MB limit for now as it's not a necessary use case. We can revisit this later when we really need this feature.

PR #300, opened 2022-11-01
feat: customization for batch jobs. Closes #259

### LINKED ISSUE #259: Customization for Job containers
# Summary

Be able to set various container settings such as annotations, labels, resources, etc. on buffer-create & buffer-cleanup Jobs.

# Use Cases

* When needing to deploy with `ResourceQuota` or `LimitRange` in the namespace one must supply & rely on defaults specified in a `LimitRange`, otherwise jobs will not run.  Don't know if leveraging existing `standardResources` or if it being configurable is best, but being able to have something set would be awesome.
* There are probably various use-cases someone might want to inject labels or annotations.  For my current setup, I hit an issue when having istio auto-injection enabled in the namespace, and needing to disable it for the pod using an `sidecar.istio.io/inject: "false"` annotation to avoid injecting a sidecar which prevents the Job from terminating.

---

dseapy:

Closes #259.

Adds support for customizing batch jobs.

### DIFF
--- a/api/json-schema/schema.json
+++ b/api/json-schema/schema.json
@@ -18220,7 +18220,7 @@
           "type": "array"
         },
         "ttlSecondsAfterFinished": {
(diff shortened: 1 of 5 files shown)


whynowy:
https://github.com/numaproj/numaflow/pull/296 merged, good to go!

whynowy (approved):

PR #304, opened 2022-11-02
feat: customize init-container resources. Closes #303

### LINKED ISSUE #303: Allow setting container resource limits and requests on init containers
# Summary

Currently `init` containers use `standardResources` defined in the code, with `numa` containers either defaulting to `standardResources` or not being set.  `numa` containers are override-able using `containerTemplates`.

Allowing users to configure the `init` container resources is useful.  A couple possibilities:
1. setting `standardResources` value, i.e. controller configmap
2. setting overrides in the k8s custom resources

# Use Cases

When deploying in namespaces with a `ResourceQuota` requiring memory limits or cpu limits be set on all containers, the pods don't run.  ResourceQuota may be set for restricting resources on namespace but also enforcing memory limits are set (i.e. https://twitter.com/thockin/status/1134193838841401345?lang=en)

A workaround is setting a `LimitRange` default in the namespace to populate limits on containers that do not specify them. 
The issues with this are:

1. This affects more containers than just the numaflow containers, so when sharing a namespace with non-numaflow pods this will affect those pods.
2. It can lead to over-provisioning since the `LimitRange` default must be as high as the container with the most resource need that doesn't specify resources.
3. Limit ranges don't have visibility into what is already set in container resources, potentially leading to unexpected behavior ("A LimitRange does not check the consistency of the default values it applies." - https://kubernetes.io/docs/concepts/policy/limit-range/#limitrange-and-admission-checks-for-pods).

---

dseapy:

Closes #303.

Allows customizing container resources for the "init" container.


whynowy (approved):
LGTM, I like these refactoring!

The init container customization looks tedious if there are many vertices in the pipeline, but at least there's something to make it doable. We can add centralized configuration support in the future if there's a high demand, but this still can remain there as a customization taking higher priority.

PR #314, opened 2022-11-04
feat: controller-wide pipeline templates. Closes #305

### LINKED ISSUE #305: Support configuring defaults for pipeline component settings
# Summary

Currently `AbstractPodTemplate` and `ContainerTemplate` need to be specified within each pipeline component (`daemon`, `job`, `vertex`).  It would be useful to be able to set defaults for the controller to use.

# Use Cases

It would be nice to simplify the responsibility of the data-scientist creating the pipeline in a Dev environment, while also simplifying the responsibility of the devops team with the (eventual) promotion of the pipeline to a Prod environment.  Allowing defaults to be set for the controller would have the following benefits:
* Streamline data-scientists writing/running successful pipelines in clusters that require specific environment-specific configuration to run well.
* Streamline devops team implementing automation to promote Pipeline resources to different environments with different constraints (i.e. fewer per-pipeline daemon replicas in Dev to save $).
* Yaml with less duplication, making it easier to work with when working with Pipelines in general.

An example of this, is that in a Prod environment one might want settings that a data-scientist making the Pipeline might not care much about in Dev.  i.e. maybe not exactly this, but something like this for `daemon` component:
```yaml
  templates:
    daemon:
      replicas: 2
      metadata:
        annotations:
          sidecar.istio.io/inject: "false"
      containerTemplate:
        resources:
          requests:
            cpu: 100m
          limits:
            memory: 200Mi
      affinity:
        podAntiAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            - labelSelector:
                matchExpressions:
                  - key: app.kubernetes.io/component
                    operator: In
                    values:
                      - daemon
                  - key: numaflow.numaproj.io/pipeline-name
                    operator: In
                    values:
                      - example-pipeline
              topologyKey: kubernetes.io/hostname
```

dseapy:

Closes #305.

Adds support for specifying controller-wide pipeline templates in the controller configmap.


pkg/apis/numaflow/v1alpha1/vertex_template.go
```diff
@@ -0,0 +1,48 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+type VertexTemplate struct {
+	// +optional
+	AbstractPodTemplate `json:",inline" protobuf:"bytes,1,opt,name=abstractPodTemplate"`
```
dseapy, line 21: initially did inline to be consistent with `daemon`/`job`, but just looked back at issue and seems like maybe `json:"podTemplate,omitempty"` is desired to be consistent with `containerTemplate`/`initContainerTemplate`.  let me know if i should make that change here & the docs.

docs/pipeline-customization.md
```diff
@@ -84,35 +96,10 @@ metadata:
...
-          - name: MY_ENV_NAME
-            value: my-env-value
-        resources:
-          limits:
-            memory: 500Mi
 ```
+
+## Vertex customization
+
+See `Component customization` described above.
```
whynowy, line 105: Is there an place to mention that it can also be done in pipeline level?

pkg/reconciler/pipeline_templates.go
```diff
@@ -0,0 +1,42 @@
...
+	"bytes"
+	"fmt"
+	"os"
+
+	"k8s.io/apimachinery/pkg/util/yaml"
+
+	dfv1 "github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
+)
+
+func LoadPipelineTemplates() (*dfv1.Templates, error) {
```
whynowy, line 29: Use the same way as `LoadConfig` to achieve hot reloading?

pkg/reconciler/pipeline/controller.go
```diff
@@ -76,11 +77,16 @@ func (r *pipelineReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 	log := r.logger.With("namespace", pl.Namespace).With("pipeline", pl.Name)
 	plCopy := pl.DeepCopy()
 	ctx = logging.WithLogger(ctx, log)
+	if err := plCopy.Spec.Templates.UpdateWithDefaultsFrom(r.templates); err != nil {
```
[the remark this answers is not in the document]
dseapy: currently there is a check in the func & unit tests for it being nil, which is not typical but done to reduce code on the caller side.
```go
	if tpl == nil {
		*tpl = *defaultTemplate.DeepCopy()
		return nil
	}
```
should i move that bit out to the couple places the func is called?

pkg/reconciler/pipeline/controller.go
```diff
@@ -76,11 +77,16 @@ func (r *pipelineReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+		r.logger.Errorw("Unable to create merged pipeline templates", err)
+		return ctrl.Result{}, err
+	}
 
 	result, reconcileErr := r.reconcile(ctx, plCopy)
 	if reconcileErr != nil {
 		log.Errorw("Reconcile error", zap.Error(reconcileErr))
 	}
+	plCopy.Spec.Templates = pl.Spec.Templates.DeepCopy()
```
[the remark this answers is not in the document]
dseapy: templates do not get changed in `reconcile()`, but they do get changed from `pl` immediately after plCopy is created in Reconcile().

Mostly I was trying to find a good way to do the cluster-wide + pipeline-wide templates merge once & not need to pass down the merged templates through the func calls.  However maybe each `job`, `daemon`, `vertex` component can handle their own merge to avoid changing it... will look at that.

pkg/reconciler/pipeline/controller.go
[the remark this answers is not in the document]
dseapy: Can also change func that takes 2 templates that doesn't modify either, but returns a separate one, and do nil checks on the caller side?

docs/pipeline-customization.md
dseapy: I'm a bit confused, as this is for the pipeline level, do you mean modify this to mention it's for the pipeline-level?  At the top of the page in the `Pipeline Customization` section I try to describe things can be done in the controller configmap and vertexes can be customized individually, but may not be doing a great job with the wording.

pkg/apis/numaflow/v1alpha1/vertex_template.go
dseapy: just to confirm, i should leave `daemon` and `job` unchanged (keep those inline)?

pkg/reconciler/pipeline_templates.go
dseapy: sure, i was struggling to get viper to unmarshal the yaml the way it does for config but I can still use viper to watch it but not unmarshal.  or if we need viper to unmarshal, I can look at why viper was unmarshaling it to empty Templates.

i'm slightly concerned with thread-safety.  does it make sense to add read/write locking?

docs/pipeline-customization.md
whynowy: With this change, there are 3 places to customize the vertex template - configmap, `pipeline spec.template.vertex`, and `pipeline.spec.vertices[*].xxx`, I don't see the 2nd one is mentioned.

dseapy: The "component customization" section in this file and the comment in the example attempt to describe that.

whynowy:
I just thought a little bit more - is it too much for the templates customization (cluster level in the configmap, pipeline level and on the vertex)? Maybe pipeline level and vertex level support is good enough, and that makes the implementation simpler. WDYT? @vigith @dseapy

dseapy:
> I just thought a little bit more - is it too much for the templates customization (cluster level in the configmap, pipeline level and on the vertex)? Maybe pipeline level and vertex level support is good enough, and that makes the implementation simpler. WDYT? @vigith @dseapy

I'm happy to do whatever is desired, for my current use-case setting on controller and vertex is sufficient.

That said, I'm not sure how to leverage that for simpler code.  At the moment, there is code for:
1. Loading from ConfigMap to a `Templates` struct
2. `AbstractVertex` struct and applying it to an `AbstractVertex`
3. Merging `Templates` struct with another one using StrategicMergePatch (cluster-wide merging with pipeline-wide)

For daemon & job we already need to do (1) and (3), and regardless of pipeline-level vertex customization support I think we need to do (2).

Maybe it's simpler because if changing approach to merging `DaemonTemplate` and `JobTemplate` (replacing `UpdateWithDefaultsFrom` which is merging the single `Templates`), then that would avoid needing to have a func for merging `VertexTemplate`?

whynowy:
> > I just thought a little bit more - is it too much for the templates customization (cluster level in the configmap, pipeline level and on the vertex)? Maybe pipeline level and vertex level support is good enough, and that makes the implementation simpler. WDYT? @vigith @dseapy
> 
> I'm happy to do whatever is desired, for my current use-case setting on controller and vertex is sufficient.
> 
> That said, I'm not sure how to leverage that for simpler code. At the moment, there is code for:
> 
> 1. Loading from ConfigMap to a `Templates` struct
> 2. `VertexTemplate` struct and applying it to an `AbstractVertex`
> 3. Merging `Templates` struct with another one using StrategicMergePatch (cluster-wide merging with pipeline-wide)
> 
> For daemon & job we already need to do (1) and (3), and regardless of pipeline-level vertex customization support I think we need to do (2).
> 
> Maybe it's simpler because if changing approach to merging `DaemonTemplate` and `JobTemplate` individually (replacing `UpdateWithDefaultsFrom` which is merging the single `Templates`), then that would avoid needing to have a func for merging `VertexTemplate`?

Ignore what I said, let's follow what we discussed before :), having 3 controls is valuable.

vigith:
> Ignore what I said, let's follow what we discussed before :), having 3 controls is valuable.

+1, we have some internal use cases where we will be sharing the controller and running multiple ML pipelines.

dseapy:
closing for now, ref: https://github.com/numaproj/numaflow/issues/331#issuecomment-1310141252

tldr: as-written this doesn't resolve need for pipeline/vertex specific info (i.e. affinity/pdb label matching).

PR #316, opened 2022-11-05
fix: jetstream build watermark progressors bug

chromevoid:
In the `jetsream.BuildWatermarkProgressors` function, we always use `fromBuffer := vertexInstance.Vertex.GetFromBuffers()[0]` as the fromBuffer because we didn't have a use case to read from multiple fromBuffers. 

Now that we've added the reduce feature, we could have multiple incoming fromBuffers. Therefore, we need to update the current `jetsream.BuildWatermarkProgressors` implementation to support multiple fromBuffers.

The way we are fixing it is to add a new input argument "fromBuffer" to the function, and the caller needs to provide the fromBuffer.

Note that in source/map_udf/sink, we still only support one incoming fromBuffer ATM, so it's safe to use `GetFromBuffers()[0]`. 
The new reduce use case has multiple pods (0~n) in one vertex, and each pod is listening on its own fromBuffer.


pkg/udf/reduce_udf.go
```diff
@@ -61,29 +61,29 @@ func (u *ReduceUDFProcessor) Start(ctx context.Context) error {
 	var reader isb.BufferReader
 	var writers map[string]isb.BufferWriter
 	var err error
+	var fromBuffer dfv1.Buffer
 	fromBuffers := u.VertexInstance.Vertex.GetFromBuffers()
-	var fromBufferName string
 	for _, b := range fromBuffers {
```
vigith, line 66:
```suggestion
	// choose the buffer that corresponds to this reduce processor because
	// reducer's incoming edge can have more than one buffer for parallelism 
	for _, b := range fromBuffers {
```

vigith (approved):

PR #321, opened 2022-11-06
feat: populate watermark settings to vertex spec. Closes #320

### LINKED ISSUE #320: Pipeline watermark config change should reflect on vertex pod
# Summary

Today, after a pipeline is created, any watermark update on the pipeline (for example, updating `maxDelay`) won't reflect on the vertex pod, this is because the watermark settings is set as ENV variables on the pod by the vertex controller,  which does not (and should not)watch the pipeline changes.

To fix this issue, the watermark configuration on the pipeline should be populated to the vertex spec, so that the watermark updates on the pipeline will be passed to the vertex by the pipeline controller, and then vertex controller is able to recreate the vertex pod with latest watermark config.

---

whynowy:

Closes: https://github.com/numaproj/numaflow/issues/320


vigith (commented):
2 concerns:

1. Users might wrongly think they can turn on and off the watermark at each Vertex level. Vn+1th vertex can have only watermark IFF Vn vertex emits it.
1. maxDelay does not have significance in any other vertex other than the source. users might wrongly think they have maxDelay at each vertex.

whynowy:
> 2 concerns:
> 
> 1. Users might wrongly think they can turn on and off the watermark at each Vertex level. Vn+1th vertex can have only watermark IFF Vn vertex emits it.
> 2. maxDelay does not have significance in any other vertex other than the source. users might wrongly think they have maxDelay at each vertex.

No. Users can not set watermark on vertex level. The populated `watermark` object is under `spec`, not `abstractVertex`.

vigith (approved):

PR #325, opened 2022-11-07
feat: support adding sidecars in vertex pods. Closes #323

### LINKED ISSUE #323: Support adding sidecar containers in udf vertex pods
# Summary

Allow the user to specify a list of containers to add to the vertex pods, similar to how init-containers can currently be provided.

# Use Cases

Certain java libraries are useful for some specialized processing needing to be done.  Would like to make those capabilities available to python udf containers written by data-scientists via an apache arrow flight (GRPC) API. Would like to avoid calling out to another pod & instead have a shared volume between the containers for this use-case as some data is quite large.

Another thing that would be nice to do is collect more custom metrics related to cost/gpu-usage.  Having a minimal sidecar container written in Go maintained by devops would be useful.

---

dseapy:

Closes #323.

Allows users to provide a list of containers to add to vertex pods.


pkg/apis/numaflow/v1alpha1/vertex_types.go
```diff
@@ -367,6 +367,9 @@ type AbstractVertex struct {
 	// More info: https://kubernetes.io/docs/concepts/workloads/pods/init-containers/
 	// +optional
 	InitContainers []corev1.Container `json:"initContainers,omitempty" protobuf:"bytes,11,rep,name=initContainers"`
+	// List of sidecar containers belonging to the pod.
+	// +optional
+	SidecarContainers []corev1.Container `json:"sidecarContainers,omitempty" protobuf:"bytes,12,rep,name=sidecarContainers"`
```
dseapy, line 372: thoughts on naming? `sidecarContainers`, `containers`, `sidecars`?

whynowy: +1 for `sidecars`.

whynowy (approved):
LGTM, thanks!

docs/sidecar-containers.md
```diff
@@ -0,0 +1,54 @@
...
+          image: busybox:latest
+          command: ["/bin/sh", "-c", "echo \"my-sidecar is running!\" && tail -f /dev/null"]
+      udf:
+        container:
+          image: my-function:latest
+```
+
+There are various use-cases for sidecars. One possible use-case is a `udf` container that needs functionality
+from a library written in a different language. The library's functionality could be made available through
+gRPC over Unix Domain Socket. The following example shows how that could be accomplished using a shared [`volume`](./volumes.md).
```
vigith, line 26:
```suggestion
gRPC over Unix Domain Socket. The following example shows how that could be accomplished using a shared [`volume`](./volumes.md).

It is the sidecar owner's responsibility to come up with a protocol that can be used with the UDF. It could be volume, gRPC, TCP, HTTP 1.x, etc.,
```

vigith (approved):
IMHO, we should not let a sidecar sit on the source vertex since we do not have a user-defined source. The moment we let that happen, users will ask for a user-defined source, but there are a few other tactical items (e.g., making watermark easy) we need to fix before we can have a user-defined source.

vigith (approved):

PR #336, opened 2022-11-10
refactor: Close watermark fetcher and publisher correctly

whynowy:

1. Close() function on watermark fetchers and publishers, handle them when application exits;
2. Shutdown reduce application with resource cleaning up.


pkg/isbsvc/jetstream_service.go
```diff
@@ -128,7 +127,6 @@ func (jss *jetStreamSvc) CreateBuffers(ctx context.Context, buffers []dfv1.Buffe
 			}
 			if _, err := js.CreateKeyValue(&nats.KeyValueConfig{
 				Bucket:       otBucket,
-				Description:  fmt.Sprintf("Offset timeline bucket of buffer [%s]", buffer.Name),
```
whynowy, line 131: The description makes each of the table row in the nats jetstream console wrap, which is hard to debug.

pkg/watermark/store/inmem/kv_store.go
```diff
@@ -79,9 +74,6 @@ func NewKVInMemKVStore(ctx context.Context, pipelineName string, bucketName stri
 		kvEntryCh:    make(chan store.WatermarkKVEntry, 10),
 		log:          logging.FromContext(ctx).With("pipeline", pipelineName).With("bucketName", bucketName),
 	}
-	bucketsLock.Lock()
-	buckets[bucketName] = s
```
vigith, line 83: @chromevoid, why did we have this mapping?

vigith (approved):

pkg/watermark/store/inmem/kv_store.go
chromevoid: I remember correctly it was added because 
1. I thought we will use in mem kv store in real production usages
2. split buckets feature we used to have

vigith: > I thought we will use in mem kv store for real production use case

why do we need this in real production use cases? How will we use it?

chromevoid: > > I thought we will use in mem kv store for real production use case
> 
> why do we need this in real production use cases? How will we use it?

No we don’t need this in real production.

PR #337, opened 2022-11-10
feat: timestamp in UI to display milliseconds. closes #280

### LINKED ISSUE #280: Display timestamp with millisecond information in the pod watermark tooltip
Currently, the tooltip displays the watermark timestamp up until second level. We'd like to have millisecond level info as well.
e.g. In the following screenshot, we expect to see something like `2022-10-26T08:54:48.788Z` for the given epoch `1666842888788`.

Note: The time is displayed using local timezone.
image

DevilAeron:
Fixes #280 

Adds milliseconds to the tooltip.


Krithika3 (approved):
Looks nice DevilAeron :)

PR #340, opened 2022-11-11
Chore: Windower interface. closes #234

### LINKED ISSUE #234: windowing strategy as an interface

ashwinidulams:
Introducing a interface for window management. each windowing strategy will then implement methods in the interface to manage the windows accordingly.


pkg/window/windower.go
```diff
@@ -20,10 +20,27 @@ import (
...
+type AlignedWindow interface {
+	StartTime() time.Time
+	EndTime() time.Time
+}
+
+// Windower manages windows
+// Will be implemented by each of the windowing strategies.
 type Windower interface {
 	// AssignWindow assigns the event to the window based on give window configuration.
 	AssignWindow(eventTime time.Time) []*IntervalWindow
```
whynowy, line 37: return `[]AlignedWindow`

pkg/window/strategy/fixed/fixed.go
```diff
@@ -23,34 +23,153 @@ limitations under the License.
...
 	return &Fixed{
-		Length: length,
+		Length:  length,
+		entries: list.New(),
+		lock:    sync.RWMutex{},
 	}
 }
 
 // AssignWindow assigns a window for the given eventTime.
 func (f *Fixed) AssignWindow(eventTime time.Time) []*window.IntervalWindow {
```
whynowy, line 67: return `[]AlignedWindow`

pkg/reduce/readloop/readloop.go
```diff
@@ -183,13 +179,14 @@ func (rl *ReadLoop) Process(ctx context.Context, messages []*isb.ReadMessage) {
 
 		// close any windows that need to be closed.
 		wm := processor.Watermark(m.Watermark)
-		closedWindows := rl.aw.RemoveWindow(time.Time(wm))
+		closedWindows := rl.windower.RemoveWindows(time.Time(wm))
 		rl.log.Debugw("closing windows", zap.Int("length", len(closedWindows)), zap.Time("watermark", time.Time(wm)))
 
 		for _, cw := range closedWindows {
-			partitions := cw.Partitions()
+			kw := cw.(*keyed.KeyedWindow)
```
vigith, line 186: why do we need this type conversion, can we avoid it? It is a little complex relationship 😄

pkg/reduce/readloop/readloop.go
```diff
@@ -241,16 +238,18 @@ func (rl *ReadLoop) upsertWindowsAndKeys(m *isb.ReadMessage) []*keyed.KeyedWindo
...
-		kw := rl.aw.GetKeyedWindow(win)
-		if kw == nil {
-			kw = rl.aw.CreateKeyedWindow(win)
-			rl.log.Debugw("Creating new keyed window", zap.Any("key", kw.Keys), zap.Int64("startTime", kw.Start.UnixMilli()), zap.Int64("endTime", kw.End.UnixMilli()))
+		w := rl.windower.GetWindow(win)
+		if w == nil {
+			w = rl.windower.CreateWindow(win)
 		}
 		// track the key to window relationship
+		kw := w.(*keyed.KeyedWindow)
```
vigith, line 249: looks like we always need to convert an AlignedWindow to KeyedWindow, is there a way to do this more naturally without type conversion?

pkg/window/keyed/keyed.go
```diff
@@ -19,10 +19,9 @@ limitations under the License.
 package keyed
 
 import (
-	"sync"
-
 	"github.com/numaproj/numaflow/pkg/pbq/partition"
 	"github.com/numaproj/numaflow/pkg/window"
+	"sync"
```
vigith, line 24: can you enable this setting? 
Screen Shot 2022-11-11 at 9 05 15 AM

pkg/window/strategy/fixed/fixed.go
```diff
@@ -23,34 +23,153 @@ limitations under the License.
...
+	// from the Head.
+	entries *list.List
+	lock    sync.RWMutex
 }
 
 var _ window.Windower = (*Fixed)(nil)
 
-// NewFixed returns a Fixed window.
+// NewFixed returns a Fixed windower.
 func NewFixed(length time.Duration) *Fixed {
```
vigith, line 58: As per the comment, you are returning `windower`; in that case, it should be`AlignedWindow`.

pkg/reduce/readloop/readloop.go
ashwinidulams: list is not typed and so any operations on node requires a cast.

pkg/reduce/readloop/readloop.go
ashwinidulams: we dont have key information in aligned window and hence the cast.

pkg/window/keyed/keyed.go
ashwinidulams: done.

pkg/window/strategy/fixed/fixed.go
ashwinidulams: done.

pkg/window/windower.go
ashwinidulams: done

pkg/window/strategy/fixed/fixed.go
ashwinidulams: done

vigith (approved):

Issue #343, opened 2022-11-14
Implement WAL with one segment that support read/write paths


xdevxy:
Implement WAL with one segment that support read/write paths.

PR #344, opened 2022-11-14
feat(wal): First pass to implement WAL and hook to PBQ store.

xdevxy:
Closes: #343

This PR has the basic implementation for WAL with one segment and hook to PBQ store. 

`Testing`: Unit tests.


pkg/pbq/store/wal/bootstrap.go
```diff
@@ -0,0 +1,198 @@
...
+	"os"
+	"time"
+
+	"github.com/numaproj/numaflow/pkg/isb"
+	"github.com/numaproj/numaflow/pkg/pbq/partition"
+)
+
+var location *time.Location
+
+func init() {
```
vigith, line 18: I think we should remove this. We just need to make sure the test data we provide is in UTC.

pkg/pbq/store/wal/bootstrap.go
```diff
@@ -0,0 +1,198 @@
...
+	var err error
+	location, err = time.LoadLocation("UTC")
+	if err != nil {
+		panic(fmt.Sprint("cannot load UTC", err))
+	}
+}
+
+// IsCorrupted checks whether the file is corrupt
+func (w *WAL) IsCorrupted() bool {
+	return w.corrupted
```
vigith, line 28: I looked through the code and I do not think we are setting the corrupted flag everywhere.

pkg/pbq/store/wal/bootstrap.go
```diff
@@ -0,0 +1,198 @@
...
+			}
+			return nil, false, err
+		}
+
+		w.rOffset += sizeRead
+		messages = append(messages, message)
+	}
+
+	if w.isEnd() {
+		w.wOffset = w.rOffset
```
vigith, line 134: this could potentially become a bug. If we start writing after the read during bootup has finished, we will wrongly assume that unless the write and read offsets match we have not finished reading.

pkg/pbq/store/wal/segment.go
```diff
@@ -0,0 +1,324 @@
...
+}
+
+// Write writes the message to the WAL. The format as follow is
+//
+//	+-------------------+----------------+-----------------+--------------+----------------+
+//	| watermark (int64) | offset (int64) | msg-len (int64) | CRC (unit32) | message []byte |
+//	+-------------------+----------------+-----------------+--------------+----------------+
+//
+// CRC will be used for detecting entry corruptions.
+func (w *WAL) Write(message *isb.ReadMessage) error {
```
vigith, line 277: I think this error handling might get tricky. let's assume that the caller retries the write again, and we might write inconsistent data to WAL because of the previous incomplete write. Perhaps we should not return any error as it might complicate error handling. Retry on error till we can succeed?

pkg/pbq/store/wal/bootstrap.go
xdevxy: It is set at `(w *WAL) Read(size int64)` when checksum doesn't match.

vigith: shouldn't it also be set if the header is bad, etc.?

pkg/pbq/store/wal/bootstrap.go
vigith: i think we can remove `location` altogether.

pkg/pbq/store/wal/bootstrap.go
vigith: yes, but I am suspecting there could be some corner case unless we might reissue a `Read` even if the read has been completed. Perhaps use a flag to return an error if such a read is attempted?

pkg/pbq/store/wal/bootstrap.go
vigith: it might be complicated because we read from disk and build time from epoch. let's keep location.

pkg/pbq/store/wal/discover.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	partitions := make([]partition.ID, 0)
+
+	for _, f := range files {
+		filePath := filepath.Join(dir, f.Name())
+		stat, err := os.Stat(filePath)
+		if err != nil {
+			return nil, err
+		}
+		fp, err := os.OpenFile(filePath, os.O_RDWR, stat.Mode())
```
vigith, line 43: why not use wal.OpenWAL?

pkg/pbq/store/wal/bootstrap.go
xdevxy: yeah I can have a check that when a Read start, it always start from bootstrap not any other cases.

pkg/pbq/store/wal/segment.go
xdevxy: Good point, so I updated it to only bump the `woffset` when `Write` is actually succeed for atomic write so we are safe for retry.

pkg/pbq/store/wal/bootstrap.go
```diff
@@ -0,0 +1,198 @@
...
+// IsCorrupted checks whether the file is corrupt
+func (w *WAL) IsCorrupted() bool {
+	return w.corrupted
+}
+
+// OpenWAL returns a WAL if present
+func OpenWAL(filePath string, o *options) (*WAL, error) {
+	stat, err := os.Stat(filePath)
+	// there could be other errors
+	if os.IsNotExist(err) {
```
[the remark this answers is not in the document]
xdevxy: Added the log, for now I think we should fail in case WAL is missing, as `OpenWAL` is only used when we bootstrap.

pkg/pbq/store/wal/bootstrap.go
```diff
@@ -0,0 +1,223 @@
...
+// IsCorrupted checks whether the file is corrupt
+func (w *WAL) IsCorrupted() bool {
+	return w.corrupted
+}
+
+// OpenWAL returns a WAL if present
+func OpenWAL(ctx context.Context, filePath string) (*WAL, error) {
+	stat, err := os.Stat(filePath)
+	// there could be other errors
+	if os.IsNotExist(err) {
```
whynowy, line 54: This is kind of weird to check `NotExist` instead `nil` first :)

pkg/pbq/store/wal/bootstrap.go
```diff
@@ -0,0 +1,223 @@
...
+		return nil, nil
+	} else if err != nil {
+		log := logging.FromContext(ctx).With("pbqStore", "FileSystem")
+		log.Errorw("Error when opening WAL file", zap.Any("file name", filePath))
```
whynowy, line 58: BTW - use single word `fileName` as the log key - it will be structured as JSON.

vigith:
@chromevoid, could you please do one dry-run to make sure we are seeing the expected behavior?

chromevoid:
Tested successfully in k8s cluster using the sample reduce pipeline: didn't see data loss, and the events were processed in the expected order.

vigith (approved):

PR #357, opened 2022-11-16
fix: context handling for persistence and ack

ashwinidulams:
context handling to ensure that all the read messages are successfully written and acked.

### DIFF
--- a/pkg/reduce/readloop/readloop.go
+++ b/pkg/reduce/readloop/readloop.go
@@ -115,7 +115,6 @@ func (rl *ReadLoop) Startup(ctx context.Context) {
 
 // Process is one iteration of the read loop.
 func (rl *ReadLoop) Process(ctx context.Context, messages []*isb.ReadMessage) {
-
 	// There is no Cap on backoff because setting a Cap will result in
 	// backoff stopped once the duration exceeds the Cap
 	var pbqWriteBackoff = wait.Backoff{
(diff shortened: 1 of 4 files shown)


ashwinidulams:
Thsi is the relevant log line that suggests that the window was closed. 

`2022-11-16T00:49:18.222Z	DEBUG	numaflow.ReduceUDF-processor	readloop/readloop.go:186	Closing Window	{"vertex": "even-odd-sum-compute-sum", "windowStart": "1970-01-01T01:42:00.000Z", "windowEnd": "1970-01-01T01:43:00.000Z"}`

pkg/reduce/readloop/readloop.go
```diff
@@ -125,67 +124,78 @@ func (rl *ReadLoop) Process(ctx context.Context, messages []*isb.ReadMessage) {
...
-				if rErr != nil {
-					rl.log.Errorw("Failed to write message", zap.Any("msgOffSet", m.ReadOffset.String()), zap.String("partitionID", partitionID.String()), zap.Any("attempt", attempt), zap.Error(rErr))
-					return false, nil
+				q := rl.associatePBQAndPnF(ctx, partitionID)
+
+				// write the message to PBQ
+				// we use a background context since these messages have to be written to pbq
+				cctx := context.Background()
+				attempt := 0
+				timeoutErr := wait.ExponentialBackoff(pbqWriteBackoff, func() (done bool, err error) {
```
whynowy, line 150: Is this going to be timeout shortly? The step number is set to `math.MaxInt`, this means, there's no way to shutdown until the pod is forced to terminate after grace shutdown period.

pkg/reduce/readloop/readloop.go
```diff
@@ -125,67 +124,78 @@ func (rl *ReadLoop) Process(ctx context.Context, messages []*isb.ReadMessage) {
...
-			if ctxClosedErr != nil {
-				rl.log.Errorw("Error while writing the message to PBQ", zap.Error(ctxClosedErr))
-				return
 			}
 
 			// Ack the message to ISB
-			attempt = 0
-			ctxClosedErr = wait.ExponentialBackoffWithContext(ctx, pbqWriteBackoff, func() (done bool, err error) {
+			attempt := 0
+			timeoutErr := wait.ExponentialBackoff(pbqWriteBackoff, func() (done bool, err error) {
```
whynowy, line 169: Same here.

pkg/reduce/reduce.go
```diff
@@ -82,19 +82,25 @@ func NewDataForward(ctx context.Context,
...
-			d.readloop.ShutDown(cctx)
 
 			if err := d.fromBuffer.Close(); err != nil {
 				d.log.Errorw("Failed to close buffer reader, shutdown anyways...", zap.Error(err))
 			} else {
 				d.log.Infow("Closed buffer reader", zap.String("bufferFrom", d.fromBuffer.GetName()))
 			}
+
+			// cancel the readloop context so that all the messages are persisted and acked.
+			rCancel()
```
whynowy, line 100: When `ctx` gets SIGTERM, the `rCtx` also has SIGTERM, which means, which means using explicit `rCancel()` is not needed.

PR #359, opened 2022-11-16
fix: add more logging and skip closed COB

vigith:
Signed-off-by: ashwinidulams <ashttk@gmail.com>

Explain what this PR does.


pkg/pbq/pbq.go
```diff
@@ -46,8 +47,8 @@ var _ ReadWriteCloser = (*PBQ)(nil)
 func (p *PBQ) Write(ctx context.Context, message *isb.ReadMessage) error {
 	// if cob we should return
 	if p.cob {
-		p.log.Errorw("Failed to write message to pbq, pbq is closed", zap.Any("ID", p.PartitionID), zap.Any("header", message.Header))
-		return ErrCOB
+		p.log.Errorw("Failed to write message to pbq, pbq is closed", zap.Any("ID", p.PartitionID), zap.Any("header", message.Header), zap.Any("message", message))
+		return nil
```
xdevxy, line 51: This is a bit confusing to me, are expect error in this case? If not, then we should not log a error?

vigith: this causes an infinite loop from which there is no return; hence we are logging and returning `nil`. This is happening because of out-of-order reads from ISB.

PR #360, opened 2022-11-16
feat: added batch sync for WAL

shashank10456:
added batch sync for WAL.

### DIFF
--- a/pkg/pbq/store/wal/segment.go
+++ b/pkg/pbq/store/wal/segment.go
@@ -26,6 +26,7 @@ import (
 	"hash/crc32"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/numaproj/numaflow/pkg/isb"
 	"github.com/numaproj/numaflow/pkg/pbq/partition"
@@ -53,6 +54,8 @@ type WAL struct {
 	wOffset int64
 	// prevSyncedWOffset is the write offset that is already synced as tracked by the writer
 	prevSyncedWOffset int64
+	// prevSyncedTime is the time when the last sync was made
+	prevSyncedTime time.Duration
 	// rOffset is the read offset as tracked when reading. Reading only
 	// happens during boostrap. It is highly unlikely to reread the data
 	// once boostrap sequence has been completed.
@@ -107,6 +110,7 @@ func openOrCreateWAL(id *partition.ID, opts *store.StoreOptions) (*WAL, error) {
 			openMode:          os.O_WRONLY,
 			wOffset:           0,
 			prevSyncedWOffset: 0,
+			prevSyncedTime:    0,
 			rOffset:           0,
 			readUpTo:          0,
 			partitionID:       id,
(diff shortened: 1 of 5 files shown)


pkg/pbq/store/wal/segment.go
```diff
@@ -124,13 +127,14 @@ func openOrCreateWAL(id *partition.ID, opts *store.StoreOptions) (*WAL, error) {
...
-			openMode:    os.O_RDWR,
-			wOffset:     0,
-			rOffset:     0,
-			readUpTo:    stat.Size(),
-			partitionID: id,
-			opts:        opts,
+			fp:                fp,
+			openMode:          os.O_RDWR,
+			wOffset:           0,
+			prevSyncedWOffset: 0,
```
xdevxy, line 133: You may need to set `prevSyncedWOffset` equal to `wOffset` when we finish bootstrap.

vigith:
can we add `age` also as a condition? else, low TPS systems will hardly sync

pkg/pbq/store/wal/segment.go
shashank10456: Fixed this. Thanks!

pkg/pbq/store/wal/segment.go
```diff
@@ -51,6 +52,10 @@ type WAL struct {
 	fp *os.File
 	// wOffset is the write offset as tracked by the writer
 	wOffset int64
+	// prevSyncedWOffset is the write offset that is already synced as tracked by the writer
+	prevSyncedWOffset int64
+	// prevSyncedTime is the time when the last sync was made
+	prevSyncedTime time.Duration
```
vigith, line 58: why is time duration? shouldn't it be `time.Time`? Because of this, you are doing a lot of conversions later on.

shashank10456: Yes. I will change this.

shashank10456: I did that because we want the duration to be more than some x nanoseconds before syncing again. So, I felt that duration might be better.  But, even i feel that I am converting a lot of times because of this. I will update this. Thanks.

xdevxy (commented):
LGTM, other than the comment Vigith mentioned. Thanks for contributing!

pkg/udf/reduce_udf.go
```diff
@@ -140,10 +140,10 @@ func (u *ReduceUDFProcessor) Start(ctx context.Context) error {
 	var pbqManager *pbq.Manager
 	// TODO: revisit to see if we need to use more storeOptions
 	if storage := u.VertexInstance.Vertex.Spec.UDF.GroupBy.Storage; storage != nil && storage.PersistentVolumeClaim != nil {
-		pbqManager, err = pbq.NewManager(ctx, pbq.WithPBQStoreOptions(store.WithPbqStoreType(dfv1.FileSystemType)))
+		pbqManager, err = pbq.NewManager(ctx, pbq.WithPBQStoreOptions(store.WithPbqStoreType(dfv1.FileSystemType), store.WithMaxBufferSize(dfv1.DefaultStoreMaxBufferSize), store.WithSyncDuration(dfv1.DefaultStoreSyncDuration)))
```
xdevxy, line 143: Can you rebase on top of the recent store provider change? Also, if it is default then better do it inside the specific store provider in this case for `FileSystemType`.

PR #372, opened 2022-11-19
fix: best effort processing during SIGTERM. Closes #371

### LINKED ISSUE #371: readloop ends up with partial acks during SIGTERM
Any error short circuits the forward loop and the ctx.Done() is also sometimes misinterpreted as as error. We should try to process all successful entries from the previous step in the subsequent steps.

vigith:
Fixes: #371 

Process all the successful entries returned by the previous step in the subsequent steps.

### DIFF
--- a/pkg/reduce/readloop/readloop.go
+++ b/pkg/reduce/readloop/readloop.go
@@ -115,10 +115,73 @@ func (rl *ReadLoop) Startup(ctx context.Context) error {
 	return nil
 }
 
-// Process is one iteration of the read loop.
+// Process is one iteration of the read loop which writes the messages to the PBQs followed by acking the messages, and
+// then closing the windows that can closed.
 func (rl *ReadLoop) Process(ctx context.Context, messages []*isb.ReadMessage) {
-	// There is no Cap on backoff because setting a Cap will result in
-	// backoff stopped once the duration exceeds the Cap
+	// write messages to windows based by PBQs.
+	successfullyWrittenMessages, err := rl.writeMessagesToWindows(ctx, messages)
+	if err != nil {
+		rl.log.Errorw("Failed to write messages", zap.Int("totalMessages", len(messages)), zap.Int("writtenMessage", len(successfullyWrittenMessages)))
+	}
+
+	// ack successful messages
+	err = rl.ackMessages(ctx, successfullyWrittenMessages)
+	if err != nil {
+		rl.log.Errorw("Failed to acks messages", zap.Error(err))
+	}
+
+	// for each successfully written message,
+	for _, m := range successfullyWrittenMessages {
+		// close any windows that need to be closed.
+		wm := processor.Watermark(m.Watermark)
+		closedWindows := rl.windower.RemoveWindows(time.Time(wm))
+		rl.log.Debugw("Windows eligible for closing", zap.Int("length", len(closedWindows)), zap.Time("watermark", time.Time(wm)))
+
+		for _, cw := range closedWindows {
+			partitions := cw.Partitions()
+			rl.closePartitions(partitions)
+			rl.log.Debugw("Closing Window", zap.Time("windowStart", cw.StartTime()), zap.Time("windowEnd", cw.EndTime()))
+		}
+	}
+}
+
+// writeMessagesToWindows write the messages to each window that message belongs to. Each window is backed by a PBQ.
+func (rl *ReadLoop) writeMessagesToWindows(ctx context.Context, messages []*isb.ReadMessage) ([]*isb.ReadMessage, error) {
+	var err error
+	var writtenMessages = make([]*isb.ReadMessage, 0, len(messages))
+
+messagesLoop:
+	for _, message := range messages {
+		// identify and add window for the message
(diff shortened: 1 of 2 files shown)


pkg/reduce/readloop/readloop.go
```diff
@@ -115,87 +115,156 @@ func (rl *ReadLoop) Startup(ctx context.Context) error {
...
+				rl.log.Errorw("Failed to ack message", zap.String("msgOffSet", m.ReadOffset.String()), zap.Error(rErr))
+				// no point retrying if ctx.Done has been invoked
+				select {
+				case <-ctx.Done():
+					// no point in retrying after we have been asked to stop.
+					return false, ctx.Err()
+				default:
+					// keep retrying
 					return false, nil
 				}
```
yhl25, line 254: This can be achieved by using ExponentialBackoffWithContext

pkg/reduce/readloop/readloop.go
```diff
@@ -115,87 +115,156 @@ func (rl *ReadLoop) Startup(ctx context.Context) error {
...
-				partitions := cw.Partitions()
-				rl.closePartitions(partitions)
-				rl.log.Debugw("Closing Window", zap.Time("windowStart", cw.StartTime()), zap.Time("windowEnd", cw.EndTime()))
-			}
+			return true, nil
+		})
+		// no point trying the rest of the message, most likely that will also fail
+		if err != nil {
+			rl.log.Errorw("Failed to ack message, exiting the loop", zap.Error(err))
+			break
```
yhl25, line 263: What if the writes to the PQB were successful and while acking those messages you received a ctx.done? Before you return you should ack all the messages in that batch right?

pkg/reduce/readloop/readloop.go
```diff
@@ -115,87 +115,156 @@ func (rl *ReadLoop) Startup(ctx context.Context) error {
...
 
-			// Ack the message to ISB
-			attempt := 0
-			timeoutErr := wait.ExponentialBackoff(pbqWriteBackoff, func() (done bool, err error) {
-				rErr := m.ReadOffset.AckIt()
-				attempt += 1
-				if rErr != nil {
-					rl.log.Errorw("Failed to ack message", zap.String("msgOffSet", m.ReadOffset.String()), zap.Int("attempt", attempt), zap.Error(rErr))
+// ackMessages acks messages. Retries until it can succeed or ctx.Done() happens.
+func (rl *ReadLoop) ackMessages(ctx context.Context, messages []*isb.ReadMessage) error {
```
ashwinidulams, line 232: does it make sense to combine this with the write operation? ack as soon as you write the message.

pkg/reduce/readloop/readloop.go
```diff
@@ -115,87 +115,156 @@ func (rl *ReadLoop) Startup(ctx context.Context) error {
...
 }
 
-// Process is one iteration of the read loop.
+// Process is one iteration of the read loop which writes the messages to the PBQs followed by acking the messages, and
+// then closing the windows that can closed.
 func (rl *ReadLoop) Process(ctx context.Context, messages []*isb.ReadMessage) {
-	// There is no Cap on backoff because setting a Cap will result in
-	// backoff stopped once the duration exceeds the Cap
+	// write messages to windows based by PBQs.
+	successfullyWrittenMessages, err := rl.writeMessagesToWindows(ctx, messages)
```
ashwinidulams, line 122: To guarantee acks, we need to handle the context Done event here in addition to write

vigith: why do we need to do that?

pkg/reduce/readloop/readloop.go
vigith: `ExponentialBackoffWithContext` will not work because it will return without executing anything if `ctx.Done()` has been closed. https://github.com/kubernetes/apimachinery/blob/v0.25.4/pkg/util/wait/wait.go#L734-L738

pkg/reduce/readloop/readloop.go
vigith: That should be fine, this `break` is to get out of the loop only if the `Ack` is failing. Otherwise, it will run to completion.

pkg/reduce/readloop/readloop.go
vigith: We check for `ctx.Done()` befeore every batch iteration in https://github.com/numaproj/numaflow/blob/a19f77c017ddd706d4e6c2e15d457e34757c604a/pkg/reduce/reduce.go#L89-L91

ashwinidulams: i guess thats fine. since we are cancelling it after we close the read, this should work.

ashwinidulams (approved):

pkg/reduce/readloop/readloop.go
yhl25: yeah makes sense

pkg/reduce/readloop/readloop.go
yhl25: okay

yhl25 (approved):

PR #377, opened 2022-11-21
unit tests for replay

yhl25:


pkg/reduce/readloop/readloop.go
```diff
@@ -102,6 +102,8 @@ func (rl *ReadLoop) Startup(ctx context.Context) error {
 		// crosses the window.
 
 		alignedKeyedWindow := keyed.NewKeyedWindow(p.Start, p.End)
+		// add key to the window
+		alignedKeyedWindow.AddKey(p.Key)
```
vigith, line 106: how was it working before?

vigith: @yhl25 please update the comment to explain why we are doing it

yhl25: Okayy

vigith:
Could you please add a unit test that invokes `ctx.Done()` path too?

pkg/reduce/readloop/readloop_test.go
```diff
@@ -0,0 +1,227 @@
...
+		},
+		Body: isb.Body{Payload: b},
+	}
+	return []*isb.Message{
+		ret,
+	}, nil
+}
+
+// mock store for test
+type TestStore struct {
```
vigith, line 61: I thought we were planning to use in-memory store to do this rather than mocking the interface. Is that still the plan?

yhl25: still the plan, I had already written this test with a mock store. I have changed it to use a memory store

pkg/reduce/reduce_test.go
```diff
@@ -672,6 +731,7 @@ func publishMessages(ctx context.Context, startTime int, messages []int, testDur
 					}
 					count = 0
 				}
+				time.Sleep(interval)
```
vigith, line 734: why do we have to sleep?

yhl25: `// publish messages with 1 millisecond delay
	go publishMessages(ctx, startTime, messages, 600, 10, p[fromBuffer.GetName()], fromBuffer, 1*time.Millisecond)
	// sleep for 100 milliseconds, so that few messages are processed and partitions are created
	time.Sleep(100 * time.Millisecond)
	cancel()`

since I am closing the context after 100 milliseconds, I have added sleep just to ensure context close happens while we are processing the data

yhl25: this sleep is only for context close tests, for other tests the sleep duration is 0

pkg/reduce/reduce_test.go
```diff
@@ -544,6 +544,65 @@ func TestReduceDataForward_SumWithDifferentKeys(t *testing.T) {
...
+
+	// start the forwarder
+	go reduceDataForward.Start(ctx)
+	// publish messages with 1 millisecond delay
+	go publishMessages(ctx, startTime, messages, 600, 10, p[fromBuffer.GetName()], fromBuffer, 1*time.Millisecond)
+	// sleep for 100 milliseconds, so that few messages are processed and partitions are created
+	time.Sleep(100 * time.Millisecond)
+	cancel()
+	// wait for a second for the messages to get persisted
+	time.Sleep(time.Second)
```
vigith, line 598: do we really need this sleep? i think we should put a `for` loop to check whether there are some messages

pkg/reduce/reduce_test.go
vigith: won't think become flaky?

pkg/reduce/reduce_test.go
yhl25: yes removed sleep

pkg/reduce/reduce_test.go
yhl25: removed sleep

vigith (approved):

PR #379, opened 2022-11-21
chore: make tickgen ID unique

vigith:

Since tickgen ID is based on time, during load tests, we are seeing duplicates

### DIFF
--- a/pkg/sources/generator/tickgen.go
+++ b/pkg/sources/generator/tickgen.go
@@ -233,7 +233,7 @@ loop:
 		select {
 		case r := <-mg.srcchan:
 			tickgenSourceReadCount.With(map[string]string{metricspkg.LabelVertex: mg.name, metricspkg.LabelPipeline: mg.pipelineName}).Inc()
-			msgs = append(msgs, newreadmessage(r.data, r.offset))
+			msgs = append(msgs, mg.newreadmessage(r.data, r.offset))
 		case <-timeout:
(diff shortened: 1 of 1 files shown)


pkg/sources/generator/tickgen.go
```diff
@@ -323,12 +323,12 @@ func (mg *memgen) generator(ctx context.Context, rate int, timeunit time.Duratio
...
 }
 
-func newreadmessage(payload []byte, offset int64) *isb.ReadMessage {
+func (mg *memgen) newreadmessage(payload []byte, offset int64) *isb.ReadMessage {
 	msg := isb.Message{
 		Header: isb.Header{
 			// TODO: insert the right time based on the generator
 			PaneInfo: isb.PaneInfo{EventTime: timefromNanos(parseTime(payload))},
-			ID:       strconv.FormatInt(offset, 10),
+			ID:       strconv.FormatInt(offset, 10) + "-" + strconv.FormatInt(int64(mg.vertexInstance.Replica), 10),
```
whynowy, line 331: use `fmt.Sprintf()`?

vigith: generally, `fmt.Sprintf()` is slower than `strconv.XXX` family, since this is `tickgen` which is purely for load testing we could go with the faster method even though the readability is compromised?

whynowy (approved):

PR #381, opened 2022-11-22
feat: add Grafana instruction and a dashboard template. Closes #287

### LINKED ISSUE #287: grafana dashboard for numaflow pipelines
# Summary

We have exposed an extensive and detailed set of metrics. We should provide grafana dashboard template and steps to install grafana to visualize the metrics.

This will improve the operability of the pipelines.

KeranYang:
Closes #287 
* Add to numaflow web page a section to introduce how to use Grafana to setup dashboards for numaflow.
* Provide a dashboard template which configures basic metrics including throughput, latency, error and buffer related metrics. User can import the template and filter on pipeline/vertex/buffer to customize the dashboard.

Explain what this PR does.


whynowy:
1. Does the `Throughput` mean `tps`? 
2. Why are there particular `UDF Inbound/outbound messages`?
3. Is `UDF processing` latency for the user defined functions and `ForwardAChunk` for an e2e? - do not use `ForwardAChunk`, which is not understood by the users.
4. Can we group the metrics by [4 Golden Signals](https://numaflow.numaproj.io/operations/metrics/metrics/#golden-signals)?

docs/operations/example-dashboard-template.json
```diff
@@ -0,0 +1,1895 @@
+{
```
xdevxy, line 1: It seems this example dashboard doesn't include the WAL metrics you just added?

KeranYang: Right. This example dashboard is for numaflow users to have a general view of the status of their pipelines. I didn't include WAL metrics in this specific dashboard because I consider WAL as implementation details of map-reduce, which don't necessarily need to be known by end users.

In the future , we can have a separate dashboard for numaflow developers which includes WAL metrics, hence I included the WAL dashboard template in the other PR for future reference.

xdevxy: Yeah, definitely not a blocker for this PR. Thanks!

xdevxy (approved):
Thanks a lot for doing this! LGTM.

whynowy (approved):

docs/operations/example-dashboard-template.json
vigith: WAL metrics can be merged with the reduce metrics once we have those available.

PR #385, opened 2022-11-24
chore: pass vertex/pipeline name to pbq wal and add a grafana dashboard template for wal metrics. Closes #365

### LINKED ISSUE #365: Add pipeline/vertex level labels to PBQ WAL metrics
# Summary

Currently PBQ WAL metrics are at global level. For future cases, we can definitely have multiple map reduce pipelines/vertices and we would like to view WAL metrics separately, hence we need to support labeling metrics at pipeline/vertex level.

---

KeranYang:
* Pass vertex/pipeline name to pbq wal and introduce pipeline and vertex label to WAL metrics.
* Add a grafana dashboard template for visualizing WAL metrics.

WAL Dashboard screenshot:

wal-dashboard-screenshot-11-30

Closes #365


pkg/pbq/store/wal/segment_test.go
```diff
@@ -90,39 +95,21 @@ func Test_encodeDecodeHeader(t *testing.T) {
...
 				return
 			}
 			result, err := decodeHeader(got)
 			assert.NoError(t, err)
 			assert.Equalf(t, tt.id, result, "encodeHeader(%v)", tt.id)
-		})
-	}
-}
-
-func Test_runeEncoding(t *testing.T) {
```
KeranYang, line 104: Deleting this test case as I don't think we need to test functionality of "encoding/binary" library.

pkg/pbq/store/wal/stores.go
```diff
@@ -35,13 +36,17 @@ type walStores struct {
 	maxBatchSize int64
 	// syncDuration timeout to sync to store
 	syncDuration time.Duration
+	pipelineName string
+	vertexName   string
 }
 
-func NewWALStores(opts ...Option) store.StoreProvider {
+func NewWALStores(pipelineName string, vertexName string, opts ...Option) store.StoreProvider {
```
vigith, line 43: Pass VertexInstance? That way we will get more info including the replica number.

vigith:
> 
> @vigith Could you elaborate more on why? My understanding is replica in this context is conceptually equal to pod, and WAL is at vertex level instead of pod level.

Pod might be recreated with different IDs due to pod migrations, but it will reuse the same WAL. We can track the WAL's metrics over time if we have a replica label.

pkg/reduce/reduce.go
```diff
@@ -51,13 +53,15 @@ type DataForward struct {
 
 func NewDataForward(ctx context.Context,
 	udf applier.ReduceApplier,
+	vertex *dfv1.Vertex,
```
ashwinidulams, line 56: can we be more explicit about what we are passing to the constructor instead of passing the Vertex object?

KeranYang: I was referring to https://github.com/numaproj/numaflow/blob/main/pkg/isb/forward/forward.go#L63 I added `vertexName` and `pipelineName` to `DataForward` such that we can use them as labels when emitting reducer metrics.

xdevxy (approved):

pkg/reduce/reduce.go
```diff
@@ -35,10 +34,13 @@ import (
 	"github.com/numaproj/numaflow/pkg/udf/applier"
 	"github.com/numaproj/numaflow/pkg/watermark/fetch"
 	"github.com/numaproj/numaflow/pkg/watermark/publish"
+	"github.com/numaproj/numaflow/pkg/window"
 )
 
 // DataForward reads data from isb and forwards them to readloop
 type DataForward struct {
+	vertexName          string
```
whynowy, line 42: Are these used anywhere?

vigith, line 42: how about we just keep `*dfv1.Vertex` and derive the rest at the usage site?

vigith: it will be required when they add reduce metrics. But I think we need the replica number too. Perhaps make more sense to keep the reference of `*dfv1.Vertex`.

KeranYang: sure will do.

pkg/pbq/store/wal/stores.go
```diff
@@ -35,13 +37,19 @@ type walStores struct {
 	maxBatchSize int64
 	// syncDuration timeout to sync to store
 	syncDuration time.Duration
+	pipelineName string
+	vertexName   string
+	replicaIndex int32
 }
 
-func NewWALStores(opts ...Option) store.StoreProvider {
+func NewWALStores(vertexInstance *dfv1.VertexInstance, opts ...Option) store.StoreProvider {
```
whynowy, line 45: If it's only for metrics purpose, is it good to only pass a mapped labels?

whynowy: No strong opinion though.

KeranYang: That's a good point. I tend to think it makes sense that when we create a WAL store, we specify under which vertex instance we are creating it. It's mostly used for emitting metrics atm but I guess could be other usages in the future.

vigith: Let's keep it as-is for consistency? I think in other places too we are passing `*dfv1.VertexInstance` instead of a map.

pkg/reduce/reduce.go
KeranYang: Done.

vigith (approved):

whynowy (approved):

PR #388, opened 2022-11-28
chore: switch to sync publish mode for jetstream isb

whynowy:

Seeing issue with async publish with errors like below, switching to sync publish by default, and keep the async mode achievable by adding an env variable.

```
"vertex":"xxxxxx","errors":{"nats: stalled with too many outstanding async published messages":500
```

### DIFF
--- a/pkg/isb/stores/jetstream/writer.go
+++ b/pkg/isb/stores/jetstream/writer.go
@@ -30,6 +30,7 @@ import (
 	jsclient "github.com/numaproj/numaflow/pkg/shared/clients/jetstream"
 	"github.com/numaproj/numaflow/pkg/shared/logging"
 	sharedqueue "github.com/numaproj/numaflow/pkg/shared/queue"
+	sharedutil "github.com/numaproj/numaflow/pkg/shared/util"
 )
 
 type jetStreamWriter struct {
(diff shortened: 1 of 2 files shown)


pkg/isb/stores/jetstream/writer.go
```diff
@@ -252,8 +260,34 @@ func (jw *jetStreamWriter) Write(_ context.Context, messages []isb.Message) ([]i
...
+	for index, msg := range messages {
+		wg.Add(1)
+		go func(message isb.Message, idx int) {
+			defer wg.Done()
+			m := &nats.Msg{
+				Header:  convert2NatsMsgHeader(message.Header),
+				Subject: jw.subject,
+				Data:    message.Payload,
+			}
+			if pubAck, err := jw.js.PublishMsg(m, nats.MsgId(message.Header.ID)); err != nil { // nats.MsgId() is for exactly-once writing
```
vigith, line 280: should we do a timeout to avoid a hanging ack?

vigith (approved):

pkg/isb/stores/jetstream/writer.go
whynowy: added `AckWait`

vigith (approved):

PR #390, opened 2022-11-29
chore: adding DeleteStore method to store provider interface

yhl25:

- removing GC from the Store interface
- adding DeleteStore to StoreProvider
- tests for StoreProvider

### DIFF
--- a/pkg/pbq/store/memory/stores.go
+++ b/pkg/pbq/store/memory/stores.go
@@ -47,6 +47,8 @@ func NewMemoryStores(opts ...Option) store.StoreProvider {
 }
 
 func (ms *memoryStores) CreateStore(ctx context.Context, partitionID partition.ID) (store.Store, error) {
+	ms.Lock()
+	defer ms.Unlock()
 	if memStore, ok := ms.partitions[partitionID]; ok {
 		return memStore, nil
 	}
@@ -59,8 +61,6 @@ func (ms *memoryStores) CreateStore(ctx context.Context, partitionID partition.I
 		log:         logging.FromContext(ctx).With("pbqStore", "Memory").With("partitionID", partitionID),
 		partitionID: partitionID,
 	}
-	ms.Lock()
-	defer ms.Unlock()
 	ms.partitions[partitionID] = memStore
 	return memStore, nil
 }
--- a/pkg/pbq/store/memory/stores.go
+++ b/pkg/pbq/store/memory/stores.go
@@ -67,7 +67,6 @@ func (ms *memoryStores) CreateStore(ctx context.Context, partitionID partition.I
 
 func (ms *memoryStores) DiscoverPartitions(ctx context.Context) ([]partition.ID, error) {
 	if ms.discoverFunc == nil {
-		println("hey")
 		partitionsIds := make([]partition.ID, 0)
 		for key := range ms.partitions {
 			partitionsIds = append(partitionsIds, key)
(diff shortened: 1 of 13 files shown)


pkg/pbq/store/memory/stores.go
```diff
@@ -28,11 +29,21 @@ import (
...
 }
 
 func NewMemoryStores(opts ...Option) store.StoreProvider {
 	s := &memoryStores{
-		storeSize: 100,
+		storeSize:  100,
+		partitions: make(map[partition.ID]*memoryStore),
+	}
+	// default discover function
+	s.discoverFunc = func(ctx context.Context) ([]partition.ID, error) {
```
whynowy, line 41: Move it to `DiscoverPartitions`?

pkg/pbq/store/memory/stores.go
```diff
@@ -60,3 +71,15 @@ func (ms *memoryStores) DiscoverPartitions(ctx context.Context) ([]partition.ID,
...
+
+func (ms *memoryStores) DeleteStore(partitionID partition.ID) error {
+	memStore, ok := ms.partitions[partitionID]
+	if !ok {
+		return errors.New("store not found")
+	}
+
+	memStore.storage = nil
+	memStore.writePos = -1
+	delete(ms.partitions, partitionID)
```
whynowy, line 83: Does it need a lock?

yhl25: No, I don't think so, because the store will be associated with only one partition

yhl25: oh you mean the map, yeah it needs a lock

pkg/pbq/store/memory/stores.go
```diff
@@ -63,11 +59,34 @@ func (ms *memoryStores) CreateStore(ctx context.Context, partitionID partition.I
...
+	ms.Lock()
+	defer ms.Unlock()
 	ms.partitions[partitionID] = memStore
-
 	return memStore, nil
 }
 
 func (ms *memoryStores) DiscoverPartitions(ctx context.Context) ([]partition.ID, error) {
+	if ms.discoverFunc == nil {
+		println("hey")
```
whynowy, line 70: `hey`

yhl25: removed :)

vigith (approved):
LGTM

pkg/pbq/store/wal/metrics.go
```diff
@@ -66,13 +66,6 @@ var entryWriteTime = promauto.NewHistogramVec(prometheus.HistogramOpts{
 	Buckets:   prometheus.ExponentialBucketsRange(1, 60, 5),
 }, []string{labelPartitionKey})
 
-var lifespan = promauto.NewHistogramVec(prometheus.HistogramOpts{
```
KeranYang, line 69: @yhl25 should we keep the lifespan metric? I think it can be a useful metric when we want to ensure not WAL file lives forever.

yhl25: we can check that using the GC error metric right?

yhl25: if there are errors, that means the file was not deleted

KeranYang: There could be cases when GC gets called very late with no errors, causing extremely long lifespan.

KeranYang: I think it should be a very rare case. Might not worth monitoring. We can remove lifespan metric for now. +1

PR #393, opened 2022-11-30
fix: e2e testing for PBQ WAL with reduce pipeline

### LINKED ISSUE #392: e2e testing for PBQ WAL with reducer pipeline

xdevxy:
Signed-off-by: Hao Hao <xdevxhao@gmail.com>

Add e2e test for PBQ WAL failover with a simple reduce pipeline. Closes #392 .


test/reducer-e2e/reducer_test.go
```diff
@@ -0,0 +1,77 @@
...
+					Status(204)
+			}
+		}
+
+		w.Expect().VertexPodLogContains("sink", "Payload -  120")
+		w.Expect().VertexPodLogContains("sink", "Payload -  240")
+	}()
+
+	// Kill one of the pod during processing to trigger failover.
+	block := func(t *testing.T, output string, err error) {
```
vigith, line 63: should we initiate a kill at a very specific point eg, at `i=x && j=y`? this will ensure that we do not kill at random points which could end up being flaky.

vigith, line 63: I am also thinking about whether this should be a utility function that can be used in other tests too.

xdevxy: This is to mimic that failure (or pod migration) can happen at any moment during pipeline execution and still get the correct result. If there is any flakiness due to this, I think we should investigate and check what is the cause.

xdevxy: Yeah, I guess only thing may be depends on the scenarios, one may want to take different actions towards the value of `output` or `err`

vigith: such random testing shouldn't be done in e2e tests. E2E tests should be predictable. Such random tests can be labeled, perhaps under Fuzz tests or something.

There are three major of pod migration scenarios
* right before processing
* during processing
* right after processing

vigith: such random testing shouldn't be done in e2e tests. E2E tests should be predictable. Such random tests can be labeled, perhaps under Fuzz tests or something.

There are three major of pod migration scenarios
* right before processing
* during processing
* right after processing

vigith: are we interpreting the result of `output` or `err` on pod kill?

xdevxy: Do you want me to rename the the test or create a new folder for fuzz tests and put it under it? I can add the other two scenarios, for `right before processing` and `right after processing` I didn't add previously as the behavior shouldn't differ with or without WAL implementation.

vigith: Let's not do fuzz testing now. Instead, let's make this E2E test complete by adding the 3 scenarios I have mentioned.

vigith: Yes, that is the expected behavior of `deleted` right? Hence I am suggesting we make it reusable. I can't think of any other checks, do you envision any?

whynowy: @xdevxy - let's do following:

1. Create the pipeline, expect both reduce pods running;
2. Kill them both;
3. Start ingesting messages, pause when half is done;
4. Expect both reduce pods running, and kill them again;
5. Finish ingesting;
6. Expect the results,
7. Kill them again, and no result changes expected.

test/reduce-e2e/reduce_test.go
```diff
@@ -0,0 +1,84 @@
...
+		VertexPodsRunning().DaemonPodsRunning().
+		VertexPodLogContains("in", LogSourceVertexStarted).
+		VertexPodLogContains("atoi", LogUDFVertexStarted, PodLogCheckOptionWithContainer("numa")).
+		VertexPodLogContains("sink", LogSinkVertexStarted).
+		DaemonPodLogContains(pipelineName, LogDaemonStarted)
+
+	defer w.VertexPodPortForward("in", 8443, dfv1.VertexHTTPSPort).
+		TerminateAllPodPortForwards()
+
+	args := "kubectl delete po `kubectl -n numaflow-system get po " +
```
whynowy: Is this used to kill both? If that the case, use following.

```shell
kubectl -n numaflow-system delete po -l numaflow.numaproj.io/pipeline-name=even-odd-sum,numaflow.numaproj.io/vertex-name=compute-sum
```

test/fixtures/util.go
```diff
@@ -48,6 +48,11 @@ var OutputRegexp = func(rx string) func(t *testing.T, output string, err error)
 	}
 }
 
+var CheckPodKillSucceed = func(t *testing.T, output string, err error) {
```
whynowy, line 51: `Succeeded`

test/reduce-e2e/reduce_test.go
```diff
@@ -0,0 +1,84 @@
...
+		"-lnumaflow.numaproj.io/pipeline-name=even-odd-sum,numaflow.numaproj.io/vertex-name=compute-sum | " +
+		"grep -v NAME | head -2 | awk '{print $1}'`"
+
+	// Kill one of the pod before processing to trigger failover.
```
whynowy, line 41: This is killing both, not one, right?

test/reduce-e2e/reduce_test.go
```diff
@@ -0,0 +1,78 @@
...
+	args := "kubectl delete po `kubectl -n numaflow-system get po " +
+		"-lnumaflow.numaproj.io/pipeline-name=even-odd-sum,numaflow.numaproj.io/vertex-name=compute-sum | " +
+		"grep -v NAME | head -1 | awk '{print $1}'`"
+
+	// Kill one of the pod before processing to trigger failover.
+	w.Exec("/bin/sh", []string{"-c", args}, CheckPodKillSucceed)
+
+	startTime := int(time.Unix(1000, 0).UnixMilli())
+	for i := 1; i <= 100; i++ {
+		for j := 0; j < 1; j++ {
```
whynowy, line 46: What is the for loop for?

test/reduce-e2e/reduce_test.go
```diff
@@ -0,0 +1,78 @@
...
+				Expect().
+				Status(204)
+		}
+	}
+
+	w.Expect().VertexPodLogContains("sink", "Payload -  120")
+	w.Expect().VertexPodLogContains("sink", "Payload -  240")
+
+	// Kill one of the pod after processing to trigger failover.
+	w.Exec("/bin/sh", []string{"-c", args}, CheckPodKillSucceed)
```
whynowy, line 71: Add a new pods running check, this is to make sure there's no result change (e.g., duplicate) after they start.

test/reduce-e2e/reduce_test.go
```diff
@@ -0,0 +1,78 @@
...
+	w.Expect().VertexPodLogContains("sink", "Payload -  120")
```
whynowy, line 72: Use `PodLogCheckOptionWithCount` option.

test/reduce-e2e/reduce_test.go
```diff
@@ -0,0 +1,78 @@
...
+	w.Expect().
+		VertexPodLogContains("in", LogSourceVertexStarted).
+		VertexPodLogContains("atoi", LogUDFVertexStarted, PodLogCheckOptionWithContainer("numa")).
+		VertexPodLogContains("sink", LogSinkVertexStarted).
+		DaemonPodLogContains(pipelineName, LogDaemonStarted)
+
+	defer w.VertexPodPortForward("in", 8443, dfv1.VertexHTTPSPort).
+		TerminateAllPodPortForwards()
+
+	args := "kubectl delete po -n numaflow-system `kubectl -n numaflow-system get po " +
```
whynowy, line 37: Is this used to kill both pods? If that's the case, use 

```shell
kubectl -n numaflow-system delete po -l numaflow.numaproj.io/pipeline-name=even-odd-sum,numaflow.numaproj.io/vertex-name=compute-sum
```

test/reduce-e2e/reduce_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	defer w.VertexPodPortForward("in", 8443, dfv1.VertexHTTPSPort).
+		TerminateAllPodPortForwards()
+
+	args := "kubectl delete po -n numaflow-system -l " +
+		"numaflow.numaproj.io/pipeline-name=even-odd-sum,numaflow.numaproj.io/vertex-name=compute-sum"
+
+	// Kill the reducer pods before processing to trigger failover.
+	w.Exec("/bin/sh", []string{"-c", args}, CheckPodKillSucceeded)
+	w.Expect().VertexPodsRunning()
```
whynowy, line 58: This line could be removed - we would expect after killing 2 reduce pods, the ingested data would still be picked up later one when they came back.

whynowy (commented):
All the reset looks good to me.

test/reduce-e2e/reduce_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	startTime := int(time.Unix(1000, 0).UnixMilli())
+	for i := 1; i <= 100; i++ {
+		eventTime := startTime + (i * 1000)
+		if i == 5 {
+			// Kill the reducer pods during processing to trigger failover.
```
whynowy, line 64: Move the line 58 here.

whynowy (approved):

PR #402, opened 2022-12-05
fix: getWatermark to  return-1 if any processor returns -1 

yhl25:

while fetching watermark return `-1`, if the offset timeline store returns `-1` for any processor.

### DIFF
--- a/pkg/watermark/fetch/edge_fetcher.go
+++ b/pkg/watermark/fetch/edge_fetcher.go
@@ -99,8 +99,9 @@ func (e *edgeFetcher) GetWatermark(inputOffset isb.Offset) processor.Watermark {
 	for _, p := range allProcessors {
 		debugString.WriteString(fmt.Sprintf("[Processor: %v] \n", p))
 		var t = p.offsetTimeline.GetEventTime(inputOffset)
-		if t == -1 { // this is a bug
+		if t == -1 { // watermark cannot be computed, perhaps a new processing unit was added or offset fell off the timeline
 			epoch = t
+			break
 		} else if t < epoch {
 			epoch = t
 		}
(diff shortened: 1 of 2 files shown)


vigith:
`TestFetcherWithSameOTBucket` is failing

vigith:
@whynowy, how do we handle this in `Source` watermark fetcher?

pkg/watermark/fetch/edge_fetcher.go
```diff
@@ -99,7 +99,10 @@ func (e *edgeFetcher) GetWatermark(inputOffset isb.Offset) processor.Watermark {
 	for _, p := range allProcessors {
 		debugString.WriteString(fmt.Sprintf("[Processor: %v] \n", p))
 		var t = p.offsetTimeline.GetEventTime(inputOffset)
-		if t != -1 && t < epoch {
+		if t == -1 { // watermark cannot be computed, perhaps a new processing unit was added or offset fell off the timeline
+			epoch = t
+			break
```
chromevoid, line 104: > return -1 for watermark if we don't find an entry in the offset timeline store in any processor

The PR description doesn't match the logic here.

For example, if we have processors A B C, if processor A's watermark for inputOffset is -1, then we break the for loop and skip processors B and C. 
The description above means if all A B C returns -1 then we return -1.

chromevoid: We already have the logic here 
https://github.com/numaproj/numaflow/pull/402/files#diff-6af877f3295078204591f48a2180bc855394b50997f764275b25ef2ffff678daL110-R116
to return -1 if no entry found in any offset timeline.

yhl25: updated the description

vigith: But we need to set the overall watermark to -1 if watermark cannot be fetched for any processing unit

chromevoid: > https://github.com/numaproj/numaflow/pull/402/files#diff-6af877f3295078204591f48a2180bc855394b50997f764275b25ef2ffff678daL110-R116

👌 

do we still need this logic then?https://github.com/numaproj/numaflow/blob/68631426b61b85fe72ca0571648dde4917d33a2e/pkg/watermark/fetch/edge_fetcher.go#L114-L116

yhl25: we don't need it

yhl25: I think we need it, if there are no processors, we will return int max

vigith (approved):

PR #409, opened 2022-12-06
chore: Windower interface refactor

yhl25:

### DIFF
--- a/pkg/reduce/readloop/readloop.go
+++ b/pkg/reduce/readloop/readloop.go
@@ -108,7 +108,7 @@ func (rl *ReadLoop) Startup(ctx context.Context) error {
 		alignedKeyedWindow.AddKey(p.Key)
 
 		// These windows have to be recreated as they are completely in-memory
-		rl.windower.CreateWindow(alignedKeyedWindow)
+		rl.windower.InsertWindow(alignedKeyedWindow)
 
 		// create and invoke process and forward for the partition
 		rl.associatePBQAndPnF(ctx, p)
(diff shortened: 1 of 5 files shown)


yhl25:
@ashwinidulams @vigith  I was also thinking should we merge the insert and get method to one method like (InsertIfNotPresent) to avoid iterating over the list twice

vigith:
> @ashwinidulams @vigith I was also thinking should we merge the insert and get method to one method like (InsertIfNotPresent) to avoid iterating over the list twice

That makes sense; let's try that?

yhl25:
> > @ashwinidulams @vigith I was also thinking should we merge the insert and get method to one method like (InsertIfNotPresent) to avoid iterating over the list twice
> 
> That makes sense; let's try that?

done

pkg/window/strategy/fixed/fixed.go
```diff
@@ -77,74 +77,50 @@ func (f *Fixed) AssignWindow(eventTime time.Time) []window.AlignedKeyedWindower
...
 		f.entries.PushFront(kw)
+		aw = kw
 	} else if !recentWindow.EndTime().After(kw.StartTime()) {
 		// early arrival
 		f.entries.PushBack(kw)
+		aw = kw
 	} else {
 		// a window in the middle
-		for e := f.entries.Back(); e != nil; e = e.Prev() {
+		for e := f.entries.Front(); e.Next() != nil; e = e.Next() {
```
vigith, line 108: Why are we reversing the order?

yhl25: we don't have to reverse the order, will change it

yhl25: done

pkg/reduce/readloop/readloop.go
```diff
@@ -103,12 +103,13 @@ func (rl *ReadLoop) Startup(ctx context.Context) error {
...
+
+		// insert the window to the list of active windows, since the active window list is in-memory
+		keyedWindow, _ := rl.windower.InsertIfNotPresent(alignedKeyedWindow)
+
 		// add key to the window, so that when a new message with the watermark greater than
 		// the window end time comes, key will not be lost and the windows will be closed as expected
-		alignedKeyedWindow.AddKey(p.Key)
-
-		// These windows have to be recreated as they are completely in-memory
-		rl.windower.CreateWindow(alignedKeyedWindow)
```
yhl25, line 111: this was a bug, while replaying partitions we used to create a new window for each partition
for ex: if we have `60-120-even` and `60-120-odd` partitions to be replayed, we used to create two separate windows 
but it should be one window(60-120) with two keys(even, odd)

I have updated the replay test to capture this case

vigith (approved):

PR #411, opened 2022-12-06
feat: nats as source

### LINKED ISSUE #355: NATS Source
# Summary

It would be nice to not only have Kafka and Webhooks as streaming sources, but also NATS. JetStream is already used as a buffer between the vertices, so why not add it as a source?

# Use Cases

We'd like to use Numaflow in combination with our existing NATS engine without having to deploy Kafka and a Kafka-NATS bridge.

---

whynowy:
Fixes #355 


docs/user-guide/autoscaling.md
```diff
@@ -110,4 +118,12 @@ spec:
 
 ## Vertical Pod Autoscaling
 
-TBD.
+`Vertical Pod Autoscaling` can be achieved by setting the `targetRed` to `Vertex` objects as following.
```
vigith, line 121:
```suggestion
`Vertical Pod Autoscaling` can be achieved by setting the `targetRef` to `Vertex` objects as following.
```

pkg/sources/nats/nats.go
```diff
@@ -0,0 +1,265 @@
...
+	n.sourcePublishWM = publish.NewPublish(ctx, processorEntity, publishWMStores, publish.IsSource(), publish.WithDelay(vertexInstance.Vertex.Spec.Watermark.GetMaxDelay()))
+
+	source := vertexInstance.Vertex.Spec.Source.Nats
+	opt := []natslib.Option{
+		natslib.MaxReconnects(-1),
+		natslib.ReconnectWait(3 * time.Second),
+		natslib.DisconnectHandler(func(c *natslib.Conn) {
+			n.logger.Info("Nats disconnected")
+		}),
+		natslib.ReconnectHandler(func(c *natslib.Conn) {
```
vigith, line 106: we are using the default reconnector from nats.go, right?

pkg/sources/nats/nats_test.go
```diff
@@ -0,0 +1,149 @@
...
+	fetchWatermark, publishWatermark := generic.BuildNoOpWatermarkProgressorsFromBufferMap(map[string]isb.BufferWriter{})
+	return New(vi, dest, fetchWatermark, publishWatermark, publishWMStores, WithReadTimeout(1*time.Second))
+}
+
+func runNatsServer(t *testing.T) *server.Server {
+	t.Helper()
+	opts := natstestserver.DefaultTestOptions
+	opts.Port = 9989
+	opts.Cluster.Name = "testing"
+	return natstestserver.RunServer(&opts)
```
vigith, line 75: this is cool, do they have a JS version too?

vigith (approved):

pkg/sources/nats/nats.go
whynowy: Yes. Our auto reconnection only works for JetStream.

pkg/sources/nats/nats_test.go
whynowy: Same thing can be used from JS.

PR #425, opened 2022-12-14
chore: use resources.limits for cpu/mem percentage calculation

whynowy:

Currently cpu/mem usage is calculated baed on the `resources.requests.cpu/mem`, which is not accurate. Changing to use `resources.limits.cpu/mem`.


vigith:
I thought this was intended since the IKS team prefers us to give the right size via `requests` and give some headroom via `limits`.

whynowy:
> I thought this was intended since the IKS team prefers us to give the right size via `requests` and give some headroom via `limits`.

Okay....

PR #426, opened 2022-12-15
chore(doc): scope UDF under a dir

vigith:

Before we start adding documents on reduce, we need to move the UDF to under a director for better scoping and namespacing.


vigith:
@whynowy should we create a `map` and `reduce` sub-dir in the doc?

whynowy:
> @whynowy should we create a `map` and `reduce` sub-dir in the doc?

I think so.

vigith:
> I think so.

@whynowy done

whynowy (approved):

PR #432, opened 2022-12-19
chore: flaky reduce e2e tests

yhl25:

### DIFF
--- a/test/reduce-e2e/reduce_test.go
+++ b/test/reduce-e2e/reduce_test.go
@@ -54,7 +54,7 @@ func (r *ReduceSuite) TestSimpleKeyedReducePipeline() {
 	go func() {
 		// publish messages to source vertex, with event time starting from 60000
 		startTime := 60000
-		for i := 0; i < 300; i++ {
+		for i := 0; true; i++ {
 			select {
 			case <-ctx.Done():
 				return
--- a/test/reduce-e2e/reduce_test.go
+++ b/test/reduce-e2e/reduce_test.go
@@ -18,6 +18,7 @@ limitations under the License.
 package reduce_e2e
 
 import (
+	"context"
 	"strconv"
 	"testing"
 	"time"
@@ -33,6 +34,8 @@ type ReduceSuite struct {
 
 // one reduce vertex (keyed)
 func (r *ReduceSuite) TestSimpleKeyedReducePipeline() {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+	defer cancel()
 	w := r.Given().Pipeline("@testdata/simple-keyed-reduce-pipeline.yaml").
 		When().
 		CreatePipelineAndWait()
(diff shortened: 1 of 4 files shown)


ashwinidulams:
what was the reason for flakyness?

yhl25:
> what was the reason for flakyness?

we used to publish 300 messages and check the sink to see if the expected output is present instead, we should keep on publishing messages until we see the output in the sink.

test/reduce-e2e/reduce_test.go
```diff
@@ -84,30 +100,42 @@ func (r *ReduceSuite) TestSimpleNonKeyedReducePipeline() {
...
-			Expect().
-			Status(204)
-	}
+	done := make(chan struct{})
+	go func() {
+		// publish messages to source vertex, with event time starting from 60000
+		startTime := 60000
+		for i := 0; true; i++ {
+			select {
+			case <-ctx.Done():
```
vigith, line 109: if this `context.Done` happens, it means the test has to be marked as a failure right?

vigith (approved):

test/reduce-e2e/reduce_test.go
```diff
@@ -218,22 +273,35 @@ func (r *ReduceSuite) TestComplexSlidingWindowPipeline() {
...
-			Expect().
-			Status(204)
-	}
+	done := make(chan struct{})
+
+	go func() {
+		// publish messages to source vertex, with event time starting from 60000
+		startTime := 60000
+
+		for i := 0; true; i++ {
```
KeranYang, line 282: nit: can we use a single for statement?
for {
}

test/reduce-e2e/reduce_test.go
```diff
@@ -33,6 +34,8 @@ type ReduceSuite struct {
 
 // one reduce vertex (keyed)
 func (r *ReduceSuite) TestSimpleKeyedReducePipeline() {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
```
KeranYang, line 37: Why 5 minutes? Please consider adding a comment on this magic number. It seems shared among all reduce test cases so consider making it a global const.

test/reduce-e2e/reduce_test.go
```diff
@@ -46,31 +49,44 @@ func (r *ReduceSuite) TestSimpleKeyedReducePipeline() {
...
+				HTTPExpect(r.T(), "https://localhost:8443").POST("/vertices/in").WithBytes([]byte("3")).WithHeader("X-Numaflow-Event-Time", eventTime).
+					Expect().
+					Status(204)
+			}
+		}
+	}()
 
 	// since the key can be even or odd and the window duration is 10s
 	// the sum should be 20(for even) and 40(for odd)
 	w.Expect().VertexPodLogContains("sink", "Payload -  40  Key -  odd  Start -  60000  End -  70000")
```
KeranYang, line 81: I am a bit new to go routine. Could you help me understand why the Expect can pass with the above go routine added?

My understanding is that the VertexPodLogContains has a one minute timeout(the default value) and within that one minute, it keeps searching for logs and finding matches. Meanwhile the go routine above keeps sending new data to the pipeline until the `w.Expect`s pass and send a signal back to `done` channel, which stops the go routine.

If my understanding is correct, will we ever reach the `case <-ctx.Done()`? (since Expect times out in one minute, which is less than the context timeout value 5 minutes)

KeranYang:
Got another flaky reduce test failure: https://github.com/numaproj/numaflow/actions/runs/3735319201/jobs/6339041738 Not sure if it's related.

test/reduce-e2e/reduce_test.go
yhl25: yeah I have missed that

test/reduce-e2e/reduce_test.go
yhl25: I add timeouts for all the tests I  write. I expect the tests to pass before the timeout.

test/reduce-e2e/reduce_test.go
yhl25: I didn't know the Expect times out after 1 min :). 
That's the thing right, the test owner doesn't know about the Expect timing out, so it's better to have their own context timeout to avoid go routine leaks or tests getting stuck.

KeranYang: > Interestingly I see the expected output in the logs 😄

Right. The failure seems coming from the test clean up. `reduce_test.go:256: Timeout after 1m0s waiting for pipeline pods terminating.` Maybe 1m is too short for terminating all pods. Let's see if such case happens more frequently and if so, adjust the timeout value.

KeranYang:
> Interestingly I see the expected output in the logs 😄

Right. The failure seems coming from the test clean up. `reduce_test.go:256: Timeout after 1m0s waiting for pipeline pods terminating.` Maybe 1m is too short for terminating all pods. Let's see if such case happens more frequently and if so, adjust the timeout value.

test/reduce-e2e/reduce_test.go
KeranYang: We already set 10 minutes for all e2e tests. https://github.com/numaproj/numaflow/blob/main/Makefile#L115

test/reduce-e2e/reduce_test.go
KeranYang: Got it. I think it makes sense to allow e2e test implementer to explicitly specify test timeout values as oppose to using a global one in Makefile. For some test cases, we can fail faster.

test/reduce-e2e/reduce_test.go
KeranYang: I am ok with keeping the 5 mins timeout. (see comments below)

PR #433, opened 2022-12-19
chore: optimizing ordered loop

ashwinidulams:
optimization to the ordered loop so that queueing and dequeuing continues to stay at O(1). 

we can delay the enqueuing of the task until the close of the book as the desired order is guaranteed.

### DIFF
--- a/pkg/reduce/readloop/readloop.go
+++ b/pkg/reduce/readloop/readloop.go
@@ -33,9 +33,10 @@ import (
 	"sync"
 	"time"
 
-	"github.com/numaproj/numaflow/pkg/metrics"
 	"go.uber.org/zap"
 	"k8s.io/apimachinery/pkg/util/wait"
+
+	"github.com/numaproj/numaflow/pkg/metrics"
 
 	"github.com/numaproj/numaflow/pkg/isb"
 	"github.com/numaproj/numaflow/pkg/isb/forward"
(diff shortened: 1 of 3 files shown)


pkg/reduce/readloop/readloop.go
```diff
@@ -362,6 +366,9 @@ func (rl *ReadLoop) closePartitions(partitions []partition.ID) {
 	for _, p := range partitions {
 		q := rl.pbqManager.GetPBQ(p)
 		rl.log.Infow("Close of book", zap.String("partitionID", p.String()))
+		rl.op.insertTask(rl.udfInvocationTracking[p])
+		partitionsInFlight.With(map[string]string{metrics.LabelVertex: rl.op.vertexName, metrics.LabelPipeline: rl.op.pipelineName}).Inc()
 		q.CloseOfBook()
```
yhl25, line 371: why are we incrementing partitionsInFlight while closing the partition? merge conflict?

ashwinidulams: i moved it here, but i guess i can leave it there itself.

yhl25 (approved):

pkg/reduce/readloop/readloop.go
ashwinidulams: since this is streaming, it makes sense to increment this as soon as the go routine is scheduled.

yhl25: yeah

vigith (approved):

PR #442, opened 2023-01-04
chore: java sdk e2e test

yhl25:


vigith (commented):
can you please add to the existing test a flatmap vertex in Java?

test/sdks-e2e/sdks_test.go
```diff
@@ -51,6 +54,43 @@ func (s *SDKsSuite) TestUDFunctionAndSink() {
 		VertexPodLogContains("go-udsink", "hello", PodLogCheckOptionWithContainer("udsink"), PodLogCheckOptionWithCount(3))
 }
 
+func (s *SDKsSuite) TestJavaSdk() {
```
KeranYang, line 57: Do we need this test case if we add to the existing test a flatmap vertex in Java?

vigith: I have asked @yhl25 to also add Java-SDK based flatmap to the e2e test. We also need to have the reduce tests too, as of today we do not have E2E reduce tests for any SDKs, Java will be the first one to have it.

vigith: I think we have to rename `TestJavaSdk` to `TestReduceSDK`. WDYT?

KeranYang: Oh I see. Reduce tests are needed because supporting reduce is a new feature in numaflow-java. +1

yhl25: yeah it makes sense to rename it

test/sdks-e2e/sdks_test.go
```diff
@@ -48,7 +51,45 @@ func (s *SDKsSuite) TestUDFunctionAndSink() {
...
+	w := s.Given().Pipeline("@testdata/simple-keyed-reduce-pipeline.yaml").
+		When().
+		CreatePipelineAndWait()
+	defer w.DeletePipelineAndWait()
+	pipelineName := "even-odd-sum"
+
+	// wait for all the pods to come up
+	w.Expect().VertexPodsRunning()
+
+	done := make(chan struct{})
```
whynowy, line 70: I guess this is not needed? Line 60 will take effect.

yhl25: yeah its not needed

yhl25: No, we need this, I want to stop publishing the messages even before deleting the pipeline.
I forgot the reason for adding it, now when I removed it the tests failed

whynowy: Oh... but you can simply call `cancel()`. Anyways, you make the call.

whynowy: Or switch the order of defers.

yhl25: I just wanted to stop publishing once I see the output in logs, so didn't want to use the context for that, so that the context can be shared with other methods and also we can add more assert statements in future.

vigith (approved):
LGTM

PR #443, opened 2023-01-04
refactor: use exact matching instead of regex to perform e2e data validation.

KeranYang:
In this PR we switch from regex matching to exact matching. By doing so we avoid using redis.KEYS api which is scanning the entire store to find matches hence improve the query performance.

The new approach uses pipelineName:sinkName as hash key. Under each **hash key**, we have **the exact data** as **field** and **no. of occurrences** of the data as **value**. Please review this PR along with the redis UDSink change PR https://github.com/numaproj/numaflow-sinks/pull/11.


test/e2e/testdata/even-odd.yaml
```diff
@@ -14,13 +14,19 @@ spec:
           image: quay.io/numaio/numaflow-go/map-even-odd
     - name: even-sink
       sink:
-        log: {}
+        udsink:
+          container:
+            image: quay.io/numaio/numaflow-go/sink-redis:test-v3
```
KeranYang, line 19: This is my test image, will switch to `quay.io/numaio/numaflow-sink/redis-e2e-test-sink:latest` once PR https://github.com/numaproj/numaflow-sinks/pull/11 gets merged.

test/e2e-api/nats.go
```diff
@@ -33,12 +33,17 @@ func init() {
 
 	http.HandleFunc("/nats/pump-subject", func(w http.ResponseWriter, r *http.Request) {
 		subject := r.URL.Query().Get("subject")
-		mf := newMessageFactory(r.URL.Query())
```
vigith, line 36: why was this replaced ?

test/e2e/testdata/even-odd.yaml
vigith: It is okay to point to `-v3`.

vigith (approved):
awesome. we no longer need the regex/glob pattern match feature, right?

test/e2e/testdata/even-odd.yaml
vigith: I didn't notice that it is pointing to the wrong quay path. It would have been fine if it was pointing to `quay.io/numaio/numaflow-sink/redis-e2e-test-sink:v3`

test/e2e-api/nats.go
KeranYang: MessageFactory is implemented to generate messages using a prefix and n, so it's generating a list of data as prefix-1, prefix-2, ... prefix-n. With us switching from regex matching to exact matching, we are no longer able to verify for prefix-* regex. I remove this and instead always publish exact same data so that we can use exact match to verify.

KeranYang:
> we no longer need the regex/glob pattern match feature, right?
@vigith I was able to switch all existing regex/glob matching data verifications to exact match. In the future, I think it would be e2e test implementors' responsibility to make sure their test udf/udsink is implemented in a way that supports exact match.

test/e2e/testdata/even-odd.yaml
KeranYang: Agree. By publishing a v3, we can ensure a smooth transfer to new version without breaking other developer's e2e tests. Will try this approach next time.

test/e2e-api/http.go
```diff
@@ -45,25 +45,22 @@ func init() {
 		reqBytes, err := io.ReadAll(r.Body)
 		if err != nil {
 			log.Println(err)
-			w.WriteHeader(400)
-			_, _ = w.Write([]byte(err.Error()))
+			http.Error(w, err.Error(), http.StatusBadRequest)
```
whynowy, line 48: 👍

PR #452, opened 2023-01-07
chore: updating watermark fetch logic for UI

yhl25:

Since we are fetching the watermark using head offset, in the UI sometimes we see the watermark value for the map vertex is less than the watermark value for the immediate reduce vertex, which is misleading.  

Screenshot 2023-01-07 at 8 55 15 PM

This is because the map vertex's offset timeline store has a lower event time for the head offset than that of the reduce vertex.

Instead, if we use the latest watermark from the offset timeline store, we will not see this issue.
Screenshot 2023-01-07 at 8 58 46 PM

### DIFF
--- a/pkg/watermark/fetch/edge_fetcher.go
+++ b/pkg/watermark/fetch/edge_fetcher.go
@@ -53,7 +53,8 @@ func NewEdgeFetcher(ctx context.Context, bufferName string, storeWatcher store.W
 	}
 }
 
-// GetHeadWatermark returns the watermark using the HeadOffset (the latest offset among all processors). This
+// GetHeadWatermark returns the latest watermark for a vertex (the latest watermark among all the processors)
+// this guarantees monotonic watermark progression in the UI.
 // can be used in showing the watermark progression for a vertex when not consuming the messages
 // directly (eg. UX, tests)
 // NOTE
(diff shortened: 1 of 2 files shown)


pkg/watermark/fetch/edge_fetcher.go
```diff
@@ -53,33 +53,25 @@ func NewEdgeFetcher(ctx context.Context, bufferName string, storeWatcher store.W
...
+	for _, p := range e.processorManager.GetAllProcessors() {
 		if !p.IsActive() {
 			continue
 		}
-		var o = p.offsetTimeline.GetHeadOffset()
-		e.log.Debugf("Processor: %v (headoffset:%d)", p, o)
-		debugString.WriteString(fmt.Sprintf("[Processor:%v] (headoffset:%d) \n", p, o))
-		if o != -1 && o > headOffset {
-			headOffset = o
-			epoch = p.offsetTimeline.GetEventTimeFromInt64(o)
```
vigith, line 78: we cannot remove this, this affects the correctness.

vigith (commented):
I don't think we can change the `GetHeadWatermark` code. The UIs problem is cosmetic because of the gap in offset-timeline entries. 

Removing `GetEventTimeFromInt64` from `GetHeadWatermark` will cause a correctness problem.

whynowy:
BTW - we are not running the latest UI code.

yhl25:
> I don't think we can change the `GetHeadWatermark` code. The UIs problem is cosmetic because of the gap in offset-timeline entries. 
> 
> 
> 
> Removing `GetEventTimeFromInt64` from `GetHeadWatermark` will cause a correctness problem.
GetHeadWatermark is only used by UI

yhl25:
> BTW - we are not running the latest UI code.
This issue is with the latest code.

vigith:
> GetHeadWatermark is only used by UI

Perhaps rename the function and move it under UIs code section?

vigith:
is this PR still valid?

PR #458, opened 2023-01-10
doc: reduce persistent store

vigith:


docs/user-guide/user-defined-functions/reduce/reduce.md
```diff
@@ -43,153 +53,62 @@ The reduce supports a parallelism value while defining the edge. This is because
...
-kubectl apply -f https://raw.githubusercontent.com/numaproj/numaflow/stable/examples/0-isbsvc-jetstream.yaml
-```
-
-## sum pipeline using fixed window
-This is a simple reduce pipeline that just does summation (sum of numbers) but uses fixed window.
-The snippet for the reduce vertex is as follows.
-
-![plot](../../../assets/simple-reduce.png)
+Reduce unlike map requires persistence. To support persistence user has to define the 
+`persistence` configuration. We replay the data stored in this store on pod startup if there has
```
whynowy, line 69: `storage`

docs/user-guide/user-defined-functions/reduce/reduce.md
```diff
@@ -43,153 +53,62 @@ The reduce supports a parallelism value while defining the edge. This is because
...
+           ....
 ```
 
-[6-reduce-fixed-window.yaml](https://github.com/numaproj/numaflow/blob/main/examples/6-reduce-fixed-window.yaml)
-has the complete pipeline definition.
+### Persistent Volume Claim (PVC)
 
-In this example we use a `parallelism` of `2`. We are setting a parallelism > 1 because it is a 
-keyed window.
+`persistentVolumeClaim` has two fields, `volumeSize` and `accessMode`. As name suggests,
```
whynowy, line 83: And `storageClassName`

docs/user-guide/user-defined-functions/reduce/reduce.md
```diff
@@ -43,153 +53,62 @@ The reduce supports a parallelism value while defining the edge. This is because
...
-2023/01/05 11:54:41 (sink)  Payload -  300  Key -  odd  Start -  60000  End -  120000
-2023/01/05 11:54:41 (sink)  Payload -  600  Key -  even  Start -  60000  End -  120000
-2023/01/05 11:54:41 (sink)  Payload -  300  Key -  odd  Start -  120000  End -  180000
-2023/01/05 11:54:41 (sink)  Payload -  600  Key -  even  Start -  120000  End -  180000
-2023/01/05 11:54:42 (sink)  Payload -  600  Key -  even  Start -  180000  End -  240000
-2023/01/05 11:54:42 (sink)  Payload -  300  Key -  odd  Start -  180000  End -  240000
-```
+We also support `emptyDir` for quick experimentation. We do not recommend this in production
+setup. If we use `emptyDir`, we will end up in data loss if there are pod migrations. `emptyDir` 
+also takes an optional `sizeLimit`.
```
whynowy, line 103: and `medium`.

docs/user-guide/user-defined-functions/reduce/reduce.md
```diff
@@ -66,7 +66,7 @@ chaining of windows, keyed streams, etc.
 ## Storage
 
 Reduce unlike map requires persistence. To support persistence user has to define the 
-`persistence` configuration. We replay the data stored in this store on pod startup if there has
+`persistence` configuration. We replay the data stored in this storage on pod startup if there has
```
whynowy, line 69: What I meant was, it's `storage` instead of `persistence`.  Maybe you are not talking about the property name, but it's confusing.

docs/user-guide/user-defined-functions/reduce/reduce.md
```diff
@@ -43,153 +53,68 @@ The reduce supports a parallelism value while defining the edge. This is because
...
+           ....
 ```
 
-[6-reduce-fixed-window.yaml](https://github.com/numaproj/numaflow/blob/main/examples/6-reduce-fixed-window.yaml)
-has the complete pipeline definition.
+### Persistent Volume Claim (PVC)
 
-In this example we use a `parallelism` of `2`. We are setting a parallelism > 1 because it is a 
-keyed window.
+`persistentVolumeClaim` has two fields, `volumeSize` and `accessMode`. As name suggests,
```
whynowy, line 83: Not `two`.

whynowy (approved):

PR #461, opened 2023-01-11
chore: move forward to top-level

vigith:

Forward like reduce is a top-level entity and not just tied with ISB.


pkg/isb/stores/redis/read_test.go
```diff
@@ -26,6 +26,7 @@ import (
 	"testing"
 	"time"
 
+	forward2 "github.com/numaproj/numaflow/pkg/forward"
```
whynowy, line 29: Why forward2?

whynowy (approved):

pkg/isb/stores/redis/read_test.go
vigith: will fix it. missed the comment. @chromevoid can you please fix in your PR when you rebase?

PR #472, opened 2023-01-12
chore: finalize doc for 0.7 release

vigith:


vigith:
Do we need to explicitly callout map-reduce in the README.md? It is implicitly mentioned through words like "data processing", "analytics", etc.

whynowy (approved):

PR #487, opened 2023-01-17
feat: Implement source data transformer and apply to all existing sources

KeranYang:
### PR description
* Implement transformer functionality for existing sources.
* Reuse an existing even-odd udf as transformer on http source to test the data gets filtered and conditional forwarded to even-sink.
* This change doesn't include changing event time.

### TODO
* <del>Merge corresponding numaflow-go change: https://github.com/numaproj/numaflow-go/pull/31
* Add MapT functionality to numaflow and switch the transformer to using MapT.
* Implement an e2e test case to cover source transformer modifying event time.
* <del>Apply transformer functionality to other sources.
* Update doc.
* Emit Metrics?

Explain what this PR does.


pkg/sources/http/http.go
```diff
@@ -192,17 +197,53 @@ func New(vertexInstance *dfv1.VertexInstance, writers []isb.BufferWriter, fetchW
 			forwardOpts = append(forwardOpts, forward.WithReadBatchSize(int64(*x.ReadBatchSize)))
 		}
 	}
-	forwarder, err := forward.NewInterStepDataForward(vertexInstance.Vertex, h, destinations, forward.All, applier.Terminal, fetchWM, publishWM, forwardOpts...)
+
+	if vertexInstance.Vertex.SpecifyUDTransformer() {
```
vigith, line 201: can't we move this logic to forwarder.go? It makes more sense to make it to the forwarder, else implementing each source can get complicated.

pkg/apis/numaflow/v1alpha1/source.go
```diff
@@ -29,10 +29,37 @@ type Source struct {
 	HTTP *HTTPSource `json:"http,omitempty" protobuf:"bytes,3,opt,name=http"`
 	// +optional
 	Nats *NatsSource `json:"nats,omitempty" protobuf:"bytes,4,opt,name=nats"`
+	// +optional
+	UdTransformer *UDTransformer `json:"udtransformer,omitempty" protobuf:"bytes,5,opt,name=udtransformer"`
```
vigith, line 33:
```suggestion
	UDTransformer *UDTransformer `json:"udtransformer,omitempty" protobuf:"bytes,5,opt,name=udtransformer"`
```

Initialism

pkg/sources/http/http.go
whynowy: Should this be in the `source.go`? Each source's `New...` function should accept an `applier` interface and `WhereTo`, and then they can be used to pass to the forwarder.

pkg/apis/numaflow/v1alpha1/vertex_types.go
```diff
@@ -71,6 +71,10 @@ func (v Vertex) IsASource() bool {
 	return v.Spec.Source != nil
 }
 
+func (v Vertex) SpecifyUDTransformer() bool {
```
whynowy, line 74: `Hasxxx`?

pkg/apis/numaflow/v1alpha1/source.go
whynowy, line 33: Can we just use `transformer` in `json` and `protobuf`?

pkg/sources/http/http.go
KeranYang: Will address.

pkg/apis/numaflow/v1alpha1/source.go
KeranYang: Currently I am using `udtransformer` everywhere as I would like to keep the wording consistent across the code base.
I am ok with either `transformer` or `udtransfomer`.

`udtransformer` emphasize the fact that it's user defined while `transformer` distinguish itself from udf, udsink and future udsource.

I am also ok with your propose where we use `udtransformer` across the code base while we expose to user the `transformer` spec keyword.

@vigith @whynowy  Let me know which approach you guys prefer. Thanks!

vigith: I am okay with `udtransformer`, makes more sense w.r.t maintainability.

KeranYang: @vigith I had a offline discussion with Derek. He raised a point that `udtransformer` is not a valid word, which might confuse end users. Mind we use `transformer` in spec and `udtransformer` in code base?

vigith: yes, please go ahead with `transformer` 😄

pkg/sources/http/http.go
```diff
@@ -256,7 +265,7 @@ func (h *httpSource) Close() error {
 
 func (h *httpSource) Stop() {
 	h.logger.Info("Stopping http reader...")
-	h.ready = false
+	defer func() { h.ready = false }()
```
whynowy, line 268: Why let it execute in the end?

pkg/sources/source.go
```diff
@@ -117,7 +122,31 @@ func (sp *SourceProcessor) Start(ctx context.Context) error {
 		return fmt.Errorf("unrecognized isb svc type %q", sp.ISBSvcType)
 	}
 
-	sourcer, err := sp.getSourcer(writers, fetchWatermark, publishWatermark, sourcePublisherStores, log)
+	var transformer *function.UdsGRPCBasedUDF
+	if sp.VertexInstance.Vertex.HasUDTransformer() {
+		transformer, err = function.NewUDSGRPCBasedUDF()
```
whynowy, line 127: You should have a dedicated implementation for transformer, right?

pkg/udf/function/uds_grpc.go
```diff
@@ -36,35 +36,35 @@ import (
 	"github.com/numaproj/numaflow/pkg/isb"
 )
 
-// udsGRPCBasedUDF applies user defined function over gRPC (over Unix Domain Socket) client/server where server is the UDF.
-type udsGRPCBasedUDF struct {
+// UdsGRPCBasedUDF applies user defined function over gRPC (over Unix Domain Socket) client/server where server is the UDF.
```
whynowy, line 39: Shouldn't reuse this implementation.

pkg/sources/http/http.go
KeranYang: Should we mark httpSource as not ready **after** instead of **before** we stop the forwarder?

whynowy: No, marking it not ready beforehand is an intended behavior.

Marking it not ready will get the health check unsuccessful, and bring it out of the traffic (if there's LB existing). Imagine what will happen that you are already in the process of shutting down backend services, but the traffic is still coming in.

pkg/reconciler/pipeline/validate.go
```diff
@@ -85,6 +85,12 @@ func ValidatePipeline(pl *dfv1.Pipeline) error {
 		return fmt.Errorf("pipeline has no sink, at least one vertex with 'sink' defined is required")
 	}
 
+	for k, s := range sources {
+		if s.Source.UDTransformer != nil && s.Source.UDTransformer.Container != nil && s.Source.UDTransformer.Container.Image == "" {
```
whynowy, line 89: `if x := s.Source.UDTransformer; x != nil && x.Container != nil && x.Image == "" {`

pkg/sources/generator/tickgen.go
```diff
@@ -122,16 +121,14 @@ func WithReadTimeout(timeout time.Duration) Option {
 	}
 }
 
-// NewMemGen function creates an instance of generator.
```
whynowy, line 125: Why remove these docs?

pkg/sources/source.go
```diff
@@ -117,7 +121,31 @@ func (sp *SourceProcessor) Start(ctx context.Context) error {
 		return fmt.Errorf("unrecognized isb svc type %q", sp.ISBSvcType)
 	}
 
-	sourcer, err := sp.getSourcer(writers, fetchWatermark, publishWatermark, sourcePublisherStores, log)
+	t, err := transformer.NewGRPCBasedTransformer()
```
whynowy, line 124: Shouldn't this be in the `HasUDTransformer()` check?

pkg/sources/source.go
```diff
@@ -117,7 +121,31 @@ func (sp *SourceProcessor) Start(ctx context.Context) error {
...
+
+	if sp.VertexInstance.Vertex.HasUDTransformer() {
+		// Readiness check
+		if err = t.WaitUntilReady(ctx); err != nil {
+			return fmt.Errorf("failed on user defined transformer readiness check, %w", err)
+		}
+		defer func() {
+			err = t.CloseConn(ctx)
+			if err != nil {
+				log.Warnw("failed to close gRPC client conn", zap.Error(err))
```
whynowy, line 137: `log.Errorw`?

`Failed to xxx` - Capital the first letter in the log message.

pkg/udf/function/uds_grpc.go
```diff
@@ -23,27 +23,21 @@ import (
...
-
-	"github.com/numaproj/numaflow/pkg/isb"
 )
 
 // udsGRPCBasedUDF applies user defined function over gRPC (over Unix Domain Socket) client/server where server is the UDF.
 type udsGRPCBasedUDF struct {
 	client functionsdk.Client
 }
 
-var _ applier.MapApplier = (*udsGRPCBasedUDF)(nil)
```
whynowy, line 44: Why removed these?

pkg/sources/source.go
```diff
@@ -117,7 +121,31 @@ func (sp *SourceProcessor) Start(ctx context.Context) error {
...
+		}()
+	}
+
+	var sourcer Sourcer
+	if sp.VertexInstance.Vertex.HasUDTransformer() {
+		sourcer, err = sp.getSourcer(writers, sp.getTransformerGoWhereDecider(), t, fetchWatermark, publishWatermark, sourcePublisherStores, log)
+	} else {
+		sourcer, err = sp.getSourcer(writers, forward.All, forward.Terminal, fetchWatermark, publishWatermark, sourcePublisherStores, log)
+	}
+
```
whynowy, line 148: Reference to https://github.com/numaproj/numaflow/blob/main/pkg/sinks/sink.go#L122-L124 to have a health check.

pkg/apis/numaflow/v1alpha1/source.go
```diff
@@ -29,10 +29,37 @@ type Source struct {
...
+		s.getMainContainer(req),
+	}
+	if s.UDTransformer != nil {
+		containers = append(containers, s.getUDTransformerContainer(req))
+	}
+	return containers, nil
+}
+
+func (s Source) getMainContainer(req getContainerReq) corev1.Container {
+	return containerBuilder{}.init(req).args("processor", "--type="+string(VertexTypeSource), "--isbsvc-type="+string(req.isbSvcType)).build()
```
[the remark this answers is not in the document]
KeranYang: Good question. I was referring to udf and udsink to explicitly declare a `getMainContainer` function (don't really understand why :)). It's only used within the classes by `getContainers`, so I agree it's ok to not wrap it as a function. Do you want me to remove the wrapper across (source, udf and udsink)?

pkg/reconciler/pipeline/validate.go
KeranYang: Cool!

pkg/sources/generator/tickgen.go
KeranYang: Because I think they are out of date. I can keep the first line though.

pkg/sources/source.go
KeranYang: Both map and reduce udf are using `log.Warnw`, do we know why?

pkg/udf/function/uds_grpc.go
KeranYang: Synced up offline. These are required to ensure udsGRPCBasedUDF implements all methods defined by MapApplier.

KeranYang:
Ready for re-review and merge. @vigith @whynowy

pkg/sources/source.go
```diff
@@ -117,7 +122,36 @@ func (sp *SourceProcessor) Start(ctx context.Context) error {
 		return fmt.Errorf("unrecognized isb svc type %q", sp.ISBSvcType)
 	}
 
-	sourcer, err := sp.getSourcer(writers, fetchWatermark, publishWatermark, sourcePublisherStores, log)
+	metricsOpts := []metrics.Option{
+		metrics.WithLookbackSeconds(int64(sp.VertexInstance.Vertex.Spec.Scale.GetLookbackSeconds())),
```
whynowy, line 126: Remove the empty line.

whynowy (approved):
It looks good to me overall.

pkg/reduce/data_forward.go
```diff
@@ -24,17 +24,16 @@ import (
 	"strconv"
```
vigith, line 24: why are we renaming this file? This does much more than just `data_forward`

pkg/sources/transformer/grpc_transformer.go
```diff
@@ -0,0 +1,123 @@
...
+				Flag:        true,
+				MainCarDown: false,
+			},
+		}
+	}
+
+	writeMessages := make([]*isb.Message, 0)
+	for i, datum := range datumList {
+		key := datum.Key
+		if datum.EventTime != nil {
```
vigith, line 106: what is the default zero value for a gRPC time datatype? Is it `nil`?

test/e2e/testdata/source-data-transform.yaml
```diff
@@ -0,0 +1,25 @@
...
+metadata:
+  name: source-data-transform
+spec:
+  vertices:
+    - name: in
+      source:
+        http: {}
+        transformer:
+          container:
+            image: quay.io/numaio/numaflow-go/map-even-odd
```
vigith, line 12: how is `map-even-odd` working here? shouldn't it be `mapT`?

pkg/sources/transformer/grpc_transformer.go
```diff
@@ -0,0 +1,123 @@
...
+			// Transformer supports changing event time.
+			parentPaneInfo.EventTime = datum.EventTime.EventTime.AsTime()
+		}
```
vigith, line 109: shouldn't there be an `else` condition so that if we get a `nil` from the source, we can assign the default value we got from the source? Basically

```
EventTime: parentPaneInfo.EventTime),
Watermark: readMessage.Watermark),
```

or do we handle that in the SDK side?

pkg/reduce/data_forward.go
KeranYang: I renamed it for 2 reasons:
1. `reduce/reduce.go` seems duplicate.
2. This file contains the following objects.
`type DataForward struct`
`func NewDataForward`
`func (d *DataForward) Start()`
`func (d *DataForward) forwardAChunk(ctx context.Context)`
They seem all tied with DataForward class. I agree it does more than just forwarding the data but I also feel renaming it to data_forward is more aligned with what we defined in the file.

pkg/sources/transformer/grpc_transformer.go
KeranYang: It's non-nil.

When `u.client.MapTFn` gets called, the mapT SDK requires user to assign a non-nil event time to the returned MessageT and then pass it to returned datumList. Therefore this nil check will always pass. 

https://github.com/numaproj/numaflow-go/blob/main/pkg/function/service.go#L145

pkg/sources/transformer/grpc_transformer.go
KeranYang: Firstly we guarantee non-nil because mapT requires user to specify an event time. Secondly even if nil, the following code (line 112) will pass the original parentPaneInfo to the `writeMessage`.

test/e2e/testdata/source-data-transform.yaml
KeranYang: You are right. I was using existing mapFn to test transformer because this PR is before my mapT SDK implementation(it's map but still it's a sidecar container). I will create follow up PRs to switch to mapT.

vigith (approved):

pkg/sources/transformer/grpc_transformer.go
whynowy: Some thoughts:

1. The current `MessageToxxx` [helper functions](https://github.com/numaproj/numaflow-go/blob/main/pkg/function/message_t.go#L16-L28) in the SDK all require `eventTime`, shall we allow the users to only return a `MessageT` without giving `eventTime`, which means they don't care about it, and use the original eventTime is okay.
2. Thinking of following scenarios, are we expecting these behaviors - if a UDF image is given to `transformer`, nothing is broken, the transformer just works as a regular UDF without having the ability to set eventTime; If a transformer image is give to UDF, it still works as a regular UDF, even though it sets the eventTime in the return, but it will be ignored. 

@vigith @KeranYang

vigith: sounds like a good idea to me.

KeranYang: +1. Thinking about how to implement this.
1. On SDK side, we can unify `Message` and `MessageT` into one generic type such that transformer and udf share exact same contract and can replace each other from user spec.
2. On numaflow side, transformer gRPC will check event time and update `writeMessage` while udf gRPC just don't check event time and pass on the parentPaneInfo.

I am gonna create a separate issue to track this change.

KeranYang: https://github.com/numaproj/numaflow/issues/504

PR #492, opened 2023-01-19
chore: optimise remove windows, adding test

yhl25:

### DIFF
--- a/pkg/reduce/readloop/readloop.go
+++ b/pkg/reduce/readloop/readloop.go
@@ -65,6 +65,7 @@ type ReadLoop struct {
 	whereToDecider        forward.ToWhichStepDecider
 	publishWatermark      map[string]publish.Publisher
 	udfInvocationTracking map[partition.ID]*task
+	latestWatermark       processor.Watermark
 }
 
 // NewReadLoop initializes  and returns ReadLoop.
@@ -94,6 +95,7 @@ func NewReadLoop(ctx context.Context,
 		whereToDecider:        whereToDecider,
 		publishWatermark:      pw,
 		udfInvocationTracking: make(map[partition.ID]*task),
+		latestWatermark:       processor.Watermark(time.Time{}),
 	}
 
 	err := rl.Startup(ctx)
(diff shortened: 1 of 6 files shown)


pkg/window/strategy/sliding/sliding.go
```diff
@@ -161,6 +161,8 @@ func (s *Sliding) RemoveWindows(wm time.Time) []window.AlignedKeyedWindower {
 		if win.EndTime().Before(wm) {
 			s.entries.Remove(e)
 			closedWindows = append(closedWindows, win)
+		} else {
```
vigith, line 164: This for loop will only run for 1 iteration, is that expected?

yhl25: this for loop will run until you find a window with end time > watermark

vigith (approved):

pkg/reduce/readloop/readloop.go
```diff
@@ -94,6 +95,7 @@ func NewReadLoop(ctx context.Context,
 		whereToDecider:        whereToDecider,
 		publishWatermark:      pw,
 		udfInvocationTracking: make(map[partition.ID]*task),
+		latestWatermark:       processor.Watermark(time.Time{}),
```
ashwinidulams, line 98: rename this to indicate that you are tracking the window boundary

pkg/reduce/readloop/readloop.go
```diff
@@ -150,8 +152,15 @@ func (rl *ReadLoop) Process(ctx context.Context, messages []*isb.ReadMessage) {
 	// for each successfully written message,
 	for _, m := range successfullyWrittenMessages {
 		// close any windows that need to be closed.
+		var closedWindows []window.AlignedKeyedWindower
 		wm := processor.Watermark(m.Watermark)
-		closedWindows := rl.windower.RemoveWindows(time.Time(wm))
+
+		// check to see if any windows can be closed, when the watermark is moved
+		if wm.After(time.Time(rl.latestWatermark)) {
```
vigith, line 159: Why are we comparing 2 watermarks? Isn't it obvious that those 2 will be diverging (the old one and the new one)?

pkg/reduce/readloop/readloop.go
yhl25: I am not tracking the window boundary, I am tracking the latest watermark we have received

pkg/reduce/readloop/readloop.go
yhl25: they can be same, I just wanted to invoke the remove windows function only when the watermark is progressed

vigith:
Let's do a benchmark to compare the performance and allocation improvements.

yhl25:
> Let's do a benchmark to compare the performance and allocation improvements.

I ran a benchmark test
observations:
I see better performance when RemoveWindows is invoked once per batch since we avoid a lot of comparisons with millisecond watermark progression
Not much difference if the watermark progresses slowly because we have a check inside remove windows which returns early.

pkg/reduce/readloop/readloop.go
```diff
@@ -144,21 +144,26 @@ func (rl *ReadLoop) Process(ctx context.Context, messages []*isb.ReadMessage) {
 		rl.log.Errorw("Failed to write messages", zap.Int("totalMessages", len(messages)), zap.Int("writtenMessage", len(successfullyWrittenMessages)))
 	}
 
+	if len(successfullyWrittenMessages) == 0 {
```
vigith, line 147: why will it be == 0? We have checked early in the reduce code to continue only if there are `Read` messages. What we read should have been written, right?

yhl25: yeah, removed the check

vigith (approved):

PR #499, opened 2023-01-21
chore: ignore udsink msgs without response or failures

whynowy:

Explain what this PR does.


vigith:
If we are going to totally ignore the status and err field, shouldn't we remove it from the data type? Won't users mistakenly set it expecting the platform to do something with it?

Shouldn't we track the errors as a metric since we only logging it? Errors as a metric could help the UDSink owners to alert on increased failures?

whynowy:
Adding metrics makes sense. Maybe we should still keep them in the API, so that an optional "retryOnError" can be provided?

PR #500, opened 2023-01-23
feat: Reduce UI Support

### LINKED ISSUE #470: Reduce UI support
# Summary

---

veds-g:
Fixes #470

Added pod level data for vertices - show metrics, watermarks, processing rates for individual pods in a vertex
Added edge level data  - show back-pressure for each edge in case of multiple edges between vertices


pkg/watermark/fetch/interface.go
```diff
@@ -30,4 +30,6 @@ type Fetcher interface {
 	GetWatermark(offset isb.Offset) processor.Watermark
 	// GetHeadWatermark returns the latest watermark based on the head offset
 	GetHeadWatermark() processor.Watermark
+	// GetPodWatermarks returns the watermarks for all the pods in a vertex
+	GetPodWatermarks() []processor.Watermark
```
vigith, line 34: I think this can be derived from other functions. Does it make sense to move it under UI since only UI is using it?

pkg/watermark/fetch/source_fetcher.go
```diff
@@ -70,6 +70,27 @@ func (e *sourceFetcher) GetHeadWatermark() processor.Watermark {
 	return processor.Watermark(time.UnixMilli(epoch))
 }
 
+// GetPodWatermarks returns the list of watermarks of all pods in a vertex
```
vigith, line 73: move under UI?

pkg/watermark/generic/noop.go
```diff
@@ -56,6 +56,10 @@ func (n NoOpWMProgressor) GetHeadWatermark() processor.Watermark {
 	return processor.Watermark{}
 }
 
+func (n NoOpWMProgressor) GetPodWatermarks() []processor.Watermark {
```
vigith, line 59: will not be required if this is moved under UI.

vigith (commented):
I am debating whether we should move PodWatermark under the UI section because it is not required otherwise. It can be derived from other functions. Personally, I do not want to change/add anything to the watermark code without a thorough review since it is very integral to data processing correctness.

@whynowy @chromevoid @ashwinidulams WDYT?

pkg/apis/proto/daemon/daemon.proto
```diff
@@ -114,7 +116,7 @@ service DaemonService {
   };
 
   rpc GetVertexMetrics (GetVertexMetricsRequest) returns (GetVertexMetricsResponse) {
-    option (google.api.http).get = "/api/v1/pipelines/{pipeline}/vertices/{vertex}/metrics";
+    option (google.api.http).get = "/api/v1/pipelines/{pipeline}/vertices/{vertex}/pods/{pods}/metrics";
```
whynowy, line 119: Should we introduce a new API, e.g. `GetVertxItemizedMetrics`, which returns metrics for each pod, and is supposed to be only used by reduce vertex? With that, we could get least changes for the backend.

pkg/reconciler/vertex/scaling/scaling.go
```diff
@@ -140,11 +141,12 @@ func (s *Scaler) scaleOneVertex(ctx context.Context, key string, worker int) err
 	log := logging.FromContext(ctx).With("worker", fmt.Sprint(worker)).With("vertexKey", key)
 	log.Debugf("Working on key: %s", key)
 	strs := strings.Split(key, "/")
-	if len(strs) != 2 {
+	if len(strs) != 3 {
```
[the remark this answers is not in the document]
whynowy: What is the reason we need to touch autoscaling?

pkg/daemon/client/daemon_client.go
```diff
@@ -87,14 +87,15 @@ func (dc *DaemonClient) GetPipelineBuffer(ctx context.Context, pipeline, buffer
 	}
 }
 
-func (dc *DaemonClient) GetVertexMetrics(ctx context.Context, pipeline, vertex string) (*daemon.VertexMetrics, error) {
+func (dc *DaemonClient) GetVertexMetrics(ctx context.Context, pipeline, vertex string, pods int64) (*daemon.GetVertexMetricsResponse, error) {
```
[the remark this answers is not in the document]
veds-g: will create / modify the response object

pkg/daemon/server/service/pipeline_watermark_query.go
```diff
@@ -142,10 +144,16 @@ func (ps *pipelineMetadataQuery) GetPipelineWatermarks(ctx context.Context, requ
 	i := 0
 	for k, vertexFetchers := range ps.watermarkFetchers {
 		var latestWatermark = int64(-1)
+		var podWatermarks []int64
 		for _, fetcher := range vertexFetchers {
 			watermark := fetcher.GetHeadWatermark().UnixMilli()
+			podWatermark := fetcher.GetPodWatermarks()
 			if watermark > latestWatermark {
 				latestWatermark = watermark
+				podWatermarks = nil
```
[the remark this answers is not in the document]
veds-g: every time we get a latest watermark for a fetcher we update the list of watermarks from all pods obtained at that instant. This is why I clear and reassign (which is assigning null and reassigning)

pkg/reconciler/vertex/scaling/scaling.go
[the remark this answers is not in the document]
veds-g: key I believe is the api endpoint in string format ('namespace/vertex/pods') which is split on '/'. Since I added pod count to the endpoint, I had to modify the function signature by accepting an additional parameter.

pkg/apis/proto/daemon/daemon.proto
veds-g: can do this

pkg/daemon/server/service/pipeline_metrics_query.go
```diff
@@ -156,61 +156,73 @@ func (ps *pipelineMetadataQuery) GetVertexMetrics(ctx context.Context, req *daem
 			Name: vertexName,
 		},
 	}
-
-	processingRates := make(map[string]float64, 0)
-	pendings := make(map[string]int64, 0)
+	podNum := int64(1)
+	// for now only reduce has parallelism might have to modify later
+	obj := ps.pipeline.GetFromEdges(req.GetVertex())
```
whynowy, line 161: I understand we do not need to check if it's reduce vertex because the parallelism is nil if it's not. Adding more comments to mention that would be appreciated - I thought a while before figuring it out. :)

pkg/reconciler/vertex/scaling/scaling.go
```diff
@@ -215,12 +215,31 @@ func (s *Scaler) scaleOneVertex(ctx context.Context, key string, worker int) err
 		return fmt.Errorf("failed to get metrics of vertex key %q, %w", key, err)
 	}
 	// Avg rate and pending for autoscaling are both in the map with key "default", see "pkg/metrics/metrics.go".
-	rate, existing := vMetrics.ProcessingRates["default"]
+	var rate float64
```
whynowy, line 218: Do not need to touch autoscaling, reduce vertices are excluded for autoscaling.

pkg/daemon/server/service/pipeline_watermark_query.go
```diff
@@ -142,18 +144,26 @@ func (ps *pipelineMetadataQuery) GetPipelineWatermarks(ctx context.Context, requ
 	i := 0
 	for k, vertexFetchers := range ps.watermarkFetchers {
 		var latestWatermark = int64(-1)
+		var latestpodWatermarks []processor.Watermark
 		for _, fetcher := range vertexFetchers {
 			watermark := fetcher.GetHeadWatermark().UnixMilli()
+			podWatermark := fetcher.GetPodWatermarks()
```
whynowy, line 150: My understanding of doing this is:

1. Update the return from `watermark` to `watermarks` (array);
2. Update `GetVertexWatermarkFetchers`, to update the `key` of the map from `vertex` to `vertex/replica` - for map vertex, the `replica` could be always `0`.
3. When querying wartermarks, use corresponding fetcher(s) from the map to do it.

pkg/daemon/server/service/pipeline_metrics_query.go
veds-g: sure

pkg/reconciler/vertex/scaling/scaling.go
veds-g: I have modified the VertexMetrics response and was summing up the processingRate values for multiple elements of returned array(multiple pods), since reduce is excluded I will only use the first entry of the array ~ 0th element (vMetrics[0].processingRates)

pkg/daemon/server/service/pipeline_watermark_query.go
veds-g: updated as mentioned

Krithika3:
Just curious - Added edge level data - show back-pressure for each edge in case of multiple edges between vertices? - Did this not already exist? @veds-g @vigith

vigith:
> Just curious - Added edge level data - show back-pressure for each edge in case of multiple edges between vertices? - Did this not already exist? @veds-g @vigith

"edges" is a misnomer, what we are doing now is back-pressure on buffers. What that means is, for a reduce vertex a logical edge is backed by multiple buffers. @veds-g is showing backpressure per partition of an edge.

pkg/daemon/server/service/pipeline_watermark_query.go
veds-g: will be moving the watermark changes to a new pr as discussed

whynowy (approved):
LGTM

PR #501, opened 2023-01-23
feat: improve reduce performance

### LINKED ISSUE #503: Memory and CPU are linearly growing with partition count, make it ~ O(log(n))
gRPC streams are consuming a lot of memory and CPU resources. We partition gRPC streams per partition; we need to revisit this strategy. Today the Memory and CPU are growing at O(N), we need to make it something like O(log(N))

ashwinidulams:
Fixes #503 

Use stream per window instead of per partition

This will reduce the memory pressure by doing the following
- the number of streaming connections
- the number of pbq, pnf and other objects.


KeranYang:
@ashwinidulams you might notice the transformer-e2e test is flaky. I think it's cause by my transformer change. Please don't worry about it and I will fix.

vigith (dismissed):

vigith:
let's do something like 

```golang
func LookupEnvInt64Or(name, defaultValue) int64 {
  ...
  if err { 
     panic(err) 
  }
}
```

pkg/shared/util/env.go
```diff
@@ -27,3 +29,15 @@ func LookupEnvStringOr(key, defaultValue string) string {
 		return defaultValue
 	}
 }
+
+func LookupEnvInt64Or(key string, defaultValue int) int {
```
whynowy, line 33: Function name does not match the type return?

vigith: made it `LookupEnvIntOr`

whynowy (approved):

vigith (approved):

PR #505, opened 2023-01-24
Add an e2e test for source data transformer

### LINKED ISSUE #498: Add Source Data Transformer E2E test
# Summary

Add E2E test cases to cover the source data transformer implementation.

---

KeranYang:
The transformer implementation can be found here:
https://github.com/numaproj/numaflow-go/blob/main/pkg/function/examples/event_time_filter/impl/filter.go

closes #498


test/transformer-e2e/transformer_test.go
```diff
@@ -0,0 +1,79 @@
+//go:build test
```
whynowy, line 1: Combine it with `sdk_test`?

KeranYang: This single transformer test case takes ~8 minutes to finish. I am concerned that if we combine it to `sdk_test`, `sdk_test` might time out. I agree that logically this change belongs to `sdk_test`.

vigith:
@KeranYang can you also add an example of MapT pipeline in this folder https://github.com/numaproj/numaflow/tree/main/examples this PR?

vigith:
> @KeranYang can you also add an example of MapT pipeline in this folder https://github.com/numaproj/numaflow/tree/main/examples this PR?

actually, you can add it via a different PR if you prefer when you add the documentation 😄

KeranYang:
@vigith Thanks! I will address it as a separate PR for documentation. Created a new issue to track:  https://github.com/numaproj/numaflow/issues/506

test/transformer-e2e/transformer_test.go
whynowy: A single transformer test case takes 8 mins does not mean combining them will double or triple the time.

KeranYang: Would it increase the time? Test cases in one file run one by one so if current sdk-test runs for 5 mins, then plus 8 mins will exceed the e2e time limit 10 mins.

vigith: True, setting up the cluster, etc., takes more time. The tests per se are < 1min.

KeranYang: `--- PASS: TestTransformerSuite/TestEventTimeFilter (369.77s)`
TestEventTimeFilter takes 6 mins.

KeranYang: It's because I have 5 `VertexPodLogNotContains` calls. To check `LogNotContains`, the default timeout is 1 minute so this is adding 5 minutes to the overall run. I am publishing a PR to a. move the test case to under `sdks-test `and b. explicitly ask VertexPodLogNotContains to timeout after 1 second. https://github.com/numaproj/numaflow/pull/508

Going forward we might eventually hit the 10 minutes limit, as we add more and more new types of sdk. Then we can think of either to re-group sdks-test cases or to divide the sdks-test to multiple sub test folders such that they can run in parallel.

whynowy: Good! It's okay to increase the 10 min limit.

PR #508, opened 2023-01-25
chore: move source transformer e2e test to sdk-test

KeranYang:
* Move source transformer e2e test to sdk-test.
* Give Expect() 1 second instead of default 1 minute to check LogNotContains. This brings down test run time from ~6 minutes to ~1 minute.

Explain what this PR does.


test/sdks-e2e/sdks_test.go
```diff
@@ -92,6 +94,46 @@ func (s *SDKsSuite) TestReduceSDK() {
...
+
+	w.Expect().VertexPodLogContains("sink-within-2022", fmt.Sprintf("EventTime -  %d", janFirst2022.UnixMilli()), PodLogCheckOptionWithCount(1)).
+		VertexPodLogContains("sink-after-2022", fmt.Sprintf("EventTime -  %d", janFirst2023.UnixMilli()), PodLogCheckOptionWithCount(2)).
+		VertexPodLogContains("sink-all", fmt.Sprintf("EventTime -  %d", janFirst2022.UnixMilli()), PodLogCheckOptionWithCount(1)).
+		VertexPodLogContains("sink-all", fmt.Sprintf("EventTime -  %d", janFirst2023.UnixMilli()), PodLogCheckOptionWithCount(2)).
+		VertexPodLogNotContains("sink-within-2022", fmt.Sprintf("EventTime -  %d", janFirst2023.UnixMilli()), PodLogCheckOptionWithTimeout(1*time.Second)).
+		VertexPodLogNotContains("sink-after-2022", fmt.Sprintf("EventTime -  %d", janFirst2022.UnixMilli()), PodLogCheckOptionWithTimeout(1*time.Second)).
+		VertexPodLogNotContains("sink-all", "Before2022", PodLogCheckOptionWithTimeout(1*time.Second)).
+		VertexPodLogNotContains("sink-within-2022", "Before2022", PodLogCheckOptionWithTimeout(1*time.Second)).
+		VertexPodLogNotContains("sink-after-2022", "Before2022", PodLogCheckOptionWithTimeout(1*time.Second))
```
vigith, line 134: what is the default value for ``PodLogCheckOptionWithTimeout`?

KeranYang: 1 minute. More details about this change in https://github.com/numaproj/numaflow/pull/505#discussion_r1086118383

vigith: Is 1 min a good default value? I think it might be better to reduce that and increase it if required?

KeranYang: I don't have strong opinions on this. On one hand, reducing the timeout could result in false positive test result when we are checking NOT EXIST. On the other hand, we don't want to wait too long if we are sure the sink already finishes its work.

I am thinking we can **keep the 1 minute default for now** and if at any point we notice tests are running slower than we expect because of the default timeout, we can adjust. (for now it's only affecting this specific test and we fix it by explicitly customize timeout value.)

whynowy: Why can't we use the new redis sink for testing?

KeranYang: Redis sink only persists `payload` to redis and the message event time gets dropped. Since in this test case we want to verify that message event time gets updated, I choose to use the log verification.

whynowy: The pod log check supports regex, maybe some of them can be combined? and then increase a little bit on the timeout, 1 second might make it flaky.

whynowy (approved):

PR #523, opened 2023-02-03
chore: update message isLate after source EventTime update

KeranYang:
At source vertex, we allow users to define a customized data transformer to update event time for each of the messages. After messages being processed by transformer, the `isLate` flag can be out-of-date because of EventTime update. This change is to update `isLate` flag before sending to output ISB.

Explain what this PR does.

### DIFF
--- a/pkg/forward/forward.go
+++ b/pkg/forward/forward.go
@@ -260,8 +260,13 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
 			isdf.opts.logger.Errorw("failed to applyUDF", zap.Error(err))
 			return
 		}
-		// update toBuffers
+
 		for _, message := range m.writeMessages {
+			// Update isLate to reflect the message event time change after being processed by source data transformer.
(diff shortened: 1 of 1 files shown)


pkg/forward/forward.go
```diff
@@ -260,8 +260,13 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
 			isdf.opts.logger.Errorw("failed to applyUDF", zap.Error(err))
 			return
 		}
-		// update toBuffers
+
 		for _, message := range m.writeMessages {
+			// Update isLate to reflect the message event time change after being processed by source data transformer.
+			if isdf.opts.vertexType == dfv1.VertexTypeSource && processorWM.After(message.EventTime) {
```
vigith, line 266: this has a problem, `processorWM` could make every data to be late if transformer does a decr operation on eventTime.

KeranYang:
Need to put more thoughts into watermark propagation with source data transformer. Closing this PR...

PR #529, opened 2023-02-07
feat: Offset time idle watermark put

chromevoid:
Add a separate put offset timeline function for assumed idle watermark


vigith (approved):
should we track metrics on how many messages are written to each edge, how many idlers, etc, in a forwardAChunk iteration?

whynowy (approved):

PR #532, opened 2023-02-08
chore: SecurityContext for user defined containers

whynowy:

Explain what this PR does.


vigith (approved):
In some places, we use `numa` and in some places, we use `main` for the numaflow container.

whynowy:
> In some places, we use `numa` and in some places, we use `main` for the numaflow container.

That's right, in the ISB service pods. Didn't want to touch them since they are not running our code.

Issue #533, opened 2023-02-08
`NewUDSGRPCBasedXXX` returns a private var


whynowy:
This is for both UDF and UDSink

PR #535, opened 2023-02-08
chore: expose UDSgRPC type because it is required in healthcheck

vigith:

Fixes: #533


whynowy:
What is the change in health check?

chromevoid:
> What is the change in health check?

Will do the health check refactor in another PR.

PR #536, opened 2023-02-09
chore: metrics server refactor

chromevoid:


pkg/sinks/sinker.go
```diff
@@ -19,10 +19,12 @@ package sinks
...
 	"github.com/numaproj/numaflow/pkg/forward"
 	"github.com/numaproj/numaflow/pkg/isb"
+	"github.com/numaproj/numaflow/pkg/shared/util"
 )
 
 // Sinker interface defines what a Sink should implement.
 type Sinker interface {
 	isb.BufferWriter
 	forward.StarterStopper
+	util.ServerHandler
```
whynowy, line 29: Instead of adding to Sinker interface, which requires all the sinks to implement it, can we do a instance of check in sink.go, then only udsink needs to do it?

pkg/shared/util/wait_until_ready.go
```diff
@@ -0,0 +1,7 @@
+package util
```
whynowy, line 1: Does it make sense to put it in `pkg/metrics` with interface name `HealthChecker`? Or put it in `pkg/shared/interfaces`?

chromevoid: The WaitUntilReady is not only used in the metrics server/metrics pkg. It's also used when we start the grpc connection. 
I renamed it to `ReadyChecker`.. 
`pkg/shared/interfaces` or `pkg/shared/util`? I prefer `util` though...

pkg/shared/interfaces/ready_checker.go
```diff
@@ -0,0 +1,7 @@
+package interfaces
+
+import "context"
+
+type ReadyChecker interface {
```
vigith, line 5: please document

pkg/shared/interfaces/ready_checker.go
```diff
@@ -0,0 +1,7 @@
...
+	WaitUntilReady(ctx context.Context) error
```
vigith, line 6: please doc

vigith (approved):

vigith (approved):

whynowy (approved):

PR #537, opened 2023-02-09
feat: edge-watermark

### LINKED ISSUE #531: Edge Watermark
# Summary

Moving the watermark from vertex to edge
---

veds-g:
Signed-off-by: veds-g <guptavedant2312@gmail.com>

Fixes #531 

Moved the watermark from vertex to edge
Watermark for all the buffers is shown as a separate tab on the edge click

image


vigith:
Can we show it on the edge itself?

veds-g:
> Can we also show it on the edge itself? For multi buffered edges, show the minimum on the edge.

Do we want just the watermark on the edge or both watermark and back-pressure? In case of both I need to explore how to render them both together in a view that makes sense. Right now the way edge-label is defined shows a single string value, we need to have custom label object for this.

vigith:
> Do we want watermark and/or back-pressure on the edge? 

Let's do both since both are very important insights?

vigith:
If it is complex, we can do it via a follow-up PR.

whynowy (approved):
LGTM

whynowy:
1. Instead of showing `Back-pressure` when the mouse hover on the edge, do `Pending`;
2. Show the human readable time when hover on the watermark timestamp;
3. Also display the the human readable time on the `watermarks` tab, `1676488319083(yyyy-MM-dd HH:mm:ss xxx)`

@veds-g

veds-g:
> 1. Instead of showing `Back-pressure` when the mouse hover on the edge, do `Pending`;
> 2. Show the human readable time when hover on the watermark timestamp, also give a little bit more space between the timestamp and the dotted line;
> 3. Also display the the human readable time on the `watermarks` tab, `1676488319083(yyyy-MM-dd HH:mm:ss xxx)`
> 4. On the same `watermarks` tab, display `Partition` instead of `Buffer`, and just use the index (0, 1, 2, xxx) as the value.
> 
> @veds-g

added the changes and updated the view in the description at the top

whynowy (approved):
LGTM

PR #555, opened 2023-02-15
chore: removing start and end time from PaneInfo

### LINKED ISSUE #547: remove Start-time and Endtime from PaneInfo and use Start-time for Event-time and WM
This will make the watermark progression conservative but will not delay the processing time.

yhl25:
- removing start and end time from pane info
- publishing window start time as the watermark
- setting window start time as the event time for the message

fixes #547


test/reduce-e2e/reduce_test.go
```diff
@@ -229,8 +229,8 @@ func (r *ReduceSuite) TestComplexSlidingWindowPipeline() {
 	}()
 
 	w.Expect().
-		SinkContains("sink", "15").
-		SinkContains("sink", "45").
+		SinkContains("sink", "30").
```
vigith, line 232: @yhl25 why will this change? because we change the watermark position?

ashwinidulams: yes that explains why this would change.

ashwinidulams: this pipeline has 5 second fixed keyed window, 10s non keyed fixed followed by 60s sliding with 10s slide. 
so every 10s, we should see 30 being emitted, but by the earlier logic, it was emitting wrong values because of the slide.

test/fixtures/expect.go
```diff
@@ -46,7 +46,7 @@ func (t *Expect) SinkContains(sinkName string, targetStr string, opts ...SinkChe
 	ctx := context.Background()
 	contains := RedisContains(ctx, t.pipeline.Name, sinkName, targetStr, opts...)
 	if !contains {
-		t.t.Fatalf("Expected redis contains target string %s written by pipeline %s, sink %s.", targetStr, t.pipeline.Name, sinkName)
+		t.t.Fatalf("Expected redis doesn't contain target string %s written by pipeline %s, sink %s.", targetStr, t.pipeline.Name, sinkName)
```
ashwinidulams, line 49: shouldn't this still be contains ?

test/reduce-e2e/reduce_test.go
yhl25: yes because the event time in the message is changed to the start time of the partition.

test/fixtures/expect.go
yhl25: this is when the string is not found in sink pod

KeranYang: I think in `expect.go`, we use the following pattern to print fatals. 
```
if Not A {
		t.t.Fatalf("Expected A")
	}
```

yhl25: reverted

KeranYang (approved):

PR #544, opened 2023-02-14
chore: move source data transformation from data forwarder to sourcer

KeranYang:
This change moves source data transformer from the data forwarder to the sourcer. By doing so we make sure we assign source watermarks according to the updated message event times.

The following related changes are included:

* Add de-duplication logics to data forwarder to ensure the offset list we sent to BufferReader Ack() doesn't have duplicates.

* Add PublishSourceWatermarks method to sourcer interface to ensure future sourcers all take care of publishing watermarks.

* Add a simple e2e test to verify watermark propagation after source data transformation.

Signed-off-by: Keran Yang <yangkr920208@gmail.com>

Explain what this PR does.

PR #557, opened 2023-02-15
feat: implement watermark propagation for source data transformer

KeranYang:
See https://github.com/numaproj/numaflow/pull/544#issuecomment-1432009459 about this change.

Explain what this PR does.


pkg/forward/forward.go
```diff
@@ -268,6 +268,25 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
...
 		}
 	}
+
+	if isdf.opts.vertexType == dfv1.VertexTypeSource {
+		var writeMessages []*isb.Message
+		var transformedReadMessages []*isb.ReadMessage
+		for _, m := range udfResults {
+			writeMessages = append(writeMessages, m.writeMessages...)
+			for _, message := range m.writeMessages {
+				transformedReadMessages = append(transformedReadMessages, message.ToReadMessage(m.readMessage.ReadOffset, m.readMessage.Watermark))
```
whynowy, line 278: Add some comments to explain why we need `ReadMessages`

pkg/forward/forward.go
```diff
@@ -268,6 +265,32 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
 			}
 		}
 	}
+
+	if isdf.opts.vertexType == dfv1.VertexTypeSource {
+		if isdf.opts.srcWatermarkPublisher == nil {
+			isdf.opts.logger.Errorw("failed to publish source watermark because no publisher is assigned")
```
whynowy, line 271: `Failed to xxx`

pkg/forward/forward.go
```diff
@@ -268,6 +265,32 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
...
+			for _, message := range m.writeMessages {
+				// we convert each writeMessage to isb.ReadMessage by providing its parent ReadMessage's ReadOffset and Watermark.
+				// which is used later for publishing source watermark.
+				transformedReadMessages = append(transformedReadMessages, message.ToReadMessage(m.readMessage.ReadOffset, m.readMessage.Watermark))
+			}
+		}
+		// Assign isLate and publish source watermark
+		for _, m := range writeMessages {
+			if processorWM.After(m.EventTime) { // Set late data at source level
+				m.IsLate = true
```
whynowy, line 288: Please test and make sure this message the same one in `messageToStep`.

pkg/forward/forward.go
```diff
@@ -207,9 +207,6 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
 	for _, m := range readMessages {
```
whynowy, line 207: Can we merge this for loop to the one in line 239?

pkg/forward/forward.go
KeranYang: Thanks! messageToStep is not holding pointers, hence the IsLate assignment doesn't take effect. Will fix. +1

KeranYang: Fixed by moving the IsLate assignments to before messageToStep assignments.

pkg/watermark/publish/publisher.go
```diff
@@ -115,8 +115,9 @@ func (p *publish) PublishWatermark(wm processor.Watermark, offset isb.Offset) {
 
 	// build value
 	var seq int64
-	if p.opts.isSource || p.opts.isSink {
-		// For source and sink publisher, we don't care about the offset, also the sequence of the offset might not be integer.
+	if p.opts.isSource || p.opts.isSink || offset == nil {
```
whynowy, line 118: Do you see any case that offset is nil, but not a source or sink?

KeranYang: No, I don't. I am adding this extra check to avoid NullPointerException below when we retrieve seq. But it could never be reached.

whynowy: Better not to add it, otherwise it will cover the problem when it really happens.

KeranYang: That's a good point. Would it be better to add the check to beginning of this method? Something like below?

```
if !p.opts.isSource && !p.opts.isSink && offset == nil {
    p.log.Errorw("Unable to publish watermark. offset is nil but the publisher is neither source nor sink.")
}
```

test/e2e/testdata/extract-event-time-from-payload.yaml
```diff
@@ -3,6 +3,9 @@ kind: Pipeline
 metadata:
   name: extract-event-time
 spec:
+  spec:
+    limits:
+      readBatchSize: 1
```
vigith, line 8: without this did you see any problem?

pkg/forward/forward.go
```diff
@@ -250,9 +257,42 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
...
+
+		var writeMessages []*isb.Message
+		var transformedReadMessages []*isb.ReadMessage
+		for _, m := range udfResults {
+			writeMessages = append(writeMessages, m.writeMessages...)
+			for _, message := range m.writeMessages {
+				// we convert each writeMessage to isb.ReadMessage by providing its parent ReadMessage's ReadOffset.
+				// since we use message event time instead of the watermark to determine and publish source watermarks, time.UnixMilli(-1) is assigned to the message watermark.
+				// transformedReadMessages are immediately used below for publishing source watermarks.
+				transformedReadMessages = append(transformedReadMessages, message.ToReadMessage(m.readMessage.ReadOffset, time.UnixMilli(-1)))
```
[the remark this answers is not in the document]
KeranYang: In this case we don't. However, for `isb.Message` type, to convert it to `isb.ReadMessage`, we need to provide two params. Since Go doesn't support method overloading, I let `ToReadMessage` take two params and let caller decide what params to pass in.

test/e2e/testdata/extract-event-time-from-payload.yaml
KeranYang: No. I added this to ensure that in E2E test, each input message is treated as a separate batch. Such that each new event time gets used when publishing watermark(not just the first message).

Based on my observation, removing this is also fine. Probably because in e2e test, I wait 2 seconds before sending next message, which automatically makes each message a separate batch.

vigith: let's not add a 2 sec delay in the data generation for test, it will slow down the test experience. i am fine with this setting, i asked because i was just curious :)

KeranYang: Adding 2 sec delay is to mitigate the watermark propagation delay. Without the 2 sec delay, daemon client getWatermark call returns -1.

vigith (approved):

test/e2e/testdata/extract-event-time-from-payload.yaml
vigith: if we add a batch size of 1, will that problem be solved? since we are propagating for every message.

KeranYang: Unfortunately no. I am still seeing -1.

KeranYang:
@yhl25 @vigith @ashwinidulams @whynowy @chromevoid I performed some manual tests to verify the functionality of this PR and got the following results, which gave me extra confidence to merge this PR. Please review and see if you are good with it and we can target to merge it some time this week. Thanks!

# Source Data Transformer Test Note

This note documents the manual tests I performed to verify the functionality of numaflow [[PR #557](https://github.com/numaproj/numaflow/pull/557)](https://github.com/numaproj/numaflow/pull/557) . The testing pipeline utilized was the [[osam-apigw-numa waterfront at pre-prod stage](https://github.intuit.com/oss-analytics/osam-apigw-numa/tree/master/spec/wf/ppd)](https://github.intuit.com/oss-analytics/osam-apigw-numa/tree/master/spec/wf/ppd).

### Test 1 - Regression Test

During this test, I verified that the functionality of the existing features was not impacted by my PR, by comparing metrics before and after applying my changes, using pipeline [[apigw-wf-go.yaml](https://github.intuit.com/oss-analytics/osam-apigw-numa/blob/master/spec/wf/ppd/apigw-wf-go.yaml)](https://github.intuit.com/oss-analytics/osam-apigw-numa/blob/master/spec/wf/ppd/apigw-wf-go.yaml).

[[Metric board](https://tinyurl.com/yckrzx5c)](https://tinyurl.com/yckrzx5c)

* Deployed the change at Friday noon.
* Monitored the continuity of metric propagations.
* No performance(memory usage/latency/throughput) downgrade.

### Test 2 - Performance Test

During this test, I moved the filtering UDF to source data transformer and compared metrics before and after the change. This is to verify saving one filtering vertex doesn't introduce any performance downgrade.

* Worked with Yashash and verified no performance downgrade.

### Test 3 - Watermark Assignment Test

The goal of this test is to verify after extracing message event time from payload, source vertex is able to use the new event time to publish watermark.

The approach I used to verify is that I created two pipeline with watermark maxDelay disabled. One pipeline is the original pipeline where we use Kafka event time, the other pipeline extract new event time from payload. With watermark maxDelay disabled, we started seeing late messages. I monitored the `no. of dropped messages` metric and observed that the event-time-from-payload pipeline dropped way more messages than original one, proving that the new event time was used to generate watermark.

**Notes**: The reason why extracting event time from payload introduces more late messages is because the event time from payload is at second level while the kafka event time is at millisecond level. With the event-time-from-payload pipeline, multiple messages are group together to share same event time and if within a group, one of the message is late, then the entire group are marked as late.

### Test 4 - Output comparison

During this test, I verified the output of the transformer pipeline is the same as the original pipeline, using Splunk. I spot-checked three [window, key] combinations and verified both pipelines are producing exact same data.

#### Splunk Search Criteria

Pipeline **apigw-wf-go** without transformer:

```
(index=iks) ((sourcetype=eventrouter AND event.involvedObject.namespace=oss-analytics-osamapigwnuma-usw2-qal) OR ( kubernetes_namespace=oss-analytics-osamapigwnuma-usw2-qal)) kubernetes_cluster=ip-msaas-ppd-usw2-k8s host="apigw-wf-go-sink-0-oh3lq" "Intuit.reporting.spreadsheetsyncadmin:Intuit.identity.authz.authzdecisionservice"
```

Pipeline **extract-event-time-from-payload** with transformer:

```
(index=iks) ((sourcetype=eventrouter AND event.involvedObject.namespace=oss-analytics-osamapigwnuma-usw2-qal) OR ( kubernetes_namespace=oss-analytics-osamapigwnuma-usw2-qal)) kubernetes_cluster=ip-msaas-ppd-usw2-k8s host="extract-event-time-from-payload-sink-0-zur9f" "Intuit.reporting.spreadsheetsyncadmin:Intuit.identity.authz.authzdecisionservice"
```

#### Results

For the same window, outputs from both pipelines are exactly same.

https://www.diffchecker.com/A8y2rbQN/

https://www.diffchecker.com/wr1h1QuY/

#### Using a different key

```
"Intuit.identity.relationship.icnccpa:Intuit.platform.servicesplatform.config:servicetoservice"
```

For the same window, outputs from both pipelineS are exactly same.

https://www.diffchecker.com/dXQxEdSz/

KeranYang:
@yhl25 @vigith @ashwinidulams @whynowy @chromevoid I performed some manual tests to verify the functionality of this PR and got the following results, which gave me extra confidence to merge this PR. Please review and see if you are good with it and we can target to merge it some time this week. Thanks!

# Source Data Transformer Test Note

This note documents the manual tests I performed to verify the functionality of numaflow [[PR #557](https://github.com/numaproj/numaflow/pull/557)](https://github.com/numaproj/numaflow/pull/557) . The testing pipeline utilized was the osam-apigw-numa waterfront pipeline at pre-prod stage

### Test 1 - Regression Test

During this test, I verified that the functionality of the existing features was not impacted by my PR, by comparing metrics before and after applying my changes.

* Deployed the change at Friday noon.
* Monitored the continuity of metric propagations.
* No performance(memory usage/latency/throughput) downgrade.

### Test 2 - Performance Test

During this test, I moved the filtering UDF to source data transformer and compared metrics before and after the change. This is to verify saving one filtering vertex doesn't introduce any performance downgrade.

* Worked with Yashash and verified no performance downgrade.

### Test 3 - Watermark Assignment Test

The goal of this test is to verify after extracing message event time from payload, source vertex is able to use the new event time to publish watermark.

The approach I used to verify is that I created two pipeline with watermark maxDelay disabled. One pipeline is the original pipeline where we use Kafka event time, the other pipeline extract new event time from payload. With watermark maxDelay disabled, we started seeing late messages. I monitored the `no. of dropped messages` metric and observed that the event-time-from-payload pipeline dropped way more messages than original one, proving that the new event time was used to generate watermark.

**Notes**: The reason why extracting event time from payload introduces more late messages is because the event time from payload is at second level while the kafka event time is at millisecond level. With the event-time-from-payload pipeline, multiple messages are group together to share same event time and if within a group, one of the message is late, then the entire group are marked as late.

### Test 4 - Output comparison

During this test, I verified the output of the transformer pipeline is the same as the original pipeline, using Splunk. I spot-checked three [window, key] combinations and verified both pipelines are producing exact same data.

vigith:
could you please resolve the conflicts? you can delete those lines, we moved those code to `serde.go`.

PR #559, opened 2023-02-21
feat: namespace search

### LINKED ISSUE #540: Ability to start Numaflow UI on Namespace scope
# Summary

Today the UI server is deployed at the cluster level, we need to support installation at Namespace scope too.

- [ ] remove listing of namespaces from UI
- [ ] ability to list all pipelines in that namespace

veds-g:
Closes #540 
Adding namespace search in the UI for namespace scope installation

- [ ] remove listing of namespaces from UI
- [ ] ability to list all pipelines in that namespace

Caches the previously searched namespace and renders its corresponding pipelines after a refresh

image

image

image

### DIFF
--- a/ui/src/components/namespaces/NamespaceRowContent.tsx
+++ b/ui/src/components/namespaces/NamespaceRowContent.tsx
@@ -10,31 +10,57 @@ interface NamespaceRowContentProps {
 
 export function NamespaceRowContent(props: NamespaceRowContentProps) {
   const { namespaceId } = props;
-  const { pipelines } = useNamespaceFetch(namespaceId);
-  return (
-    <div className={"NamespaceRowContent"} data-testid="namespace-row-content">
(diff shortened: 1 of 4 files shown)


whynowy:
1. Can we make `enter/return` also trigger namespace search?
2. Please add the history namespaces in cookie, and make it be available as a dropdown for quick access.

whynowy:
Some comments:

1. Can we make the left border of the dropdown aligned with the input box:
2. Currently the dropdown items are filtered by the text in the box, can we remove the filtering, which means no matter what is in the text box, the dropdown shows the full history.

whynowy:
@veds-g - it looks pretty good, there's only one thing needs to be addressed, when picking an item from the drop down, it should also trigger a search action, it requires to click the query icon now, which is not convenient.

PR #574, opened 2023-02-28
chore: edge label

veds-g:
Fixing styling of edge label so as to get a consistent view across pipelines

image

image

### DIFF
--- a/ui/src/components/pipeline/graph/CustomEdge.css
+++ b/ui/src/components/pipeline/graph/CustomEdge.css
@@ -4,15 +4,12 @@
 
 .edge-watermark {
     width: fit-content;
-    /*height: 12px;*/
     background: white;
-    /*border: 1px solid #8fa4b1;*/
     position: absolute;
-    /*border-radius: 5px;*/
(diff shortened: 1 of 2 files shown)


vigith:
In the second figure, the labels are on the edge when there are multiple edges. Isn't that inconsistent?

veds-g:
> In the second figure, the labels are on the edge when there are multiple edges. Isn't that inconsistent?

So for vertical edges, are we expecting it to be left and right? If yes, then I need to take a look about how to position the watermark and pending in a custom way, because right now i am moving the pieces up and down based on an initial coordinate that is defined by the graph algo  - which I am not sure we can modify. Since right now there is no specific way to identify if an edge is vertical / horizontal, I need to think of a way to identify them and then place the pieces accordingly.

veds-g:
> In the second figure, the labels are on the edge when there are multiple edges. Isn't that inconsistent?

Also initially the label was just a single element on the center co-ordinate of the edge so I went ahead with that view in mind

vigith (approved):

PR #575, opened 2023-02-28
fix: Idle watermark fix for read batch size > 0 and partial idle outgoing edges

chromevoid:
In the case when a pod doesn't publish to all the out edges, we send an idle watermark whose watermark is assigned as the read batch's watermark.

Note that this solution is when the read in batch size is > 0. Meaning the reduce vertex is receiving incoming messages.

### DIFF
--- a/pkg/watermark/fetch/offset_timeline.go
+++ b/pkg/watermark/fetch/offset_timeline.go
@@ -171,6 +171,27 @@ func (t *OffsetTimeline) GetHeadOffsetWatermark() OffsetWatermark {
 	return t.watermarks.Front().Value.(OffsetWatermark)
 }
 
+// GetReferredWatermark returns the largest offset with the largest watermark.
+func (t *OffsetTimeline) GetReferredWatermark(eventTime int64) OffsetWatermark {
+	t.lock.RLock()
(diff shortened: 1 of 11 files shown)


pkg/watermark/fetch/offset_timeline.go
```diff
@@ -171,6 +171,28 @@ func (t *OffsetTimeline) GetHeadOffsetWatermark() OffsetWatermark {
...
+// GetReferredWatermark returns the watermark that will be used to replace
+// the idle watermark value
+func (t *OffsetTimeline) GetReferredWatermark(eventTime int64) OffsetWatermark {
+	t.lock.RLock()
+	defer t.lock.RUnlock()
+	if eventTime == 0 {
+		// if eventTime is empty, use the head offset watermark
+		return t.watermarks.Front().Value.(OffsetWatermark)
+	}
+	// otherwise, find the OffsetWatermark which watermark is <= the eventTime
```
vigith, line 183:
```suggestion
	// otherwise, find the OffsetWatermark whose referredWM is >= the watermark
```

vigith (approved):

PR #576, opened 2023-03-01
chore: clean up unused routes and rbac

whynowy:
Explain what this PR does.

### DIFF
--- a/server/apis/v1/handler.go
+++ b/server/apis/v1/handler.go
@@ -99,25 +99,6 @@ func (h *handler) GetPipeline(c *gin.Context) {
 	c.JSON(http.StatusOK, pl)
 }
 
-// ListNamespaces is used to provide all the namespaces that have numaflow pipelines running
-func (h *handler) ListNamespaces(c *gin.Context) {
-	ns := c.GetString("namespace")
-	l, err := h.numaflowClient.Pipelines(ns).List(context.Background(), metav1.ListOptions{})
-	if err != nil {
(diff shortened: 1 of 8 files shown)


server/cmd/start.go
```diff
@@ -42,9 +42,6 @@ func Start(insecure bool, port int, namespaced bool, managedNamespace string) {
 	router.Use(gin.Logger())
 	router.RedirectTrailingSlash = true
 	router.Use(static.Serve("/", static.LocalFile("./ui/build", true)))
-	if namespaced {
```
whynowy, line 45: With this, when querying a namespace which is not allowed, it always shows the pipelines of the installation namespace, which is confusing on the UI, and breaks the following up features.

server/apis/v1/handler.go
```diff
@@ -99,25 +99,6 @@ func (h *handler) GetPipeline(c *gin.Context) {
 	c.JSON(http.StatusOK, pl)
 }
 
-// ListNamespaces is used to provide all the namespaces that have numaflow pipelines running
```
vigith, line 102: For cluster scoped, can we use this for type-ahead search?

vigith (approved):

server/apis/v1/handler.go
whynowy: It can be.

However the type-ahead experience might be conflicted with existing dropdown from cookie, it needs some effort to figure out the best user experience if we want to support that. We can bring it back when we implement it.

PR #581, opened 2023-03-02
chore: adding nil check for vertex metrics endpoint and rendering UI with partial data

veds-g:
Fixes metrics api to prevent daemon pod crashes
Fixes UI to render for partial data.

Vertex with some pods not running grays out and shows message on hover, also the processing rates becomes -1.
image

### DIFF
--- a/pkg/daemon/server/service/pipeline_metrics_query.go
+++ b/pkg/daemon/server/service/pipeline_metrics_query.go
@@ -180,6 +180,10 @@ func (ps *pipelineMetadataQuery) GetVertexMetrics(ctx context.Context, req *daem
 		url := fmt.Sprintf("https://%s-%v.%s.%s.svc.cluster.local:%v/metrics", vertexName, i, headlessServiceName, ps.pipeline.Namespace, v1alpha1.VertexMetricsPort)
 		if res, err := ps.httpClient.Get(url); err != nil {
 			log.Debugf("Error reading the metrics endpoint, it might be because of vertex scaling down to 0: %f", err.Error())
+			metricsArr[i] = &daemon.VertexMetrics{
+				Pipeline: &ps.pipeline.Name,
+				Vertex:   req.Vertex,
+			}
 		} else {
 			// expfmt Parser from prometheus to parse the metrics
 			textParser := expfmt.TextParser{}
@@ -225,15 +229,7 @@ func (ps *pipelineMetadataQuery) GetVertexMetrics(ctx context.Context, req *daem
 		}
 	}
 
-	var metricsArrFilter []*daemon.VertexMetrics
-
-	// in case any vertex queried for metrics is not running, this check prevents daemon crashes
-	for _, v := range metricsArr {
-		if v != nil {
-			metricsArrFilter = append(metricsArrFilter, v)
-		}
-	}
-	resp.VertexMetrics = metricsArrFilter
+	resp.VertexMetrics = metricsArr
 	return resp, nil
 }
 
--- a/pkg/daemon/server/service/pipeline_metrics_query.go
+++ b/pkg/daemon/server/service/pipeline_metrics_query.go
@@ -225,14 +225,15 @@ func (ps *pipelineMetadataQuery) GetVertexMetrics(ctx context.Context, req *daem
 		}
 	}
 
+	var metricsArrFilter []*daemon.VertexMetrics
+
 	// in case any vertex queried for metrics is not running, this check prevents daemon crashes
 	for _, v := range metricsArr {
-		if v == nil {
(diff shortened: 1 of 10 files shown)


sandangel (approved):

pkg/daemon/server/service/pipeline_metrics_query.go
```diff
@@ -225,6 +225,13 @@ func (ps *pipelineMetadataQuery) GetVertexMetrics(ctx context.Context, req *daem
 		}
 	}
 
+	// in case any vertex queried for metrics is not running, this check prevents daemon crashes
+	for _, v := range metricsArr {
+		if v == nil {
+			return nil, fmt.Errorf("error while retrieving metrics for the %s vertex", req.GetVertex())
```
vigith, line 231: this will break the UI; let's make sure the UI experience is not broken.

sandangel: I think UI should show no metrics available banner instead of crashing when not getting metrics

sandangel: Or even better UI can say the vertex pod is not running

sandangel: @vigith I think the UI experience is already broken when there is no metrics. This check will not change that

vigith: that is correct, I have asked @veds-g to fix the UI experience too.

sandangel: Is this PR good to merge?

veds-g: i will be adding a UI fix in the same PR

sandangel (approved):

pkg/daemon/server/service/pipeline_metrics_query.go
```diff
@@ -225,7 +225,15 @@ func (ps *pipelineMetadataQuery) GetVertexMetrics(ctx context.Context, req *daem
 		}
 	}
 
-	resp.VertexMetrics = metricsArr
+	var metricsArrFilter []*daemon.VertexMetrics
+
+	// in case any vertex queried for metrics is not running, this check prevents daemon crashes
+	for _, v := range metricsArr {
+		if v != nil {
+			metricsArrFilter = append(metricsArrFilter, v)
```
whynowy, line 233: This is like we do following in line 172, and use `append` to add items.

```
var metricsArr []*daemon.VertexMetrics
```

With this approach, if for some reason, one of the reduce pods fails to get data, then the returned list will miss and skip it, thus it makes the items displayed on the UI incorrect. Can we still return something special in this case, and the UI will determine what to display based on the response, so that it does not lose the position on the UI? For example, set `Pendings`  to -1, and empty `ProcessingRates`.

veds-g: so like I mentioned above, when 1 reduce pod misses the data, the graph shows processing rate for remaining pods and it becomes un-clickable

sandangel:
```sh
{"level":"error","ts":1678258535.5001812,"logger":"numaflow.autoscaler","caller":"scaling/scaling.go:118","msg":"Failed to scale a vertex","vertex":"na-107/conversion-converter","error":"failed to get metrics of vertex key \"na-107/conversion-converter\", rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing dial tcp 172.20.25.202:4327: connect: connection refused\"","stacktrace":"github.com/numaproj/numaflow/pkg/reconciler/vertex/scaling.(*Scaler).scale\n\t/home/runner/work/numaflow/numaflow/pkg/reconciler/vertex/scaling/scaling.go:118"}
```

I think the controller also need to be checked. It's failing to scale vertex

pkg/daemon/server/service/pipeline_metrics_query.go
whynowy: The concern here is not about the UI, but the API itself, the API is not returning the correct data.

Let's assume there are 4 partitions (0, 1, 2, 3) for a reduce vertex (thus there are 4 pods), if the pod 2 has an issue for whatever reason, and no metrics is retrieved. The API expects to see the returned metrics list contains items ordered by the index, however, with this approach, only the data of 0, 1, 3 returned, and the client has no idea which one is missing. On the UI, we have some magic not to display any information at all if we detect the data is incomplete, but it would be misinterpreted by other clients (even though we don't have other clients today).

veds-g: i modified the api to return empty processing rates for partitions that are not available. This way the api has data of all partitions for the clients to act upon.

whynowy (approved):
LGTM, thanks!

PR #597, opened 2023-03-09
chore: use dateparse for extracting event time from payload

### LINKED ISSUE #589: Support extracting event time from an epoch value from payload
# Summary

Our builtin event time extractor uses [time.Parse](https://pkg.go.dev/time#Parse) to convert a time string representation to a time.Time object. Currently, the `time.Parse` only supports converting the time string using a user-provided format string, e.g. `2006-01-02T15:04:05Z07:00`. It doesn't support converting an epoch to time.

A common use case is that the event time extracted from payload is in format of epoch instead of string. We should support parsing epoch in our builtin extractor.

What change needs making?
* Update the argument map of builtin extractor to allow specifying the time string representation is expected in "epoch" format.
* Unit Tests and manual E2E tests,

---

KeranYang:
Closes #589 

This change uses https://github.com/araddon/dateparse for parsing event time string to `time.Time` object. The benefit of using this library is that it can parse many date strings without knowing format in advance, which makes our job easier when we want to parse epoch time string, because we don't need to implement parsing logics for different granularities.

This change also marks the `format` specification as optional. If specified, we will continue using time.Parse to parse the data. But if user doesn't specify `format`, it's also fine as we use `dateparse` to find the format based on time string.

The only uncertainty is that we haven't tested the performance of `dateparse` under a high-traffic numaflow environment. Based on https://github.com/araddon/dateparse/blob/master/bench_test.go, it seems performing pretty well. If we ever find `dateparse` being the bottleneck of performance, we should fallback and implement our own approach to parse epoch.


whynowy:
Does it have a problem for string like `02/03/2023`? It has different meanings in UK and US.

whynowy:
> 

Overall, this is great, remember to have it well ducumented.

KeranYang:
> Does it have a problem for string like `02/03/2023`? It has different meanings in UK and US.

@whynowy Yes, good callout Derek, according to dataparse,

> MM/DD/YYYY VS DD/MM/YYYY Right now this uses mm/dd/yyyy WHEN ambiguous if this is not desired behavior, use ParseStrict which will fail on ambiguous date strings.

I think we should use ParseStrict as we prefer failing the message than parsing it wrong. If user is using such format, we should recommend them to specify format to avoid ambiguity.

whynowy (approved):
LGTM.

Also, add an example when working on the docs.

vigith (approved):
https://github.com/araddon/dateparse is not actively maintained, but it should be okay.

PR #599, opened 2023-03-10
feat: add API for pipeline status check. Fixes #407.

### LINKED ISSUE #407: API/metrics for pipeline health check
# Summary

In terms of the health status of a pipeline, there's nothing to indicate it directly. It would be good to have a health check page or API giving the health status, to make monitoring and alerting easier.

The API/metrics should give a response including:

- Pipeline overall status
- Which vertex or component has what an issue if not healthy.

This API should still be working even if the pipeline is unhealthy, daemon service might be a good place to host this.

---

dpadhiar:
This PR adds an API for a pipeline status check. 

The information returned currently is if the pipeline is "healthy" or not and a message accompanying this. 

Pipeline health here is determined by checking the vertices of the pipeline and comparing their pending message counts to their processing rates. If the processing rate of a vertex of 0 but there are multiple pending messages left, this indicates there may be an issue with the vertex that should be checked.


pkg/daemon/server/service/pipeline_metrics_query.go
```diff
@@ -233,6 +233,40 @@ func (ps *pipelineMetadataQuery) GetVertexMetrics(ctx context.Context, req *daem
...
+	vertices := ps.pipeline.Spec.Vertices
+
+	// loop over vertices and get metrics to check pending messages vs processing rate
+	for _, vertex := range vertices {
+		vertexReq := new(daemon.GetVertexMetricsRequest)
+		vertexReq.Pipeline = req.Pipeline
+		vertexReq.Vertex = &vertex.Name
+		vertexResp, err := ps.GetVertexMetrics(ctx, vertexReq)
+		if err != nil {
+			return nil, err
```
whynowy, line 250: If err is not nil, should not return an error, it probably just because autoscaling down to 0 and then metrics are not available. In this case, we could just return the status as `Unknown`.

pkg/daemon/server/service/pipeline_metrics_query.go
```diff
@@ -233,6 +233,40 @@ func (ps *pipelineMetadataQuery) GetVertexMetrics(ctx context.Context, req *daem
...
+		}
+		// check default pending msg and processing rates
+		if vertexResp.VertexMetrics[0].Pendings["default"] > 0 && vertexResp.VertexMetrics[0].ProcessingRates["default"] == 0 {
```
whynowy, line 253: Should not only check [0], reduce vertex has more items in the array.

Based on the `GetVertexMetrics()` implementation, `default` is not always available in the map, so also need to check if it's existing.

pkg/daemon/server/service/pipeline_metrics_query.go
```diff
@@ -233,6 +233,40 @@ func (ps *pipelineMetadataQuery) GetVertexMetrics(ctx context.Context, req *daem
...
+			resp.Status = &daemon.PipelineStatus{
+				Health:  pointer.Bool(false),
+				Message: pointer.String("Pipeline may have issue."),
```
whynowy, line 256: More information in the message, such as which vertex has what issue.

pkg/metrics/metrics.go
```diff
@@ -46,6 +46,10 @@ const (
 
 	VertexProcessingRate  = "vertex_processing_rate"
 	VertexPendingMessages = "vertex_pending_messages"
+
+	PipelineStatusOK      = "OK"
```
whynowy, line 50: Can we move them to pipeline_metrics_query.go for now?

pkg/apis/proto/daemon/daemon.proto
```diff
@@ -110,4 +124,9 @@ service DaemonService {
   rpc GetPipelineWatermarks (GetPipelineWatermarksRequest) returns (GetPipelineWatermarksResponse) {
     option (google.api.http).get = "/api/v1/pipelines/{pipeline}/watermarks";
   };
+
+  // GetPipelineStatus () returns (PipelineStatus) // status: bool and string
```
whynowy, line 128: Comment is not up to date.

pkg/daemon/server/service/pipeline_metrics_query.go
```diff
@@ -233,6 +239,57 @@ func (ps *pipelineMetadataQuery) GetVertexMetrics(ctx context.Context, req *daem
...
+				if processingRate, ok := vertexMetrics.ProcessingRates["default"]; ok {
+					if pending > 0 && processingRate == 0 {
+						resp.Status = &daemon.PipelineStatus{
+							Status:  pointer.String(PipelineStatusError),
+							Message: pointer.String(fmt.Sprintf("Pipeline has an error. Vertex %s is not processing pending messages.", vertex.Name)),
+						}
+					}
+				}
+			}
+			if vertexMetrics.Pendings["1m"] > 0 && vertexMetrics.ProcessingRates["1m"] == 0 {
```
whynowy, line 276: What I meant was, we don't need to check `1m`.

dpadhiar: Got it, for future reference, what are the cases "default" won't be in the map and why?

whynowy: The `default` will be always in the map if `1m` is there. But if you look at the `GetVertexMetrics` implementation, there's chance the map is empty when the target pod is scaled down to 0 or something is wrong with that pod.

pkg/daemon/server/service/pipeline_metrics_query.go
```diff
@@ -233,6 +239,57 @@ func (ps *pipelineMetadataQuery) GetVertexMetrics(ctx context.Context, req *daem
...
+				if processingRate, ok := vertexMetrics.GetProcessingRates()["default"]; ok {
+					if pending > 0 && processingRate == 0 {
+						resp.Status = &daemon.PipelineStatus{
+							Status:  pointer.String(PipelineStatusError),
+							Message: pointer.String(fmt.Sprintf("Pipeline has an error. Vertex %s is not processing pending messages.", vertex.Name)),
+						}
+						return resp, nil
+					}
+				}
+			} else {
```
whynowy, line 275: We don't need this `else`.

It is expected sometimes there's no `pending` information:

1. Source like `HTTP`, it does not have `pending` at all;
2. Based on the `GetVertexMetrics` implementation, when scaling down to 0 replica, the `pending` is also not available.

For these cases, we should consider it's ok.

whynowy:
@dpadhiar - please also test this api with real pipelines to make sure the output is expected (find the pipeline spec in examples).

dpadhiar:
> @dpadhiar - please also test this api with real pipelines to make sure the output is expected (find the pipeline spec in examples).

Will do, I will test the API with most of them tomorrow morning.

whynowy (approved):

PR #607, opened 2023-03-13
feat: handle watermark barrier for map vertex

chromevoid:
Issue: our current implementation won't propagate the watermark when no events flow into the processing pods. This issue causes the pipeline to stop/stuck and also creates latency in data processing. 

We have added some new concepts:
- WMB: short for watermark barrier. A WMB is a watermark with the "Idle=true/false" information. We use the new "idle" field to determine if the watermark is idle (meaning no incoming or outgoing events) or not.
- control message: a control message is a empty message with it's messageKind set to "WMB".  We use control message to identify if the processing pod is idle or not. 

In this PR, we solve the following use case for map vertex.
1. We read a batch, and the len(batch) != 0.
image
2. We read a batch, and the len(batch) == 0. 
(Note: if the WMB fetched doesn't pass the validation, the flow will return early)
image

Test
In this pipeline, we have both even and odd branches.
The source vertex only produces odd number, so there's no data flow in the even branch.
We can see that the watermark is still propagating in the even branch, and the odd sink logs the messages as expected.
Then we let the source vertex send only string values instead of odd numbers, the odd branch also gets no data now. We can still observe watermarks getting updated.
Later we resume source vertex, and odd sink starts to log again.
Screenshot 2023-03-08 at 2 27 48 PM
image

TODO: 
1. solve the idling situation for reduce vertex.
    - a partition is idle
    - reduce is withholding the watermark (this is the first problem that triggered all of this)
3. solve the idling situation for source vertex.


pkg/watermark/wmb/wmb.go
```diff
@@ -58,3 +58,55 @@ func DecodeToWMB(b []byte) (WMB, error) {
...
+
+type WMBChecker struct {
+	counter int
+	max     int
+	w       WMB
+}
+
+// NewWMBChecker returns a WMBChecker to check if the wmb is idle.
+// If all the iterations get the same wmb, the wmb is
+func NewWMBChecker(numOfIteration int) WMBChecker {
```
yhl25, line 70: the wmb is?

pkg/forward/forward.go
```diff
@@ -564,6 +600,48 @@ func (isdf *InterStepDataForward) whereToStep(writeMessage *isb.Message, message
...
 }
 
+// publishIdleWatermark publishes a ctrl message with isb.Kind set to WMB when write length is zero
+// A WMB is only created if this a new
+func (isdf *InterStepDataForward) publishIdleWatermark(toBuffer isb.BufferWriter, wm wmb.Watermark) {
+	var bufferName = toBuffer.GetName()
+	var wmbMessage = []isb.Message{{Header: isb.Header{Kind: isb.WMB}}}
+
+	if isdf.wmbOffset[bufferName] == nil {
+		if isdf.opts.vertexType == dfv1.VertexTypeSink {
```
yhl25, line 610: ```var wmbMessage = []isb.Message{{Header: isb.Header{Kind: isb.WMB}}}```
you can move this inside the if condition

yhl25: does the compiler take care of that?

yhl25:
I know that pr is WIP, was just understanding the code, so left a few comments.

pkg/watermark/wmb/wmb.go
chromevoid: WMB is short for "watermark barrier" ... which is basically a watermark with idle=true/false information. Open for better name. :-)

yhl25: No, I meant in line 69 the comment is not complete

chromevoid: updated :-D

pkg/forward/forward.go
```diff
@@ -203,23 +207,54 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
...
+
+	// store the offsets of the messages we read from ISB
+	var readOffsets = make([]isb.Offset, len(readMessages))
+	for idx, m := range readMessages {
+		readOffsets[idx] = m.ReadOffset
+		if m.Kind == isb.Data {
+			dataMessages = append(dataMessages, m)
+		}
+	}
+
```
yhl25, line 243: do we need to ack the control messages?

pkg/reduce/readloop/readloop.go
```diff
@@ -149,6 +160,10 @@ func (rl *ReadLoop) Process(ctx context.Context, messages []*isb.ReadMessage) {
 
 	// ack successful messages
 	rl.ackMessages(ctx, successfullyWrittenMessages)
+	if len(ctrlMessages) != 0 {
+		rl.ackMessages(ctx, ctrlMessages)
+	}
+
 	// close any windows that need to be closed.
```
yhl25, line 167: Can we ack these control messages immediately after reading? because if ```len(successfullyWrittenMessages) == 0```  we return early.

yhl25: or ack them before this if check ```len(successfullyWrittenMessages) == 0```

pkg/forward/forward.go
yhl25: nvm we ack it

pkg/reduce/readloop/readloop.go
yhl25: I've fixed it, now the e2e tests should go through fine.

pkg/watermark/fetch/edge_fetcher.go
```diff
@@ -70,12 +102,12 @@ func (e *edgeFetcher) GetHeadWatermark() wmb.Watermark {
...
-		debugString.WriteString(fmt.Sprintf("[Processor:%v] (headoffset:%d) \n", p, o))
-		if o != -1 && o > headOffset {
-			headOffset = o
-			epoch = p.offsetTimeline.GetEventTimeFromInt64(o)
+		var w = p.offsetTimeline.GetHeadWMB()
+		e.log.Debugf("Processor: %v (headOffset:%d) (headWatermark:%d) (headIdle:%t)", p, w.Offset, w.Watermark, w.Idle)
+		debugString.WriteString(fmt.Sprintf("[Processor:%v] (headOffset:%d) (headWatermark:%d) (headIdle:%t) \n", p, w.Offset, w.Watermark, w.Idle))
+		if w.Offset != -1 && w.Offset > headOffset {
+			headOffset = w.Offset
+			epoch = w.Watermark
```
chromevoid, line 110: Note that we change how we get the head watermark:
we directly use the head watermark's epoch value now
previously, we use the head watermark's offset to fetch the watermark

pkg/watermark/fetch/interface.go
```diff
@@ -30,4 +30,6 @@ type Fetcher interface {
 	GetWatermark(offset isb.Offset) wmb.Watermark
 	// GetHeadWatermark returns the latest watermark based on the head offset
 	GetHeadWatermark() wmb.Watermark
```
vigith, line 32: `GetHeadWatermark`, we do not use head offset anymore, right?

pkg/watermark/fetch/interface.go
```diff
@@ -30,4 +30,6 @@ type Fetcher interface {
...
+	// GetHeadWMB returns the latest idle WMB from Vn-1 processors
+	GetHeadWMB() wmb.WMB
```
vigith, line 34: In `GetHeadWMB`'s comment, we explicitly call Vn-1 processors; isn’t it true for `GetHeadWatermark` too?

pkg/watermark/fetch/offset_timeline.go
```diff
@@ -115,25 +107,78 @@ func (t *OffsetTimeline) Put(node OffsetWatermark) {
...
-			if node.offset > elementNode.offset {
+		var elementNode = e.Value.(wmb.WMB)
+		if elementNode.Idle {
+			if node.Watermark < elementNode.Watermark {
+				// should never happen
+				return
+			}
+			if elementNode.Offset == node.Offset {
+				e.Value = wmb.WMB{
+					Idle:      true,
```
vigith, line 125:
```suggestion
					Idle:      node.Idle,
```

why not do this?

pkg/watermark/fetch/offset_timeline.go
```diff
@@ -115,25 +107,78 @@ func (t *OffsetTimeline) Put(node OffsetWatermark) {
...
+				}
+				return
+			}
+		} else {
+			return
+		}
+	}
+}
+
+// PutReferred inserts the referred WMB which replaces the idle watermark into list. It ensures that the list will remain sorted after the insert.
```
vigith, line 162: do you want to comment on what `referred WMB` mean?

pkg/watermark/fetch/offset_timeline.go
chromevoid: because node.Idle here is always true

pkg/watermark/fetch/edge_fetcher.go
```diff
@@ -53,9 +53,40 @@ func NewEdgeFetcher(ctx context.Context, bufferName string, storeWatcher store.W
...
+	var allProcessors = e.processorManager.GetAllProcessors()
+	for _, p := range allProcessors {
+		debugString.WriteString(fmt.Sprintf("[Processor: %v] \n", p))
+		var t = p.offsetTimeline.GetEventTime(inputOffset)
+		if t == -1 { // watermark cannot be computed, perhaps a new processing unit was added or offset fell off the timeline
+			epoch = t
+		} else if t < epoch {
+			epoch = t
+		}
+		if p.IsDeleted() && (offset > p.offsetTimeline.GetHeadOffset()) {
```
whynowy, line 74: Should this go before line 68 with `continue`, and then line 70 could be followed by a `break`?

vigith (approved):

pkg/forward/forward.go
```diff
@@ -95,13 +99,13 @@ func NewInterStepDataForward(vertex *dfv1.Vertex,
...
+		wmbOffset:    make(map[string]isb.Offset, len(toSteps)),
+		wmbChecker:   wmb.NewWMBChecker(2), // TODO: make configurable
 		Shutdown: Shutdown{
 			rwlock: new(sync.RWMutex),
 		},
 		opts: *options,
-	}
-
-	// Add logger from parent ctx to child context.
+	} // Add logger from parent ctx to child context.
```
whynowy, line 108: Accidentally moved the comment here?

PR #616, opened 2023-03-16
fix: corrected reduce vertex replica number. Fixes #593

### LINKED ISSUE #593: Incorrect reduce vertex replica number
**Describe the bug**

Vertex replica information is incorrect for reduce ones with parallelism > 1.

```
kubectl get vtx
NAME                       PHASE     REASON   MESSAGE   DESIRED   CURRENT
even-odd-sum-in            Running                      1         1
even-odd-sum-atoi          Running                      1         1
even-odd-sum-compute-sum   Running                      1         2
even-odd-sum-sink          Running                      1         1
```

dpadhiar:
Fixes #593 

Currently, reduce vertex replica number is incorrect when parallelism is greater than 1. 

This is what is shown for even-odd-sum pipeline from `examples/6-reduce-fixed-window.yaml` where `even-odd-sum-compute-sum` is a reduce vertex.

```
kubectl get vertex
NAME                       PHASE     REASON   MESSAGE   DESIRED   CURRENT
even-odd-sum-in            Running                      1         1
even-odd-sum-atoi          Running                      1         1
even-odd-sum-compute-sum   Running                      1         2
even-odd-sum-sink          Running                      1         1
```

The issue arises from the vertex's spec not accurately being updated to reflect the number of desired replicas.
By updating the spec for reduce vertices, the correct information will be displayed.

```
kubectl get vertex                                                              
NAME                       PHASE     REASON   MESSAGE   DESIRED   CURRENT
even-odd-sum-in            Running                      1         1
even-odd-sum-atoi          Running                      1         1
even-odd-sum-compute-sum   Running                      2         2
even-odd-sum-sink          Running                      1         1
```

### DIFF
--- a/pkg/reconciler/pipeline/controller.go
+++ b/pkg/reconciler/pipeline/controller.go
@@ -444,6 +444,11 @@ func buildVertices(pl *dfv1.Pipeline) map[string]dfv1.Vertex {
 				replicas = *x.Max
 			}
 		}
+		for _, edge := range fromEdges {
+			if edge.Parallelism != nil {
+				replicas = *edge.Parallelism
+			}
+		}
 		spec := dfv1.VertexSpec{
 			AbstractVertex:             *vCopy,
 			PipelineName:               pl.Name,
--- a/pkg/reconciler/pipeline/controller.go
+++ b/pkg/reconciler/pipeline/controller.go
@@ -435,6 +435,12 @@ func buildVertices(pl *dfv1.Pipeline) map[string]dfv1.Vertex {
 		replicas := int32(1)
 		if pl.Status.Phase == dfv1.PipelinePhasePaused {
 			replicas = int32(0)
+		} else if v.UDF != nil && v.UDF.GroupBy != nil {
+			for _, edge := range fromEdges {
+				if edge.Parallelism != nil {
+					replicas = *edge.Parallelism
+				}
+			}
 		} else {
 			x := vCopy.Scale
 			if x.Min != nil && *x.Min > 1 && replicas < *x.Min {
@@ -444,11 +450,7 @@ func buildVertices(pl *dfv1.Pipeline) map[string]dfv1.Vertex {
 				replicas = *x.Max
 			}
 		}
-		for _, edge := range fromEdges {
-			if edge.Parallelism != nil {
-				replicas = *edge.Parallelism
-			}
-		}
+
 		spec := dfv1.VertexSpec{
 			AbstractVertex:             *vCopy,
 			PipelineName:               pl.Name,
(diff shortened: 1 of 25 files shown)


whynowy (changes requested):

pkg/reconciler/vertex/controller.go
```diff
@@ -159,6 +160,13 @@ func (r *vertexReconciler) reconcile(ctx context.Context, vertex *dfv1.Vertex) (
 	currentReplicas := int(vertex.Status.Replicas)
 	if currentReplicas != desiredReplicas || vertex.Status.Selector == "" {
 		log.Infow("Replicas changed", "currentReplicas", currentReplicas, "desiredReplicas", desiredReplicas)
+		if vertex.IsReduceUDF() {
```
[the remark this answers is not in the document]
dpadhiar: In this case, it should in the logic of `func buildVertices(pl *dfv1.Pipeline) map[string]dfv1.Vertex` in the pipeline controller. Working on this now.

pkg/reconciler/pipeline/controller.go
```diff
@@ -435,6 +435,12 @@ func buildVertices(pl *dfv1.Pipeline) map[string]dfv1.Vertex {
 		replicas := int32(1)
 		if pl.Status.Phase == dfv1.PipelinePhasePaused {
 			replicas = int32(0)
+		} else if v.UDF != nil && v.UDF.GroupBy != nil {
```
KeranYang, line 438: @dpadhiar Hey Dillen, I see we are looping through the fromEdges, does it mean if all fromEdges have parallelism, we will always take the last edge's number? I am not quite familiar with reduce replica but that doesn't sound right to me intuitively.

Also two nitpickings: 1. I think you can use `v.IsReduceUDF() instead of `v.UDF != nil && v.UDF.GroupBy != nil` and 2. could you add a unit test case to cover this change?

dpadhiar: @KeranYang Hi Keran, thanks for taking a look. For `v.IsReduceUDF()`, I also tried to use this but since `v` is a `dfv1.AbstractVertex`, I couldn't use this function but I did use the logic outright to check for a reduce vertex.

You are correct that looping through all the from edges and taking the last `Parallelism` value is incorrect.  I am still getting familiar with Numa so I have not seen a UDF with multiple from edges that have different `Parallelism` values. If this was the case, @whynowy is the largest `Parallelism` value used?

As for unit test, this PR is closed but I will open another to include that after we confirm the above.

whynowy: The reason I dismissed the for loop logic is, there's only 1 from edge for a vertex, technically we should only take the 1st item of the array to make logic look right. @dpadhiar - please address this together with the unit test.

dpadhiar: Will take care of this now, should be a quick turnaround.

PR #617, opened 2023-03-17
feat: track and expose oldest work yet to be done to the reduce loop

yhl25:
Track the oldest work yet to be completed in the reduce-loop so that both read-loop and the process-and-forward are in agreement with the watermark.

### DIFF
--- a/pkg/reduce/pbq/pbqmanager.go
+++ b/pkg/reduce/pbq/pbqmanager.go
@@ -27,8 +27,8 @@ import (
 	"github.com/numaproj/numaflow/pkg/metrics"
 	"github.com/numaproj/numaflow/pkg/reduce/pbq/partition"
 	"github.com/numaproj/numaflow/pkg/reduce/pbq/store"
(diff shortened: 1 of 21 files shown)


pkg/reduce/pbq/pbqmanager.go
```diff
@@ -216,6 +233,18 @@ func (m *Manager) register(partitionID partition.ID, p *PBQ) {
 func (m *Manager) deregister(partitionID partition.ID) error {
 	m.Lock()
 	defer m.Unlock()
+
+	ww := &RegisteredWindow{
+		AlignedKeyedWindower: m.pbqMap[partitionID.String()].kw,
+	}
+
+	ww, _ = m.yetToBeClosed.InsertIfNotPresent(ww)
```
vigith, line 241: this is counter-intuitive that it looks like we insert in deregister :)

vigith (approved):

whynowy (approved):

PR #619, opened 2023-03-17
test: add unit test for reduce vtx replica number

dpadhiar:
Add unit test and remove for loop from implementation of #616

### DIFF
--- a/pkg/forward/forward.go
+++ b/pkg/forward/forward.go
@@ -56,6 +56,10 @@ type InterStepDataForward struct {
 	opts             options
 	vertexName       string
 	pipelineName     string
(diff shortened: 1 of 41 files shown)


dpadhiar:
@KeranYang take a look when you can. Thanks!

KeranYang (approved):
Thanks Dillen for making the change. LGTM. It would be better if we can add a comment on why we are taking the `Parallelism` of the **first** fromEdge at this moment. Approving.

dpadhiar:
> Thanks Dillen for making the change. LGTM. It would be even better if we can add a comment on why we are taking the `Parallelism` of the **first** fromEdge at this moment. Approving.

This is a good idea Keran, adding that now!

whynowy (approved):

Issue #524, opened 2023-02-03
kill switch to drop messages when buffer.isFull == true


vigith:
# Summary

We need to have an edge level setting to drop the messages if the `buffer.isFull == true`. Even if the UDF or UDSink drops a message due to some internal error in the user-defined code, the processing latency will spike up causing a natural back pressure. A kill switch to drop messages can help alleviate/avoid any repercussions on the rest of the DAG. 

This setting is an edge-level setting and can be enabled by `dropOnFull` to `true` (default is `false`).
```yaml
  edges:
    - from: a
      to: b
      dropOnFull: true
```

# Use Cases

**NOTE**: This is a data loss scenario but can be useful in cases where we are doing user-introduced experimentations, like A/B testing, on the pipeline. It is totally okay for the experimentation side of the DAG to have data loss while the production is unaffected.

# Considerations
We could use add a `onFull` (instead of `dropOnFull`) flag which could take different values like `drop`, `retry` (current behavior), etc. This will enable us to support other experiences like writing to dlq, etc in the future if we ever need those.

KeranYang:
Adding initial investigation notes.

### Current Behaviour

I checked 3 ISB implementations, **JetStream**, **Redis** and **In-Memory** Buffer. They all use the same logic when writing to a full buffer. It's as follow:

When forwarder(applying to both `reduce` and `non-reduce`) tries to write-to-buffer a batch of messages, if the ToBuffer is full, the entire batch will be marked as failed with `BufferWriteErr`. Then it will keep retrying the batch until all the messages are successfully processed. The forwarder **DO NOT acknowledge** the batch read offsets until all of them are successfully processed.

In short, as vigith mentioned above, current `onFull` is `retryUntilSuccess`.

### The DropOnFull behaviour

From my understanding, the **DropOnFull** behaviour, translating to `onFull`, is `dropAndAcknowledge`. If we drop messages without acknowledging, the batch will still get re-processed later, which contradict with the purpose of this task.

### Implementation Plan

* I will introduce the `onFull` edge level specification as Enum. (It will also be available at pipeline level and allow edge level to override. Like `bufferUsageLimit `)
  * Two enums for now: `retryUntilSuccess` and `dropAndAcknowledge`.
* All 3 existing buffer writer implementations support `options`, I will add a `WithOnFullAction(action enum)` to each of them such that writer can use it to determine how to handle a full buffer.
[...]

PR #634, opened 2023-03-24
feat: enable edge-level kill switch to drop messages when buffer is full, for the non-reduce forwarder

KeranYang:
* Introduce a new pipeline edge-level specification to define write behaviours when a buffer is full.
* Solve the `map` part of issue #524 

I will create follow up PRs for reduce forwarder once this one gets approved.


Makefile
```diff
@@ -135,7 +136,7 @@ Test%:
 	kubectl -n numaflow-system delete po -lapp.kubernetes.io/component=controller-manager,app.kubernetes.io/part-of=numaflow
 	kubectl -n numaflow-system delete po e2e-api-pod  --ignore-not-found=true
 	cat test/manifests/e2e-api-pod.yaml |  sed 's@quay.io/numaproj/@$(IMAGE_NAMESPACE)/@' | sed 's/:$(BASE_VERSION)/:$(VERSION)/' | kubectl -n numaflow-system apply -f -
-	-go test -v -timeout 10m -count 1 --tags test -p 1 ./test/e2e  -run='.*/$*'
+	-go test -v -timeout 10m -count 1 --tags test -p 1 ./test/e2e-suite-1  -run='.*/$*'
```
whynowy, line 139: Instead of dividing to 2, try `-p 2`?

pkg/apis/numaflow/v1alpha1/edge_types.go
```diff
@@ -26,10 +26,15 @@ type Edge struct {
...
-	// if it's provided, the default value is set to "1".
+	// if it's not provided, the default value is set to "1".
 	// Parallelism is ignored when the "to" vertex is not a reduce vertex.
 	// +optional
 	Parallelism *int32 `json:"parallelism" protobuf:"bytes,5,opt,name=parallelism"`
+	// OnFull specifies the behaviour for the write actions when the inter step buffer is full.
+	// if not provided, the default value is set to "", which currently has the same effect as "retryUntilSuccess".
+	// +kubebuilder:validation:Enum="";retryUntilSuccess;dropAndAckLatest
+	// +optional
+	OnFull string `json:"onFull" protobuf:"bytes,6,opt,name=onFull"`
```
whynowy, line 37: 1. Use pointer?
2. Rename `dropAndAckLatest` to `discardLatest`?

pkg/forward/forward.go
```diff
@@ -478,53 +478,73 @@ func (isdf *InterStepDataForward) writeToBuffers(ctx context.Context, messageToS
...
+
+		for idx, msg := range messages {
+			if err := errs[idx]; err != nil {
+				// ATM there are no user defined errors during write, all are InternalErrors.
+
+				// using `strings.Contains` to check BufferFull error type is not a good practice for two reasons:
+				// 1. it's an O(n) operation which, if being called frequently, can introduce performance issue.
+				// 2. it assumes that all the buffer implementations implement err.Error() to use "Buffer full" as error message content.
+				// TODO - a better way could be to declare a BufferFullError type and use type check instead of checking error message.
+				needRetry = !(isdf.onFullActions[toBuffer.GetName()] == dfv1.DropAndAckLatest && strings.Contains(err.Error(), "Buffer full"))
```
whynowy, line 504: Added a function like `OnFullStrategy()` in the model, and defaults to `retry`, then use `==retry`  check here.

pkg/forward/forward.go
```diff
@@ -478,53 +478,73 @@ func (isdf *InterStepDataForward) writeToBuffers(ctx context.Context, messageToS
...
-				needRetry = true
-				// we retry only failed messages
-				failedMessages = append(failedMessages, messages[idx])
-				writeMessagesError.With(map[string]string{metrics.LabelVertex: isdf.vertexName, metrics.LabelPipeline: isdf.pipelineName, "buffer": toBuffer.GetName()}).Inc()
-				// a shutdown can break the blocking loop caused due to InternalErr
-				if ok, _ := isdf.IsShuttingDown(); ok {
-					err := fmt.Errorf("writeToBuffer failed, Stop called while stuck on an internal error with failed messages:%d, %v", len(failedMessages), errs)
-					platformError.With(map[string]string{metrics.LabelVertex: isdf.vertexName, metrics.LabelPipeline: isdf.pipelineName}).Inc()
-					return writeOffsets, err
+		failedMessages := make([]isb.Message, 0, len(messages))
```
whynowy, line 493: This is initializing the length and capacity?

pkg/forward/forward.go
```diff
@@ -478,53 +478,73 @@ func (isdf *InterStepDataForward) writeToBuffers(ctx context.Context, messageToS
...
 			time.Sleep(isdf.opts.retryInterval)
-			goto retry
 		} else {
 			break
 		}
 	}
 
-	writeMessagesCount.With(map[string]string{metrics.LabelVertex: isdf.vertexName, metrics.LabelPipeline: isdf.pipelineName, "buffer": toBuffer.GetName()}).Add(float64(len(messages)))
-	writeBytesCount.With(map[string]string{metrics.LabelVertex: isdf.vertexName, metrics.LabelPipeline: isdf.pipelineName, "buffer": toBuffer.GetName()}).Add(totalBytes)
+	dropMessagesCount.With(map[string]string{metrics.LabelVertex: isdf.vertexName, metrics.LabelPipeline: isdf.pipelineName, "buffer": toBuffer.GetName()}).Add(float64(totalCount - writeCount))
```
whynowy, line 544: > 0 check?

pkg/forward/forward.go
```diff
@@ -478,53 +478,73 @@ func (isdf *InterStepDataForward) writeToBuffers(ctx context.Context, messageToS
...
+	dropBytesCount.With(map[string]string{metrics.LabelVertex: isdf.vertexName, metrics.LabelPipeline: isdf.pipelineName, "buffer": toBuffer.GetName()}).Add(dropBytes)
```
whynowy, line 545: > 0 check?

pkg/isb/interfaces.go
```diff
@@ -51,6 +51,7 @@ type BufferWriter interface {
 	BufferWriterInformation
 	io.Closer
 	Write(context.Context, []Message) ([]Offset, []error)
+	IsFull() bool
```
whynowy, line 54: Do wee need to make it part of the interface? this means all the writers need to implement it, but it's not used anywhere else.

pkg/sinks/udsink/sink.go
```diff
@@ -78,7 +78,7 @@ func NewUserDefinedSink(vertex *dfv1.Vertex, fromBuffer isb.BufferReader, fetchW
 		return nil, fmt.Errorf("failed to create gRPC client, %w", err)
 	}
 	s.udsink = udsink
-	isdf, err := forward.NewInterStepDataForward(vertex, fromBuffer, map[string]isb.BufferWriter{vertex.GetToBuffers()[0].Name: s}, forward.All, applier.Terminal, fetchWatermark, publishWatermark, forwardOpts...)
+	isdf, err := forward.NewInterStepDataForward(vertex, fromBuffer, map[string]isb.BufferWriter{vertex.GetToBuffers()[0].Name: s}, forward.All, map[string]string{vertex.GetToBuffers()[0].Name: dfv1.RetryUntilSuccess}, applier.Terminal, fetchWatermark, publishWatermark, forwardOpts...)
```
whynowy, line 81: Is there a way to make the new arg an option?

pkg/apis/numaflow/v1alpha1/const.go
```diff
@@ -122,6 +122,10 @@ const (
 	DefaultBufferUsageLimit = 0.8
 	DefaultReadBatchSize    = 500
 
+	// Edge-level specification to define the writing behaviour when the ToBuffer is full
+	RetryUntilSuccess = "retryUntilSuccess"
```
whynowy, line 126: Use a `type`?

Makefile
KeranYang: Synced up offline, I tried running E2E tests in parallel, which introduced multiple errors, like port fowarding issue, pipeline creation timeout etc. We agree to keep sequential executions for now.

pkg/forward/forward.go
KeranYang: Done

pkg/forward/forward.go
KeranYang: Addressed.

Yes. I think if retrying is frequent , pre-allocate memory could be more efficient because it help us avoid resizing the slice repeatedly and minimize the number of allocations and copies needed during the append operations.

But I think for now we are not expecting unhappy path, I changed it back.

pkg/forward/forward.go
KeranYang: Synced up offline. I agree it's ok to emit metrics with value 0.

pkg/isb/interfaces.go
KeranYang: Addressed.

pkg/apis/numaflow/v1alpha1/const.go
KeranYang: Addressed.

KeranYang:
Next Step
* Apply same logic to reduce.
* Add doc.

pkg/apis/numaflow/v1alpha1/const.go
```diff
@@ -122,6 +125,10 @@ const (
 	DefaultBufferUsageLimit = 0.8
 	DefaultReadBatchSize    = 500
 
+	// Edge-level specification to define the writing behaviour when the ToBuffer is full
```
whynowy, line 128: Minor - These definitions could stay in the edge_types.go.

pkg/apis/numaflow/v1alpha1/edge_types.go
```diff
@@ -46,3 +52,17 @@ type EdgeLimits struct {
 	// +optional
 	BufferUsageLimit *uint32 `json:"bufferUsageLimit,omitempty" protobuf:"varint,2,opt,name=bufferUsageLimit"`
 }
+
+func (e Edge) OnFullWritingStrategy() OnFullWritingStrategy {
+	if e.OnFull == nil {
+		return RetryUntilSuccess
+	}
+	switch *e.OnFull {
```
whynowy, line 60:
```golang
switch *e.OnFull {
  case RetryUntilSuccess, DiscardLatest:
    return *e.OnFull
  default:
    return RetryUntilSuccess
}
```

pkg/forward/forward.go
```diff
@@ -478,53 +473,67 @@ func (isdf *InterStepDataForward) writeToBuffers(ctx context.Context, messageToS
...
-				// a shutdown can break the blocking loop caused due to InternalErr
-				if ok, _ := isdf.IsShuttingDown(); ok {
-					err := fmt.Errorf("writeToBuffer failed, Stop called while stuck on an internal error with failed messages:%d, %v", len(failedMessages), errs)
-					platformError.With(map[string]string{metrics.LabelVertex: isdf.vertexName, metrics.LabelPipeline: isdf.pipelineName}).Inc()
-					return writeOffsets, err
+		needRetry := false
+		for idx, msg := range messages {
+			if err := errs[idx]; err != nil {
+				// ATM there are no user defined errors during write, all are InternalErrors.
+				if _, ok := err.(isb.NoRetryableBufferWriteErr); ok {
```
whynowy, line 493: Use `errors.Is()`

pkg/apis/numaflow/v1alpha1/edge_types.go
```diff
@@ -26,10 +26,16 @@ type Edge struct {
...
+	// if it's not provided, the default value is set to "1".
 	// Parallelism is ignored when the "to" vertex is not a reduce vertex.
 	// +optional
 	Parallelism *int32 `json:"parallelism" protobuf:"bytes,5,opt,name=parallelism"`
+	// OnFull specifies the behaviour for the write actions when the inter step buffer is full.
+	// There are currently two options, retryUntilSuccess and discardLatest.
+	// if not provided, the default value is set to "retryUntilSuccess"
+	// +kubebuilder:validation:Enum=retryUntilSuccess;discardLatest
+	// +optional
+	OnFull *string `json:"onFull,omitempty" protobuf:"bytes,6,opt,name=onFull"`
```
whynowy, line 38: `OnFull *OnFullWritingStrategy`

pkg/apis/numaflow/v1alpha1/const.go
```diff
@@ -122,6 +125,10 @@ const (
...
+	RetryUntilSuccess OnFullWritingStrategy = "retryUntilSuccess"
```
whynowy, line 129: Use `BufferFullWritingStrategy` to make it more straightforward?

pkg/apis/numaflow/v1alpha1/const.go
KeranYang: done.

pkg/apis/numaflow/v1alpha1/edge_types.go
KeranYang: Cool! done.

pkg/forward/forward.go
KeranYang: Nice, I didn't know about this. I changed it to use `errors.As()`, the difference is that `Is()` checks the error and target error as exactly same `==`, while `As()` checks that `the error's concrete value is assignable to the value pointed to by target error`, which is more close to what I want to do here - type check.

pkg/apis/numaflow/v1alpha1/edge_types.go
KeranYang: done.

pkg/apis/numaflow/v1alpha1/const.go
```diff
@@ -24,6 +24,8 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
+type BufferFullWritingStrategy string
```
whynowy, line 27: Move to edge_types.go?

pkg/forward/forward.go
```diff
@@ -478,53 +474,67 @@ func (isdf *InterStepDataForward) writeToBuffers(ctx context.Context, messageToS
...
-				// a shutdown can break the blocking loop caused due to InternalErr
-				if ok, _ := isdf.IsShuttingDown(); ok {
-					err := fmt.Errorf("writeToBuffer failed, Stop called while stuck on an internal error with failed messages:%d, %v", len(failedMessages), errs)
-					platformError.With(map[string]string{metrics.LabelVertex: isdf.vertexName, metrics.LabelPipeline: isdf.pipelineName}).Inc()
-					return writeOffsets, err
+		needRetry := false
+		for idx, msg := range messages {
+			if err := errs[idx]; err != nil {
+				// ATM there are no user defined errors during write, all are InternalErrors.
+				if ok := errors.As(err, &isb.NoRetryableBufferWriteErr{}); ok {
```
whynowy, line 494: Minor - removing `ok`?

pkg/apis/numaflow/v1alpha1/const.go
```diff
@@ -122,6 +122,9 @@ const (
 	DefaultBufferUsageLimit = 0.8
 	DefaultReadBatchSize    = 500
 
+	RetryUntilSuccess BufferFullWritingStrategy = "retryUntilSuccess"
```
whynowy, line 125: Move these as well?

whynowy (approved):
LGTM

PR #635, opened 2023-03-24
chore: fix TestBuiltinEventTimeExtractor e2e test

### LINKED ISSUE #632: Flaky E2E test - TestBuiltinEventTimeExtractor
TestBuiltinEventTimeExtractor is flaky.

---

KeranYang:
With idle watermark in place, the watermark propagation becomes faster than before, that it seems the edge watermark can be the event time of the latest message. Since the purpose of this specific test case is to verify the event time is extracted from the message payload, I am updating the test to pass if the watermark is the event time extracted from the latest message.

Closes #632


chromevoid:
We have a fix in the reduce wmb PR for the UI's GetHeadWatermark
https://github.com/numaproj/numaflow/pull/627/files#diff-6af877f3295078204591f48a2180bc855394b50997f764275b25ef2ffff678daR107-R114

I think this e2e test fix PR won't be needed once we merged PR627

vigith:
#627 should fix this.

Issue #375, opened 2022-11-21
Numaflow UI access path


martiningka:
# Summary

What change needs making?

I need to be able to access the UI under path different than the root path ("/")

# Use Cases

When would you use this?

I'm installing numaflow in a k8s cluster and I need to access the numaflow UI under <some_host>/numaflow. The current implementation makes this impossible for me, since I can create a k8s ingress for <some_host>/numaflow but not directly to <some_host>

---

**Message from the maintainers**:
It will be nice if you enhance BrowserRouter(index.tsx) with PUBLIC_URL environment variable
`<BrowserRouter basename= { process.env.PUBLIC_URL } >` 

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

dseapy:
in case it's helpful I did some initial work [here](https://github.com/dseapy/numaflow/commit/058a5ae8788c98ecd0436618e29dd904c06284d7) 
 for https://github.com/numaproj/numaflow/issues/229

whynowy:
@dseapy - would you mind raising a PR with your code change?

dseapy:
> @dseapy - would you mind raising a PR with your code change?

done, put it up as a draft PR. have had other higher priority tasks recently, so haven't had a chance to work on fixing the browserrouter issue yet.

louisnow:
@whynowy @dseapy Will we merge the feature to main sometime? This is very useful to avoid having to spin up something like nginx just to change the base path

PR #638, opened 2023-03-27
feat: namespace scope api and disable namespace search on UI

### LINKED ISSUE #578: Remove namespace input box for namespace scoped installation [UI]
# Summary

When the UI server is installed in namespace scope, there's no need to ask the user to enter the namespace in the input box, instead, it should display the namespace in the page, and use it for following up operations.

This will require the UI provide a new API to give the information of:

- Is it installed in namespace scoped;
- What is the the namespace if it's true.

# Use Cases

When would you use this?

---

veds-g:
Closes #578 
Added an api to return whether the UI server is installed in namespace scope
Disable the namespace search on UI when installed in namespace scope

image

### DIFF
--- a/server/cmd/start.go
+++ b/server/cmd/start.go
@@ -42,8 +42,7 @@ func Start(insecure bool, port int, namespaced bool, managedNamespace string) {
 	router.Use(gin.Logger())
 	router.RedirectTrailingSlash = true
 	router.Use(static.Serve("/", static.LocalFile("./ui/build", true)))
-	router.Use(NamespaceStatus(namespaced, managedNamespace))
-	routes.Routes(router)
+	routes.Routes(router, routes.SystemInfo{ManagedNamespace: &managedNamespace, Namespaced: &namespaced})
 	router.Use(UrlRewrite(router))
 	server := http.Server{
 		Addr:    fmt.Sprintf(":%d", port),
@@ -84,14 +83,6 @@ func UrlRewrite(r *gin.Engine) gin.HandlerFunc {
 			c.Request.URL.Path = "/"
 			r.HandleContext(c)
 		}
-		c.Next()
-	}
-}
-
-func NamespaceStatus(namespaced bool, managedNamespace string) gin.HandlerFunc {
-	return func(c *gin.Context) {
-		c.Set("namespaced", namespaced)
-		c.Set("managedNamespace", managedNamespace)
 		c.Next()
 	}
 }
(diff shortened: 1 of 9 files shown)


server/routes/routes.go
```diff
@@ -24,11 +24,24 @@ import (
...
+}
+
 func Routes(r *gin.Engine) {
 	r.GET("/healthz", func(c *gin.Context) {
 		c.Status(http.StatusOK)
 	})
-	v1Routes(r.Group("/api/v1"))
+	rGroup := r.Group("/api/v1")
+	v1Routes(rGroup)
+	rGroup.GET("/namespaces/status", func(c *gin.Context) {
```
whynowy, line 38: The full URI path could be `/api/v1/sysinfo`

dpadhiar:
> @dpadhiar - please look into this implementation, it could be used to solve #375.

Thank you Derek, will take a look.

server/cmd/start.go
```diff
@@ -86,3 +87,11 @@ func UrlRewrite(r *gin.Engine) gin.HandlerFunc {
 		c.Next()
 	}
 }
+
+func NamespaceStatus(namespaced bool, managedNamespace string) gin.HandlerFunc {
```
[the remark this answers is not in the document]
veds-g: just wanted to keep the routes in a common file, can do this instead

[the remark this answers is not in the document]
whynowy: > thought of keeping the routes in a common file, can do this instead. Or instead of setting parameters we can pass them as argument ~ `routes.Routes(router, managedNamespace, namespaced)`. wdyt?

This works too. Or pass the wrapped `SystemInfo`.

server/routes/routes.go
```diff
@@ -24,11 +24,21 @@ import (
 	v1 "github.com/numaproj/numaflow/server/apis/v1"
 )
 
-func Routes(r *gin.Engine) {
+type SystemInfo struct {
+	ManagedNamespace *string `json:"managedNamespace"`
```
whynowy, line 28: Minor - can we use `string` and `bool` directly?

server/routes/routes.go
```diff
@@ -24,11 +24,21 @@ import (
...
+
+func Routes(r *gin.Engine, sysinfo SystemInfo) {
 	r.GET("/healthz", func(c *gin.Context) {
 		c.Status(http.StatusOK)
 	})
-	v1Routes(r.Group("/api/v1"))
+	rGroup := r.Group("/api/v1")
+	v1Routes(rGroup)
+	rGroup.GET("/sysinfo", func(c *gin.Context) {
+		resp := sysinfo
```
whynowy, line 39: Minor - no need to define `resp`.

veds-g:
@whynowy i have modified the `interface.go` along with the changes requested, please have a look at that as well.

whynowy:
fix the test cases.

whynowy (approved):
LGTM

Issue #630, opened 2023-03-23
Vertex keeps scaling out even there is 0 pending (ackpending > 0)


sandangel:
**Describe the bug**

**To Reproduce**

```yaml
# Pipeline spec
limits:
    readBatchSize: 1
    bufferMaxLength: 500
    bufferUsageLimit: 50

    # Vertex spec:
      scale:
        min: 0
        max: 200
        lookbackSeconds: 420
        cooldownSeconds: 60
        zeroReplicaSleepSeconds: 60
        targetBufferAvailability: 100
        replicasPerScale: 20
```

You produce new message that take vertex 7 mins to process, then you'll see even there is only 4 items in the ackpending, and 0 pending item, the vertex still scale out to 73

More information in Slack thread: https://numaproj.slack.com/archives/C03GQ3AMHE1/p1679303465992269

**Expected behavior**

It should stay at 4 pods instead of scale out to 73

**Screenshots**

<img src="https://numaproj.slack.com/files/U04N16MMEA1/F04UST0JGGL/image.png" alt="Pipeline" style="height: 100px; width:100px;" />

**Environment (please complete the following information):**

- Kubernetes: [e.g. v1.18.6]
- Numaflow: [e.g. v0.5.1]

**Additional context**

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

PR #639, opened 2023-03-27
fix: desired replicas should not be greater than pending

whynowy:
1. Excluding `ackpending` from Pending() will cause unexpected scalding down to 0 when it takes a long time to process a message, and it happens there's no real pending messages in the buffer. Reverted that logic.
2. Let the desired replicas <= pending.

### DIFF
--- a/pkg/reconciler/vertex/scaling/scaling.go
+++ b/pkg/reconciler/vertex/scaling/scaling.go
@@ -297,7 +297,7 @@ func (s *Scaler) desiredReplicas(ctx context.Context, vertex *dfv1.Vertex, rate 
 		return 0
 	}
 	if pending == 0 || rate == 0 {
-		// No pending messages, we don't do anything.
+		// Pending is 0 and rate is not 0, or rate is 0 and pending is not 0, we don't do anything.
 		// Technically this would not happen because the pending includes ackpending, which means rate and pending are either both 0, or both > 0.
 		// But we still keep this check here for safety.
 		return int32(vertex.Status.Replicas)
--- a/pkg/reconciler/vertex/scaling/scaling.go
+++ b/pkg/reconciler/vertex/scaling/scaling.go
@@ -296,13 +296,10 @@ func (s *Scaler) desiredReplicas(ctx context.Context, vertex *dfv1.Vertex, rate 
 	if rate == 0 && pending == 0 { // This could scale down to 0
 		return 0
 	}
-	if pending == 0 {
+	if pending == 0 || rate == 0 {
 		// No pending messages, we don't do anything.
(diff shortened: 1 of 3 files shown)


vigith:
this will reopen https://github.com/numaproj/numaflow/issues/630, right?

whynowy:
> this will reopen #630, right?

There's not a perfect solution for all the use cases, we will not do anything for that issue - having something to fix one corner case makes trouble for some other cases. Simply applying a `max` will mitigate that issue.

whynowy:
> > this will reopen #630, right?
> 
> There's not a perfect solution for all the use cases, we will not do anything for that issue - having something to fix one corner case makes trouble for some other cases. Simply applying a `max` will mitigate that issue.

Actually there's already a 200 max in place, in this case, the parameters should be fined tuned.

whynowy:
> > > this will reopen #630, right?
> > 
> > 
> > There's not a perfect solution for all the use cases, we will not do anything for that issue - having something to fix one corner case makes trouble for some other cases. Simply applying a `max` will mitigate that issue.
> 
> Actually there's already a 200 max in place, in this case, the parameters should be fined tuned.

Or maybe some simple logic we can add, like the replica number should not be more than the calculated `pending`.

vigith (approved):

pkg/reconciler/vertex/scaling/scaling.go
```diff
@@ -294,31 +296,40 @@ func (s *Scaler) desiredReplicas(ctx context.Context, vertex *dfv1.Vertex, rate
 	if rate == 0 && pending == 0 { // This could scale down to 0
 		return 0
 	}
+	if pending == 0 {
+		// No pending messages, we don't do anything.
+		// Technically this would not happen because the pending includes ackpending, which means rate and pending are either both 0, or both > 0.
+		// But we still keep this check here for safety.
```
whynowy, line 302: This line is generated by Copilot.

pkg/reconciler/vertex/scaling/scaling.go
```diff
@@ -294,31 +296,40 @@ func (s *Scaler) desiredReplicas(ctx context.Context, vertex *dfv1.Vertex, rate
...
+			desired = int32(vertex.Status.Replicas) + int32(vertex.Spec.Scale.GetReplicasPerScale())
 		}
 		singleReplicaContribution := float64(totalBufferLength-pending) / float64(vertex.Status.Replicas)
 		desired := int32(math.Round(float64(targetAvailableBufferLength) / singleReplicaContribution))
 		if desired == 0 {
 			desired = 1
 		}
-		return desired
 	}
+	if desired > int32(pending) { // For some corner cases, we don't want to scale up to more than pending.
```
whynowy, line 329: `we don't want to scale up to more than pending.` is generated by Copilot.

PR #644, opened 2023-03-29
chore: enable BufferFullWritingStrategy for reduce data forwarder

### LINKED ISSUE #524: kill switch to drop messages when buffer.isFull == true

KeranYang:
* Enable BufferFullWritingStrategy for reduce data forwarder.
* Re-write BufferFullWritingStrategy unit tests for map data forwarder.
* Fix minor typos.

Closes #524

### DIFF
--- a/pkg/reconciler/vertex/scaling/scaling.go
+++ b/pkg/reconciler/vertex/scaling/scaling.go
@@ -40,9 +40,10 @@ import (
 )
 
 type Scaler struct {
-	client    client.Client
+	client client.Client
+	// Key is the vertex namespaced name, Value is the corresponding *list.Element in the vertexList
 	vertexMap map[string]*list.Element
-	// List of the vertex namespaced name, format is "namespace/name"
+	// List of the vertex namespaced name, format is "namespace/vertex-name"
 	vertexList *list.List
 	lock       *sync.RWMutex
 	options    *options
@@ -77,7 +78,7 @@ func (s *Scaler) Contains(key string) bool {
 	return ok
 }
 
-// Length returns how many vetices are being watched for autoscaling
+// Length returns how many vertices are being watched for autoscaling
 func (s *Scaler) Length() int {
 	s.lock.RLock()
 	defer s.lock.RUnlock()
@@ -133,8 +134,8 @@ func (s *Scaler) scale(ctx context.Context, id int, keyCh <-chan string) {
 //	desiredReplicas = targetAvailableBufferLength / singleReplicaContribution
 //
 // Back pressure factor
-// When desiredReplicas > currentReplics:
-// If there's back pressure in the directly connectied vertices, desiredReplicas = currentReplicas-1;
+// When desiredReplicas > currentReplicas:
+// If there's back pressure in the directly connected vertices, desiredReplicas = currentReplicas-1;
 // If there's back pressure in the downstream vertices (not connected), desiredReplicas remains the same.
 func (s *Scaler) scaleOneVertex(ctx context.Context, key string, worker int) error {
 	log := logging.FromContext(ctx).With("worker", fmt.Sprint(worker)).With("vertexKey", key)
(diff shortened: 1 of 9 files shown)


pkg/reconciler/vertex/scaling/scaling.go
```diff
@@ -273,14 +273,14 @@ func (s *Scaler) scaleOneVertex(ctx context.Context, key string, worker int) err
 		directPressure, downstreamPressure := s.hasBackPressure(*pl, *vertex)
 		if directPressure {
 			if current > 1 {
-				log.Debugf("Vertex %s has direct back pressure from connected vertices, decreasing one replica")
+				log.Debugf("Vertex %s has direct back pressure from connected vertices, decreasing one replica", key)
```
vigith, line 276: interesting, i thought linter would catch these :)

vigith (approved):
The control messages write directly using the `xxx.Write` methods; they will fail if the buffer is full. Perhaps we should make them follow the same path? /cc @chromevoid

pkg/reduce/pnf/processandforward.go
```diff
@@ -213,9 +219,16 @@ func (p *ProcessAndForward) writeToBuffer(ctx context.Context, bufferID string,
 		offsets, writeErrs = p.toBuffers[bufferID].Write(ctx, writeMessages)
 		for i, message := range writeMessages {
 			if writeErrs[i] != nil {
-				failedMessages = append(failedMessages, message)
+				if errors.As(writeErrs[i], &isb.NoRetryableBufferWriteErr{}) {
+					// If toBuffer returns us a NoRetryableBufferWriteErr, we drop the message.
+					dropBytes += float64(len(message.Payload))
+					p.log.Debugw("Dropped message", zap.String("bufferID", bufferID), zap.Any("message", message), zap.Error(writeErrs[i]))
```
whynowy, line 225: Do we do the same logging in map? One concern is this will generate lots of logs for a batch.

KeranYang: No we don't. I think I added this line because in reduce forwarder, when writing successfully, we have such log enabled.

```
p.log.Debugw("Forwarded message", zap.String("bufferID", bufferID), zap.Any("message", message))
```

Do we know why we have it for reduce but not map? I also noticed the log is at debug level so maybe it's configured this way intentionally?

KeranYang:
> The control messages write directly using the xxx.Write methods; they will fail if the buffer is full. Perhaps we should make them follow the same path? /cc @chromevoid

@vigith @chromevoid I see that to write ctrl messages, we use `writeOffsets, err := isdf.writeToBuffer(isdf.ctx, toBuffer, ctrlMessage)`, which means the writing will follow the user-specified BufferFullWritingStrategy for the `toBuffer`. It should follow the same path as data messages.

One question, are we sharing buffers between data and ctrl messages? If so, the buffer-level metrics like byte_counts might be messed up...

chromevoid:
> If so, the buffer-level metrics like byte_counts might be messed up as we add bytes for both data and ctrl messages.

you're right

whynowy (approved):

Issue #626, opened 2023-03-21
Goroutine leak?


whynowy:
All following returns will end up with a new loop with a new message reading, and there's no acknowledgement on the previous read messages, thus the `InProgress` [Goroutines](https://github.com/numaproj/numaflow/blob/main/pkg/isb/stores/jetstream/reader.go#L289) won't be cancelled. Also, because those read messages are always `InProgress`, they won't be read by next reading.

https://github.com/numaproj/numaflow/blob/main/pkg/forward/forward.go#L340
https://github.com/numaproj/numaflow/blob/main/pkg/forward/forward.go#L346
https://github.com/numaproj/numaflow/blob/main/pkg/forward/forward.go#L355

PR #647, opened 2023-03-30
chore: fix goroutine leak at messages forwarding via jetstream

xdevxy:
Closes https://github.com/numaproj/numaflow/issues/626

This PR enables go routine leak detector in `forward` package and `jetstream` package, and fixes a bunch of go routine leak  failure detect by it. It also fixes the goroutine leak at messages forwarding via jetstream by cancel the `workInProgess` goroutine during message processing failure cases.


pkg/isb/interfaces.go
```diff
@@ -92,6 +94,10 @@ type Offset interface {
 	// This is often used when the BufferReader can not simply use the offset identifier to ack the message,
 	// then the work can be done in this function, and call it in BufferReader Ack() function implementation.
 	AckIt() error
+	// NoAck to indicate the offset no longer needs to be acknowledged
+	// It is used when error occur, and we want to reprocess the batch to indicate acknowledgement no
+	// longer needed.
+	NoAck()
```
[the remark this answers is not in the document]
xdevxy: Yeah, I also thought about that approach, but considering we are launching a goroutine for each offset to cancel the `workInProgress` call, if we do it in `forward.go` then for any source we are launching that much goroutines and do nothing, seems to be a waste of resource and not ideal as different source may want to have different behavior for `NoAck()`.

pkg/forward/forward.go
```diff
@@ -332,16 +332,20 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
 	// let's figure out which vertex to send the results to.
 	// update the toBuffer(s) with writeMessages.
 	for _, m := range udfResults {
-		// look for errors in udf processing, if we see even 1 error let's return. handling partial retrying is not worth ATM.
+		// look for errors in udf processing, if we see even 1 error flag the error boolean and
```
vigith, line 335:
```suggestion
		// look for errors in udf processing, if we see even 1 error 
```

didn't understand what the "flag the error boolean" meant.

pkg/forward/forward_test.go
```diff
@@ -63,6 +65,10 @@ func (t *testForwardFetcher) Close() error {
 	return nil
 }
 
+func TestMain(m *testing.M) {
+	goleak.VerifyTestMain(m)
```
vigith, line 69: 👍

pkg/isb/stores/jetstream/reader.go
```diff
@@ -322,6 +339,15 @@ func (o *offset) AckIt() error {
 	return nil
 }
 
+func (o *offset) NoAck() {
+	if o.cancelFunc != nil {
+		o.cancelFunc()
+	}
+	// Ignore the returned error as the worst case the message will
+	// take longer to be redelivered.
+	_ = o.msg.Nak()
```
vigith, line 348: should we log it?

pkg/isb/stores/jetstream/reader.go
```diff
@@ -270,6 +270,23 @@ func (jr *jetStreamReader) Ack(_ context.Context, offsets []isb.Offset) []error
...
+		go func(o isb.Offset) {
+			defer wg.Done()
+			o.NoAck()
+		}(o)
+	}
+	go func() {
+		wg.Wait()
+		close(done)
+	}()
+	<-done
```
vigith, line 287:
```suggestion
	wg.Wait()
```

Won't this do? Do we need go routine? You can avoid the done channel too.

pkg/isb/stores/jetstream/reader.go
```diff
@@ -270,6 +270,23 @@ func (jr *jetStreamReader) Ack(_ context.Context, offsets []isb.Offset) []error
 	return errs
 }
 
+func (jr *jetStreamReader) NoAck(_ context.Context, offsets []isb.Offset) {
+	done := make(chan struct{})
```
whynowy, line 274: this is not needed.

pkg/forward/forward.go
xdevxy: Ah sorry, typo here, was intended for a logic I removed later. Will remove

pkg/isb/stores/jetstream/reader.go
```diff
@@ -339,13 +334,16 @@ func (o *offset) AckIt() error {
...
+func (o *offset) NoAck(ctx context.Context) {
 	if o.cancelFunc != nil {
 		o.cancelFunc()
 	}
 	// Ignore the returned error as the worst case the message will
 	// take longer to be redelivered.
-	_ = o.msg.Nak()
+	log := logging.FromContext(ctx).With("Offset", o.String())
+	if err := o.msg.Nak(); err != nil {
+		log.Debug("Failed to nak the message")
```
vigith, line 345: this should not be in Debug log level, Error level?, and also please print the error message too.

@whynowy adding log here has value?, since we have to get ctx into this.

whynowy: Letting `offset.NoAck()` return an error, and get the logged by `reader.NoAck()` might be better.

xdevxy: Makes sense, updated.

whynowy (approved):
LGTM

Issue #652, opened 2023-03-31
flaky: TestReduceDataForward_{Max,Sum,NonKeyed}


vigith:
https://github.com/numaproj/numaflow/actions/runs/4570493173/jobs/8067881582?pr=648#step:7:4561

vigith:
TestReduceDataForward_NonKeyed is also failing https://github.com/numaproj/numaflow/actions/runs/4570731921/jobs/8068367055

PR #653, opened 2023-03-31
fix: unit test

chromevoid:
fixes https://github.com/numaproj/numaflow/issues/652


vigith (approved):

vigith (dismissed):

chromevoid:
Root cause:
The size of the in-mem kvStore is 10. In the test we don't have any watchers for the toBuffer..
Now that we publish idle watermark, it quickly occupies all the spaces in the in-mem kvStore. Therefore, no more idle watermark can be written to the `kvWatchCh`

vigith (approved):

PR #654, opened 2023-03-31
feat: incremental search and namespace preview in search bar

### LINKED ISSUE #606: Numaflow UI Search Bar Improvement
# Summary

What change needs making?
The current search bar needs the exact namespace name to find the result. Could we add, for example, incremental search?
Also it would be great if we can have a preview of all available namespaces.

# Use Cases

When would you use this?
I can enter partial name, for example, "test", to get all the available namespaces that have the word "test" in the name.
When I open the UI I can see a list of available namespaces and click the one I want from the list to enter the pipeline view.

veds-g:
Closes #606 

enhanced the namespace search bar to provide incremental search
provided a view of all available cluster namespaces on the search dropdown

image
image

### DIFF
--- a/server/apis/v1/handler.go
+++ b/server/apis/v1/handler.go
@@ -322,6 +322,25 @@ func (h *handler) GetPipelineStatus(c *gin.Context) {
 	c.JSON(http.StatusOK, l)
 }
 
+// ListNamespaces is used to provide all the namespaces that have numaflow pipelines running
+func (h *handler) ListNamespaces(c *gin.Context) {
+	ns := c.GetString("namespace")
+	l, err := h.numaflowClient.Pipelines(ns).List(context.Background(), metav1.ListOptions{})
+	if err != nil {
+		c.JSON(http.StatusInternalServerError, err.Error())
+		return
+	}
+	m := make(map[string]bool)
+	for _, pl := range l.Items {
+		m[pl.Namespace] = true
+	}
+	var namespaces []string
+	for k := range m {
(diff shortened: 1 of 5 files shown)


server/apis/v1/handler.go
```diff
@@ -322,6 +322,25 @@ func (h *handler) GetPipelineStatus(c *gin.Context) {
 	c.JSON(http.StatusOK, l)
 }
 
+// ListNamespaces is used to provide all the namespaces that have numaflow pipelines running
+func (h *handler) ListNamespaces(c *gin.Context) {
+	ns := c.GetString("namespace")
```
whynowy, line 327: Do we need this?

server/apis/v1/handler.go
```diff
@@ -322,6 +322,25 @@ func (h *handler) GetPipelineStatus(c *gin.Context) {
...
+	l, err := h.numaflowClient.Pipelines(ns).List(context.Background(), metav1.ListOptions{})
```
whynowy, line 328: The `ns` here should be always empty string.

server/apis/v1/handler.go
veds-g: do you mean the api? it is to provide a preview of all available namespace to the user

server/apis/v1/handler.go
veds-g: right, it was the previous implementation i didn't modify it

whynowy (approved):

PR #661, opened 2023-04-03
chore: generator : max inflight go routines should be 1

ashwinidulams:
while testing back pressure scenarios, we need to make sure that the source handles back pressure appropriately.  while handling back pressure, we also need to make sure that all the records are generated and nothing is skipped.


docs/user-guide/user-defined-functions/reduce/reduce.md
vigith, line 1: are these changes intended?

pkg/sources/generator/tickgen.go
```diff
@@ -312,12 +312,60 @@ func (mg *memgen) Start() <-chan struct{} {
...
+}
+
 // generator fires once per time unit and generates records and writes them to the channel
 func (mg *memgen) generator(ctx context.Context, rate int, timeunit time.Duration) {
 	go func() {
-		var rCount int32 = 0
-		// we are capping the limit at 10000 msgs / second
-		var limit = int32(10000 / rate)
+		// capping the rate to 10000 msgs/sec
+		if rate > 10000 {
```
vigith, line 355: why? if we are doing this, we need to at least log it so the user will know about this.

pkg/sources/generator/tickgen.go
```diff
@@ -312,12 +312,60 @@ func (mg *memgen) Start() <-chan struct{} {
...
+			rate = 10000
+		}
+
+		tickChan := make(chan time.Time, math.MaxInt32)
+		doneChan := make(chan struct{})
+		childCtx, cancel := context.WithCancel(ctx)
```
vigith, line 361:
```suggestion
		childCtx, cancel := context.WithCancel(ctx)
		defer cancel()
```
you can remove the explicit cancel in https://github.com/numaproj/numaflow/pull/661/files#diff-1145efea11df28ccc255ebaef330c426040ffeeaef5a9e6a7978713e837ec837R377

ashwinidulams: wanted to wait until the other routine exits  cleanly.

pkg/sources/generator/tickgen.go
ashwinidulams: done.

vigith (approved):

PR #663, opened 2023-04-04
docs: add metric docs for dropping messages and some of the reduce me…

KeranYang:
…trics

Explain what this PR does.

### DIFF
--- a/pkg/reduce/pnf/metrics.go
+++ b/pkg/reduce/pnf/metrics.go
@@ -60,8 +60,8 @@ var dropBytesCount = promauto.NewCounterVec(prometheus.CounterOpts{
 
 // platformError is used to indicate the number of Internal/Platform errors
 var platformError = promauto.NewCounterVec(prometheus.CounterOpts{
-	Subsystem: "platform",
-	Name:      "error_total",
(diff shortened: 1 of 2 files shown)


pkg/reduce/pnf/metrics.go
```diff
@@ -60,8 +60,8 @@ var dropBytesCount = promauto.NewCounterVec(prometheus.CounterOpts{
 
 // platformError is used to indicate the number of Internal/Platform errors
 var platformError = promauto.NewCounterVec(prometheus.CounterOpts{
-	Subsystem: "platform",
```
KeranYang, line 63: To be consistent with and also distinguished from map forwarder's platform error metric.

vigith: do we need to change the grafana dashboard?

vigith (approved):

KeranYang:
> do we need to change the grafana dashboard?

Good point Vigith. We don't have reduce metrics in example grafana dashboard template. We can work on enriching the template later.

I made this change assuming there is no existing dashboard having this metric. If my assumption is wrong, users need to make dashboard updates accordingly.

Issue #538, opened 2023-02-09
Enhance message passing to UDF with metadata for tracking and custom logics


sandangel:
# Summary

What change needs making?
I'm wondering if we could add some metadata to the message to make it easier to handle logic inside UDF. 
I'm thinking of
- message-id, 
- last proceeded pod name, from which vertex
- how many times the pipeline has proceeded the same message

# Use Cases

When would you use this?

One important aspect of the pipeline is failure handling and retry. Right now, the workflow will continuously retry the message that has not been returned by UDF. Which mean if there is a message that make the pod crash (e.g. OOM killed), it will continuously be proceeded and crash the pods. So I'm thinking a solution for this problem:

- First, I think we can expose the unique message ID that user can use for tracking how many times the message has failed to proceed before decide to continue or forward it to another sink (like reporting this has tried many times and failed)
- Second, I think the pipeline can even more robust and track by itself how many times it has passed the same message to a pod. That will make it much easier for users and simplify the logic inside UDF. User can just check the retried times, add a key (labels) to the message and let the pipeline decide what to do next based on conditional forwarding in pipeline config.
- Last but not least, the last proceeded pod name or the executor that processed the message could be used for debugging purpose where the user can find the exact pods that had errors and check their logs/metrics

Hopefully these use cases are strong enough that the team can consider adding to the code. If you have any question, please let me know. I'm happy to answer because I'm also building a data process pipeline for our team and have a good understanding of what do we need from the orchestration.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #664, opened 2023-04-04
chore: add `numDelivered` field in message

xdevxy:
As part of https://github.com/numaproj/numaflow/issues/538, this PR adds numDelivered field in the message header as a metadata information to be exposed later to UDF via Datum field. It also adds the field when `applyUDF`. But as Sink we only have write message, we cannot inject this metadata information ATM.

Note that now only jetstream ISB has the `numDelivered` information exposed to indicate how many times the message has been delivered, as redis stream ISB doesn't expose such information when read the message. This can be distinguished by the value whether or not is `0`.


pkg/isb/message.go
```diff
@@ -61,6 +61,8 @@ type Header struct {
 	ID string
 	// Key is (key,value) in the map-reduce paradigm which will be used for conditional forwarding.
 	Key string
+	// NumDelivered is the number of times the message has been delivered.
+	NumDelivered uint64
```
whynowy, line 65: Considering it's not a major property, shall we put it in something like `metadata`?

whynowy (approved):

whynowy:
@vigith - can you take a look as well?

pkg/isb/serde_test.go
```diff
@@ -363,9 +371,10 @@ func TestReadMessage(t *testing.T) {
 				t.Errorf("UnmarshalBinary() error = %v, wantUnmarshalError %v", err, tt.wantMarshalError)
 				return
 			}
-			if !reflect.DeepEqual((*newRM).Message, tt.wantData.Message) &&
-				!reflect.DeepEqual((*newRM).Watermark, tt.wantData.Watermark) &&
-				!reflect.DeepEqual((*newRM).ReadOffset.String(), tt.wantData.ReadOffset.String()) {
+			if !reflect.DeepEqual((*newRM).Message, tt.wantData.Message) ||
+				!reflect.DeepEqual((*newRM).Watermark, tt.wantData.Watermark) ||
+				!reflect.DeepEqual((*newRM).ReadOffset.String(), tt.wantData.ReadOffset.String()) ||
+				!reflect.DeepEqual((*newRM).Metadata, tt.wantData.Metadata) {
```
vigith, line 377: 👍

PR #666, opened 2023-04-05
fix: skip empty Kafka partitions when calculating pending count

### LINKED ISSUE #642: Kafka source pod keep recreating for multi-partition kafka topic
**Describe the bug**

Kafka source pod fails to `Terminate` on default heartbeat and keeps recreating the pods when configured with Kafka topic which is having multiple partitions.

**To Reproduce**

*Local environment setup:*

* Installed Minikube locally
* Added Numaflow to the cluster from https://numaflow.numaproj.io/quick-start/#installation 
* Added helm chart of Kafka from here: https://bitnami.com/stack/kafka/helm
* Created a Kafka topic with the name `input-topic-partition` and the config looks like the below (having 2 Partitions).
Screenshot 2023-03-29 at 10 40 40 AM

*Pipeline:*

Installed a sample Kafka pipeline consisting of 3 vertices with the name `sample-kafka-pipeline`,
* Kafka source vertex to read the message from a Kafka topic
* UDF to do some processing over the message (set the expression to always true)
* Kafka sink to push message to a different Kafka topic
```apiVersion: numaflow.numaproj.io/v1alpha1
kind: Pipeline
metadata:
  name: sample-kafka-pipeline
spec:
  # Kafka Pipeline
  vertices:
    # Sources
    - name: input
      source:
        kafka:
          brokers:
            - 'local-kafka-0.local-kafka-headless.default.svc.cluster.local:9092'
          topic: input-topic-partition
          consumerGroup: numaflow-consumer-group

    # User defined functions
    - name: filter-messages
      udf:
        builtin:
          name: filter
          kwargs:
            expression: 1 < 100

    # Sinks
    - name: output-sink
      sink:
        kafka:
          brokers:
            - 'local-kafka-0.local-kafka-headless.default.svc.cluster.local:9092'
          topic: output-topic

  # The relationship between the vertices
  edges:
    - from: input
      to: filter-messages
    - from: filter-messages
      to: output-sink

```
**Steps to reproduce the behaviour:**
* After installing the pipeline in the cluster (in default namespace), add a few messages to the topic `input-topic-partition`.
* Wait for some time so that numaflow-controller will kill those vertices pods because of 10 heartbeats.
* Trigger a new message again and watch the pods in the cluster.
* Can be seen that multiple pods keep recreating with the name prefix `sample-kafka-pipeline-input-*` 

**Expected behaviour**
There should be one pod for the source as configured. but the pods keep creating whenever the controller tries to Terminate the pods and create those back.

**Screenshots**

Pods: 
Screenshot 2023-03-29 at 11 08 56 AM

Logs:

```kubectl logs sample-kafka-pipeline-input-0-jahju```
image

```kubectl logs sample-kafka-pipeline-daemon-9cb96c99f-gwhrb```
image

- kubeletVersion: v1.25.3
- Minikube: v1.28.0
- Numaflow: v0.7.2

**Additional context**

---

### PR DESCRIPTION
**Pull request by KeranYang**
closes #642 

### The problem
When source Kafka topic has multiple partitions and user is sending very few messages, there can be cases when some of the partitions are empty.

For partitions that are empty (let's call those idle partitions), `sarama.ClusterAdmin` `ListConsumerGroupOffsets` API returns us block offset of `-1` to indicate there is no offset associated with that specific partition. Meanwhile when we call `r.saramaClient.GetOffset(r.topic, partition, sarama.OffsetNewest)` to get the newest available offset for idle partitions, it returns us 0. Eventually when we calculate total pending messages using `totalPending += partitionOffset - block.Offset`, for idle partitions, we are always counting (0-(-1)=) 1 for each idle partitions.

Therefore even if we have no input messages at all, if input topic has multiple partitions, the source Kafka vertex will never scale down to 0 pod.

### The fix
When calculating pending count for Kafka, always skip the idle partitions.

### Testing
* Before this change, kafka source vertex doesn't scale down when there is no traffic.
* After this change, I verified it can scale down to 0 pods and scale up when I start sending messages.

### DIFF
--- a/pkg/sources/kafka/reader.go
+++ b/pkg/sources/kafka/reader.go
@@ -183,7 +183,6 @@ func (r *KafkaSource) Ack(_ context.Context, offsets []isb.Offset) []error {
 			r.logger.Errorw("Unable to extract partition offset of type int64 from the supplied offset. skipping and continuing", zap.String("suppliedoffset", offset.String()), zap.Error(err))
 			continue
 		}
-		r.logger.Infof("KeranTest - Marking offset %d as consumed for topic %s and partition %d", poffset, topic, partition)
 
 		// we need to mark the offset of the next message to read
 		r.handler.sess.MarkOffset(topic, partition, poffset+1, "")
@@ -246,19 +245,12 @@ func (r *KafkaSource) Pending(ctx context.Context) (int64, error) {
 		return isb.PendingNotAvailable, nil
 	}
 	partitions, err := r.saramaClient.Partitions(r.topic)
-	r.logger.Infof("KeranTest - Partitions: %v", partitions)
-	// Always [0,1]
 	if err != nil {
-		// Never reached
-		r.logger.Errorf("KeranTest - Failed to get partitions: %v", err)
 		return isb.PendingNotAvailable, fmt.Errorf("failed to get partitions, %w", err)
 	}
 	totalPending := int64(0)
 	rep, err := r.adminClient.ListConsumerGroupOffsets(r.groupName, map[string][]int32{r.topic: partitions})
-	r.logger.Infof("KeranTest - OffsetFetchResponse: %v", rep)
 	if err != nil {
-		// Never reached
-		r.logger.Errorf("KeranTest - Failed to list consumer group offsets: %v", err)
 		return isb.PendingNotAvailable, fmt.Errorf("failed to list consumer group offsets, %w", err)
 	}
 	for _, partition := range partitions {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by whynowy** [comment]
Nice!

**Comment by whynowy** [comment]
> closes #642
> 
> ### The problem
> When source Kafka topic has multiple partitions and next vertex has one single pod, only one of the partitions is actively being consumed.

- Do you mean some of the partitions has no data?

> 
> For partitions that are not consumed (let's call those idle partitions), `sarama.ClusterAdmin` `ListConsumerGroupOffsets` API returns us block offset of `-1` to indicate there is no offset associated with that specific partition. Meanwhile when we call `r.saramaClient.GetOffset(r.topic, partition, sarama.OffsetNewest)` to get the newest available offset for idle partitions, it returns us 0. Eventually when we calculate total pending messages using `totalPending += partitionOffset - block.Offset`, for idle partitions, we are always counting (0-(-1)=) 1 for each idle partitions.
> 
> Therefore even if we have no input messages at all, if input topic has multiple partitions, the source Kafka vertex will never scale down to 0 pod.
> 
> ### The fix
> When calculating pending count for Kafka, always skip the idle partitions.
> 
> ### Testing
> * Before this change, kafka source vertex doesn't scale down when there is no traffic.
> * After this change, I verified it can scale down to 0 pods and scale up when I start sending messages.

**Comment by vigith** [comment]
Nice work! 

> When source Kafka topic has multiple partitions and next vertex has one single pod, only one of the partitions is actively being consumed.

Are you saying if we have 2 partitions with one consumer pod in the source vertex, then there will be a partition whose data will never be consumed?

**Comment by KeranYang** [comment]
> Do you mean some of the partitions have no data?
@whynowy Yes. I updated the PR description to reflect the fact.

> Are you saying if we have 2 partitions with one consumer pod in the source vertex, then there will be a partition whose data will never be consumed?
@vigith Sorry I put some mis-leading descriptions there. I have updated the PR description. The real case is that there can be partitions with no data.

**Review by vigith — APPROVED** [review]

File: pkg/sources/kafka/reader.go [hunk 1]
```diff
@@ -254,11 +254,16 @@ func (r *KafkaSource) Pending(ctx context.Context) (int64, error) {
 		return isb.PendingNotAvailable, fmt.Errorf("failed to list consumer group offsets, %w", err)
 	}
 	for _, partition := range partitions {
+		block := rep.GetBlock(r.topic, partition)
+		if block.Offset == -1 {
+			// This partition has not been processed yet.
```
whynowy, line 259: This comment is still misleading, update it together with some other PRs.

vigith: I thought it was valid, thinking what the comment meant was this partition had never been processed; hence it is -1. If it ever read something, it wouldn't be -1.

@KeranYang is this understanding correct?

KeranYang: @vigith Yes your understanding is correct. I think the word `process` can be misleading as people can think that `not been processed`  means the partition has data but not being consumed yet. I can add more details and explicitly mention the partition has no data.

PR #668, opened 2023-04-06
feat: introducing tags for conditional forwarding

### LINKED ISSUE #665: Conditional forwarding with tags
Using `keys` for both conditional forwarding and aggregation causes problems, it would be good to introduce a new filed `tags` to the SDK udf/transformer interface, and use it for conditional forwarding only.

- “tags” could be a list of strings.
- The lifecycle of the “tags” starts from the returning of the UDF/Transformer function to pre-forwarding to the next ISB.
“- Tags” will be a property of UDF [Message](https://github.com/numaproj/numaflow-go/blob/main/pkg/function/message.go#L11-L14), when getting the response from UDF, we need to introduce a new struct to wrap the converted ISB Message and the tags, this new struct will be used in the forwarder for conditional forwarding.
.
- Some helper functions in UDF [Message](https://github.com/numaproj/numaflow/blob/main/pkg/isb/message.go):

```golang
// Most map functions do not need keys.
// Do not need ALL any more, by default it’s to all unless user define
// forward conditions in the spec.
func New(value []byte) Message {
  return Message{value: value}
}

// New a message with keys and value
func NewWithKeys(keys []string, value []byte) Message {
  return Message{keys: keys, value: value}
}


func MessageToDrop() Message {
  return Message{tags: []string{DROP}, value: []byte{}}
}


func (m Message) WithKeys(keys []string) Message {
  m.keys = keys
  return m
}

func (m Message) WithTags(tags []string) Message {
  m.tags = tags
  return m
}
```

- User spec change. Remove “keyIn”.

```yaml
conditions:
  tags:
    operator: and/or
    values: ["ca", "education"]
````

- [x] Python
- [x] Golang
- [x] Java
- [x] examples and docs

### PR DESCRIPTION
**Pull request by yhl25**
closes #665

### CODE REVIEW

File: pkg/apis/numaflow/v1alpha1/edge_types.go [hunk 1]
```diff
@@ -39,7 +39,16 @@ type Edge struct {
 }
 
 type ForwardConditions struct {
-	KeyIn []string `json:"keyIn" protobuf:"bytes,1,rep,name=keyIn"`
+	Tags *TagConditions `json:"tags" protobuf:"bytes,1,rep,name=tags"`
+}
+
+type TagConditions struct {
+	// Operator specifies the type of operation that should be used for conditional forwarding
+	// value could be "and", "or", "not"
```
KeranYang, line 47: nit: you can use kubebuilder to validate enum early.
Example: https://github.com/numaproj/numaflow/blob/main/pkg/apis/numaflow/v1alpha1/edge_types.go#L36

pkg/apis/numaflow/v1alpha1/edge_types.go
```diff
@@ -39,7 +39,17 @@ type Edge struct {
...
-	KeyIn []string `json:"keyIn" protobuf:"bytes,1,rep,name=keyIn"`
+	// Tags used to specify tags for conditional forwarding
+	Tags *TagConditions `json:"tags" protobuf:"bytes,1,rep,name=tags"`
+}
+
+type TagConditions struct {
+	// Operator specifies the type of operation that should be used for conditional forwarding
+	// value could be "and", "or", "not"
+	// +optional
+	Operator string `json:"operator" protobuf:"bytes,1,opt,name=operator"`
```
whynowy, line 50: Define a type e.g. `type LogicOperator string`, and some constants.

Use `Operator *LogicOperator xxxx`

Add a function and use it in other places.

```golang
func (tc TagConditions) GetOperator() LogicOperator {

}
```

pkg/shared/util/string.go
```diff
@@ -48,3 +48,31 @@ func StringSliceContains(list []string, str string) bool {
 	}
 	return false
 }
+
+func CompareSlice(operator string, sa []string, sb []string) bool {
```
whynowy, line 52: doc.

KeranYang:
Could you also add an E2E test case for tags?

pkg/apis/numaflow/v1alpha1/const.go
```diff
@@ -152,7 +152,7 @@ const (
 )
 
 var (
-	MessageKeyDrop = fmt.Sprintf("%U__DROP__", '\\') // U+005C__DROP__
+	MessageTagDrop = fmt.Sprintf("%U__DROP__", '\\') // U+005C__DROP__
 	MessageKeyAll  = fmt.Sprintf("%U__ALL__", '\\')  // U+005C__ALL__
```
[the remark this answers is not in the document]
yhl25: yeah we are using it 
https://github.com/numaproj/numaflow/blob/15c66c43e6d0f073b66b05aa3614401e216964a7/pkg/forward/interfaces.go#L24

pkg/forward/forward.go
```diff
@@ -182,9 +182,9 @@ func (isdf *InterStepDataForward) Start() <-chan struct{} {
 
 // readWriteMessagePair represents a read message and its processed (via UDF) write message.
 type readWriteMessagePair struct {
-	readMessage   *isb.ReadMessage
-	writeMessages []*isb.Message
-	udfError      error
+	readMessage    *isb.ReadMessage
+	taggedMessages []*isb.TaggedMessage
```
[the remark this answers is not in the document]
yhl25: @vigith @whynowy thoughts?

pkg/reconciler/pipeline/validate_test.go
```diff
@@ -328,16 +328,46 @@ func TestValidatePipeline(t *testing.T) {
 		assert.Contains(t, err.Error(), "not supported")
 	})
 
-	t.Run("good conditional forwarding", func(t *testing.T) {
+	t.Run("or conditional forwarding", func(t *testing.T) {
```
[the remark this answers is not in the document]
yhl25: I am yet to add tests

yhl25:
> Could you also add an E2E test case for tags?

I will update one of the E2E tests to test the tags, we already have a lot of E2E tests.

pkg/sources/transformer/builtin/event_time/event_time_extractor.go
```diff
@@ -80,8 +80,8 @@ func (e eventTimeExtractor) apply(et time.Time, payload []byte) (functionsdk.Mes
 		newEventTime, err = dateparse.ParseStrict(timeStr)
 	}
 	if err != nil {
-		return functionsdk.MessageTToAll(et, payload), err
+		return functionsdk.NewMessageT(payload).WithEventTime(et), err
```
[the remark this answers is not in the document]
yhl25: the use case of the transformer is not just to change the event time of the message right? if the udf developer doesn't set the event time (that means they don't care about the event time, they could be using the transformer for other purposes) for such cases we should use the parent message's event time, right? My understanding is transformers are used to transform the input message in the source vertex itself to avoid one more map vertex, it's not just limited to setting new event time.

pkg/apis/numaflow/v1alpha1/edge_types.go
```diff
@@ -39,7 +39,17 @@ type Edge struct {
 }
 
 type ForwardConditions struct {
-	KeyIn []string `json:"keyIn" protobuf:"bytes,1,rep,name=keyIn"`
+	// Tags used to specify tags for conditional forwarding
+	Tags *TagConditions `json:"tags" protobuf:"bytes,1,rep,name=tags"`
```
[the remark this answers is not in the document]
yhl25: done

pkg/apis/numaflow/v1alpha1/edge_types.go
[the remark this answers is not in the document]
yhl25: done

pkg/forward/forward.go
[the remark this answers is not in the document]
yhl25: changed it to writeMessage

pkg/sources/transformer/builtin/event_time/event_time_extractor.go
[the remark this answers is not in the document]
KeranYang: Discussed offline, will use public constructors to enforce specifying event time.

pkg/shared/util/string.go
```diff
@@ -48,3 +48,31 @@ func StringSliceContains(list []string, str string) bool {
...
+	switch operator {
+	case "and":
+		for _, val := range sa {
+			if !StringSliceContains(sb, val) {
+				return false
+			}
+		}
+		return true
```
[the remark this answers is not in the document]
KeranYang: If we construct a string hash set of slice b before `switch`, we can achieve O(n) time with O(n) space. Not sure if that could be a performance improvement. 😃

pkg/shared/util/string_test.go
```diff
@@ -20,6 +20,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
```
KeranYang, line 23: This is a bit of a nitpicking but I feel it's important from a code structure perspective - string utils shall not have knowledge of upper level logics like `v1alpha1`.

This limits `CompareSlice` in a way that it can only be used by `edge_types.go` while in fact it can be a util function that can be used everywhere.

I would suggest updating it to `func CompareSlice(operator string, sa []string, sb []string) bool` and switch on the string value instead of `v1alpha1.LogicOperator`.

vigith: +1, we might unknowingly end up in a cyclic-dependency

yhl25: lets keep it as is, we can change it later.

pkg/reduce/data_forward_test.go
```diff
@@ -114,15 +114,15 @@ type CounterReduceTest struct {
 }
 
 // Reduce returns a result with the count of messages
-func (f CounterReduceTest) ApplyReduce(_ context.Context, partitionID *partition.ID, messageStream <-chan *isb.ReadMessage) ([]*isb.Message, error) {
+func (f CounterReduceTest) ApplyReduce(ctx context.Context, partitionID *partition.ID, messageStream <-chan *isb.ReadMessage) ([]*isb.WriteMessage, error) {
```
vigith, line 117:
```suggestion
func (f CounterReduceTest) ApplyReduce(_ context.Context, partitionID *partition.ID, messageStream <-chan *isb.ReadMessage) ([]*isb.WriteMessage, error) {
```

pkg/reduce/readloop/readloop_test.go
```diff
@@ -48,7 +48,7 @@ func (s *SumReduceTest) WhereTo(_ []string) ([]string, error) {
 	return []string{"reduce-buffer"}, nil
 }
 
-func (s SumReduceTest) ApplyReduce(_ context.Context, partitionID *partition.ID, messageStream <-chan *isb.ReadMessage) ([]*isb.Message, error) {
+func (s SumReduceTest) ApplyReduce(ctx context.Context, partitionID *partition.ID, messageStream <-chan *isb.ReadMessage) ([]*isb.WriteMessage, error) {
```
vigith, line 51:
```suggestion
func (s SumReduceTest) ApplyReduce(_ context.Context, partitionID *partition.ID, messageStream <-chan *isb.ReadMessage) ([]*isb.WriteMessage, error) {
```

pkg/sources/transformer/builtin/event_time/event_time_extractor.go
```diff
@@ -69,7 +69,7 @@ func New(args map[string]string) (functionsdk.MapTFunc, error) {
 func (e eventTimeExtractor) apply(et time.Time, payload []byte) (functionsdk.MessageT, error) {
 	timeStr, err := expr.EvalStr(e.expression, payload)
 	if err != nil {
-		return functionsdk.MessageTToAll(et, payload), err
+		return functionsdk.NewMessageT(payload).WithEventTime(et), err
```
vigith, line 72: update SDK

pkg/shared/util/string.go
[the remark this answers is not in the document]
KeranYang: @yhl25 Please consider my proposal of saving time complexity by adding space complexity (GoLang GC periodically reclaims the allocated memory).

`CompareSlice` is on a frequent path that every message goes through it hence its performance matters.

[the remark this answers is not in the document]
vigith: +1, also, the tags are constant; we can build that hash at the initialization phase. 
Let's do that via a follow-up separate PR, as this has a lot of breaking changes.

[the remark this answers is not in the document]
vigith: #672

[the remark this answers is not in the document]
yhl25: yeah @ashwinidulams and I had discussed using a hashmap for tags, but since the number of keys is very less we thought don't need it. But since compareSlice gets invoked for each message the optimisation might actually help, will fix it in a follow-up PR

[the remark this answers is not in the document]
KeranYang: Giving it a second thought, it might not help much to introduce a hash map. O(n) matters when n is large but for `CompareSlice`, we expect both input slices to be small. We can dive deeper in #672 . Thank you Vigith for creating the issue. :)

pkg/apis/numaflow/v1alpha1/edge_types.go
```diff
@@ -39,7 +39,33 @@ type Edge struct {
...
+	Operator *LogicOperator `json:"operator" protobuf:"bytes,1,opt,name=operator"`
+	// Values tag values for conditional forwarding
+	Values []string `json:"values" protobuf:"bytes,2,rep,name=values"`
+}
+
+func (tc TagConditions) GetOperator() LogicOperator {
+	if tc.Operator == nil {
+		return LogicOperatorOr
+	}
+	return *tc.Operator
```
whynowy, line 68: Better to have a logic to return the valid operator, similar to line 87-91 in this file.

yhl25: we have Kubebuilder's Enum validation check, the only check we need is nil check

whynowy: Kubebuilder's enum validation only works when full CRD is installed.

yhl25: Okayy updated.

docs/user-guide/conditional-forwarding.md
```diff
@@ -1,27 +1,37 @@
 # Conditional Forwarding
 
-After processing the data, conditional forwarding is doable based on the `Key` returned in the result.
+After processing the data, conditional forwarding is doable based on the `Tags` returned in the result. 
+Below is list of different logic operations that can be done on tags.
+- **and** - forwards the message if all the tags specified are present in Message's tags.
+- **or** - forwards the message if one of the tags specified is present in Message's tags.
+- **not** - forwards the message if all the tags specified is not present in Message's tags.
```
KeranYang, line 7: nit: `all the tags specified are`

pkg/sources/source.go
```diff
@@ -237,35 +237,29 @@ func (sp *SourceProcessor) getTransformerGoWhereDecider() forward.GoWhere {
...
 		}
 
 		for _, edge := range sp.VertexInstance.Vertex.Spec.ToEdges {
-			// If returned keys is not "DROP", and there's no conditions defined in the edge, treat it as "ALL"?
-			if edge.Conditions == nil || len(edge.Conditions.KeyIn) == 0 || sharedutil.StringSliceContains(edge.Conditions.KeyIn, keys[len(keys)-1]) {
+			// If returned tags is not "DROP", and there's no conditions defined in the edge, treat it as "ALL"?
+			if edge.Conditions == nil || edge.Conditions.Tags == nil || len(edge.Conditions.Tags.Values) == 0 {
 				if edge.Parallelism != nil && *edge.Parallelism > 1 { // Need to shuffle
-					result = append(result, shuffleFuncMap[fmt.Sprintf("%s:%s", edge.From, edge.To)].Shuffle(keys))
+					result = append(result, shuffleFuncMap[fmt.Sprintf("%s:%s", edge.From, edge.To)].Shuffle(tags))
```
whynowy, line 251: Should we use `tags` to shuffle? Not `keys`?

yhl25: no, we should use keys :). will update it..

docs/user-guide/conditional-forwarding.md
```diff
@@ -1,27 +1,37 @@
...
 
 ```yaml
 edges:
   - from: p1
     to: even-vertex
     conditions:
-      keyIn:
-        - even
+      tags:
+        operator: or
```
whynowy, line 18: Need some opinions - do we encourage ppl to put `operator` all the time? Do we need to put a comment `# Optional, defaults to "or".`?

whynowy:
All the rest looks good to me.

docs/user-guide/conditional-forwarding.md
yhl25: +1, I think it's better to add that comment

pkg/forward/forward.go
```diff
@@ -27,6 +27,7 @@ import (
 	"sync"
 	"time"
 
+	sharedutil "github.com/numaproj/numaflow/pkg/shared/util"
```
vigith, line 30: ordering is not right

pkg/forward/forward_test.go
```diff
@@ -548,7 +548,7 @@ func (f mySourceForwardTest) ApplyMap(ctx context.Context, message *isb.ReadMess
 				Payload: result,
 			},
 		}
-		return []*isb.Message{writeMessage}, nil
+		return []*isb.WriteMessage{{Message: writeMessage}}, nil
```
vigith, line 551: 👍 i didn't know that this can auto-convert to a pointer

pkg/forward/interfaces.go
```diff
@@ -21,24 +21,24 @@ import (
 )
 
 var (
-	All  = GoWhere(func([]string) ([]string, error) { return []string{dfv1.MessageKeyAll}, nil })
-	Drop = GoWhere(func([]string) ([]string, error) { return []string{dfv1.MessageKeyDrop}, nil })
+	All  = GoWhere(func([]string, []string) ([]string, error) { return []string{dfv1.MessageTagAll}, nil })
```
vigith, line 24: we should standardize the args order (tags, keys); in SDK i see it (keys, tags)

vigith (approved):

whynowy (approved):
LGTM

PR #669, opened 2023-04-06
feat: Redis Streams source fixes

### LINKED ISSUE #662: Redis Streams Source - enable data from a single message to stay collocated
# Summary

Redis Streams actually allow multiple key/value pairs on a given incoming message. So, you could have a message added like this:
`XADD * mystream temperature 30 humidity 20`
In this case, "temperature" is a key whose value is "30" and "humidity" is a key whose value is "20".
The way I implemented it was to have this result in two outgoing messages, where the "key" is the key and the "value" is the value.

But this makes it so the outgoing message no longer contains collocated data. I am thinking about an alternative approach in which there is just one single outgoing message which has no Key and instead has a value which is a JSON serialized version of the message.

Proposal: if there is only 1 key/value pair then key and value are separated, else key will be none and value will be a json serialized array of key, value pairs

---

juliev0:
Fixes #662 
Also:
1. Use the incoming ID as the published Event Time. 
2. Document metrics

### DIFF
--- a/pkg/shared/clients/redis/redis_reader.go
+++ b/pkg/shared/clients/redis/redis_reader.go
@@ -44,8 +44,8 @@ type RedisStreamsRead struct {
 type Metrics struct {
 	ReadErrorsInc metricsIncrementFunc
 	ReadsAdd      metricsAddFunc // number of actual messages read in
-	ProducedAdd   metricsAddFunc // number of outgoing messages produced based on the input (an incoming message can result in multiple)
 	AcksAdd       metricsAddFunc
+	AckErrorsAdd  metricsAddFunc
 }
 
 // need a function type which increments a particular counter
@@ -103,10 +103,6 @@ func (br *RedisStreamsRead) Read(_ context.Context, count int64) ([]*isb.ReadMes
 	}
 	// Generate messages from the XStream
 	msgs, err := br.XStreamToMessages(xstreams, messages, labels)
-	// Update metric for number of messages produced
-	if br.Metrics.ProducedAdd != nil {
-		br.Metrics.ProducedAdd(len(messages))
-	}
 	br.Log.Debugf("Received %d messages over Redis Streams Source, err=%v", len(msgs), err)
 	return msgs, err
 }
@@ -126,6 +122,7 @@ func (br *RedisStreamsRead) processReadError(xstreams []redis.XStream, messages 
 	return messages, fmt.Errorf("XReadGroup failed, %w", err)
 }
 
+// Todo: we can actually probably put the original code back here - we don't really need dedup
 // Ack acknowledges the offset to the read queue. Ack is always pipelined, if you want to avoid it then
 // send array of 1 element.
 func (br *RedisStreamsRead) Ack(_ context.Context, offsets []isb.Offset) []error {
--- a/pkg/shared/clients/redis/redis_reader.go
+++ b/pkg/shared/clients/redis/redis_reader.go
@@ -122,13 +122,12 @@ func (br *RedisStreamsRead) processReadError(xstreams []redis.XStream, messages 
 	return messages, fmt.Errorf("XReadGroup failed, %w", err)
 }
 
-// Todo: we can actually probably put the original code back here - we don't really need dedup
 // Ack acknowledges the offset to the read queue. Ack is always pipelined, if you want to avoid it then
 // send array of 1 element.
 func (br *RedisStreamsRead) Ack(_ context.Context, offsets []isb.Offset) []error {
 	errs := make([]error, len(offsets))
-	// in the case of n messages produced from 1 incoming message, we could have
-	// the same offset more than once: need to deduplicate
+	// if we were to have n messages produced from 1 incoming message, we could have
+	// the same offset more than once: just in case, we can deduplicate
 	dedupOffsets := make(map[string]struct{}) // essentially a Set
 	strOffsets := []string{}
 	for _, o := range offsets {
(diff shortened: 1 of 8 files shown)


pkg/shared/clients/redis/redis_reader.go
```diff
@@ -94,11 +95,14 @@ func (br *RedisStreamsRead) Read(_ context.Context, count int64) ([]*isb.ReadMes
...
 
-	// for each XMessage in []XStream
-	msgs, err := br.XStreamToMessages(xstreams, messages, labels)
-	if br.Metrics.ReadsAdd != nil {
-		br.Metrics.ReadsAdd(len(messages))
+	// Update metric for number of messages read in
+	for _, xstream := range xstreams {
+		if br.Metrics.ReadsAdd != nil {
+			br.Metrics.ReadsAdd(len(xstream.Messages))
+		}
```
juliev0, line 102: this is technically the actual number of messages read in. The old code represents the number of messages published which theoretically could be different.

pkg/shared/clients/redis/redis_reader.go
```diff
@@ -122,17 +126,28 @@ func (br *RedisStreamsRead) processReadError(xstreams []redis.XStream, messages
...
-			errs[i] = err
+			errs[i] = err // 'errs' is indexed the same as 'offsets'
+		}
+		if br.Metrics.AckErrorsAdd != nil {
+			br.Metrics.AckErrorsAdd(len(strOffsets))
+		}
+	} else {
+		if br.Metrics.AcksAdd != nil {
+			br.Metrics.AcksAdd(len(strOffsets))
 		}
```
juliev0, line 150: this dedup logic isn't really being used at the moment since we're publishing 1 message for every incoming message, but if in the future we were to configure an option to publish multiple messages in the case of multiple key/value pairs then we would need this

pkg/sources/redisstreams/redisstream.go
```diff
@@ -228,6 +229,93 @@ func newRedisClient(sourceSpec *dfv1.RedisStreamsSource) (*redisclient.RedisClie
...
+		},
+		Body: isb.Body{Payload: []byte(jsonSerialized)},
+	}
+
+	return &isb.ReadMessage{
+		ReadOffset: isb.SimpleStringOffset(func() string { return readOffset }),
+		Message:    isbMsg,
+	}, nil
+}
+
```
juliev0, line 268: @vigith How do you feel about this being a json map? (since you specifically mentioned a "json serialized array of key, value pairs") This should work just as well, right?

vigith (approved):
LGTM

pkg/sources/redisstreams/redisstream.go
```diff
@@ -228,6 +229,93 @@ func newRedisClient(sourceSpec *dfv1.RedisStreamsSource) (*redisclient.RedisClie
...
 
+// incoming message should have multiple key/value pairs: produce a message with no Key, with value set to JSON serialization of the pairs
+func produceUnkeyedJSONMsg(inMsg redis.XMessage) (*isb.ReadMessage, error) {
+	var readOffset = inMsg.ID
+
+	// JSON serialize our key/value pairs
+	// need to first convert map[string]interface to map[string]string
+	kvMap := make(map[string]string)
+	for k, v := range inMsg.Values {
+		strValue, castOk := v.(string)
```
whynowy, line 240: Why can't we convert the original `map[string]interface{}` to json directly? I believe all the value types are okay to marshal.

pkg/sources/redisstreams/redisstream.go
```diff
@@ -164,35 +169,31 @@ func New(
...
-					isbMsg := isb.Message{
-						Header: isb.Header{
-							MessageInfo: isb.MessageInfo{EventTime: time.Now()}, //doesn't seem like Redis offers a timestamp
-							ID:          id,                                     // assumption is that this only needs to be unique for this source vertex
-							Keys:        []string{f},
-						},
-						Body: isb.Body{Payload: []byte(v.(string))},
+				var outMsg *isb.ReadMessage
+				var err error
+				if len(message.Values) > 1 {
```
whynowy, line 180: This is going to generate different behaviors when someone sends data with single or multiple key/value pairs, one output is a single string, the other one is a json. 

I believe it is possible that both single and multiple key/value pairs will happen in one streamed data, and this implementation will make troubles. I suggest we alway use json.

@vigith @juliev0

juliev0: It's a good point. Here's what I'm thinking: how about for consistency we always Key on the first Key, and always use JSON as well?

whynowy: We do `keys []string` now, so all the keys can be there.

pkg/sources/redisstreams/redisstream.go
juliev0: You're right - I'd need to do a `[]byte assertion` and then cast to string. So, apparently the underlying type must be a string in the tests I've done so far. Hmmm...

pkg/sources/redisstreams/redisstream.go
juliev0: oh, that's the "multi-key" feature I guess? I haven't looked at how that works yet. Cool, will do that then.

juliev0:
@whynowy this is ready for re-review now. For consistency, we always use the combination of Keys and we always serialize to map of keys to values.

We simply serialize the original map[string]interface{}. In general, I assume this will be a map of string to string, but if for some reason it's not, like you said the json serialization should do the right thing. (BTW, if I directly call ` XADD test-stream * mykey myvalue` then I confirmed that `myvalue` is in fact a string and not a []byte)

whynowy (approved):
LGTM

PR #670, opened 2023-04-06
chore: fix typos and update Kafka reader pending count comments

KeranYang:
Explain what this PR does.


vigith (approved):
can you delete this line too https://github.com/numaproj/numaflow/blob/main/pkg/isb/serde.go#L19

PR #671, opened 2023-04-12
chore: update liveness probe config for vertex pods

whynowy:
Current liveness probe is too aggressive, update it to a more reasonable config.


pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -233,9 +233,9 @@ func (p Pipeline) GetDaemonDeploymentObj(req GetDaemonDeploymentReq) (*appv1.Dep
 				Scheme: corev1.URISchemeHTTPS,
 			},
 		},
-		InitialDelaySeconds: 3,
-		PeriodSeconds:       3,
-		TimeoutSeconds:      1,
+		InitialDelaySeconds: 90,
```
vigith, line 236: isn't this too long a wait?

vigith (approved):

PR #674, opened 2023-04-12
chore: remove health check to udf/udsink containers

whynowy:
Remove the health check from the `numa` container to udf/udsink containers as it does not work as expected.

If the liveness check on `numa` container fails, it only restarts the `numa` container, the udf container stays as is.

### DIFF
--- a/pkg/sinks/sink.go
+++ b/pkg/sinks/sink.go
@@ -111,16 +111,7 @@ func (u *SinkProcessor) Start(ctx context.Context) error {
 			return
 		}
 	}()
-	var metricsOpts []metrics.Option
-	if udSink := u.VertexInstance.Vertex.Spec.Sink.UDSink; udSink != nil {
-		if serverHandler, ok := sinker.(*udsink.UserDefinedSink); ok {
-			metricsOpts = metrics.NewMetricsOptions(ctx, u.VertexInstance.Vertex, serverHandler, reader, nil)
-		} else {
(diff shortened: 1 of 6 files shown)


pkg/sinks/sink.go
```diff
@@ -111,16 +111,7 @@ func (u *SinkProcessor) Start(ctx context.Context) error {
...
-	if udSink := u.VertexInstance.Vertex.Spec.Sink.UDSink; udSink != nil {
-		if serverHandler, ok := sinker.(*udsink.UserDefinedSink); ok {
-			metricsOpts = metrics.NewMetricsOptions(ctx, u.VertexInstance.Vertex, serverHandler, reader, nil)
-		} else {
-			return fmt.Errorf("unable to get the metrics options for the udsink")
-		}
-	} else {
-		metricsOpts = metrics.NewMetricsOptions(ctx, u.VertexInstance.Vertex, nil, reader, nil)
-	}
+	metricsOpts := metrics.NewMetricsOptions(ctx, u.VertexInstance.Vertex, nil, reader, nil)
```
vigith, line 114: should we change the signature of `NewMetricsOptions` since we are passing `nil` for the checker.

vigith (approved):

pkg/sinks/sink.go
whynowy: Technically, that still could be used for other kinds of health check related to the platform if needed, let's keep it there for now.

Issue #675, opened 2023-04-12
Investigate how to do liveness probe on udf/udsink containers


whynowy:
# Summary

When the udf application exits explicitly without an error, the udf container goes to `completed` status, in this case, we should get it restarted. We need to figure out a way to configure liveness probe on the udf/udsink containers.

Starting from k8s 1.23, there's a native gRPC liveness probe supported, but that is not able to support gRPC over Unix Domain Socket. Need to think about other approaches.

PR #676, opened 2023-04-13
chore: enable liveness check for both numa and sidecar containers

whynowy:
Closes: #675 

1. Expose a new endpoint through the `numa` container, which is dedicated for sidecar health check, and configure liveness probes for all the sidecars.
2. Original `numa` container liveness check simply tells its own status.


vigith (approved):
will this be able to show the side-car status such that kubelet will know the side-car is down and restart the sidecar?

PR #677, opened 2023-04-13
chore: increase InvokeE2EAPIPOST retry attempts from 3 to 6

KeranYang:
With our recent changes to Numaflow SDKs, it takes longer than before for UDF servers to boot up. Based on my observation, the first 2 http post attempts are always failing with 500 Errors. Sometimes the third one also fails, failing the test case. This code change increase the retry attempts from 3 to 6 to prevent e2e test flakiness.

To reproduce the issue, I checked out the latest numaflow-java sdk and built the event-time-filter to publish a new version of that image, then ran the e2e test.

Also simplified the dropOnFull test to make it less flaky.


pkg/isb/serde.go
```diff
@@ -16,8 +16,6 @@ limitations under the License.
 
 package isb
 
-// TODO: integrate with the WAL segment and JetStream
```
KeranYang, line 19: https://github.com/numaproj/numaflow/pull/670#pullrequestreview-1375579969

test/e2e-suite-2/functional_test.go
```diff
@@ -52,9 +52,12 @@ func (s *FunctionalSuite) TestDropOnFull() {
 
 	w.SendMessageTo(pipelineName, "in", NewHttpPostRequest().WithBody([]byte("1")))
 	// give buffer writer some time to update the isFull attribute.
-	// 5s is a carefully chosen number to create a stable buffer full scenario, it is a bit higher than the buffer writer's refresh interval.
-	time.Sleep(time.Second * 5)
+	// 10s is a carefully chosen number to create a stable buffer full scenario.
+	time.Sleep(time.Second * 10)
 	w.SendMessageTo(pipelineName, "in", NewHttpPostRequest().WithBody([]byte("2")))
+	// there could be a race condition between the buffer write and the sink scaling up, if the buffer write happens after the sink scaling up, message 2 will not be dropped.
```
whynowy, line 58: Check the metrics to verify it has n message dropped? And in this case, we might not need to do the rest?

KeranYang: Thanks Derek! I didn't know there's a way to retrieve number of dropped message, is it through port-forwarding with VertexMetricsPort? And if we are able to verify n message dropped through metrics, should the rest checks guarantee to succeed?

vigith (approved):

test/e2e-suite-2/functional_test.go
whynowy: That's what I'm thinking, verifying the dropped messages and processed messages in the metics, and then skip the rest.

whynowy (approved):

Issue #688, opened 2023-04-16
Support running as non-root user


sandangel:
# Summary

What change needs making?

Currently numaflow is running as root

# Use Cases

When would you use this?

The cluster is enforcing running container as non-root, but there is an issue with numaflow when running as non-root. It was not able to write to mounted volumes, even though securityContext fsGroup is set.

More info: https://numaproj.slack.com/archives/C03GQ3AMHE1/p1681546238208929

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

whynowy:
Errors.


{"level":"warn","ts":1681541906.1529396,"logger":"numaflow.ReduceUDF-processor","caller":"readloop/readloop.go:400","msg":"Failed to create pbq during startup, retrying","vertex":"conversion-reporter","attempt":1,"partitionID":"1681541520000-1681541580000-slot-0","error":"failed to create a PBQ store, stat /var/numaflow/pbq/wals/segment_1681541520.1681541580.slot-0: permission denied"}
{"level":"warn","ts":1681541914.2472534,"logger":"numaflow.ReduceUDF-processor","caller":"readloop/readloop.go:400","msg":"Failed to create pbq during startup, retrying","vertex":"conversion-reporter","attempt":1,"partitionID":"1681541520000-1681541580000-slot-0","error":"failed to create a PBQ store, stat /var/numaflow/pbq/wals/segment_1681541520.1681541580.slot-0: permission denied"}
{"level":"warn","ts":1681541926.675691,"logger":"numaflow.ReduceUDF-processor","caller":"readloop/readloop.go:400","msg":"Failed to create pbq during startup, retrying","vertex":"conversion-reporter","attempt":1,"partitionID":"1681541520000-1681541580000-slot-0","error":"failed to create a PBQ store, stat /var/numaflow/pbq/wals/segment_1681541520.1681541580.slot-0: permission denied"}
{"level":"warn","ts":1681541944.6003408,"logger":"numaflow.ReduceUDF-processor","caller":"readloop/readloop.go:400","msg":"Failed to create pbq during startup, retrying","vertex":"conversion-reporter","attempt":1,"partitionID":"1681541520000-1681541580000-slot-0","error":"failed to create a PBQ store, stat /var/numaflow/pbq/wals/segment_1681541520.1681541580.slot-0: permission denied"}
[...]

PR #689, opened 2023-04-16
fix: add wal dir x permission

whynowy:
Fixes: #688

`/var/numaflow/pbq/wal` was created with `0644`, there's no issue with it when the vertex pod/containers run as root. If they run as non-root, since it has no `x` permission, there's no way to `cd` to the dir and create wal files.

### DIFF
--- a/pkg/reduce/pbq/store/wal/stores.go
+++ b/pkg/reduce/pbq/store/wal/stores.go
@@ -61,7 +61,7 @@ func (ws *walStores) CreateStore(_ context.Context, partitionID partition.ID) (s
 	// Create wal dir if not exist
 	var err error
 	if _, err = os.Stat(ws.storePath); os.IsNotExist(err) {
-		err = os.Mkdir(ws.storePath, 0644)
+		err = os.Mkdir(ws.storePath, 0744)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/reduce/pbq/store/wal/stores.go
```diff
@@ -61,7 +61,7 @@ func (ws *walStores) CreateStore(_ context.Context, partitionID partition.ID) (s
 	// Create wal dir if not exist
 	var err error
 	if _, err = os.Stat(ws.storePath); os.IsNotExist(err) {
-		err = os.Mkdir(ws.storePath, 0644)
+		err = os.Mkdir(ws.storePath, 0744)
```
vigith, line 64: should we do 755?

vigith (approved):

pkg/reduce/pbq/store/wal/stores.go
whynowy: Makes sense.

sandangel:
Is this change now included in v0.8.0-rc.1?

whynowy:
> Is this change now included in v0.8.0-rc.1?

nightly build

PR #695, opened 2023-04-18
chore:update conditional-forwarding.md

yhl25:


docs/user-guide/reference/conditional-forwarding.md
```diff
@@ -2,31 +2,32 @@
...
+* **not** - forwards the message if all the tags specified are not present in Message's tags.
 
 For example, there's a UDF used to process numbers, and forward the result to different vertices based on the number is even or odd. In this case, you can set the `tag` to `even-tag` or `odd-tag` in each of the returned messages,
 and define the edges as below:
 
 ```yaml
 edges:
   - from: p1
-    to: even-vertex
+    to: vertex-1
```
whynowy, line 16: I understand this example is to showcase `or`, `and` and `not`, but somehow it does not make sense.

docs/user-guide/reference/conditional-forwarding.md
```diff
@@ -2,31 +2,32 @@
...
       tags:
         operator: or # Optional, defaults to "or".
         values:
           - even-tag
   - from: p1
-    to: odd-vertex
+    to: vertex-2
     conditions:
       tags:
         operator: not
```
whynowy, line 26: `not` odd-tag means it's even....

docs/user-guide/reference/conditional-forwarding.md
```diff
@@ -2,31 +2,32 @@
...
         values:
           - odd-tag
   - from: p1
-    to: all
+    to: vertex-3
     conditions:
       tags:
         operator: and
```
whynowy, line 33: `and` means `even-tag  && odd-tag`, then no data meets.

PR #696, opened 2023-04-18
fix: allow late message as long as window is not closed

vigith:
Even if a message is late, we can push it in as long as the window is open.

### DIFF
--- a/pkg/reduce/readloop/readloop.go
+++ b/pkg/reduce/readloop/readloop.go
@@ -210,18 +210,31 @@ func (rl *ReadLoop) writeMessagesToWindows(ctx context.Context, messages []*isb.
 
 messagesLoop:
 	for _, message := range messages {
-		// drop the late messages
+		// drop the late messages only if there is no window open
 		if message.IsLate {
-			rl.log.Warnw("Dropping the late message", zap.Time("eventTime", message.EventTime), zap.Time("watermark", message.Watermark))
-			droppedMessagesCount.With(map[string]string{
-				metrics.LabelVertex:             rl.vertexName,
-				metrics.LabelPipeline:           rl.pipelineName,
-				metrics.LabelVertexReplicaIndex: strconv.Itoa(int(rl.vertexReplica)),
-				LabelReason:                     "late"}).Inc()
+			// we should be able to get the late message in as long as there is an open window
+			nextWin := rl.pbqManager.NextWindowToBeClosed()
+			if nextWin != nil && message.EventTime.Before(nextWin.StartTime()) {
+				rl.log.Warnw("Dropping the late message", zap.Time("eventTime", message.EventTime), zap.Time("watermark", message.Watermark), zap.Time("nextWindowToBeClosed", nextWin.StartTime()))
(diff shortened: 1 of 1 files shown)


chromevoid (approved):

whynowy:
How come later message comes earlier than the watermark? I remember we discussed before, that the late message is always later than the wartermark.

chromevoid:
> How come late message comes earlier than the watermark? I remember we discussed before, that the late message is always later than the wartermark.

context is https://github.com/numaproj/numaflow/issues/697 
we'd like to add latency to COB.

vigith:
> How come late message comes earlier than the watermark? I remember we discussed before, that the late message is always later than the wartermark.

Watermark and lateness gets decoupled at source, so it is totally possible that an element with isLate field set to true reach Vn vertex  before the watermark which caused it to late has arrived the same vertex.

whynowy:
> > How come late message comes earlier than the watermark? I remember we discussed before, that the late message is always later than the wartermark.
> 
> context is #697 we'd like to add latency to COB, so the late message could come earlier than the `watermark minus latency` time.
> 
> This PR is the preparation for the above task.



> > How come late message comes earlier than the watermark? I remember we discussed before, that the late message is always later than the wartermark.
> 
> Watermark and lateness get decoupled at the source, so it is totally possible that an element with isLate field set to true reaches Vn vertex before the watermark, which caused it to be late, has arrived at the same vertex.

Even without the context https://github.com/numaproj/numaflow/issues/697? I remember we discussed before about this, you said that was impossible :)

vigith:
> Even without the context #697? I remember we discussed before about this, you said that was impossible :)


lol, maybe I had got the question wrong earlier 😁. But, it does make sense right?

whynowy:
> > Even without the context #697? I remember we discussed before about this, you said that was impossible :)
> 
> lol, maybe I had got the question wrong earlier 😁. But, it does make sense right?

On the other hand, if this is possible, then a "good" data can also come later than the wartermark, if that's the case, is our watermark propagation still trustable?

vigith:
> On the other hand, if this is possible, then a "good" data can also come later than the wartermark, if that's the case, is our watermark propagation still trustable?

The above condition will never happen. I think there is a confusion. There a no watermark propagation problem. Watermark will always trail behind the data. This is the expected behavior.

whynowy (approved):

ashwinidulams:
only thought is that the same event stream may or may not drop the late events for subsequent runs with the same data set. Since the condition of watermark trailing late events is arbitrary and is not guaranteed always.

ashwinidulams:
> only thought is that the same event stream may or may not drop the late events for subsequent runs with the same data set. Since the condition of watermark trailing late events is arbitrary and is not guaranteed always.

however, i think this is ok as there is no determinism in watermark progression and is totally dependent on how the previous vertices go about processing the data stream.

ashwinidulams (approved):

yhl25 (approved):

vigith:
When we tackle #697 we will introduce a flag to make this optional. There could be a case where user does not want late data to be processed.

Issue #694, opened 2023-04-18
Ability to configure different late data in the same pipeline


tannaga:
# Summary

What change needs making?
I would like to define a DAG as depicted in the below diagram. 
<img width="1172" alt="image" src="https://user-images.githubusercontent.com/19876069/232674872-16afa95c-6943-4044-9940-28e24bd939d3.png">

# Use Cases

When would you use this?
I have a topic produced by different sources with different arrival time SLA. i want to be able to materialize data independently for each of these producers without slowing down the whole pipeline.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #703, opened 2023-04-21
feat: allowedLateness to support late data ingestion

### LINKED ISSUE #697: allowedLateness in reduce vertex
# Summary

In the pipeline DAG, there could be multiple processing reduce vertices where we might have to support different delays for close-of-book based on the lateness of data. We can support "allowedLateness" feature where if the pipeline author knows that there is skewness in the event time in the source payload, then they can potentially allow late data to be processed by slowing the down the close-of-book operation of particular reduce vertex by adding a fixed delay.

# Use Cases
* Low latency quick approximate materialization vs. relatively higher latency (still in the order of single-digit minutes) but more accurate results.
* #694

vigith:
fixes #697 #694

- [ ] documentation
- [x] unit test


whynowy (approved):
LGTM

PR #704, opened 2023-04-22
docs(proposal): edges, buffers and buckets

whynowy:
Explain what this PR does.


docs/specifications/edges-buffers-buckets.md
```diff
@@ -0,0 +1,75 @@
...
+      to: sink
+```
+
+Each `edge` could have a name for internal usage, the naming convention is `{pipeline-name}-{from-vertex-name}-{to-vertex-name}`.
+
+## Buffers
+
+`Buffer` is `InterStepBuffer`. Each buffer has an owner, which is the vertex who reads from it. Each `udf` and `sink` vertex in a pipeline owns a group of partitioned buffers. Each buffer has a name with the naming convention `{pipeline-name}-{vertex-name}-{index}`, where the `index` is the partition index, starting from 0. This naming convention applies to the buffers of both map and reduce udf vertices.
+
+When multiple vertices connecting to the same vertex, if the to vertex is a map, the data from all the from vertices will be forwarded to the group of partitoned buffers round-robinly. If the to vertex is a reduce, the data from all the from vertices will be forwarded to the group of partitoned buffers based on the partitioning key.
```
whynowy, line 55: We will be able to easily support diamond dags.

docs/specifications/edges-buffers-buckets.md
```diff
@@ -0,0 +1,75 @@
...
+
+A source vertex does not have any owned buffers. But a pipeline may have multiple source vertices, followed by one vertex. Same as above, if the following vertex is a map, the data from all the source vertices will be forwarded to the group of partitoned buffers round-robinly. If it is a reduce, the data from all the source vertices will be forwarded to the group of partitoned buffers based on the partitioning key.
```
whynowy, line 57: Be able to support multiple sources.

docs/specifications/edges-buffers-buckets.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Buckets
+
+`Bucket` is a K/V store (or a pair of stores) used for watermark propagation.
+
+There are 3 types of buckets in a pipeline:
+
+- `Edge Bucket`: used for edge watermark propagation. For Map vertices or Sink vertices, each of them has one bucket; for Reduce vertices, each of them has N buckets, where N is the partitions.
```
whynowy, line 65: Carefully review this: 

1. Use 1 bucket (2 K/V stores) for a Map with multi-partitioned buffers;
2. Use N buckets (2*N K/V stores) for a multi-partitioned reduce.

Are we able to do the same as Map for Reduce?

docs/specifications/edges-buffers-buckets.md
```diff
@@ -0,0 +1,75 @@
...
+      parallelism: 2
+    - from: compute-sum
+      to: sink
+```
+
+Each `edge` could have a name for internal usage, the naming convention is `{pipeline-name}-{from-vertex-name}-{to-vertex-name}`.
+
+## Buffers
+
+`Buffer` is `InterStepBuffer`. Each buffer has an owner, which is the vertex who reads from it. Each `udf` and `sink` vertex in a pipeline owns a group of partitioned buffers. Each buffer has a name with the naming convention `{pipeline-name}-{vertex-name}-{index}`, where the `index` is the partition index, starting from 0. This naming convention applies to the buffers of both map and reduce udf vertices.
```
whynowy, line 53: `edge.parallelism` should be moved to vertex.

docs/specifications/edges-buffers-buckets.md
whynowy: `edge.parallelism` should be moved to vertex, otherwise it does not make sense for diamond dags.

vigith (commented):
Discussed offline

vigith:
* Buffer name will be just `{to}` for both Map and Reduce
* Buffer Bucket for WM and Metadata will be `{from}-{to}`, and there will be only one bucket for an edge, immaterial of how many partitions we will have, for both Map and Reduce.

whynowy:
Updated.

vigith (approved):

docs/specifications/edges-buffers-buckets.md
```diff
@@ -0,0 +1,79 @@
...
+      to: out
+```
+
+Each `edge` could have a name for internal usage, the naming convention is `{pipeline-name}-{from-vertex-name}-{to-vertex-name}`.
+
+## Buffers
+
+`Buffer` is `InterStepBuffer`. Each buffer has an owner, which is the vertex who reads from it. Each `udf` and `sink` vertex in a pipeline owns a group of partitioned buffers. Each buffer has a name with the naming convention `{pipeline-name}-{vertex-name}-{index}`, where the `index` is the partition index, starting from 0. This naming convention applies to the buffers of both map and reduce udf vertices.
+
+When multiple vertices connecting to the same vertex, if the `to` vertex is a Map, the data from all the from vertices will be forwarded to the group of partitoned buffers round-robinly. If the `to` vertex is a Reduce, the data from all the from vertices will be forwarded to the group of partitoned buffers based on the partitioning key.
```
KeranYang, line 59: Before the proposal, if the target vertex is Map/Sink, there is one partition buffer, if Reduce, the number of partitions is `parallelism`. With this proposal, how do we determine how many partitioned buffers are allocated to the vertex?

KeranYang: Synced offline. I missed comments about making `parallelism` as vertex level attribute.

KeranYang (approved):

whynowy:
@numaproj/numaflow-dev - Please review.

docs/specifications/edges-buffers-buckets.md
```diff
@@ -0,0 +1,79 @@
...
+
+A Source vertex does not have any owned buffers. But a pipeline may have multiple Source vertices, followed by one vertex. Same as above, if the following vertex is a map, the data from all the Source vertices will be forwarded to the group of partitoned buffers round-robinly. If it is a reduce, the data from all the Source vertices will be forwarded to the group of partitoned buffers based on the partitioning key.
```
ashwinidulams, line 61: i see that we propose that map vertex could own multiple partitions. I understand from offline conversation that we dont necessarily intend to develop multi-stream readers and so therefore, will there be a logic to determine which partition needs to be allocated to which map pod?

ashwinidulams: if the above statements are true, then during autoscaling, we may end up having multiple streams processed by a single vertex? or is the expectation that there will be no more than 1 stream/partition per vertex-pod?

docs/specifications/edges-buffers-buckets.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Buckets
+
+`Bucket` is a K/V store (or a pair of stores) used for watermark propagation.
+
+There are 3 types of buckets in a pipeline:
+
+- `Edge Bucket`: Each edge has a bucket, used for edge watermark propagation, no matter if the vertex that the edge leads to is a Map or a Reduce. The naming convention of an edge bucket is `{pipeline-name}-{from-vertex-name}-{to-vertex-name}`.
```
ashwinidulams, line 69: so if i understand correctly, this bucket will contain information for all the partitions (streams) ?

docs/specifications/edges-buffers-buckets.md
whynowy: The proposal for multi partitioned buffer is, one vertex pod will start N Goroutines (N=partitions) to read messages, each on is responsible for 1 partition.

ashwinidulams: makes sense

ashwinidulams (approved):

docs/specifications/edges-buffers-buckets.md
whynowy: > here, partition refers to a stream (jetstream) as of today?

yes

docs/specifications/edges-buffers-buckets.md
xdevxy: Just to be clear, for `Edge Bucket`, no matter how many edges the vertex has there would be one bucket for map, and N buckets for reduce?

whynowy: > Just to be clear, for `Edge Bucket`, no matter how many edges the vertex has there would be one bucket for map, and N buckets for reduce?

No, it's been updated to 1 bucket for both use cases.

xdevxy (commented):
Just for my own understanding, in summary 1) each vertex owns the incoming buffer no matter how many incoming edges there are. 2) each edge will have one corresponding edge bucket. Is that correct?

whynowy:
> Just for my own understanding, in summary 1) each vertex owns the incoming buffer no matter how many incoming edges there are. 2) each edge will have one corresponding edge bucket no matter how many edges there are between two vertices. Is that correct?

Good summary!

PR #709, opened 2023-04-25
feat: enable streaming message to next vertex when batch size is 1

### LINKED ISSUE #623: A way to push message early in numaflow instead of waiting until function return
# Summary

What change needs making?

Add a code in sdk so user can push the message early in the numaflow map/reduce function instead of at the end of functions 

# Use Cases

When would you use this?

In flat map situation (1 message generate multiple messages to the downstream vertices),
When we completely generate 1 new message from flat map, we would like to push it to the buffer ASAP because it would take a few minutes until the next message.

Some pseudocode:

```python
import functions from numaflow

...
for f in as_completed(futures_of items):
    functions.Messages.push(Message('success', f.result().as_bytes()))
```

---

xdevxy:
Closes #623


xdevxy:
@whynowy this PR is good for review for initial feedback, even though the E2E test is blocked by example stream UDF which I'm working on. Thanks!

pkg/forward/forward.go
```diff
@@ -416,6 +435,91 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
...
+			errs, ctx := errgroup.WithContext(ctx)
+			errs.Go(func() error {
+				return isdf.UDF.ApplyMapStream(ctx, m, writeMessageCh)
+			})
+
+			// Stream the message to the next vertex. First figure out which vertex
+			// to send the result to. Then update the toBuffer(s) with writeMessage.
+			for writeMessage := range writeMessageCh {
+				// if we do not get a time from UDF, we set it to the time from (N-1)th vertex
+				if writeMessage.EventTime.IsZero() {
```
whynowy, line 476: The eventTime is not supposed to be changed in the map udf, it's already specified in pkg/udf/function/uds_grpc.go, line 178.

pkg/forward/forward.go
```diff
@@ -416,6 +435,91 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
...
+				return true
+			}
+
+			udfProcessingTime.With(map[string]string{metrics.LabelVertex: isdf.vertexName, metrics.LabelPipeline: isdf.pipelineName,
+				"buffer": isdf.fromBuffer.GetName()}).Observe(float64(time.Since(start).Microseconds()))
+		}
+	}
+
+	// forward the message to the edge buffer (could be multiple edges)
+	err := isdf.writeToBuffers(ctx, messageToStep, writeOffsets)
```
whynowy, line 512: This means we only write to next buffers after receiving all the returned messages from the channel?

xdevxy: Good catch! Initially I have inside the loop for processing each received messages. Then realized non data messages also need forwarding. I guess I need to add it for both cases.

pkg/forward/forward.go
xdevxy: This is a replication of non streaming case at L695. Do you mean it is not needed? Then we are doing it for non streaming case?

pkg/forward/forward.go
```diff
@@ -281,80 +275,100 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
...
-		return
+		// forward the message to the edge buffer (could be multiple edges)
+		err = isdf.writeToBuffers(ctx, messageToStep, &writeOffsets)
+		if err != nil {
+			isdf.opts.logger.Errorw("failed to write to toBuffers", zap.Error(err))
+			isdf.fromBuffer.NoAck(ctx, readOffsets)
+			return
+		}
+	} else {
+		failure := isdf.streamMessage(ctx, dataMessages, processorWM, readOffsets, &writeOffsets, messageToStep)
```
whynowy, line 368: 1. No need to give the pointer of a map;
2. Can we keep original mode, that returns `(writerOffsets, error)`?

pkg/forward/forward.go
```diff
@@ -256,17 +261,6 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
 	udfResults := make([]readWriteMessagePair, len(dataMessages))
```
whynowy, line 261: Do we need to move `udfCh` and `udfResults` into the `if` section?

pkg/forward/forward.go
whynowy: Technically it's not needed I think.

pkg/apis/numaflow/v1alpha1/udf.go
```diff
@@ -43,6 +43,10 @@ type UDF struct {
 	Builtin *Function `json:"builtin" protobuf:"bytes,2,opt,name=builtin"`
 	// +optional
 	GroupBy *GroupBy `json:"groupBy" protobuf:"bytes,3,opt,name=groupBy"`
+	// MapStream specifies whether to enable streaming in map udf. If it is
+	// not provided, the default value is set to false.
+	// +optional
+	MapStream bool `json:"mapStream" protobuf:"bytes,4,opt,name=mapStream"`
```
whynowy, line 49: Thinking of exposing it through `vertex.metadata.annotation[].xxxx`, let's discuss it.

pkg/forward/forward.go
```diff
@@ -416,6 +432,87 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
...
+		errMsg := "stream is not supported for source data transformer"
+		isdf.opts.logger.Errorw(errMsg)
+		return nil, fmt.Errorf(errMsg)
+	}
+
+	if len(dataMessages) > 1 {
+		errMsg := "data message size is not 1 with UDF streaming"
+		isdf.opts.logger.Errorw(errMsg)
+		return nil, fmt.Errorf(errMsg)
+	} else if len(dataMessages) == 1 {
```
whynowy, line 458: nit: this else if not needed.

xdevxy: just extra check in case some bug is introduced to allow batch size larger than 1 in future.

pkg/forward/forward.go
```diff
@@ -467,17 +571,19 @@ func (isdf *InterStepDataForward) ackFromBuffer(ctx context.Context, offsets []i
 
 // writeToBuffers is a blocking call until all the messages have be forwarded to all the toBuffers, or a shutdown
 // has been initiated while we are stuck looping on an InternalError.
-func (isdf *InterStepDataForward) writeToBuffers(ctx context.Context, messageToStep map[string][]isb.Message) (writeOffsetsEdge map[string][]isb.Offset, err error) {
+func (isdf *InterStepDataForward) writeToBuffers(
```
xdevxy, line 574: Had to update `writeToBuffers` signature to keep track of previous `writeOffsets`

whynowy: I think it might be better to keep the original signature, because it clearly means the returned offsets are the results of writing the input of messageToStep.

In this case, we merge the returned in offsets in line 488 to the one in line 452 as the function return.

pkg/forward/forward.go
```diff
@@ -416,6 +436,90 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
...
+			if ok, _ := isdf.IsShuttingDown(); ok {
+				isdf.opts.logger.Errorw("UDF.Apply, Stop called while stuck on an internal error", zap.Error(err))
+				platformError.With(map[string]string{metrics.LabelVertex: isdf.vertexName, metrics.LabelPipeline: isdf.pipelineName}).Inc()
+			}
+			return nil, err
+		}
+
+		udfProcessingTime.With(map[string]string{metrics.LabelVertex: isdf.vertexName, metrics.LabelPipeline: isdf.pipelineName,
+			"buffer": isdf.fromBuffer.GetName()}).Observe(float64(time.Since(start).Microseconds()))
+	} else {
```
whynowy, line 511: I'm confused about this else section...

whynowy:
@jy4096 @yhl25 @KeranYang - please also review.

pkg/forward/forward.go
whynowy: This is for ctrl message.

pkg/udf/function/uds_grpc_test.go
```diff
@@ -272,6 +272,240 @@ func TestHGRPCBasedUDF_ApplyWithMockClient(t *testing.T) {
 	assert.Equal(t, expectedKeys, resultKeys)
 }
 
+func TestGRPCBasedUDF_BasicApplyStreamWithMockClient(t *testing.T) {
```
KeranYang, line 275: Since there is goroutines in these tests, I think we can apply the goleak library you found to ensure no leakage. I found it very useful. :)

```
func TestMain(m *testing.M) {
	goleak.VerifyTestMain(m)
}
```

pkg/sources/transformer/grpc_transformer.go
```diff
@@ -127,3 +127,7 @@ func (u *gRPCBasedTransformer) ApplyMap(ctx context.Context, readMessage *isb.Re
 	}
 	return taggedMessages, nil
 }
+
+func (u *gRPCBasedTransformer) ApplyMapStream(_ context.Context, _ *isb.ReadMessage, _ chan<- isb.WriteMessage) error {
+	return fmt.Errorf("method ApplyMapStream not implemented")
```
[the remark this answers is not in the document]
xdevxy: I don't think this matters, since this is not supported and we won't even allow transformer type with forwarder to be initialized. See [here](https://github.com/numaproj/numaflow/pull/709/files#diff-72611cd85ccc44e7cd2fcc37a91e5ac5c4973178be377d0a00ff41f65aee25acR116)

pkg/apis/numaflow/v1alpha1/vertex_types.go
```diff
@@ -355,6 +356,15 @@ func (v Vertex) GetReplicas() int {
 	return int(*v.Spec.Replicas)
 }
 
+func (v Vertex) MapUdfStreamEnabled() (bool, error) {
+	if v.Spec.Metadata != nil && v.Spec.Metadata.Annotations != nil {
+		if mapUdfStream := v.Spec.Metadata.Annotations[MapUdfStreamKey]; mapUdfStream != "" {
```
whynowy, line 361:
```golang
if mapUdfStream, existing := v.Spec.Metadata.Annotations[MapUdfStreamKey]; existing && mapUdfStream == "true" {
  return true
}
```

pkg/forward/forward.go
```diff
@@ -281,80 +264,116 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
...
+		// forward the message to the edge buffer (could be multiple edges)
+		writeOffsets, err = isdf.writeToBuffers(ctx, messageToStep)
+		if err != nil {
+			isdf.opts.logger.Errorw("failed to write to toBuffers", zap.Error(err))
+			isdf.fromBuffer.NoAck(ctx, readOffsets)
+			return
+		}
+	} else {
+		writeOffsets, err = isdf.streamMessage(ctx, dataMessages, processorWM)
+		if err != nil {
```
whynowy, line 372: log it?

pkg/forward/forward.go
```diff
@@ -416,6 +435,95 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
...
+
+			// Forward the message to the edge buffer (could be multiple edges)
+			curWriteOffsets, err := isdf.writeToBuffers(ctx, messageToStep)
+			if err != nil {
+				isdf.opts.logger.Errorw("failed to write to toBuffers", zap.Error(err))
+				return nil, err
+			}
+			// Merge curWriteOffsets into writeOffsets
+			for k, v := range curWriteOffsets {
+				writeOffsets[k] = v
```
whynowy, line 494: Should it be merging the offset array?

pkg/forward/forward.go
xdevxy: Already log all error cases respectively in `streamMessage`, so skipped here, or you want it to move here and have a general error msg for `streamMessage`?

pkg/apis/numaflow/v1alpha1/vertex_types.go
xdevxy: Hmm, `strconv.ParseBool` actually covers a boarder [true cases](https://pkg.go.dev/strconv) which should be better?

pkg/forward/forward.go
whynowy: Logging in the invoker might be easier and less code, then in the called function only need to do something like `return nil, fmt.Errorf("failed at xxxxx, error: %w", err)`, so that we can locate the error place.

whynowy:
It looks good to me, @numaproj/numaflow-dev - can you take a look as well to see if there's anything missing since it's a pretty complex change.

pkg/udf/function/uds_grpc.go
```diff
@@ -129,6 +130,67 @@ func (u *UDSgRPCBasedUDF) ApplyMap(ctx context.Context, readMessage *isb.ReadMes
...
+		Value:     payload,
+		EventTime: &functionpb.EventTime{EventTime: timestamppb.New(parentMessageInfo.EventTime)},
+		Watermark: &functionpb.Watermark{Watermark: timestamppb.New(message.Watermark)},
+		Metadata: &functionpb.Metadata{
+			Id:           id,
+			NumDelivered: numDelivered,
+		},
+	}
+
+	datumCh := make(chan *functionpb.DatumResponse)
```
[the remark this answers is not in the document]
KeranYang: Is the sender us(platform) or the UDF implementor? How can we ensure this channel is closed after `ApplyMapStream` is done?

KeranYang (approved):

pkg/udf/function/uds_grpc.go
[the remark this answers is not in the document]
xdevxy: Go sdk client, which is the sender of the datum responses, will close the channel .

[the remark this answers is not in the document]
xdevxy: Without closing the channel, it would be blocked at L172.

Issue #707, opened 2023-04-25
Update Native Redis ISBSVC to Redis 7


juliev0:
# Summary

Decision to update the Native Redis ISBSVC to version 7

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

juliev0:
[This](https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/VersionManagementConsiderations.html) is the list of incompatible changes in Redis 7.0 btw (and I noticed we're currently deploying 6.2.6 by default):

> SCRIPT LOAD and SCRIPT FLUSH are no longer propagated to replicas. If you need to have some durability for scripts, we recommend you consider using [Redis functions](https://redis.io/docs/manual/programmability/functions-intro/).
> 
> Pubsub channels are now blocked by default for new ACL users.
> 
> STRALGO command was replaced with the LCS command.
> 
> The format for ACL GETUSER has changed so that all fields show the standard access string pattern. If you had automation using ACL GETUSER, you should verify that it will handle either format.
> 
> The ACL categories for SELECT, WAIT, ROLE, LASTSAVE, READONLY, READWRITE, and ASKING have changed.
> 
> The INFO command now shows command stats per sub-command instead of in the top level container commands.
> 
> The return values of LPOP, RPOP, ZPOPMIN and ZPOPMAX commands have changed under certain edge cases. If you use these commands, you should check the release notes and evaluate if you are impacted.
> 
> The SORT and SORT_RO commands now require access to the entire keyspace in order to use the GET and BY arguments.

Release notes [here](https://raw.githubusercontent.com/redis/redis/7.0/00-RELEASENOTES)

juliev0:
Here are the [redis.conf](https://raw.githubusercontent.com/redis/redis/7.0/redis.conf) and [sentinel.conf](https://raw.githubusercontent.com/redis/redis/7.0/sentinel.conf) settings for 7.0. I will confirm that our numaflow-controller-config.yaml settings are still valid.

juliev0:
After speaking with Derek yesterday, my process is this:
1. run latest Helm chart to see what's deployed wrt StatefulSet, ConfigMaps
2. compare to what we're running currently to see what needs to be modified as far as the StatefulSet and the ConfigMaps both.

[...]

PR #717, opened 2023-04-30
feat: Redis7 as an ISB svc

juliev0:
Fixes: #707 

Update the hosted version of Redis to Redis 7.

Tested a Redis 7 ISBSVC with a simple pipeline running:
1. Deleted Leader Pod and observed Sentinel Pods elect new Leader. Simple pipeline was able to recover after new Leader elected.
2. Deleted both Slave Pods and observed them get recreated and rediscovered. Afterward, deleted Leader Pod. Simple pipeline was able to recover after new Leader elected.


pkg/reconciler/isbsvc/installer/assets/redis/config/replica.conf
```diff
@@ -1,5 +1,4 @@
 dir /data
-slave-read-only yes
```
whynowy, line 2: Remove this because it's the default behavior?

pkg/reconciler/isbsvc/installer/assets/redis/health/ping_readiness_master.sh
```diff
@@ -1,14 +1,21 @@
-#!/bin/bash
 
-[[ -f $REDIS_MASTER_PASSWORD_FILE ]] && export REDIS_MASTER_PASSWORD="$(< "${REDIS_MASTER_PASSWORD_FILE}")"
-[[ -n "$REDIS_MASTER_PASSWORD" ]] && export REDISCLI_AUTH="$REDIS_MASTER_PASSWORD"
+#!/bin/bash
```
whynowy, line 2: Should this be the first line?

pkg/reconciler/isbsvc/installer/assets/redis/config/replica.conf
juliev0: Yes, that and the fact that the default Helm chart doesn't set it.

juliev0: And it actually looks like it was explicitly taken out - the older chart (deploying v6.2) does include it and the newer chart (deploying v7) doesn't.

whynowy:
@juliev0 - can we make the redis default settings like below?

```
      redis:
        # Default Redis settings, could be overridden by InterStepBufferService specs
        settings:
          # Redis config shared by both master and replicas
          redis: |
            min-replicas-to-write 1
            # Disable RDB persistence, AOF persistence already enabled.
            save ""
            # Enable AOF https://redis.io/topics/persistence#append-only-file
            appendonly yes
            auto-aof-rewrite-percentage 100
            auto-aof-rewrite-min-size 64mb
            maxmemory 512mb
            maxmemory-policy allkeys-lru
          # Special config only used by master
          master: ""
          # Special config only used by replicas
          replica: ""
          # Sentinel config
          sentinel: |
            sentinel down-after-milliseconds mymaster 10000
            sentinel failover-timeout mymaster 2000
            sentinel parallel-syncs mymaster 1
```

pkg/apis/numaflow/v1alpha1/redis_buffer_service.go
```diff
@@ -321,9 +321,9 @@ func (nr NativeRedis) GetStatefulSetSpec(req GetRedisStatefulSetSpecReq) appv1.S
 				Image:           req.MetricsExporterImage,
 				ImagePullPolicy: metricsContainerPullPolicy,
 				Command: []string{"/bin/bash", "-c", `if [[ -f '/secrets/redis-password' ]]; then
-export REDIS_PASSWORD=$(cat /secrets/redis-password)
-fi
-redis_exporter`},
+			export REDIS_PASSWORD=$(cat /secrets/redis-password)
```
whynowy, line 324: Remove the indents?

pkg/apis/numaflow/v1alpha1/redis_buffer_service.go
```diff
@@ -374,8 +405,8 @@ redis_exporter`},
 		runAsUser := int64(1001)
 		fsGroup := int64(1001)
 		runAsUser0 := int64(0)
-		spec.Template.Spec.Containers[0].SecurityContext = &corev1.SecurityContext{RunAsUser: &runAsUser}
-		spec.Template.Spec.Containers[1].SecurityContext = &corev1.SecurityContext{RunAsUser: &runAsUser}
+		spec.Template.Spec.Containers[0].SecurityContext = &corev1.SecurityContext{RunAsUser: &runAsUser, RunAsGroup: &fsGroup}
+		spec.Template.Spec.Containers[1].SecurityContext = &corev1.SecurityContext{RunAsUser: &runAsUser, RunAsGroup: &fsGroup}
 		spec.Template.Spec.InitContainers = []corev1.Container{
 			{
```
juliev0, line 411: if our /data directory is 1001:1001, it seemed to me that maybe we should run as 1001:1001 instead of as 1001:root, although maybe it doesn't matter.

whynowy (approved):
LGTM

whynowy:
@juliev0 - could you resolve the conflict?

juliev0:
> @juliev0 - could you resolve the conflict?

resolved and confirmed everything still works

PR #726, opened 2023-05-07
feat: Autoscale for Redis Streams Source

### LINKED ISSUE #651: Redis Streams Source should implement "Pending" function
# Summary

We can do Autoscaling of a Redis Streams Vertex if we implement the Pending() function. This function will return the number of messages pending for the entire ConsumerGroup.

Can probably call `XINFO GROUPS <stream>` from [here](https://redis.io/docs/data-types/streams-tutorial/) and use "lag".

---

juliev0:
Fixes #651 
(If we merge this, we don't need to separately merge #728)

Pending() returns the total lag for the entire Consumer Group.

Note that I tested this by temporarily modifying the Read() method to only read one message at a time as well as Sleep. This resulted in the Pods being autoscaled upward.

### DIFF
--- a/docs/user-guide/sources/redis-source.md
+++ b/docs/user-guide/sources/redis-source.md
@@ -2,6 +2,8 @@
 
 A Redis Streams source is used to ingest messages from [Redis Streams](https://redis.io/docs/data-types/streams-tutorial/).
 
+It is recommended to use this with Redis versions >= 7.0 (in order for autoscaling to work).
+
 ## Example:
 
 ```yaml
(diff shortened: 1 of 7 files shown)


docs/user-guide/sources/redis-source.md
```diff
@@ -18,9 +20,9 @@ spec:
 ```
 
 Please see [API](https://github.com/numaproj/numaflow/blob/main/docs/APIs.md#redisstreamssource) for details on how to optionally do the following:
-- Define TLS
-- Define username/password
-- Connect to Redis Sentinel
```
juliev0, line 23: not really related, but I noticed the dashes didn't effectively turn into bullets, but the asterisks do

vigith (approved):

PR #734, opened 2023-05-10
feat: use metrics to calculate vertex processing rate v0

KeranYang:
This PR is the initial implementation of using metrics to calculate vertex processing rate.

At a high level, the approach I am taking is that for each of the vertices, I assign a RateCalculator. When a daemon server starts running, the list of RateCalculator also start. The RateCalculator maintains a map holding vertex processing rate for default, 1m, 5m and 15m lookback time. It also maintains a queue holding timestamped total counts. RateCalculator runs every refresh interval to grab total count metrics from pods, calculate new total count for the vertex, add a new entry to the timestamped total count queue and update the processing rate map.

On daemon server end, when receiving getMetrics API calls, it uses the RateCalculator GetRates() function to retrieve the latest processing rates for the corresponding vertex.

As a v0 implementation, this PR is **NOT** production ready. For the following reasons:
1. It doesn't work when a vertex has high number of pods. Currently the way we grab message count from pods are sequential. Need to make it run in parallel. I plan to use followup PRs to address this issue.
2. This hasn't been fully tested to cover all cases. The tests that I did are 1. unit tests and 2. a manual test using the simple-pipeline to observe the UI and verify the rate is ~5/s


pkg/daemon/server/daemon_server.go
```diff
@@ -137,7 +150,7 @@ func (ds *daemonServer) newGRPCServer(isbSvcClient isbsvc.ISBService, wmFetchers
 	}
 	grpcServer := grpc.NewServer(sOpts...)
 	grpc_prometheus.Register(grpcServer)
-	pipelineMetadataQuery, err := service.NewPipelineMetadataQuery(isbSvcClient, ds.pipeline, wmFetchers)
+	pipelineMetadataQuery, err := service.NewPipelineMetadataQuery(isbSvcClient, ds.pipeline, wmFetchers, rateCalculators, true)
```
KeranYang, line 153: I will change this line from using `true` to `false` before merging the PR such that the new rate calculation approach stays turned off until it's production ready.

pkg/daemon/server/daemon_server.go
```diff
@@ -87,6 +89,14 @@ func (ds *daemonServer) Run(ctx context.Context) error {
 		}
 	}()
 
+	rateCalculators := make(map[string]*server.RateCalculator, len(ds.pipeline.Spec.Vertices))
```
whynowy, line 92: Can we use one single `rateCalculator` for all the vertices? And we just new a calculator here, and pass it to `newGRPCServer()`.

The place to start the calculator should be before `<-ctx.Done()`.

whynowy:
Few points:

1. Maintain a structure (A) for available pods for all the vertices;
2. Detecting available pods should be separated from fetching the data, it's only used to update the structure A;
3. Fetching the data should be in parallel, it should only take the candidates in the structure mentioned above;
4. The fetched data is maintained in another structure B;
5. Calculating the rates should be separated from fetching the data, it only uses the data in structure B. It's okay either calculate the rates at the time it receives a request, or do it asynchronously (maintain) a cache.

KeranYang:
I addressed the comments in a new PR: https://github.com/numaproj/numaflow/pull/743 Closing this one.

Issue #725, opened 2023-05-05
Pipeline failed to restart though it couldn't connect to and create the consumergroup on kafka


louisnow:
**Describe the bug**
Pipeline failed to restart though it couldn't connect to and create the consumer group on kafka.
We have 4 pipelines, each with their own kafka topic source vertex.
Out of 4 pipelines, three restarted till they successfully connected but one stayed up despite being unable to restart at source.

**To Reproduce**

1. Our kafka/zookeeper instance went down
2. One of the source vertexes received a `kafka server: Replication-factor is invalid.` error but continued staying up instead of restarting. 

For some background, we scale down our clusters and our nodes when not in use during off business hours and scale it back up when needed to reduce our infra cost.

**Expected behavior**
- Pipelines Source vertex ideally restarts until success instead of staying up unsuccessfully

**Screenshots**

**Environment (please complete the following information):**

- Kubernetes: [e.g. v1.24]
- Numaflow: [e.g. v0.8.0]

**Additional context**
```
"numaflow.Source-processor","caller":"kafka/reader.go:394","msg":"Initialization of consumer failed with error: ","vertex":"atlan-nf-pipe-webhooks-kafka-source","error":"kafka server: Replication-factor is invalid."}

{"level":"info","ts":1683256265.8245962,"logger":"numaflow.Source-processor","caller":"fetch/processor_manager.go:145","msg":"Processor has been inactive for 10 heartbeats, deleting...","vertex":"atlan-nf-pipe-webhooks-kafka-source","key":"atlan-nf-pipe-webhooks-kafka-source-1","atlan-nf-pipe-webhooks-kafka-source-1":"atlan-nf-pipe-webhooks-kafka-source-1 status:active, timeline: [-1:-1] -> [-1:-1] -> [-1:-1] -> [-1:-1] -> [-1:-1] -> [-1:-1] -> [-1:-1] -> [-1:-1] -> [-1:-1] -> [-1:-1]"}
```

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

whynowy:
@louisnow - do you mean 3 of the pipeline sources restarted automatically and got back running, but one of them didn't?

whynowy:
> @louisnow - do you mean 3 of the pipeline sources restarted automatically and got back running, but one of them didn't?

And I assume manually killing the pod to get it restarted fixed the problem?

louisnow:
[...]

PR #735, opened 2023-05-11
fix: let kafka source crash and restart when there is any server side error

whynowy:
Fixes: #725

Explain what this PR does.

### DIFF
--- a/pkg/sources/kafka/reader.go
+++ b/pkg/sources/kafka/reader.go
@@ -324,18 +324,19 @@ func NewKafkaSource(
 		}
 	}
 	kafkasource.config = config
-	// Best effort to initialize the clients for pending messages calculation
-	adminClient, err := sarama.NewClusterAdmin(kafkasource.brokers, config)
-	if err != nil {
-		kafkasource.logger.Warnw("Problem initializing sarama admin client", zap.Error(err))
-	} else {
-		kafkasource.adminClient = adminClient
-	}
 	client, err := sarama.NewClient(kafkasource.brokers, config)
 	if err != nil {
-		kafkasource.logger.Warnw("Problem initializing sarama client", zap.Error(err))
+		return nil, fmt.Errorf("failed to create sarama client, %w", err)
 	} else {
 		kafkasource.saramaClient = client
+	}
+	adminClient, err := sarama.NewClusterAdmin(kafkasource.brokers, config)
+	if err != nil {
+		_ = kafkasource.saramaClient.Close()
+		// Does it require any special privileges to create a cluster admin client?
+		return nil, fmt.Errorf("failed to create cluster sarama admin client, %w", err)
+	} else {
+		kafkasource.adminClient = adminClient
 	}
 
 	ctx, cancel := context.WithCancel(context.Background())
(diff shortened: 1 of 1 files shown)


pkg/sources/kafka/reader.go
```diff
@@ -324,18 +324,21 @@ func NewKafkaSource(
...
 	if err != nil {
-		kafkasource.logger.Warnw("Problem initializing sarama admin client", zap.Error(err))
+		return nil, fmt.Errorf("failed to create sarama client, %w", err)
 	} else {
-		kafkasource.adminClient = adminClient
+		kafkasource.saramaClient = client
 	}
-	client, err := sarama.NewClient(kafkasource.brokers, config)
+	// Does it require any special privileges to create a cluster admin client?
+	adminClient, err := sarama.NewClusterAdminFromClient(client)
```
whynowy, line 334: Reference to https://github.com/kedacore/keda/blob/cf3cd1dc49bd77e86d57e8a7b12689f70dd91348/pkg/scalers/kafka_scaler.go#L393-L399

pkg/sources/kafka/reader.go
```diff
@@ -391,7 +394,8 @@ func (r *KafkaSource) startConsumer() {
 			// server-side rebalance happens, the consumer session will need to be
 			// recreated to get the new claims
 			if err := client.Consume(r.lifecyclectx, []string{r.topic}, r.handler); err != nil {
-				r.logger.Warnw("Initialization of consumer failed with error: ", zap.Error(err))
+				// Panic on errors to let it crash and restart the process
+				r.logger.Panicw("Consumer failed with error: ", zap.Error(err))
```
whynowy, line 398: Reference to https://github.com/Shopify/sarama/blob/main/examples/consumergroup/main.go#L109.

vigith: Does this return error and cause panic due to repartitioning on the server side? If yes, we shouldn't panic. We should do a manual repartition on a local kafka cluster and try it out.

I skimmed through the Sarama code, but it wasn't quite obvious to me.

whynowy: > Does this return error and cause panic due to repartitioning on the server side? If yes, we shouldn't panic. We should do a manual repartition on a local kafka cluster and try it out.
> 
> I skimmed through the Sarama code, but it wasn't quite obvious to me.

I didn't find any `Initialization of consumer failed with error` logs in our prod pipelines in the last 15 days, I assume repartitioning won't cause an error.

pkg/sources/kafka/reader.go
```diff
@@ -196,6 +196,23 @@ func (r *KafkaSource) Ack(_ context.Context, offsets []isb.Offset) []error {
 func (r *KafkaSource) NoAck(_ context.Context, _ []isb.Offset) {}
 
 func (r *KafkaSource) Start() <-chan struct{} {
+	client, err := sarama.NewClient(r.brokers, r.config)
```
whynowy, line 199: Move to `Start()` to make the test cases work.

vigith (approved):

PR #742, opened 2023-05-23
feat: using one bucket for partitioned reduce watermark propagation

### LINKED ISSUE #741: Use one bucket for partitioned reduce watermark propagation
# Summary

After refactoring [buffers and buckets](https://github.com/numaproj/numaflow/pull/733), the buckets used for partitioned reduce watermark propagation needs to be updated to one.

Once this is done, clean up all the code marked with `TODO(one bucket)`.

---

yhl25:
fixes #741 

* share OT buckets for offset timelines with partition IDs
* share PROCESSOR bucket between different partitions IDS
* Refactor
  * moved `processormanager` and heartbeat from edge_fetcher to processor
  * timeline  also moved from to timeline package


pkg/watermark/generic/jetstream/generic.go
```diff
@@ -53,9 +53,9 @@ func BuildWatermarkProgressors(ctx context.Context, vertexInstance *v1alpha1.Ver
 	pipelineName := vertexInstance.Vertex.Spec.PipelineName
 	fromBucket := vertexInstance.Vertex.GetFromBuckets()[0]
 
-	// TODO(one bucket): remove this after we combine the buckets to one for reduce.
-	if vertexInstance.Vertex.IsReduceUDF() {
-		fromBucket = fmt.Sprintf("%s-%d", fromBucket, vertexInstance.Replica)
+	println("from buckets: ")
```
jy4096, line 56: clean up? or make it log?

pkg/watermark/generic/jetstream/generic.go
```diff
@@ -53,9 +53,9 @@ func BuildWatermarkProgressors(ctx context.Context, vertexInstance *v1alpha1.Ver
...
+	for _, bucket := range vertexInstance.Vertex.GetFromBuckets() {
+		println(bucket)
```
jy4096, line 58: clean up? or make it log?

jy4096 (commented):
Just a note that... I think previously we put everything in fetch pkg because only fetch uses these datatypes and functions (if I remember correctly)

pkg/watermark/generic/jetstream/generic.go
yhl25: removed

pkg/forward/forward_test.go
```diff
@@ -1170,12 +1170,14 @@ func buildPublisherMapAndOTStore(toBuffers map[string]isb.BufferWriter) (map[str
 	processorEntity := processor.NewProcessorEntity("publisherTestPod")
 	publishers := make(map[string]publish.Publisher)
 	otStores := make(map[string]wmstore.WatermarkKVStorer)
+	index := int32(0)
```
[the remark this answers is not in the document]
yhl25: its not needed

examples/6-reduce-fixed-window.yaml
```diff
@@ -39,5 +39,6 @@ spec:
       to: atoi
     - from: atoi
       to: compute-sum
+#      parallelism: 2
```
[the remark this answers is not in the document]
yhl25: deleted

yhl25:
> Just a note that... I think previously we put everything in fetch pkg because only fetch uses these datatypes and functions (if I remember correctly)

Since I moved the processor manager out of the edge fetcher, it will be used in different packages, so created a separate package for processor stuff. That way it's easy to test and avoids confusion. Because the fetcher should not create the processor manager it should just use the processor manager that was passed to get the active processors. I think it's better to separate these two components to avoid confusion.

pkg/watermark/generic/jetstream/generic.go
```diff
@@ -99,44 +99,28 @@ func BuildWatermarkProgressors(ctx context.Context, vertexInstance *v1alpha1.Ver
...
+				return nil, nil, fmt.Errorf("failed at new HB Publish JetStreamKVStore, HeartbeatPublisherBucket: %s, %w", hbPublisherBucketName, err)
+			}
+
+			otStoreBucketName := isbsvc.JetStreamOTBucket(toBucket)
+			otStore, err := jetstream.NewKVJetStreamKVStore(ctx, pipelineName, otStoreBucketName, jsclient.NewInClusterJetStreamClient())
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed at new OT Publish JetStreamKVStore, OTBucket: %s, %w", otStoreBucketName, err)
+			}
+			opts := []publish.PublishOption{}
+			if vertexInstance.Vertex.IsASink() {
```
whynowy, line 114: This can be remove, there is already a branch for `IsASink()` above.

pkg/watermark/publish/options.go
```diff
@@ -34,6 +34,10 @@ type publishOptions struct {
 	// Whether it is sink publisher or not
 	// isSource and isSink should not be both true
 	isSink bool
+	// isReduce to indicate it is reduce publisher
+	isReduce bool
```
whynowy, line 38: Use `isToVertexReduce` to void confusion?

pkg/watermark/publish/options.go
```diff
@@ -34,6 +34,10 @@ type publishOptions struct {
...
+	// toVertexPartition to indicate the partition of toVertex buffer
```
whynowy, line 39: Also mention it's only useful when `isToVertexReduce` is true.

whynowy:
Also search `TODO(one bucket)` and clean up them all.

pkg/watermark/publish/options.go
vigith: let's delete this option.

pkg/watermark/publish/options.go
vigith: this should be true for everything. that means we will be adding partition info for map too.

whynowy (approved):

PR #743, opened 2023-05-23
feat: use metrics to calculate vertex processing rate

KeranYang:
@whynowy @numaproj/numaflow-dev , team, although I am publishing it as a draft PR, this change is code-wise ready to be reviewed. The functionality has been tested using a simple pipeline as follow.

Screenshot 2023-05-16 at 11 42 12 AM

Please review it and leave comments. While I collect feedbacks from PR review, I will be manually testing it with more use cases like reduce, auto-scaling, running on IKS etc.

### More test notes
* Case 1: 1k TPS

Screenshot 2023-05-23 at 3 13 54 PM

* Case 2: auto-scaling

Verified with high throughput, vertex p1 can scale up and down according to the traffic change.

* Case 3: more pods within a vertex

Screenshot 2023-05-23 at 4 06 55 PM

* Case 4: reduce pipeline

Screenshot 2023-05-23 at 4 44 56 PM


pkg/apis/numaflow/v1alpha1/const.go
```diff
@@ -131,6 +131,7 @@ const (
 	DefaultCooldownSeconds          = 90  // Default cooldown seconds after a scaling operation
 	DefaultZeroReplicaSleepSeconds  = 180 // Default sleep time in seconds after scaling down to 0, before peeking
 	DefaultMaxReplicas              = 50  // Default max replicas
+	MaxReplicasLimit                = 100 // The highest number of replicas allowed
```
whynowy, line 134: Can we use https://github.com/numaproj/numaflow/blob/ba1f493dc57acd2ba080a8107adbaa6b7c27fa7a/pkg/apis/numaflow/v1alpha1/vertex_types.go#L568?

pkg/daemon/server/service/rater/pod_tracker.go
```diff
@@ -0,0 +1,132 @@
...
+		ticker := time.NewTicker(pt.refreshInterval)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				for _, v := range pt.pipeline.Spec.Vertices {
+					var limit int
+					if max := v.Scale.Max; max != nil {
```
whynowy, line 85: Use `v.Scale.GetMaxReplicas()`

pkg/daemon/server/service/rater/rater.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+// Function monitor() defines each of the worker's job.
+// It waits for keys in the channel, and starts a monitoring job
+func (r *Rater) monitor(ctx context.Context, id int, keyCh <-chan string) {
+	r.log.Debugf("Started monitoring worker %v", id)
+	for {
+		select {
+		case <-ctx.Done():
+			r.log.Debugf("Stopped monitoring worker %v", id)
```
whynowy, line 92: `Infof`

pkg/apis/numaflow/v1alpha1/const.go
KeranYang: I see. By using `GetMaxReplicas`, if user doesn't specify max, we will use `DefaultMaxReplicas` 50. I am ok with that.

Regarding the `MaxReplicasLimit`, do we have a limit on how many replicas are allowed? The concern is if user specifies a very high max, the rator might break because of sending too many http requests within short period of time.

whynowy: What `too many http requests` do you mean? Live pod check with `HEAD`, or the other one?

KeranYang: Live pod check with head

whynowy: I don't think you check up to the `MaxReplicasLimit` right? even if a very high `max` is specified, you only check up to the one is live, right? BTW - if that really happens, then you should enhance the live pod check logic, instead of doing tick triggering, do something similar as you did for count check.

KeranYang: Yes. I think we can use tick triggering for now and revisit the decision once we have use case of high replicas. Will address the comment.

KeranYang:
Addressed early comments.

I found one issue during manual tests: the processing rate for reduce vertex showing on UI is incorrect. I created a separate issue to dive deeper into it. https://github.com/numaproj/numaflow/issues/746

This PR is now officially ready for review

pkg/daemon/server/service/pipeline_metrics_query.go
```diff
@@ -196,15 +214,17 @@ func (ps *pipelineMetadataQuery) GetVertexMetrics(ctx context.Context, req *daem
...
-			// Check if the resultant metrics list contains the processingRate, if it does look for the period label
-			if value, ok := result[metrics.VertexProcessingRate]; ok {
-				metricsList := value.GetMetric()
-				for _, metric := range metricsList {
-					labels := metric.GetLabel()
-					for _, label := range labels {
-						if label.GetName() == metrics.LabelPeriod {
-							lookback := label.GetValue()
-							processingRates[lookback] = metric.Gauge.GetValue()
+			if !ps.useNewRateCalculation {
```
vigith, line 217: will this part of the code be deleted once we make `useNewRateCalculation` the standard? if yes, could you please add a TODO?

pkg/daemon/server/service/rater/rater.go
```diff
@@ -0,0 +1,245 @@
...
+			log.Debugf("Pod %s does not exist, removing it from the active pod list...", podName)
+			activePods.Remove(key)
+		}
+	} else {
+		log.Debugf("Pod %s does not exist, updating it with CountNotAvailable...", podName)
+		count = CountNotAvailable
+	}
+	// track the total counts using a 10-second time window
+	// e.g. if the current time is 12:00:07, the count will be tracked in the 12:00:00-12:00:10 time window using 12:00:10 as the timestamp
+	now := time.Now().Add(time.Second * 10).Truncate(time.Second * 10).Unix()
```
vigith, line 126: I see a lot of `time.Second * 10`.. can we make it a const so it is easy to update if need be?

pkg/daemon/server/service/rater/pod_tracker.go
```diff
@@ -0,0 +1,126 @@
...
+					var vType string
+					if v.IsReduceUDF() {
+						vType = "reduce"
+					} else {
+						vType = "non_reduce"
+					}
+					for i := 0; i < int(v.Scale.GetMaxReplicas()); i++ {
+						podName := fmt.Sprintf("%s-%s-%d", pt.pipeline.Name, v.Name, i)
+						// podKey is used as a unique identifier for the pod, it is used by worker to determine the count of processed messages of the pod.
+						// "*" is used as a separator such that the worker can split the key to get the pipeline name, vertex name, pod index and vertex type.
```
vigith, line 93: Chose "*" because it is not a valid DNS char? just curious :).. perhaps make it a const so we can use it at split side too.

pkg/shared/logging/log.go
```diff
@@ -33,6 +34,7 @@ func NewLogger() *zap.SugaredLogger {
 		config = zap.NewProductionConfig()
 	}
 	// Config customization goes here if any
+	config.EncoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
```
vigith, line 37: ODL follows `RFC3339xxx` any particular reason for `ISO8601`. either ways i am fine :)

vigith (approved):

pkg/shared/logging/log.go
```diff
@@ -34,7 +34,7 @@ func NewLogger() *zap.SugaredLogger {
 		config = zap.NewProductionConfig()
 	}
 	// Config customization goes here if any
-	config.EncoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
+	config.EncoderConfig.EncodeTime = zapcore.RFC3339TimeEncoder
```
vigith, line 37: will this show milli-seconds?

KeranYang: no

vigith: use RFC3339Nano?

pkg/daemon/server/service/rater/pod_tracker.go
```diff
@@ -0,0 +1,131 @@
...
+						vType = "reduce"
+					} else {
+						vType = "non_reduce"
+					}
+					for i := 0; i < int(v.Scale.GetMaxReplicas()); i++ {
+						podName := fmt.Sprintf("%s-%s-%d", pt.pipeline.Name, v.Name, i)
+						// podKey is used as a unique identifier for the pod, it is used by worker to determine the count of processed messages of the pod.
+						podKey := strings.Join([]string{pt.pipeline.Name, v.Name, fmt.Sprintf("%d", i), vType}, PodInfoSeparator)
+						if pt.isActive(v.Name, podName) {
+							pt.activePods.PushBack(podKey)
```
whynowy, line 101: Do we need to PushBack? Will that be a problem with the pod metrics monitoring, that both of them are doing some `PushBack` operations.

whynowy: Ignore it, `PushBack` is not `MoveBack`.

whynowy:
As a follow up, please also clean up the `rate()` function in ISB readers and writers.

pkg/shared/logging/log.go
KeranYang: Sure will do. +1

KeranYang:
> As a follow up, please also clean up the rate() function in ISB readers and writers.

@whynowy yes, will do when I create PR to remove the `useNewRateCalculation` flag.

whynowy (approved):

PR #751, opened 2023-05-25
feat: support multi-partitioned edges

### LINKED ISSUE #748: Enhance the watermark to support multi-partitioned edges.

### LINKED ISSUE #766: Vertex to be able to process messages from multiple partitions, read/write

yhl25:
pr changes include

- map and sink vertex support to read from multiple partitions
- map and source vertex support to write to multiple partitions in round-robin fashion
- watermark enhancements to support multi-partitioned edges.
- metrics server changes to support multi-partitioned edges.

fixes #748 
fixes #766 


pkg/watermark/processor/processor_manager.go
```diff
@@ -186,7 +190,14 @@ func (v *ProcessorManager) startHeatBeatWatcher() {
...
 					var entity = NewProcessorEntity(value.Key())
-					var fromProcessor = NewProcessorToFetch(v.ctx, entity, 10)
+					var fromProcessor *ProcessorToFetch
+					// if the processor is a reduce or source processor, then we only need one fromProcessor
+					// because the reduce or source will read from only one partition.
+					if v.opts.isReduce || v.opts.isSource {
+						fromProcessor = NewProcessorToFetch(v.ctx, entity, 10, 1)
+					} else {
+						fromProcessor = NewProcessorToFetch(v.ctx, entity, 10, v.fromBufferPartitionCount)
+					}
```
vigith, line 200:
```suggestion
					partitionCnt := 0
					if v.opts.isReduce || v.opts.isSource {
						partitionCnt = 1
					} else {
						partitionCnt = v.fromBufferPartitionCount
					}
					fromProcessor = NewProcessorToFetch(v.ctx, entity, 10, partitionCnt)
```

pkg/watermark/processor/processor_manager.go
```diff
@@ -263,11 +274,20 @@ func (v *ProcessorManager) startTimeLineWatcher() {
 					continue
 				}
 				if otValue.Idle {
-					p.offsetTimeline.PutIdle(otValue)
+					if v.opts.isReduce {
```
vigith, line 277: what about source? answer: #633

pkg/watermark/processor/processor_manager.go
```diff
@@ -263,11 +274,20 @@ func (v *ProcessorManager) startTimeLineWatcher() {
...
+						p.offsetTimelines[0].PutIdle(otValue)
+					} else {
+						p.offsetTimelines[otValue.Partition].PutIdle(otValue)
+					}
 					v.log.Debugw("TimelineWatcher- Updates", zap.String("bucket", v.otWatcher.GetKVName()), zap.Int64("idleWatermark", otValue.Watermark), zap.Int64("idleOffset", otValue.Offset))
 				} else {
 					// NOTE: currently, for source edges, the otValue.Idle is always false
-					p.offsetTimeline.Put(otValue)
+					// for reduce we will always have only one partition
+					if v.opts.isReduce {
```
vigith, line 286: what about source?

yhl25: source always publishes the wmb with partition set to 0

vigith (approved):

jy4096:
Could you also add a description to explain what the PR solves in the PR description or in the issue itself?

xdevxy:
> Could you also add a description to explain what the PR solves in the PR description or in the issue itself? e.g. example questions to cover: what are the changes, what is the benefit, comparison before and after this PR, what is the outcome... etc.

+1, would appreciate high level background description of the change. Thanks!

pkg/watermark/fetch/edge_fetcher.go
```diff
@@ -71,15 +71,27 @@ func (e *edgeFetcher) GetWatermark(inputOffset isb.Offset) wmb.Watermark {
...
+		headOffset := int64(-1)
+		// iterate over all the timelines of the processor and get the smallest watermark
 		debugString.WriteString(fmt.Sprintf("[Processor: %v] \n", p))
-		var t = p.GetOffsetTimeline().GetEventTime(inputOffset)
-		if t == -1 { // watermark cannot be computed, perhaps a new processing unit was added or offset fell off the timeline
-			epoch = t
-		} else if t < epoch {
-			epoch = t
+		for _, tl := range p.GetOffsetTimelines() {
+			var t = tl.GetEventTime(inputOffset)
```
whynowy, line 79: `inputOffset` is from a particular partition, while `GetOffsetTimelines()` returns OT for different partitions, is that correct? If yes, is this logic correct?

pkg/forward/forward.go
```diff
@@ -236,7 +236,7 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
 			return
 		}
```
juliev0, line 237: So, my understanding is that we are getting the smallest watermark from all publishers and using that as our Idle Watermark. Why do we restrict it to just the current partition?

pkg/watermark/processor/processor_to_fetch.go
```diff
@@ -48,38 +49,48 @@ func (s status) String() string {
...
-	offsetTimeline *timeline.OffsetTimeline
-	lock           sync.RWMutex
-	log            *zap.SugaredLogger
+	ctx             context.Context
+	entity          ProcessorEntitier
+	status          status
+	offsetTimelines []*timeline.OffsetTimeline
+	lock            sync.RWMutex
+	log             *zap.SugaredLogger
 }
```
juliev0, line 60: might be helpful to have a comment for `offsetTimelines` that it's one per partition?

vigith:
closing and reopening to get the latest changes

pkg/watermark/fetch/source_fetcher.go
```diff
@@ -67,8 +67,8 @@ func (e *sourceFetcher) GetWatermark(_ isb.Offset) wmb.Watermark {
 		if !p.IsActive() {
 			continue
 		}
-		if p.GetOffsetTimeline().GetHeadWatermark() < epoch {
-			epoch = p.GetOffsetTimeline().GetHeadWatermark()
+		if p.GetOffsetTimelines()[fromPartitionIdx].GetHeadWatermark() < epoch {
```
KeranYang, line 70: please add nil check for array out of bound exception.

pkg/watermark/processor/processor_manager_test.go
```diff
@@ -19,11 +19,11 @@ package processor
 import (
```
KeranYang, line 19: Changes to this file reverted a previous goroutine leak fix. Please apply the fix in https://github.com/numaproj/numaflow/pull/774

KeranYang (changes requested):
Thanks for making the changes Yashash. This is indeed a **BIG** PR. I feel to review such PR, one needs to have a thorough understanding of multiple pieces of the system, watermark, metrics, partitions, forwarder etc. This is quite challenging to me as I context-switch from files to files while maintaining the high level picture in mind.

I wonder if it would be better next time that we publish small PRs one at a time, each focusing on one piece of the functionalities. It helps detect potential issues early and also make both implementor/review more focused. See more benefits: https://artsy.github.io/blog/2021/03/09/strategies-for-small-focused-pull-requests/

pkg/watermark/fetch/edge_fetcher.go
```diff
@@ -71,25 +85,45 @@ func (e *edgeFetcher) GetWatermark(inputOffset isb.Offset) wmb.Watermark {
...
+	// update the last processed watermark for the partition
+	e.lastProcessedWm[fromPartitionIdx] = epoch
+	// get the smallest watermark among all the partitions
+	// since we cannot compare the offset of different partitions, we get the smallest among the last processed watermarks of all the partitions
+	minEpoch := e.getMinFromLastProcessed(epoch)
 	e.log.Debugf("%s[%s] get watermark for offset %d: %+v", debugString.String(), e.bucketName, offset, epoch)
 
-	return wmb.Watermark(time.UnixMilli(epoch))
+	return wmb.Watermark(time.UnixMilli(minEpoch))
 }
```
juliev0, line 127: This is a bit of a nitpick, but I'm thinking about whether we should consider separating this into two methods. If I look at the method signature itself `GetWatermark(inputOffset isb.Offset, fromPartitionIdx int32)` it seems to suggest to me that we're getting the watermark just from this partition, but we are actually getting it from all of the partitions. I'm wondering what if we broke it into two functions: something like `ProcessPartitionOffset(inputOffset isb.Offset, fromPartitionIdx int32)` and `GetWatermark()`? What do you think?

pkg/udf/map_udf.go
```diff
@@ -45,149 +46,176 @@ func (u *MapUDFProcessor) Start(ctx context.Context) error {
...
+				shuffleFuncMap[fmt.Sprintf("%s:%s", edge.From, edge.To)] = s
+			}
 		}
-	}
 
-	conditionalForwarder := forward.GoWhere(func(keys []string, tags []string) ([]forward.VertexBuffer, error) {
-		var result []forward.VertexBuffer
+		// create a conditional forwarder for each partition
+		getVertexPartition := GetPartitionedBufferName()
+		conditionalForwarder := forward.GoWhere(func(keys []string, tags []string) ([]forward.VertexBuffer, error) {
```
juliev0, line 113: could all partitions use the same conditionalForwarder? I was trying to see what was different for each, but I may have missed it.

pkg/metrics/metrics.go
```diff
@@ -142,15 +148,15 @@ func NewMetricsServer(vertex *dfv1.Vertex, opts ...Option) *metricsServer {
...
+	if m.lagReaders != nil {
 		m.pendingInfo = sharedqueue.New[timestampedPending](1800)
 	}
 	return m
 }
 
 // Enqueue pending information
 func (ms *metricsServer) buildupPendingInfo(ctx context.Context) {
-	if ms.lagReader == nil {
+	if ms.lagReaders == nil {
```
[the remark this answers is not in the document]
jy4096: > || length <= 0?

length == 0 should be enough? there won't be a case for length to be less than zero right?

pkg/sources/source.go
```diff
@@ -340,3 +325,14 @@ func (sp *SourceProcessor) getTransformerGoWhereDecider(shuffleFuncMap map[strin
 	})
 	return fsd
 }
+
+// GetPartitionedBufferName returns a function that returns a partitioned buffer name based on the toVertex name and the partition count
```
jy4096, line 329: `a function that returns a partitioned buffer name` ?? is it buffer name or partition index?

pkg/sources/source.go
```diff
@@ -340,3 +325,14 @@ func (sp *SourceProcessor) getTransformerGoWhereDecider(shuffleFuncMap map[strin
...
+// it distributes the messages evenly to the partitions of the toVertex based on the message count(round robin)
+func GetPartitionedBufferName() func(toVertex string, toVertexPartitionCount int) int32 {
+	messagePerPartitionMap := make(map[string]int)
+	return func(toVertex string, toVertexPartitionCount int) int32 {
+		vertexPartition := (messagePerPartitionMap[toVertex] + 1) % toVertexPartitionCount
+		messagePerPartitionMap[toVertex] = vertexPartition
+		return int32(vertexPartition)
+	}
+}
```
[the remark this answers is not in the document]
vigith: i think it is ok, func programming uses [currying](https://en.wikipedia.org/wiki/Currying) a lot.

pkg/udf/function/uds_grpc.go
```diff
@@ -160,13 +159,12 @@ func (u *UDSgRPCBasedUDF) ApplyMap(ctx context.Context, readMessage *isb.ReadMes
...
 
 	writeMessages := make([]*isb.WriteMessage, 0)
-	for i, datum := range datumList {
+	for _, datum := range datumList {
 		keys := datum.Keys
 		taggedMessage := &isb.WriteMessage{
 			Message: isb.Message{
 				Header: isb.Header{
 					MessageInfo: parentMessageInfo,
-					ID:          fmt.Sprintf("%s-%d", offset.String(), i),
```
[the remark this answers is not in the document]
vigith: yes, primarily based of readoffset

pkg/watermark/fetch/edge_fetcher.go
```diff
@@ -44,24 +45,37 @@ type edgeFetcher struct {
...
 	}
 }
 
-// GetWatermark gets the smallest watermark for the given offset
-func (e *edgeFetcher) GetWatermark(inputOffset isb.Offset) wmb.Watermark {
+// GetWatermark gets the smallest watermark for the given offset and partition from all the active processors.
+// deletes the processor if it's not active.
+func (e *edgeFetcher) GetWatermark(inputOffset isb.Offset, fromPartitionIdx int32) wmb.Watermark {
+	e.Lock()
+	defer e.Unlock()
```
vigith, line 78: do we need a coarse lock, can't it be granular only where are updating the critical section.

pkg/watermark/fetch/edge_fetcher.go
```diff
@@ -125,8 +164,10 @@ func (e *edgeFetcher) GetHeadWatermark() wmb.Watermark {
 	return wmb.Watermark(time.UnixMilli(headWatermark))
 }
 
-// GetHeadWMB returns the latest idle WMB with the smallest watermark among all processors
-func (e *edgeFetcher) GetHeadWMB() wmb.WMB {
+// GetHeadWMB returns the latest idle WMB with the smallest watermark among all processors for the given partition.
+func (e *edgeFetcher) GetHeadWMB(fromPartitionIdx int32) wmb.WMB {
+	e.Lock()
```
vigith, line 169: coarse?

pkg/watermark/generic/noop.go
```diff
@@ -37,7 +37,7 @@ func NewNoOpWMProgressor() *NoOpWMProgressor {
 }
 
 // GetWatermark returns the default watermark.
-func (n NoOpWMProgressor) GetWatermark(_ isb.Offset) wmb.Watermark {
+func (n NoOpWMProgressor) GetWatermark(offset isb.Offset, partition int32) wmb.Watermark {
```
vigith, line 40:
```suggestion
func (n NoOpWMProgressor) GetWatermark(_ isb.Offset, _ int32) wmb.Watermark {
```

pkg/watermark/fetch/edge_fetcher.go
jy4096: I would say

`GetCurrentPartitionWatermark(inputOffset isb.Offset, partitionIdx int32)`
and
`GetEdgeWatermark()`

and we can refactor this in the next PR

pkg/watermark/processor/processor_manager.go
yhl25: source idling is not yet supported.

pkg/forward/forward.go
```diff
@@ -495,7 +495,11 @@ func (isdf *InterStepDataForward) streamMessage(
 
 		// Stream the message to the next vertex. First figure out which vertex
 		// to send the result to. Then update the toBuffer(s) with writeMessage.
+		msgIndex := 0
 		for writeMessage := range writeMessageCh {
+			// add partition to the ID, this is to make sure that the ID is unique across partitions
+			writeMessage.ID = fmt.Sprintf("%s-%d-%d", dataMessages[0].ReadOffset.String(), isdf.fromBuffer.GetPartition(), msgIndex)
```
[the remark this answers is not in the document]
yhl25: we were deduping inside `uds_grpc.go,` I moved it here since we don't have the partition information inside `uds_grpc.go`

pkg/isbsvc/jetstream_service.go
```diff
@@ -119,46 +119,51 @@ func (jss *jetStreamSvc) CreateBuffersAndBuckets(ctx context.Context, buffers, b
...
-					TTL:          v.GetDuration("otBucket.ttl"),
-					MaxBytes:     v.GetInt64("otBucket.maxBytes"),
-					Storage:      nats.StorageType(v.GetInt("otBucket.storage")),
-					Replicas:     v.GetInt("otBucket.replicas"),
-					Placement:    nil,
-				}); err != nil {
-					return fmt.Errorf("failed to create offset timeline bucket %q, %w", otBucket, err)
-				}
+		//TODO: remove sleep and use a better way to wait for the stream to be ready
+		//time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
yhl25: just removed it to get the e2e tests to pass. Will increase the test timeout.

pkg/metrics/metrics.go
```diff
@@ -142,15 +148,15 @@ func NewMetricsServer(vertex *dfv1.Vertex, opts ...Option) *metricsServer {
 			opt(m)
 		}
 	}
-	if m.lagReader != nil {
+	if m.lagReaders != nil {
```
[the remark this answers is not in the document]
yhl25: there won't be a case where len can be zero.

pkg/metrics/metrics.go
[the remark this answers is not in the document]
yhl25: there won't be a case where len can be zero.

pkg/metrics/metrics.go
```diff
@@ -161,21 +167,26 @@ func (ms *metricsServer) buildupPendingInfo(ctx context.Context) {
...
+				ms.pendingInfo.Append(ts)
+			}
 		}
 	}
 }
 
 // Expose pending metrics
 func (ms *metricsServer) exposePendingMetrics(ctx context.Context) {
-	if ms.lagReader == nil {
+	if ms.lagReaders == nil {
```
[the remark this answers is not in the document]
yhl25: there won't be a case where len can be zero.

pkg/metrics/metrics.go
```diff
@@ -187,7 +198,7 @@ func (ms *metricsServer) exposePendingMetrics(ctx context.Context) {
 	for {
 		select {
 		case <-ticker.C:
-			if ms.lagReader != nil {
+			if ms.lagReaders != nil {
```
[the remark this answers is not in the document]
yhl25: there won't be a case where len can be zero.

pkg/udf/map_udf.go
yhl25: ` GetPartitionedBufferName()` is separate for each partition

pkg/sources/source.go
yhl25: will update the name.

pkg/watermark/fetch/source_fetcher.go
yhl25: that will never happen.

pkg/watermark/processor/processor_manager_test.go
yhl25: may be got removed while resolving conflicts, will add that back.

pkg/watermark/fetch/edge_fetcher.go
yhl25: I don't think we need separate methods.

pkg/udf/map_udf.go
juliev0: It looks like `GetPartitionedBufferName()` is set to the same function each time? It seems to me that in order for there to be some difference in all of these ConditionalForwarders, I'd need to see `index` or `bufferPartition` (which we're iterating through) used somewhere on those lines. Now, on the other hand, if `conditionalForwarder` requires state (it's something that gets modified in the course of running), then it's reasonable to have separate instances of them. Anyway, it's not a big deal either way.

juliev0: never mind, I see the state is the map defined in GetPartitionedBufferName()

pkg/watermark/fetch/source_fetcher.go
yhl25: Since these methods are only used within the platform, we can avoid these unnecessary if checks, since we control the input parameters. It makes sense to have those checks if the methods are exposed to users. And also if we are getting an invalid partitionCount it is better to catch those instead of ignoring them using an if check otherwise we will have a bigger problem.

pkg/watermark/fetch/edge_fetcher.go
```diff
@@ -44,24 +45,37 @@ type edgeFetcher struct {
 	bucketName       string
 	storeWatcher     store.WatermarkStoreWatcher
 	processorManager *processor.ProcessorManager
+	lastProcessedWm  []int64
```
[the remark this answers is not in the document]
yhl25: we do integer comparison while fetching the watermark so its easy if we use array of integers and also avoids unnecessary type conversions.

pkg/watermark/fetch/source_fetcher.go
KeranYang: I don't fully agree with `Since these methods are only used within the platform, we can avoid these unnecessary if checks, since we control the input parameters.` I think each class/struct should take the responsibility of ensuring no NullPointer instead of the entire project/platform, because otherwise there could be very few cases when we need to do null check.

I agree with the second point that we'd rather to naturally panic instead of a if check. Resolving this comment.

pkg/forward/forward.go
```diff
@@ -223,9 +223,9 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
 		// situation as idling.
 		// In order to continue propagating watermark, we will set watermark idle=true and publish it.
 		// We also publish a control message if this is the first time we get this idle situation.
-
-		// we use the HeadWMB as the watermark for the idle
-		var processorWMB = isdf.wmFetcher.GetHeadWMB()
+		// we use the HeadWMB as the watermark for the idle\
```
jy4096, line 226:
```suggestion
		// we use the HeadWMB as the watermark for the idle
```

pkg/watermark/fetch/edge_fetcher.go
juliev0: I still think the message signature's a little confusing and I'm a fan of making it more modular, but I'm okay if we do it at a later time, as it would be good to merge this now.

jy4096: > I still think the message signature's a little confusing and I'm a fan of making it more modular, but I'm okay if we do it at a later time, as it would be good to merge this now.

The signature is actually good IMO since it's a method of edge, edge.GetWatermark is the watermark for the edge which includes all the partitions

I just submitted a suggestion to add how we get the watermark in the godoc for more details.

pkg/watermark/fetch/edge_fetcher.go
```diff
@@ -125,8 +164,8 @@ func (e *edgeFetcher) GetHeadWatermark() wmb.Watermark {
 	return wmb.Watermark(time.UnixMilli(headWatermark))
 }
 
-// GetHeadWMB returns the latest idle WMB with the smallest watermark among all processors
-func (e *edgeFetcher) GetHeadWMB() wmb.WMB {
+// GetHeadWMB returns the latest idle WMB with the smallest watermark among all processors for the given partition.
```
jy4096, line 167:
```suggestion
// GetHeadWMB returns the latest idle WMB with the smallest watermark among all processors from all partitions for the given partition.
// We first calculate the watermark for the given partition and update the lastProcessedWatermark of the given partition using this watermark.
// Then, we compare this watermark with all the lastProcessedWatermark of other partitions and return the idle watermark as the edge watermark only if it's the minimum among all of the lastProcessedWatermark.
```

pkg/metrics/metrics.go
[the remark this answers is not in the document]
KeranYang: I personally prefer programming defensively, assuming that anything that could go wrong will go wrong and I feel that's a good practice most of the time. :)

PR #755, opened 2023-05-27
fix: pipeline view fix

### LINKED ISSUE #739: Pipeline view is broken
**Describe the bug**

After the refactor of https://github.com/numaproj/numaflow/pull/733, the pipeline view in the UI is broken.

veds-g:
Fixes #739 

- Moved BufferInfo from Edge to Vertex
- Renamed Edge Label, and it contains Watermark and Conditions
- Used Partitions instead of Parallelism to query for pod level metrics for a vertex
- Updated the watermark fetcher map to contain fetcher for each pod of a vertex

image
image


whynowy (approved):
LGTM, thanks!

PR #757, opened 2023-05-30
chore: refactoring code to support multi partitioned edges

yhl25:


pkg/sources/source.go
```diff
@@ -190,7 +205,7 @@ func (sp *SourceProcessor) Start(ctx context.Context) error {
 		}
 	}()
 
-	metricsOpts := metrics.NewMetricsOptions(ctx, sp.VertexInstance.Vertex, readyChecker, sourcer, writers[0])
+	metricsOpts := metrics.NewMetricsOptions(ctx, sp.VertexInstance.Vertex, readyChecker, sourcer, nil)
```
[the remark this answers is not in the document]
KeranYang: @jy4096 You are right, we don't need the rater any more. I plan to work on removing the argument as part of my legacy rater cleanup PR.

pkg/forward/interfaces.go
```diff
@@ -16,28 +16,24 @@ limitations under the License.
...
 
-import (
-	dfv1 "github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
-)
-
-var (
-	All  = GoWhere(func([]string, []string) ([]string, error) { return []string{dfv1.MessageTagAll}, nil })
-	Drop = GoWhere(func([]string, []string) ([]string, error) { return []string{dfv1.MessageTagDrop}, nil })
-)
+type Step struct {
```
whynowy, line 19: How about

```golang
type VertexPartition struct {
  Name string
  Partition int32
}
```

pkg/sinks/blackhole/blackhole.go
```diff
@@ -80,6 +83,18 @@ func NewBlackhole(vertex *dfv1.Vertex, fromBuffer isb.BufferReader, fetchWaterma
 	return bh, nil
 }
 
+func (b *Blackhole) getSinkGoWhereDecider() forward.GoWhere {
```
whynowy, line 86: Each sink needs to implement something like this?

pkg/forward/forward.go
```diff
@@ -62,18 +62,20 @@ type InterStepDataForward struct {
 	// idleManager manages the idle watermark status.
 	idleManager *wmb.IdleManager
 	// wmbChecker checks if the idle watermark is valid.
-	wmbChecker wmb.WMBChecker
+	wmbChecker           wmb.WMBChecker
+	toVertexPartitionMap map[string]int
```
whynowy, line 66: Do we need to have a map to track the partitions? Can we use the `len(val)` of `toBuffers`?

I can imagine it's different for a sink, but does it matter?

yhl25: removed it.

pkg/sinks/blackhole/blackhole.go
yhl25: done

whynowy (approved):

pkg/forward/forward.go
```diff
@@ -98,7 +98,7 @@ func NewInterStepDataForward(vertex *dfv1.Vertex,
 		vertexName:   vertex.Spec.Name,
 		pipelineName: vertex.Spec.PipelineName,
 		idleManager:  wmb.NewIdleManager(len(toSteps)),
-		wmbChecker:   wmb.NewWMBChecker(2), // TODO: make configurable
+		wmbChecker:   wmb.NewWMBChecker(2), // TODO: make configurable\
```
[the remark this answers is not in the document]
yhl25: yes

pkg/sources/source.go
```diff
@@ -146,14 +149,26 @@ func (sp *SourceProcessor) Start(ctx context.Context) error {
...
+				bufferWriters = append(bufferWriters, writer)
 			}
+			writersMap[e.To] = bufferWriters
 		}
 	default:
 		return fmt.Errorf("unrecognized isb svc type %q", sp.ISBSvcType)
 	}
 	var sourcer Sourcer
 	var readyChecker metrics.HealthChecker
+	// Populate shuffle function map
```
KeranYang, line 161: > there was a bug in source vertex code, if there was a reduce vertex after the source vertex we were not shuffling the messages.

@yhl25 , I don't think it is a bug. Before this PR, the contract of shuffling at source is that if the source vertex has a UDTransformer, then we shuffle using `getTransformerGoWhereDecider`, this is because no UDTransformer means no map operation, hence we don't expect an immediate reduce as next vertex simply because we haven't assign any keys to source messages. I have two questions:

1. With this change, we indicate that if the source doesn't have a UDTransformer, we can still have a reduce vertex immediately after it. Is it correct?

2. With multi-partitioned edge, we are going to support shuffling when the next vertex is a map. Assuming that's in place, would we be able to generalize `sp.getTransformerGoWhereDecider(shuffleFuncMap)` and `sp.getSourceGoWhereDecider(shuffleFuncMap)` into one?

PR #758, opened 2023-05-30
chore: use metrics to calculate rate for reducer partitions

### LINKED ISSUE #746: Implement per-partition rate calculation and display it on UI
As we migrate to metric-based rate calculation, we no longer have partition level processing rates for reduce vertex. The processing rates we delivered through daemon are ALL per vertex. Therefore we no longer support the per-partition rate feature:

Screenshot 2023-05-24 at 5 59 23 PM

In this task, we implement per-partition rate calculation and display it on UI.

---

KeranYang:
Fixes #746 

Showing the processing rate for each of the partitions is important as it helps us identify hot key issue. This change enables the rater to calculate rate for a single pod, such that we can retrieve per-partition rates and display them on UI.

### How was this change tested
**Before this change**

before

**After**

after

### DIFF
--- a/pkg/daemon/server/service/pipeline_metrics_query.go
+++ b/pkg/daemon/server/service/pipeline_metrics_query.go
@@ -253,12 +253,12 @@ func (ps *pipelineMetadataQuery) GetVertexMetrics(ctx context.Context, req *daem
 					Pendings: pendings,
 				}
 				if podNum == 1 {
-					// non-reduce vertex, or reduce vertex with single partition
-					// the processing rate of this pod is exactly same as the vertex level rate
+					// the vertex is either non-reduce, or single-partition reducer.
+					// in this case the processing rate of this pod is exactly same as the vertex level rate.
 					vm.ProcessingRates = vertexLevelRates
 				} else {
-					// reduce vertex with multiple partitions
-					// the processing rate of this pod is the rate of this partition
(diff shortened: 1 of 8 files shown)


pkg/daemon/server/service/pipeline_metrics_query.go
```diff
@@ -238,22 +207,23 @@ func (ps *pipelineMetadataQuery) GetVertexMetrics(ctx context.Context, req *daem
...
+				vm.ProcessingRates = ps.rater.GetPodRates(req.GetVertex(), int(i))
 			} else {
-				metricsArr[i] = &daemon.VertexMetrics{
-					Pipeline:        &ps.pipeline.Name,
-					Vertex:          req.Vertex,
-					ProcessingRates: vertexLevelRates,
-					Pendings:        pendings,
-				}
+				// if the vertex is not a reduce udf, then the processing rate is the sum of all pods in this vertex.
+				// TODO - change this to display the processing rate of each partition when we finish multi-partition support for non-reduce vertices.
```
vigith, line 222:
```suggestion
				// TODO (multi-partition) - change this to display the processing rate of each partition when we finish multi-partition support for non-reduce vertices.
```

we have too many TODOs, tagging with `multi-partition` might help while searching :)

vigith (approved):

PR #762, opened 2023-05-31
chore: fix idle watermark bug

yhl25:
Screenshot 2023-06-01 at 2 08 00 PM

fixes https://github.com/numaproj/numaflow/issues/748#issuecomment-1570702145


pkg/forward/forward.go
```diff
@@ -384,45 +386,46 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
...
-					activeWatermarkBuffers[toVertexName] = true
+					activeWatermarkBuffers[toVertexName][index] = true
 					// reset because the toBuffer is no longer idling
 					isdf.idleManager.Reset(toVertexName)
 				}
 			}
 		}
 	}
 	// TODO(multi-partition): handle idle watermark publishing for multi partitioned buffer
-	if len(dataMessages) > 0 && len(activeWatermarkBuffers) < len(isdf.publishWatermark) {
```
vigith, line 412: shouldn't we have a different variant of the same check?

KeranYang (commented):
I think we should add E2E tests to cover multi-partition

pkg/sources/generator/tickgen.go
```diff
@@ -233,7 +233,8 @@ func (mg *memgen) buildSourceWatermarkPublisher(publishWMStores store.WatermarkS
 	// for tickgen, it can be the name of the replica
 	entityName := fmt.Sprintf("%s-%d", mg.vertexInstance.Vertex.Name, mg.vertexInstance.Replica)
 	processorEntity := processor.NewProcessorEntity(entityName)
-	return publish.NewPublish(mg.lifecycleCtx, processorEntity, publishWMStores, publish.IsSource(), publish.WithDelay(mg.vertexInstance.Vertex.Spec.Watermark.GetMaxDelay()))
+	// source publisher toVertexPartitionCount will be 1
```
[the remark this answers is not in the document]
yhl25: its within the source itself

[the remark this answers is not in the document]
KeranYang: Oh, you mean source publisher is publishing to the "left" side of source vertex which is the source watermark KV store, hence toVertexPartitionCount is always 1?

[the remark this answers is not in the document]
vigith: let's add a comment in the code so it is easy for future readers.

pkg/forward/forward.go
yhl25: > I think we should add E2E tests to cover multi-partition. #763

multi-partition is not supported yet. I have just made the refactoring so that it's easy to support multi-partition. Multi-partition support and tests related to it will be a separate pr.

xdevxy (commented):
I don't see any new test case is added to ensure the bug is covered? :)

pkg/watermark/publish/publisher.go
```diff
@@ -242,7 +248,13 @@ func (p *publish) loadLatestFromStore() wmb.Watermark {
 
 // GetLatestWatermark returns the latest watermark for that processor.
 func (p *publish) GetLatestWatermark() wmb.Watermark {
-	return p.GetHeadWM()
+	var latestWatermark = wmb.Watermark(time.UnixMilli(-1))
```
xdevxy, line 251: Same here, we may want to create a const for `-1` and use it for any watermark initialization.

pkg/watermark/publish/publisher.go
```diff
@@ -242,7 +248,13 @@ func (p *publish) loadLatestFromStore() wmb.Watermark {
...
+	for _, wm := range p.headWatermarks {
+		if wm.After(time.Time(latestWatermark)) {
+			latestWatermark = wm
+		}
```
xdevxy, line 255: How to compare the watermark will be updated later when we finalize the design?

yhl25: this method is only used in tests.

pkg/watermark/publish/publisher.go
```diff
@@ -48,22 +48,25 @@ type Publisher interface {
 	GetLatestWatermark() wmb.Watermark
 }
 
+var initialWatermark = wmb.Watermark(time.UnixMilli(-1))
```
KeranYang, line 51:
```suggestion
const InitialWatermark = wmb.Watermark(time.UnixMilli(-1))
```

Can we also move this constant to pkg/watermark/wmb/wmb.go? I think it can be used across files under watermark.

KeranYang (approved):

jy4096 (commented):
Looks good

pkg/forward/forward.go
```diff
@@ -385,45 +388,49 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
...
-				if p, ok := isdf.publishWatermark[bufferName]; ok {
-					idlehandler.PublishIdleWatermark(ctx, isdf.toBuffers[bufferName][0], p, isdf.idleManager, 0, isdf.opts.logger, isdf.opts.vertexType, processorWM)
+	// - condition1 "len(dataMessages) > 0" :
+	//   Meaning, we do have some data messages, but we may not have written to all out buffers or its partitions.
+	//   It could be all data messages are dropped, or conditional forwarding to part of the out buffers.
+	//   If we don't have this condition check, when dataMessages is zero but ctrlMessages > 0, we will
+	//   wrongly publish an idle watermark without the ctrl message and the ctrl message tracking map.
+	// - condition 2 "len(activeWatermarkBuffers) < len(isdf.wmPublishers)" :
+	//   send idle watermark only if we have idle out buffers
+	if len(dataMessages) > 0 {
```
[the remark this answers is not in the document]
jy4096: We've agreed that, in the condition where the readMessage contains only control messages, we don't execute extra steps: because if the idling situation continues, the next `readMessages, err := isdf.fromBuffer.Read(ctx, isdf.opts.readBatchSize)` will have `len(readMessages)==0`, and we eventually will handle the idling.

pkg/forward/forward.go
```diff
@@ -385,45 +388,49 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
...
-				// same as read len==0 because there's no event published to the buffer
-				if p, ok := isdf.publishWatermark[bufferName]; ok {
-					idlehandler.PublishIdleWatermark(ctx, isdf.toBuffers[bufferName][0], p, isdf.idleManager, 0, isdf.opts.logger, isdf.opts.vertexType, processorWM)
+	// - condition1 "len(dataMessages) > 0" :
+	//   Meaning, we do have some data messages, but we may not have written to all out buffers or its partitions.
+	//   It could be all data messages are dropped, or conditional forwarding to part of the out buffers.
+	//   If we don't have this condition check, when dataMessages is zero but ctrlMessages > 0, we will
+	//   wrongly publish an idle watermark without the ctrl message and the ctrl message tracking map.
+	// - condition 2 "len(activeWatermarkBuffers) < len(isdf.wmPublishers)" :
+	//   send idle watermark only if we have idle out buffers
```
jy4096, line 424:
```suggestion
	//   send idle watermark only if we have idle out buffers
	// Note: When the len(dataMessages) is 0, meaning all the readMessages are control messages, we choose not to do extra steps
	// This is because, if the idle continues, we will eventually handle the idle watermark when we read the next batch where the len(readMessages) will be zero
```

pkg/forward/forward.go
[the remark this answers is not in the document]
jy4096: I think in reduce, we return early if len(dataMessages) is zero
so removing the `p.result>0` is the situation where we do have data messages, but after applying the reduce UDF, we don’t have any valid outputs, all messages are dropped, and we send idle watermark in this situation...

so the change looks good to me
no need to add the check

PR #775, opened 2023-06-02
chore: move go client code to numaflow repo

### LINKED ISSUE #692: remove SDK dependency from Numaflow
# Problem statement

Currently, the client code is in `numaflow-go` repo, so every time we have to make a change on the client part of the code we need to change it in` numaflow-go` repo and tag a new version just to update the `go.mod` entry in `numaflow`.

# Proposal

Move the client code to `numaflow` repo, that way we can remove the SDK dependency from `numaflow` and we only keep the server related code in Go SDK repo similar to Java and Python SDKs

kohlisid:
Moves the client code from the go-SDK to the numaflow repository.
Fixes #692


pkg/func-client/client/client.go
vigith, line 1: do we need `func-client`? can't it be `pkg/client/`

whynowy: pkg/sdkclient?

vigith: let's go with `pkg/sdkclient` ?

vigith:
Can we have sink-client also inside sdkclient? Something like

```
sdkclient
  udf/{client,clienttest}
  sink/{client,clienttest}
```

vigith (approved):

Issue #768, opened 2023-06-01
Performance Test for Numaflow Raptor


KeranYang:
# Summary

Rate calculation using metrics is feature-wise ready and merged. We need to performance test it to ensure auto-scaling works as expected with a 100+ pod pipeline.

* Daemon server should not crash.
* Vertex should be able to scale up and down.
* Rates should display accurately on UI.

# Use Cases

When would you use this?

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

KeranYang:
## Performance Test summary

### Start from Simple - 4k tps with 10 pods per vertex

![4kTps-10Pods](https://github.com/numaproj/numaflow/assets/13165977/bc773b20-7e2f-4677-b9c3-754b39c61af7)

### Add more pods - 4k tps with 50 pods per vertex

![4kTps-50Pods](https://github.com/numaproj/numaflow/assets/13165977/e29469a3-b167-4f83-9998-b7469f130493)

### Finally, 100 pods per vertex

![4kTps-100Pods](https://github.com/numaproj/numaflow/assets/13165977/ae542d7b-0cae-4d1f-9c1d-c94fb2f6bc3d)

For each case, I monitored the pipeline for 30 minutes and verified:
1. Daemon server didn't crash.
2. The processing rates are stable per 1m, 5m and 15m.

### A bug found

Fixed in PR: https://github.com/numaproj/numaflow/pull/780/

PR #780, opened 2023-06-07
fix: unexpected high processing rates

KeranYang:
During performance test, it was observed that sometimes we calculate extremely high processing rate. Following is a 4k TPS pipeline.

issue-1

The issue arises from the fact that our task interval mirrors our rate count window, which sometimes results in the omission of a pod. In these instances, when computing the difference with the subsequent window, we incorrectly consider the total processed count as the delta for that specific pod. This ultimately results in a falsely inflated processing rate.

In this PR, two changes are made to ensure that each count window NEVER misses a single pod. More details in the code changes.

This PR is tested by https://github.com/numaproj/numaflow/issues/768#issuecomment-1581525023


pkg/daemon/server/service/rater/options.go
```diff
@@ -28,14 +28,15 @@ type Option func(*options)
 func defaultOptions() *options {
 	// A simple example of how these numbers work together:
 	// Assuming we have 200 tasks, we have 20 workers, each worker will be responsible for approximately 10 tasks during one iteration.
-	// The task interval is 10 seconds, which means each task need to be picked up by a worker every 10 seconds.
-	// Hence a worker needs to finish processing 1 task in 1 second.
-	// Translating to numaflow language, for a 200-pod pipeline, a worker needs to finish scraping 1 pod in 1 second, which is a reasonable number.
-	// TODO - need to adjust the numbers once we have a better idea of the performance
+	// The task interval is 5 seconds, which means each task need to be picked up by a worker every 5 seconds.
```
xdevxy, line 31: How do we know 5 seconds is a better number?

pkg/daemon/server/service/rater/pod_tracker.go
```diff
@@ -125,6 +125,10 @@ func (pt *PodTracker) isActive(vertexName, podName string) bool {
 	// example for 0th pod : https://simple-pipeline-in-0.simple-pipeline-in-headless.default.svc.cluster.local:2469/metrics
 	url := fmt.Sprintf("https://%s.%s.%s.svc.cluster.local:%v/metrics", podName, pt.pipeline.Name+"-"+vertexName+"-headless", pt.pipeline.Namespace, v1alpha1.VertexMetricsPort)
 	if _, err := pt.httpClient.Head(url); err != nil {
+		// during performance test (100 pods per vertex), we never saw a false negative, meaning every time isActive returns false,
+		// it truly means the pod doesn't exist.
+		// in reality, we can imagine that a pod can be active but the Head request times out for some reason and returns an incorrect false,
```
xdevxy, line 130: I guess that is a very legitimate case. Do we have the time out configured some how?

pkg/daemon/server/service/rater/options.go
KeranYang: We use a 10s time window to collect count metrics. To guarantee that we include ALL pods in each of the 10s window, we need to ensure each pod gets picked up and scrapped every 5 seconds(10/2). I won't say it's a better number. It's our current configuration which is tested to be able to handle 100-pod vertices' use case.

pkg/daemon/server/service/rater/timestamped_counts.go
```diff
@@ -45,7 +46,14 @@ func (tc *TimestampedCounts) Update(podName string, count float64) {
 	tc.lock.Lock()
 	defer tc.lock.Unlock()
 	if count == CountNotAvailable {
-		delete(tc.podCounts, podName)
```
vigith, line 48: won't this cause a leak? can't we garbage collect by adding a last-seen value? if not seen for the last x seconds (say 180), we can delete it.

KeranYang: Good question. I don't think this will cause a leak. If a pod is turned down, the PodTracker will remove it from the activePod list. If a pod doesn't exist in the activePod list, then a task will never pick it up, meaning when we start the next round of populating TimestampedCounts, that pod won't be included.

pkg/daemon/server/service/rater/options.go
vigith: even though we show wrong rates, it will converge if two consecutive fetches succeed.

vigith (approved):

pkg/daemon/server/service/rater/options.go
KeranYang: The way we calculate rates is that we calculate total delta by summing up all deltas among each of the consecutive TimestampedCounts pairs within lookback seconds. If the lookback second is 300s, there will be ~30 deltas calculated. If we have a pod wrongly missed in one of the TimestampedCounts, the calculated rate for lookback 300s will continue being off until that TimestampedCounts passes the 300s timeline, meaning the full converge takes 300 seconds.

vigith: fair point

xdevxy (approved):

PR #789, opened 2023-06-13
feat: support UI for multipartition edges

### LINKED ISSUE #767: Multi partitioning UI support

veds-g:
Fixes #767 
- updated GetHeadWatermark() api to take partitionIdx. That way we will have a head watermark for each partition.
- updated watermark fetcher map to include single fetcher for non-reduce vertex, and a fetcher for each partition of reduce vertex
- updated metrics query loop for non-reduce vertex, to return once, since it is the same aggregated value across all pods of the non-reduce vertex


pkg/daemon/server/service/pipeline_metrics_query.go
```diff
@@ -173,8 +173,15 @@ func (ps *pipelineMetadataQuery) GetVertexMetrics(ctx context.Context, req *daem
 	abstractVertex := ps.pipeline.GetVertex(req.GetVertex())
 	vertexLevelRates := ps.rater.GetRates(req.GetVertex())
 
-	metricsArr := make([]*daemon.VertexMetrics, abstractVertex.GetPartitionCount())
-	for i := int64(0); i < int64(abstractVertex.GetPartitionCount()); i++ {
+	metricsCount := 1
+	// currently metrics is an aggregation per vertex, so same across each pod for non-reduce vertex
+	// once multi-partition metrics are in - need to modify to per partition for every vertex
```
vigith, line 178: is this a TODO?

pkg/daemon/server/service/pipeline_watermark_query.go
```diff
@@ -68,7 +60,7 @@ func (ps *pipelineMetadataQuery) GetPipelineWatermarks(ctx context.Context, requ
 		watermarkArr := make([]*daemon.EdgeWatermark, len(ps.watermarkFetchers))
 		i := 0
 		for k := range ps.watermarkFetchers {
-			edgeName := k
+			edgeName := k.From + "-" + k.To
```
vigith, line 63: since `k` is `v1alpha1.Edge`, perhaps we should write a `Stringer` for `v1alpha1.Edge`?

pkg/watermark/fetch/edge_fetcher_test.go
```diff
@@ -308,7 +308,7 @@ func TestBuffer_GetWatermarkWithMultiplePartition(t *testing.T) {
 			}
 			// this will always be 27 because the timeline has been populated ahead of time
```
vigith, line 309:
```suggestion
			// this will always be 26 because the timeline has been populated ahead of time
```

do you know why it became 26 from 27?

pkg/watermark/fetch/edge_fetcher_test.go
```diff
@@ -341,7 +341,7 @@ func Test_edgeFetcher_GetHeadWatermark(t *testing.T) {
 		{
 			name:             "some pods idle and skip an idle WMB",
 			processorManager: processorManager2,
-			want:             16,
+			want:             17,
```
vigith, line 344: how did it become 17?

vigith (approved):

pkg/watermark/fetch/edge_fetcher_test.go
yhl25: because before for head watermark we were considering the max watermark across the partitions(27), now I updated it to consider only the incoming partition(26).

pkg/daemon/server/service/pipeline_metrics_query.go
veds-g: yes

pkg/daemon/server/service/pipeline_watermark_query.go
veds-g: added that

pkg/watermark/fetch/edge_fetcher_test.go
xdevxy: +1

xdevxy (approved):

pkg/daemon/server/service/pipeline_metrics_query.go
```diff
@@ -173,8 +173,15 @@ func (ps *pipelineMetadataQuery) GetVertexMetrics(ctx context.Context, req *daem
 	abstractVertex := ps.pipeline.GetVertex(req.GetVertex())
 	vertexLevelRates := ps.rater.GetRates(req.GetVertex())
 
-	metricsArr := make([]*daemon.VertexMetrics, abstractVertex.GetPartitionCount())
-	for i := int64(0); i < int64(abstractVertex.GetPartitionCount()); i++ {
+	metricsCount := 1
+	// currently metrics is an aggregation per vertex, so same across each pod for non-reduce vertex
```
vigith, line 177:
```suggestion
	// TODO(multi-partition): currently metrics is an aggregation per vertex, so same across each pod for non-reduce vertex
```

pkg/watermark/fetch/edge_fetcher_test.go
vigith: because the test is now per partition

PR #790, opened 2023-06-13
chore: unit tests for multi partitioned buffers.

yhl25:


pkg/forward/forward_test.go
```diff
@@ -1071,6 +1193,81 @@ func TestSourceInterStepDataForward(t *testing.T) {
...
+
+	stopped := f.Start()
+	count := int64(4)
+	// write some data
+	_, errs := fromStep.Write(ctx, writeMessages[0:count])
+	assert.Equal(t, make([]error, count), errs)
+
+	time.Sleep(time.Second)
+	// read some data
+	// since we have produced two messages, both the partitions should have one message each.(we write in round-robin fashion)
```
KeranYang, line 1238:
```suggestion
	// since we have produced four messages, both the partitions should have two messages each.(we write in round-robin fashion)
```

pkg/reduce/data_forward_test.go
```diff
@@ -1095,6 +1108,103 @@ func TestDataForward_WithContextClose(t *testing.T) {
...
+	// since the window duration is 5 minutes, the output should be
+	// 100 * 300(for key even) and 99 * 300(for key odd)
+	// we cant guarantee the order of the output
+	assert.Contains(t, []int{30000, 29700}, readMessagePayload0.Value)
+	assert.Contains(t, []string{"even", "odd"}, readMessagePayload0.Key)
+
+	var readMessagePayload1 PayloadForTest
+	_ = json.Unmarshal(msgs1[0].Payload, &readMessagePayload1)
+	assert.Contains(t, []int{30000, 29700}, readMessagePayload1.Value)
+	assert.Contains(t, []string{"even", "odd"}, readMessagePayload1.Key)
```
KeranYang, line 1204: We initialized both `ordered processor` and `reduce data forwarder` with `myForwardTestRoundRobin`. Should we expect the reduce vertex output to be send to toBuffers round-robinly? Why do we expect that both buffer contain same data? Is it because they reflect different windows but same window size?

pkg/watermark/fetch/edge_fetcher_test.go
```diff
@@ -1342,3 +1363,400 @@ func TestFetcherWithSameOTBucket(t *testing.T) {
...
+	assert.NoError(t, err)
+	defer hbStore.Close()
+
+	// create otStore
+	otStore, err := jetstream.NewKVJetStreamKVStore(ctx, "testFetch", keyspace+"_OT", defaultJetStreamClient)
+	assert.NoError(t, err)
+	defer otStore.Close()
+
+	// put values into otStore
+
```
KeranYang, line 1434:
```suggestion
```

pkg/watermark/fetch/edge_fetcher_test.go
```diff
@@ -1342,3 +1363,400 @@ func TestFetcherWithSameOTBucket(t *testing.T) {
...
+	}
+
+	// publish an idle watermark: simulate map
+	otValueByteOne, err = otValueToBytes(107, epoch+700, true, 0)
+	assert.NoError(t, err)
+	otValueByteTwo, err = otValueToBytes(107, epoch+700, true, 1)
+	assert.NoError(t, err)
+	otValueByteThree, err = otValueToBytes(107, epoch+700, true, 2)
+	assert.NoError(t, err)
+
```
KeranYang, line 1736:
```suggestion
```

pkg/watermark/fetch/edge_fetcher_test.go
```diff
@@ -1342,3 +1363,400 @@ func TestFetcherWithSameOTBucket(t *testing.T) {
...
+	}
+
+	allProcessors = processorManager.GetAllProcessors()
+	assert.Equal(t, 2, len(allProcessors))
+	assert.True(t, allProcessors["p1"].IsDeleted())
+	assert.True(t, allProcessors["p2"].IsActive())
+
+	_ = fetcher.GetWatermark(isb.SimpleStringOffset(func() string { return strconv.FormatInt(testOffset, 10) }), 0)
+	_ = fetcher.GetWatermark(isb.SimpleStringOffset(func() string { return strconv.FormatInt(testOffset, 10) }), 1)
+	_ = fetcher.GetWatermark(isb.SimpleStringOffset(func() string { return strconv.FormatInt(testOffset, 10) }), 2)
```
KeranYang, line 1576: Should we verify the return value of GetWatermark? If not, what are we checking here?

pkg/watermark/fetch/edge_fetcher_test.go
```diff
@@ -1342,3 +1363,400 @@ func TestFetcherWithSameOTBucket(t *testing.T) {
...
+
+	err = otStore.PutKV(ctx, "p1", otValueByteOne)
+	assert.NoError(t, err)
+	err = otStore.PutKV(ctx, "p1", otValueByteTwo)
+	assert.NoError(t, err)
+	err = otStore.PutKV(ctx, "p1", otValueByteThree)
+	assert.NoError(t, err)
+
+	// p1 should get the head offset watermark from p2
+	for allProcessors["p1"].GetOffsetTimelines()[0].Dump() != "[IDLE 1651161660700:107] -> [IDLE 1651161660600:106] -> [1651161660500:103] -> [-1:-1] -> [-1:-1] -> [-1:-1] -> [-1:-1] -> [-1:-1] -> [-1:-1] -> [-1:-1]" &&
```
KeranYang, line 1745: Searching around our code base I notice that `GetOffsetTimelines` is used only by `GetHeadWatermark` and `GetWatermark`. It looks the `GetOffsetTimelines` is the implementation detail of `GetHeadWatermark` and `GetWatermark`, which should be hidden from unit tests. Unit tests should focus on test **what** instead of **how**. Do you think we can replace the `Dump() != "..."` check with `GetHeadWatermark() != "..."` or `GetWatermark() != "..."`?

pkg/forward/forward_test.go
```diff
@@ -259,7 +319,11 @@ func TestNewInterStepDataForward(t *testing.T) {
 			otKeys2, _ := otStores["to2"].GetAllKeys(ctx)
```
xdevxy, line 319: > // NOTE: in this test we only have one processor to publish
> // so len(otKeys) should always be 1

Does this comment still holds?

pkg/forward/forward_test.go
```diff
@@ -259,7 +319,11 @@ func TestNewInterStepDataForward(t *testing.T) {
...
 			otValue2, _ := otStores["to2"].GetValue(ctx, otKeys2[0])
 			otDecode2, _ := wmb.DecodeToWMB(otValue2)
-			assert.False(t, otDecode2.Idle)
+			if tt.batchSize > 1 {
+				assert.False(t, otDecode2.Idle)
```
xdevxy, line 323: Do you mind adding a comment on why `batchSize <= 1`  then `otDecode2.Idle` would be true?

pkg/forward/forward_test.go
```diff
@@ -1071,6 +1193,81 @@ func TestSourceInterStepDataForward(t *testing.T) {
...
+	time.Sleep(time.Second)
+	// read some data
+	// since we have produced two messages, both the partitions should have one message each.(we write in round-robin fashion)
+	readMessages, err := to11.Read(ctx, 2)
+	assert.NoError(t, err, "expected no error")
+	assert.Len(t, readMessages, 2)
+	assert.Equal(t, []interface{}{writeMessages[0].Header.Keys, writeMessages[2].Header.Keys}, []interface{}{readMessages[0].Header.Keys, readMessages[1].Header.Keys})
+	assert.Equal(t, []interface{}{writeMessages[0].Header.ID + "-0-0", writeMessages[2].Header.ID + "-0-0"}, []interface{}{readMessages[0].Header.ID, readMessages[1].Header.ID})
+	for _, m := range readMessages {
+		// verify new event time gets assigned to messages.
```
xdevxy, line 1245: L1239-L1249 and L1253-L1263 seems duplicated verification logic, may be cleaner to wrap in a common block and just pass in different `buffers`/`writeMessages`.

pkg/forward/forward_test.go
yhl25: yes..

pkg/forward/forward_test.go
yhl25: done

pkg/forward/forward_test.go
yhl25: comparing with the same set of messages but with different index

pkg/forward/forward_test.go
yhl25: done

pkg/reduce/data_forward_test.go
```diff
@@ -1139,12 +1249,12 @@ func buildPublisherMapAndOTStore(ctx context.Context, toBuffers map[string][]isb
...
 	// create publisher for to Buffers
 	index := int32(0)
-	for key := range toBuffers {
+	for key, partitionedBuffers := range toBuffers {
 		publishEntity := processor.NewProcessorEntity(key)
 		hb, hbKVEntry, _ := inmem.NewKVInMemKVStore(ctx, pipelineName, key+"_PROCESSORS")
 		ot, otKVEntry, _ := inmem.NewKVInMemKVStore(ctx, pipelineName, key+"_OT")
 		otStores[key] = ot
-		p := publish.NewPublish(ctx, publishEntity, wmstore.BuildWatermarkStore(hb, ot), 1, publish.WithAutoRefreshHeartbeatDisabled(), publish.WithPodHeartbeatRate(1))
+		p := publish.NewPublish(ctx, publishEntity, wmstore.BuildWatermarkStore(hb, ot), int32(len(partitionedBuffers)), publish.WithAutoRefreshHeartbeatDisabled(), publish.WithPodHeartbeatRate(1))
```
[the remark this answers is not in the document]
yhl25: one test should be good enough.. because if we increase the heartbeat rate it will increase the test time.

pkg/reduce/data_forward_test.go
yhl25: yes because we are producing more than one message, so we are expecting both buffers to have some data.

pkg/watermark/fetch/edge_fetcher_test.go
yhl25: we are deleting the processor using the GetWatermark method, if you look at the GetWatermark implementation in some cases we are deleting the processor. So we are testing those cases here.

pkg/watermark/fetch/edge_fetcher_test.go
yhl25: this is fine I guess.

pkg/watermark/fetch/edge_fetcher_test.go
KeranYang: Got it. L1514-1521 verified that if inputOffset is less than head offset of a processor, the processor won't get deleted.

pkg/watermark/fetch/edge_fetcher_test.go
```diff
@@ -1342,3 +1320,402 @@ func TestFetcherWithSameOTBucket(t *testing.T) {
...
+
+	_ = fetcher.GetWatermark(isb.SimpleStringOffset(func() string { return strconv.FormatInt(testOffset, 10) }), 0)
+	_ = fetcher.GetWatermark(isb.SimpleStringOffset(func() string { return strconv.FormatInt(testOffset, 10) }), 1)
+	_ = fetcher.GetWatermark(isb.SimpleStringOffset(func() string { return strconv.FormatInt(testOffset, 10) }), 2)
+
+	allProcessors = processorManager.GetAllProcessors()
+	assert.Equal(t, 2, len(allProcessors))
+	assert.True(t, allProcessors["p1"].IsDeleted())
+	assert.True(t, allProcessors["p2"].IsActive())
+	// "p1" should be deleted after this GetWatermark offset=101
```
KeranYang, line 1522:
```suggestion
	// "p1" should be deleted after this GetWatermark offset=103
```

pkg/forward/forward_test.go
```diff
@@ -1071,6 +1184,81 @@ func TestSourceInterStepDataForward(t *testing.T) {
...
+
+	stopped := f.Start()
+	count := int64(4)
+	// write some data
+	_, errs := fromStep.Write(ctx, writeMessages[0:count])
+	assert.Equal(t, make([]error, count), errs)
+
+	time.Sleep(time.Second)
+	// read some data
+	// since we have produced two messages, both the partitions should have two messages each.(we write in round-robin fashion)
```
KeranYang, line 1229:
```suggestion
	// since we have produced four messages, both the partitions should have two messages each.(we write in round-robin fashion)
```

KeranYang (commented):
Flaky TestFetcherWithSameOTBucket_InMem ?

KeranYang (approved):

pkg/watermark/fetch/edge_fetcher_test.go
KeranYang: I have strong opinions on not exposing implementation details in unit tests. I approved the PR but we should address this later.

PR #791, opened 2023-06-14
chore: update e2e tests with multi partitions

yhl25:


KeranYang (approved):

vigith (approved):
LGTM. only question I have is whether we should keep some tests the same with both single and multi-partition.

yhl25:
> LGTM. only question I have is whether we should keep some tests the same with both single and multi-partition.

I have kept some tests with just a single partition, and a few tests with both single and multi partition.

PR #792, opened 2023-06-14
chore: metrics changes for multi partition edges

yhl25:


pkg/reduce/data_forward.go
```diff
@@ -179,21 +179,21 @@ func (df *DataForward) ReplayPersistedMessages(ctx context.Context) error {
...
+	readMessages, err := df.fromBufferPartition.Read(ctx, df.opts.readBatchSize)
 	totalBytes := 0
 	if err != nil {
 		df.log.Errorw("Failed to read from isb", zap.Error(err))
 		readMessagesError.With(map[string]string{
 			metrics.LabelVertex:             df.vertexName,
 			metrics.LabelPipeline:           df.pipelineName,
 			metrics.LabelVertexReplicaIndex: strconv.Itoa(int(df.vertexReplica)),
-			"buffer":                        df.fromBuffer.GetName()}).Inc()
+			"partition_idx":                 df.fromBufferPartition.GetName()}).Inc()
```
vigith, line 190:
```suggestion
			"partition_name":                 df.fromBufferPartition.GetName()}).Inc()
```

vigith (changes requested):
`idx` means it should be an integer, you are using name.

pkg/reduce/data_forward.go
yhl25: done

vigith (approved):

pkg/forward/forward_test.go
```diff
@@ -1280,7 +1280,7 @@ func validateMetrics(t *testing.T, batchSize int64) {
 		# TYPE forwarder_read_total counter
 		`
 	expected := `
-		forwarder_read_total{buffer="from",pipeline="testPipeline",vertex="testVertex"} ` + fmt.Sprintf("%f", float64(batchSize)) + `
+		forwarder_read_total{partition_idx="from",pipeline="testPipeline",vertex="testVertex"} ` + fmt.Sprintf("%f", float64(batchSize)) + `
```
KeranYang: We might also need to update the doc metrics.md

pkg/forward/metrics.go
```diff
@@ -28,77 +28,77 @@ var readMessagesCount = promauto.NewCounterVec(prometheus.CounterOpts{
 	Subsystem: "forwarder",
 	Name:      "read_total",
 	Help:      "Total number of Messages Read",
-}, []string{metrics.LabelVertex, metrics.LabelPipeline, "buffer"})
+}, []string{metrics.LabelVertex, metrics.LabelPipeline, "partition_idx"})
```
KeranYang: I am wondering if we care about the metric per in-coming edge. With this dimension definition, we can't tell how much data is read for a specific in-coming edge.

pkg/metrics/metrics.go
```diff
@@ -168,18 +168,17 @@ func (ms *metricsServer) buildupPendingInfo(ctx context.Context) {
...
-			for _, pending := range ms.lagReaders {
-				if p, err := pending.Pending(ctx); err != nil {
+			for _, pendingLag := range ms.lagReaders {
+				if p, err := pendingLag.Pending(ctx); err != nil {
 					log.Errorw("Failed to get pending messages", zap.Error(err))
 				} else {
-					totalPending += p
+					if p != isb.PendingNotAvailable {
+						totalPending += p
+					}
```
KeranYang, line 177: Under what circumstance can we have some of the pendingLag return pending information and others return PendingNotAvailable? And if that happens, why do we choose to calculate partial data instead of skipping the current ticker?

pkg/reduce/pnf/processandforward.go
```diff
@@ -246,7 +246,7 @@ func (p *processAndForward) writeToBuffer(ctx context.Context, edgeName string,
 				metrics.LabelVertex:             p.vertexName,
 				metrics.LabelPipeline:           p.pipelineName,
 				metrics.LabelVertexReplicaIndex: strconv.Itoa(int(p.vertexReplica)),
-				"buffer":                        p.toBuffers[edgeName][partition].GetName()}).Add(float64(len(failedMessages)))
+				"partition_idx":                 p.toBuffers[edgeName][partition].GetName()}).Add(float64(len(failedMessages)))
```
KeranYang: nit: maybe move "partition_idx" to metrics.Label* ?

vigith:
look like we have a flaky unit test
```
=== RUN   TestFetcherWithSameOTBucketWithSinglePartition
{"level":"info","ts":"2023-06-14T16:10:58.171016269Z","logger":"numaflow","caller":"processor/processor_manager.go:219","msg":"Deleting","key":"p2","p2":"p2 status:active, timelines:  [1651161660000:105] -> [-1:-1] -> [-1:-1] -> [-1:-1] -> [-1:-1] -> [-1:-1] -> [-1:-1] -> [-1:-1] -> [-1:-1] -> [-1:-1]\n "}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x183344b]

goroutine 82 [running]:
github.com/numaproj/numaflow/pkg/watermark/store/inmem.(*inMemWatch).Watch.func2()
	/home/runner/work/numaflow/numaflow/pkg/watermark/store/inmem/kv_watch.go:87 +0x18b
created by github.com/numaproj/numaflow/pkg/watermark/store/inmem.(*inMemWatch).Watch
	/home/runner/work/numaflow/numaflow/pkg/watermark/store/inmem/kv_watch.go:75 +0x2e6
FAIL	github.com/numaproj/numaflow/pkg/watermark/fetch	1.724s
?   	github.com/numaproj/numaflow/pkg/watermark/generic	[no test files]
?   	github.com/numaproj/numaflow/pkg/watermark/generic/jetstream	[no test files]
=== RUN   TestEntity```

pkg/reduce/data_forward.go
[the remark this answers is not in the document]
yhl25: please review the latest changes.

pkg/forward/metrics.go
yhl25: since the edge is owned by the vertex now, it's the same as how much data is read in the vertex.

pkg/metrics/metrics.go
yhl25: since each partition is a separate stream, it's possible to have pendingNotAvailable for some of them.

yhl25: +1 to skip instead of sending partial information, will update the pr

pkg/forward/metrics.go
KeranYang: Sorry let me re-phase my comment, when we have diamond shape pipeline, we will have multiple edges owned by the same vertex. In that case, with the current metric definition, we won't be able to tell the read_total per edge. I guess for this PR, diamond shape is out of scope.

pkg/metrics/metrics.go
vigith: since it is a gauge does it really matter? there are 2 scenarios
* current pending gauge is > partial value (here sending a partial value will not help)
* current pending gauge is < partial value  (here sending a partial value will help)

isn't there a 50/50 chance for anything to happen? 😄

pkg/reduce/data_forward.go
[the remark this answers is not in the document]
KeranYang: Based on the latest change, it seems the answer to my question is YES. Could you confirm?

pkg/forward/metrics.go
yhl25: we will still be reading from the same set of partitions.. diamond DAG is out of scope for this pr.

pkg/reduce/data_forward.go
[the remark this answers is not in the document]
yhl25: vertex replica and partition_idx will have one-to-one mapping while reading while writing its one to many.

pkg/metrics/metrics.go
KeranYang: Adding some context here to refresh my memory. `pending` is a metric used by UI to display and by controller to autoscale. For both case, I think we support calculating it use a user-specified look back seconds. We calculate `pending` of last n seconds by collecting last n-second `totalPending`s, adding them up, and then dividing them by n.

Yes, it's a gauge. If we only care about the current value, it's ok to send partial data, because if partial data is lower, we just don't scale up and otherwise, it could lead to a scale-up operation. However, we are looking back for n data points to calculate average, and under that case, I think accuracy become more important. We would rather collect n/2 accurate data points, sum and divide by n/2 than collect n data points with n/2 partial data points and divide by n.

pkg/reduce/data_forward.go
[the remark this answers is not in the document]
KeranYang: By `while writing`, do you mean that one replica writing to multiple out-going partitions? I thought in this metric, the partition_idx means in-coming partitions.

pkg/metrics/metrics.go
vigith: If we are storing the data and then dividing it, we shouldn't store partial data. thank you, Keran for the explanation.

pkg/reduce/data_forward.go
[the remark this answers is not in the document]
yhl25: yes, one replica writing to multiple toBufferPartitions.  I updated `parition_idx` to `partition_name`.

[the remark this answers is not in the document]
KeranYang: Since this metric is `readMessagesError`, `LabelVertexReplicaIndex ` and `partition_name` are 1:1. It seems unnecessary to adding the extra dimension for reduce.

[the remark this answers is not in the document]
yhl25: I prefer keeping both because we will be able to filter either using the replica or the stream name.

KeranYang (approved):

PR #793, opened 2023-06-15
refactor: buffers to partitions

yhl25:
Explain what this PR does.


cmd/commands/isbsvc_create.go
```diff
@@ -79,7 +79,7 @@ func NewISBSvcCreateCommand() *cobra.Command {
 				return fmt.Errorf("unsupported isb service type %q", isbSvcType)
 			}
 
-			if err = isbsClient.CreateBuffersAndBuckets(ctx, buffers, buckets, opts...); err != nil {
+			if err = isbsClient.CreatePartitionsAndBuckets(ctx, buffers, buckets, opts...); err != nil {
```
jy4096, line 82: `isbsClient.CreatePartitionsAndBuckets(ctx, partitions, buckets, opts...)` ??

cmd/commands/isbsvc_delete.go
```diff
@@ -62,7 +62,7 @@ func NewISBSvcDeleteCommand() *cobra.Command {
 				cmd.HelpFunc()(cmd, args)
 				return fmt.Errorf("unsupported isb service type %q", isbSvcType)
 			}
-			if err = isbsClient.DeleteBuffersAndBuckets(ctx, buffers, buckets); err != nil {
+			if err = isbsClient.DeletePartitionsAndBuckets(ctx, buffers, buckets); err != nil {
```
jy4096, line 65: isbsClient.DeletePartitionsAndBuckets(ctx, buffers, buckets)

same as above... is it still `buffers` or `partitions` ?

cmd/commands/isbsvc_create.go
```diff
@@ -79,7 +79,7 @@ func NewISBSvcCreateCommand() *cobra.Command {
...
 				logger.Errorw("Failed to create buffers and buckets.", zap.Error(err))
```
jy4096, line 83: `logger.Errorw("Failed to create partitons and buckets.", zap.Error(err))` also for other logs
e.g.
`logger.Errorw("Failed on buffers and buckets deletion.", zap.Error(err))` -> partitions

juliev0:
Can we change [this](https://github.com/numaproj/numaflow/blob/main/docs/specifications/edges-buffers-buckets.md) document (per [this](https://intuit-teams.slack.com/archives/C02P1NKMBHQ/p1685643897058979) slack thread)?

api/json-schema/schema.json
```diff
@@ -17382,7 +17382,7 @@
           "type": "object"
         },
         "partitions": {
-          "description": "Number of partitions of the vertex owned buffers. It applies to udf and sink vertices only.",
+          "description": "Number of partitions of the vertex owned partitions. It applies to udf and sink vertices only.",
```
juliev0, line 17385: this one can probably remain "buffers"

docs/APIs.md
```diff
@@ -394,8 +394,8 @@ List of sidecar containers belonging to the pod.
 <td>
 <em>(Optional)</em>
 <p>
-Number of partitions of the vertex owned buffers. It applies to udf and
-sink vertices only.
+Number of partitions of the vertex owned partitions. It applies to udf
+and sink vertices only.
```
juliev0, line 398: keep the same

pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -332,7 +332,7 @@ func (p Pipeline) getDaemonPodInitContainer(req GetDaemonDeploymentReq) corev1.C
 		Resources:       standardResources,
 		Args:            []string{"isbsvc-validate", "--isbsvc-type=" + string(req.ISBSvcType)},
 	}
-	c.Args = append(c.Args, "--buffers="+strings.Join(p.GetAllBuffers(), ","))
+	c.Args = append(c.Args, "--buffers="+strings.Join(p.GetAllPartitions(), ","))
```
juliev0, line 335: you don't think we should also change the name of the flag?

pkg/daemon/server/service/pipeline_metrics_query.go
```diff
@@ -91,13 +91,13 @@ func (ps *pipelineMetadataQuery) ListBuffers(ctx context.Context, req *daemon.Li
 	resp := new(daemon.ListBuffersResponse)
 
 	buffers := []*daemon.BufferInfo{}
-	for _, buffer := range ps.pipeline.GetAllBuffers() {
-		bufferInfo, err := ps.isbSvcClient.GetBufferInfo(ctx, buffer)
+	for _, buffer := range ps.pipeline.GetAllPartitions() {
```
juliev0, line 94: I think we should change "buffers" and "buffer" as well

juliev0: unless you prefer to do this with multiple PRs - I feel like when this effort is done there should probably only be a few uses of the word "buffer" left in the code altogether - do you agree or no?

pkg/daemon/server/service/pipeline_metrics_query.go
```diff
@@ -125,11 +125,11 @@ func (ps *pipelineMetadataQuery) ListBuffers(ctx context.Context, req *daemon.Li
 
 // GetBuffer is used to obtain one buffer information of a pipeline
 func (ps *pipelineMetadataQuery) GetBuffer(ctx context.Context, req *daemon.GetBufferRequest) (*daemon.GetBufferResponse, error) {
-	bufferInfo, err := ps.isbSvcClient.GetBufferInfo(ctx, *req.Buffer)
+	bufferInfo, err := ps.isbSvcClient.GetPartitionInfo(ctx, *req.Buffer)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get information of buffer %q:%v", *req.Buffer, err)
 	}
-	v := ps.pipeline.FindVertexWithBuffer(*req.Buffer)
+	v := ps.pipeline.FindVertexWithPartition(*req.Buffer)
```
juliev0, line 132: more instances of the word "buffer"

pkg/daemon/server/service/pipeline_metrics_query.go
```diff
@@ -288,14 +288,14 @@ func (ps *pipelineMetadataQuery) GetPipelineStatus(ctx context.Context, req *dae
...
-		if x.BufferMaxLength != nil {
-			bufferLength = int64(*x.BufferMaxLength)
+		if x.PartitionMaxLength != nil {
+			bufferLength = int64(*x.PartitionMaxLength)
 		}
-		if x.BufferUsageLimit != nil {
-			bufferUsageLimit = float64(*x.BufferUsageLimit) / 100
+		if x.PartitionUsageLimit != nil {
+			bufferUsageLimit = float64(*x.PartitionUsageLimit) / 100
 		}
```
juliev0, line 299: more instances of the word "buffer"

api/json-schema/schema.json
xdevxy: +1

api/json-schema/schema.json
```diff
@@ -18767,13 +18767,13 @@
     },
     "io.numaproj.numaflow.v1alpha1.PipelineLimits": {
       "properties": {
-        "bufferMaxLength": {
-          "description": "BufferMaxLength is used to define the max length of a buffer Only applies to UDF and Source vertices as only they do buffer write. It can be overridden by the settings in vertex limits.",
+        "partitionMaxLength": {
```
xdevxy, line 18770: Will `isbPartitionMaxLength` be more clear that this is a ISB partition? Same for other properties below.

api/json-schema/schema.json
```diff
@@ -18767,13 +18767,13 @@
...
+          "description": "PartitionMaxLength is used to define the max length of a partition Only applies to UDF and Source vertices as only they do partition write. It can be overridden by the settings in vertex limits.",
```
xdevxy, line 18771: ISB partition.

api/json-schema/schema.json
```diff
@@ -19074,13 +19074,13 @@
           "format": "int64",
           "type": "integer"
         },
-        "targetBufferAvailability": {
-          "description": "TargetBufferAvailability is used to define the target percentage of the buffer availability. A valid and meaningful value should be less than the BufferUsageLimit defined in the Edge spec (or Pipeline spec), for example, 50. It only applies to UDF and Sink vertices because only they have buffers to read.",
+        "targetBufferUsage": {
+          "description": "TargetBufferUsage is used to define the target percentage of the buffer availability. A valid and meaningful value should be less than the PartitionUsageLimit defined in the Edge spec (or Pipeline spec), for example, 50. It only applies to UDF and Sink vertices because only they have buffers to read. Deprecated: use targetBufferAvailability instead. Will be removed in v0.9",
```
xdevxy, line 19078: Do you mean `targetBufferUsage` will be deprecated and use `targetPartitionAvailability`? Or `targetBufferAvailability`. If the later, I don't see it defined after the change?

vigith:
We are blindly assuming that we can replace buffer with partition, which is not right. Eg, we cannot set per-partition limits, but we can set a limit across all the partitions. Hence i think we should have a concept of a buffer where a buffer consists of partitions. A good analogy will be Kafka, where we have a topic, partitions of a topic, and brokers similar to ISBSvc which manages the topic partitions.

PR #795, opened 2023-06-15
refactor: remove redundant delta calculations for rater

KeranYang:
### Description
When we calculate processing rate, we calculate deltas between all pairs of timestamped counts within lookback seconds. If two consecutive timestamped counts all finished collecting metrics, then the delta calculation should only execute once. This PR introduce the concept of `window`. When a timestamp counts finish collecting counts, we mark the it as window-closed and calculate the delta and save the delta. Such that future GetRate calls won't need to re-calculate the delta and instead directly use the delta attribute of that closed window. This brings down the time complexity of calculating vertex level rate from O(lookback seconds * no. of pods) to O(lookback seconds)

### Testing

Tested with 4k TPS and 100-pod vertices

Screenshot 2023-06-15 at 3 56 04 PM


xdevxy (approved):

KeranYang:
> Does it mean we must iterate through all the deltas during metric fetching? IIRC, fetching just updated the queue without doing any other computation.

@vigith I assume `metric fetching` means scrape count metric from a single pod. Yes, before this change, fetching will update **one of** the items in the queue without doing any other computation. After this change, in addition to the update, if fetching falls into a new window, it also closes the previous window by calculating the delta for that window.

We don't iterate through all the deltas during the metric fetching, just the most recent one.

vigith:
Has the runtime complexity moved from the `Get` call to the `Fetch` call?

vigith (approved):

PR #797, opened 2023-06-15
docs: map stream udf feature

### LINKED ISSUE #770: docs: map stream udf feature
# Summary

What change needs making?
Document the map stream udf feafure.

# Use Cases

When would you use this?

---

xdevxy:
Closes #770

### DIFF
--- a/docs/user-guide/user-defined-functions/map/map.md
+++ b/docs/user-guide/user-defined-functions/map/map.md
@@ -47,6 +47,30 @@ spec:
           image: my-python-udf-example:latest
 ```
 
+### Streaming Mode
+In cases the map function generates more than one outputs (e.g. flat map), the UDF can be
+configured to run in a streaming mode instead of batching which is the default mode.
+In streaming mode, the messages will be pushed to the downstream vertices once generated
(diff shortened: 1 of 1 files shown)


docs/user-guide/user-defined-functions/map/map.md
```diff
@@ -47,6 +47,29 @@ spec:
           image: my-python-udf-example:latest
 ```
 
+### Streaming Mode
+In cases the Map generates more than one outputs (e.g. flat map), the Map UDF can be configured to run in a
+streaming mode instead of batching which is the default mode. In streaming mode, the messages
+will be pushed to the downstream vertices once generated instead of at the end in the map function.
+The streaming mode can be enabled by setting the annotation `numaflow.numaproj.io/map-stream` to `true`
+in the vertex spec. Note that to maintain data orderliness, we restrict the read batch size to be `1`
```
vigith, line 55:
```suggestion
in the vertex spec. 

Note that to maintain data orderliness, we restrict the read batch size to be `1`.
```

vigith (approved):

PR #804, opened 2023-06-22
fix: segmentation fault in daemon server

yhl25:


vigith (approved):
why do we have to do this?

yhl25:
> why do we have to do this?

https://github.com/numaproj/numaflow/blob/939d40a13778c4ab84df464c5d4f5db20602ca0e/pkg/daemon/server/service/rater/helper.go#L29

https://github.com/numaproj/numaflow/blob/939d40a13778c4ab84df464c5d4f5db20602ca0e/pkg/daemon/server/service/rater/helper.go#L44

Instead of using the length of the local object we were using the length of the shared one, which is wrong. Because the shared could be updated by other go routines.

PR #806, opened 2023-06-23
feat: autoscaling changes to support multi partition

yhl25:
- cache daemon clients to avoid creating new clients each time. (LRU based)
- supports autoscaling when there are multiple partitions by considering processing rates and pending information of all the partitions


pkg/reconciler/vertex/scaling/client_cache.go
```diff
@@ -0,0 +1,65 @@
...
+
+	daemonclient "github.com/numaproj/numaflow/pkg/daemon/client"
+)
+
+type entry struct {
+	address string
+	client  *daemonclient.DaemonClient
+}
+
+// ClientCache is a cache for daemon clients.
```
vigith, line 15: please comment what kind of cache you have implemented. also.. can we make it generic?

pkg/reconciler/vertex/scaling/scaling.go
```diff
@@ -61,6 +62,8 @@ func NewScaler(client client.Client, opts ...Option) *Scaler {
 	}
 	vertexMetricsCache, _ := lru.New(10000)
 	s.vertexMetricsCache = vertexMetricsCache
+	// cache top 100 daemon clients
+	s.daemonClientsCache = NewClientCache(100)
```
vigith, line 66: make 100 optional

pkg/reconciler/vertex/scaling/client_cache.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	return nil
+}
+
+// DeleteAll function closes all the daemon clients in the cache and removes them from the cache.
+func (c *ClientCache) DeleteAll() {
+	c.Lock()
+	defer c.Unlock()
+	for _, element := range c.clientsMap {
+		_ = element.Value.(*entry).client.Close()
```
vigith, line 61: this cannot be generic as is.. so, pass a func that can clean it up.

pkg/reconciler/vertex/scaling/client_cache.go
```diff
@@ -0,0 +1,65 @@
...
+	daemonclient "github.com/numaproj/numaflow/pkg/daemon/client"
+)
+
+type entry struct {
+	address string
+	client  *daemonclient.DaemonClient
+}
+
+// ClientCache is a cache for daemon clients.
+type ClientCache struct {
```
whynowy, line 16: Why don't use a simple LRU cache?

vigith: I think it is because he has that DeleteAll func that closes the client.

whynowy: There's a `NewWithEvict` function which can give an eviction callback, the callback is also called in `Purge()`.

```golang
// NewWithEvict constructs a fixed size cache with the given eviction
// callback.
func NewWithEvict(size int, onEvicted func(key interface{}, value interface{})) (*Cache, error) {
	lru, err := simplelru.NewLRU(size, simplelru.EvictCallback(onEvicted))
	if err != nil {
		return nil, err
	}
	c := &Cache{
		lru: lru,
	}
	return c, nil
}
```

yhl25: didn't know about this.. updated it to use simple LRU.

vigith (approved):

pkg/reconciler/vertex/scaling/scaling.go
```diff
@@ -202,50 +208,66 @@ func (s *Scaler) scaleOneVertex(ctx context.Context, key string, worker int) err
...
-	if !existing || rate < 0 { // Rate not available
-		log.Debugf("Vertex %s has no rate information, skip scaling", vertex.Name)
-		return nil
-	}
-	pending, existing := vMetrics[0].Pendings["default"]
-	if !existing || pending < 0 || pending == isb.PendingNotAvailable {
-		// Pending not available, we don't do anything
-		log.Debugf("Vertex %s has no pending messages information, skip scaling", vertex.Name)
-		return nil
+	// vmMetrics is a map which contains metrics of all the partitions of a vertex.
```
whynowy, line 227: Minor: `vMetrics`.

whynowy (approved):

PR #808, opened 2023-06-27
chore: fix dropped messages count metric

yhl25:


pkg/reduce/data_forward.go
```diff
@@ -409,7 +409,7 @@ messagesLoop:
 		// This could be due to a couple of problem, eg. ack was not registered, etc.
 		// Please do not confuse this with late data! This is a platform related problem causing the watermark inequality
 		// to be violated.
-		if message.EventTime.Before(message.Watermark.Add(-1 * df.opts.allowedLateness)) {
+		if !message.IsLate && message.EventTime.Before(message.Watermark.Add(-1*df.opts.allowedLateness)) {
```
vigith, line 412: this is wrong

PR #810, opened 2023-06-28
chore: optimise rate calculation

### LINKED ISSUE #788: Processing Rate calculation optimization
Currently to calculate the processing rate of a vertex, we sum the count deltas between each of the timestamp interval pairs, which is an O(length of look back period) operation. If we assume no pod crash, then we can view the count-to-time relationship as a slop and quickly estimate the processing rate by only using the first and last data points. Rate = delta(count)/delta(t)

Considering that in PROD pod crash is a rare case, we can further assume within a single lookback period, the number of pod crash is constant. In that case, if we keep track of under which timestamp a pod crash was detected, then we just need 3 data points to calculate rates, first, last and the one when the pod crashed. This can be an O(3) operation.

This task is to investigate and experiment on this approach and see if it works in practice.

---

yhl25:
Closes #788 

- Optimized rate calculation as a slope ~ **delta(count)/delta(t)**, instead of summing the count deltas between each of the timestamp interval pairs
- Optimized the search for startIndex of TimestampedCounts from linear to binary
- Rate calculation remains linear in case of pod restart, which is a rare case


pkg/daemon/server/service/rater/helper.go
```diff
@@ -74,41 +74,76 @@ func CalculateRate(q *sharedqueue.OverflowQueue[*TimestampedCounts], lookbackSec
...
-			delta += calculatePartitionDelta(counts[i+1], partitionName)
+		if counts[i] != nil && counts[i+1] != nil && counts[i].IsWindowClosed() && counts[i+1].IsWindowClosed() {
+			delta += calculatePartitionDelta(counts[i], counts[i+1], partitionName)
 		}
 	}
 	return delta / float64(timeDiff)
 }
 
+func getDiffBetweenTimestampedCounts(t1, t2 *TimestampedCounts, partitionName string) float64 {
+	prevPodReadCount := t1.PodReadCountSnapshot()
```
yhl25, line 94: getDeltaBetweenTimestampedCounts ?

pkg/daemon/server/service/rater/helper.go
```diff
@@ -74,41 +74,76 @@ func CalculateRate(q *sharedqueue.OverflowQueue[*TimestampedCounts], lookbackSec
...
 		return IndexNotFound
 	}
 
 	startIndex := n - 2
-	for i := n - 2; i >= 0; i-- {
-		if now-counts[i].timestamp <= lookbackSeconds && counts[i].IsWindowClosed() {
-			startIndex = i
+	b := 0
+	e := n - 2
+	lastTimestamp := now - lookbackSeconds
```
yhl25, line 137: left and right instead of b and e?

whynowy:
Can anybody explain to me why we introduced the concept of `WidowClosed`?

vigith:
> Can anybody explain to me why we introduced the concept of `WidowClosed`?

https://github.com/numaproj/numaflow/pull/795. Ideally, this should be removed by #788

whynowy:
> > Can anybody explain to me why we introduced the concept of `WidowClosed`?
> 
> #795. Ideally, this should be removed by #788

I was expecting the delta calculation happens during `GetRate()` which is memory operation only, would not cause any performance issue, there's no need to introduce the complexity to calculate it during metrics fetching.

vigith:
> 
> I was expecting the delta calculation happens during `GetRate()` which is memory operation only, would not cause any performance issue, there's no need to introduce the complexity to calculate it during metrics fetching.

Won't GetRate be called frequently for auto-scaling?

whynowy:
> 

3 times in 30 seconds.

whynowy:
> > I was expecting the delta calculation happens during `GetRate()` which is memory operation only, would not cause any performance issue, there's no need to introduce the complexity to calculate it during metrics fetching.
> 
> Won't GetRate be called frequently for auto-scaling?

3 times in 30 seconds.

veds-g:
> > > Can anybody explain to me why we introduced the concept of `WidowClosed`?
> > 
> > 
> > #795. Ideally, this should be removed by #788
> 
> I was expecting the delta calculation happens during `GetRate()` which is memory operation only, would not cause any performance issue, there's no need to introduce the complexity to calculate it during metrics fetching.

Delta calculation only happens once GetRates() is called, prior to that just the counts are fetched and updated in the queue.

pkg/daemon/server/service/rater/timestamped_counts.go
```diff
@@ -85,57 +82,19 @@ func (tc *TimestampedCounts) PodReadCountSnapshot() map[string]map[string]float6
...
-			if _, ok := podPartitionDelta[podName]; !ok {
-				podPartitionDelta[podName] = make(map[string]float64)
-			}
-			podPartitionDelta[podName][partitionName] = delta
-		}
-	}
-
-	// finalize the window by setting isWindowClosed to true and delta to the calculated value
+// CloseWindow closes the window
+func (tc *TimestampedCounts) CloseWindow() {
```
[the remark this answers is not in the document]
veds-g: we can remove this since the window is not closed only for the latest timestamp (last element of the queue) which we skip for rate calculation for the same reason.

pkg/daemon/server/service/rater/helper.go
```diff
@@ -25,31 +25,31 @@ import (
...
 	case 0:
 	// if the queue is empty, we just append the new timestamped count
 	case 1:
 		// if the queue has only one element, we close the window for this element
-		items[0].CloseWindow(nil)
+		items[0].CloseWindow()
 	default:
 		// if the queue has more than one element, we close the window for the most recent element
-		items[n-1].CloseWindow(items[n-2])
+		items[n-1].CloseWindow()
```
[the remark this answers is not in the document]
yhl25: removed

vigith (approved):

pkg/daemon/server/service/rater/helper.go
```diff
@@ -74,52 +64,75 @@ func CalculateRate(q *sharedqueue.OverflowQueue[*TimestampedCounts], lookbackSec
 		// this should not happen in practice because we are using a 10s interval
 		return 0
 	}
-	// TODO: revisit this logic, we can just use the slope (counts[endIndex] - counts[startIndex] / timeDiff) to calculate the rate.
+	rate := getDeltaBetweenTimestampedCounts(counts[startIndex], counts[endIndex], partitionName) / float64(timeDiff)
+
+	// positive slope, meaning there was no restart in the last lookback seconds
```
KeranYang, line 69: > positive slope, meaning there was no restart in the last lookback seconds

@yhl25 @veds-g , does this statement hold true when a pod restarted and then processed more messages than before restart?

Consider a single pod single partition scenario and let's assume the pod process 5 messages per second.
t1 -> 5
t2-> 10
t3 -> 0
t4 -> 5
t5 -> 10
t6 -> 15

in this case, `getDeltaBetweenTimestampedCounts(t1, t6)` returns a positive delta but that doesn't mean no restart, right?

KeranYang: https://github.com/numaproj/numaflow/issues/846

whynowy: My opinion is, still use the approach of calculating each delta, again, it's just memory operation, won't have performance issue.

KeranYang: @whynowy +1. I will use a separate PR to address.

PR #812, opened 2023-06-28
chore: some code refactor and docs change

whynowy:
Explain what this PR does.


pkg/udf/function/uds_grpc.go
```diff
@@ -19,24 +19,24 @@ package function
...
-	functionpb "github.com/numaproj/numaflow-go/pkg/apis/proto/function/v1"
-	functionsdk "github.com/numaproj/numaflow-go/pkg/function"
-	"github.com/numaproj/numaflow/pkg/udferr"
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/protobuf/types/known/emptypb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 	"k8s.io/apimachinery/pkg/util/wait"
 
+	functionpb "github.com/numaproj/numaflow-go/pkg/apis/proto/function/v1"
```
vigith, line 32: this doesn't look right

pkg/sources/transformer/grpc_transformer.go
```diff
@@ -19,18 +19,18 @@ package transformer
...
-	"github.com/numaproj/numaflow/pkg/sdkclient/udf/client"
 	"time"
 
+	"google.golang.org/protobuf/types/known/emptypb"
+	"google.golang.org/protobuf/types/known/timestamppb"
+
 	functionpb "github.com/numaproj/numaflow-go/pkg/apis/proto/function/v1"
 	"github.com/numaproj/numaflow/pkg/forward/applier"
 	"github.com/numaproj/numaflow/pkg/isb"
+	udferr "github.com/numaproj/numaflow/pkg/sdkclient/error"
```
vigith, line 30: sdkerr?

pkg/udf/function/uds_grpc.go
whynowy: means?

vigith: `github.com/numaproj/numaflow-go` and `github.com/numaproj/numaflow` shouldn't be together since `numaflow-go` is outside of `numaflow`.

vigith (approved):
NIT: on package grouping

pkg/udf/function/uds_grpc.go
whynowy: Does your IDE has the feature to auto group? And what's the behavior? I thought packages from the same github org could stay together.

PR #828, opened 2023-07-05
chore: consider all the partitions separately while autoscaling

yhl25:
when there is skewness between the partitions, pods won't autoscale which results in a deadlock state (pods not scaling up and the previous vertex not writing because one of the partitions is full). To fix this we need to consider the processing rate and pending information of each partition separately and calculate the desired replicas and take the max desired replicas across all the partitions.

### DIFF
--- a/pkg/apis/numaflow/v1alpha1/const.go
+++ b/pkg/apis/numaflow/v1alpha1/const.go
@@ -127,9 +127,9 @@ const (
 	DefaultReadBatchSize    = 500
 
 	// Auto scaling
-	DefaultLookbackSeconds          = 180 // Default lookback seconds for calculating avg rate and pending
+	DefaultLookbackSeconds          = 120 // Default lookback seconds for calculating avg rate and pending
 	DefaultCooldownSeconds          = 90  // Default cooldown seconds after a scaling operation
-	DefaultZeroReplicaSleepSeconds  = 180 // Default sleep time in seconds after scaling down to 0, before peeking
+	DefaultZeroReplicaSleepSeconds  = 120 // Default sleep time in seconds after scaling down to 0, before peeking
 	DefaultMaxReplicas              = 50  // Default max replicas
 	DefaultTargetProcessingSeconds  = 20  // Default targeted time in seconds to finish processing all the pending messages for a source
 	DefaultTargetBufferAvailability = 50  // Default targeted percentage of buffer availability
--- a/pkg/reconciler/vertex/scaling/scaling.go
+++ b/pkg/reconciler/vertex/scaling/scaling.go
@@ -228,6 +228,9 @@ func (s *Scaler) scaleOneVertex(ctx context.Context, key string, worker int) err
 	// vMetrics is a map which contains metrics of all the partitions of a vertex.
 	// We need to aggregate them to get the total rate and pending of the vertex.
 	// If any of the partition doesn't have the rate or pending information, we skip scaling.
+	// we need both aggregated and partition level metrics for scaling, because we use aggregated metrics to
+	// determine whether we can scale down to 0 and for calculating back pressure, and partition level metrics to determine
+	// the max desired replicas among all the partitions.
 	partitionRates := make([]float64, 0)
 	partitionPending := make([]int64, 0)
 	totalRate := float64(0)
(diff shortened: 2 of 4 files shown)


pkg/apis/numaflow/v1alpha1/const.go
```diff
@@ -127,9 +127,9 @@ const (
 	DefaultReadBatchSize    = 500
 
 	// Auto scaling
-	DefaultLookbackSeconds          = 180 // Default lookback seconds for calculating avg rate and pending
```
xdevxy, line 130: Why the value changes? How do we evaluate that?

pkg/reconciler/vertex/scaling/scaling.go
```diff
@@ -314,39 +332,46 @@ func (s *Scaler) scaleOneVertex(ctx context.Context, key string, worker int) err
...
+			// then we figure out how many replicas are needed to keep the available buffer length at target level.
+			if pending >= partitionBufferLengths[i] {
+				// Simply return current replica number + max allowed if the pending messages are more than available buffer length
+				desired = int32(vertex.Status.Replicas) + int32(vertex.Spec.Scale.GetReplicasPerScale())
+			} else {
+				singleReplicaContribution := float64(partitionBufferLengths[i]-pending) / float64(vertex.Status.Replicas)
+				desired = int32(math.Round(float64(partitionAvailableBufferLengths[i]) / singleReplicaContribution))
+			}
+		}
+		if desired == 0 {
```
xdevxy, line 363: Can you add a comment on why we need to set to `1` when `desired == 0 `?

pkg/apis/numaflow/v1alpha1/const.go
```diff
@@ -127,9 +127,9 @@ const (
...
+	DefaultLookbackSeconds          = 120 // Default lookback seconds for calculating avg rate and pending
 	DefaultCooldownSeconds          = 90  // Default cooldown seconds after a scaling operation
-	DefaultZeroReplicaSleepSeconds  = 180 // Default sleep time in seconds after scaling down to 0, before peeking
+	DefaultZeroReplicaSleepSeconds  = 120 // Default sleep time in seconds after scaling down to 0, before peeking
```
whynowy, line 132: If default values change, don't forget to update the doc.

https://numaflow.numaproj.io/user-guide/reference/autoscaling/

pkg/reconciler/vertex/scaling/scaling.go
```diff
@@ -130,6 +130,7 @@ func (s *Scaler) scale(ctx context.Context, id int, keyCh <-chan string) {
 // scaleOneVertex implements the detailed logic of scaling up/down a vertex.
 //
 // For source vertices which have both rate and pending message information,
+// if there are multiple partitions, we consider the max desired replicas among all partitions.
```
whynowy, line 133: Do we need to update the logic in back pressure check?

yhl25: https://github.com/numaproj/numaflow/blob/49a61569e637c2350152a693fd78a5a5a8b69fc7/pkg/reconciler/vertex/scaling/scaling.go#L255

pkg/reconciler/vertex/scaling/scaling.go
```diff
@@ -227,14 +228,18 @@ func (s *Scaler) scaleOneVertex(ctx context.Context, key string, worker int) err
 	// vMetrics is a map which contains metrics of all the partitions of a vertex.
 	// We need to aggregate them to get the total rate and pending of the vertex.
 	// If any of the partition doesn't have the rate or pending information, we skip scaling.
+	partitionRates := make([]float64, 0)
```
vigith, line 231: please also add a comment on why we are also keeping individual partition data points along with the aggregates.

vigith (approved):

vigith:
i have approved, please fix the other comments before you merge it in.

pkg/apis/numaflow/v1alpha1/const.go
yhl25: done

pkg/reconciler/vertex/scaling/scaling.go
```diff
@@ -314,39 +332,46 @@ func (s *Scaler) scaleOneVertex(ctx context.Context, key string, worker int) err
...
+	// We calculate the max desired replicas based on the pending messages and processing rate for each partition.
+	for i := 0; i < len(partitionPending); i++ {
+		rate := partitionProcessingRate[i]
+		pending := partitionPending[i]
+		var desired int32
+		if pending == 0 || rate == 0 {
+			// Pending is 0 and rate is not 0, or rate is 0 and pending is not 0, we don't do anything.
+			// Technically this would not happen because the pending includes ackpending, which means rate and pending are either both 0, or both > 0.
+			// But we still keep this check here for safety.
+			continue
```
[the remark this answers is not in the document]
yhl25: we cannot determine the desired in this case..

pkg/reconciler/vertex/scaling/scaling.go
yhl25: added

pkg/apis/numaflow/v1alpha1/const.go
yhl25: for higher throughput pipelines I am seeing better autoscaling with 120, if we use 180 it will take 3 mins to reflect which will cause high back pressure in the pipeline, anyway it's a configuration change users can update it.

PR #832, opened 2023-07-06
feat: implement optional validation webhook. Fixes #817.

### LINKED ISSUE #817: Introduce an optional validation webhook
# Summary

An optional validation webhook is needed to prevent changes such as:

1. Most of the ISBSvc spec changes;
2. Update a pipeline with a different `interStepBufferServiceName`;
3. Update a pipeline topology;
4. xxx

---

dpadhiar:
Fixes: #817

This PR introduces an optional validation webhook for Numaflow. Currently, this prevents users from modifying the spec of an InterStepBufferService from Redis to Jetstream or vice versa. Previous behavior would allow a user to apply this change and have both running at the same time. 

```
Error from server (BadRequest): error when applying patch:
{"metadata":{"annotations":{"kubectl.kubernetes.io/last-applied-configuration":"{\"apiVersion\":\"numaflow.numaproj.io/v1alpha1\",\"kind\":\"InterStepBufferService\",\"metadata\":{\"annotations\":{},\"name\":\"default\",\"namespace\":\"numaflow-system\"},\"spec\":{\"redis\":{\"native\":{\"version\":\"7.0.11\"}}}}\n"}},"spec":{"jetstream":null,"redis":{"native":{"version":"7.0.11"}}}}
to:
Resource: "numaflow.numaproj.io/v1alpha1, Resource=interstepbufferservices", GroupVersionKind: "numaflow.numaproj.io/v1alpha1, Kind=InterStepBufferService"
Name: "default", Namespace: "numaflow-system"
for: "redis.yaml": error when patching "redis.yaml": admission webhook "webhook.numaflow.numaproj.io" denied the request: Can not change ISB Service type from Jetstream to Redis
```


config/install-validating-webhook.yaml
```diff
@@ -0,0 +1,120 @@
+apiVersion: v1
```
whynowy, line 1: Use `validating-webhook-install.yaml` as the file name?

whynowy:
Overall it looks good.

Could you add Pipeline validation, as well as Pipeline updating validating logic? We could start with disallowing `interStepBufferServiceName` change.

dpadhiar:
> Overall it looks good.
> 
> Could you add Pipeline validation, as well as Pipeline updating validating logic? We could start with disallowing `interStepBufferServiceName` change.

I have this ready to add today, was just adding test cases yesterday and regenerating manifests. For preserving the topology of the pipeline, I also have a check that the edges of the old and new pipeline are the same. Is this the correct way to do this?

whynowy:
> > Overall it looks good.
> > Could you add Pipeline validation, as well as Pipeline updating validating logic? We could start with disallowing `interStepBufferServiceName` change.
> 
> I have this ready to add today, was just adding test cases yesterday and regenerating manifests. For preserving the topology of the pipeline, I also have a check that the edges of the old and new pipeline are the same. Is this the correct way to do this?

Let's skip the edge change validation for now, it's ok for some edge change use cases.

webhook/validator/validator.go
```diff
@@ -0,0 +1,82 @@
...
+
+type Validator interface {
+	ValidateCreate(context.Context) *admissionv1.AdmissionResponse
+	ValidateUpdate(context.Context) *admissionv1.AdmissionResponse
+}
+
+// GetValidator returns a Validator instance
+func GetValidator(ctx context.Context, client kubernetes.Interface, ISBSVCClient v1alpha1.InterStepBufferServiceInterface, PipelineClient v1alpha1.PipelineInterface, kind metav1.GroupVersionKind, oldBytes []byte, newBytes []byte) (Validator, error) {
+	log := logging.FromContext(ctx)
+	switch kind.Kind {
```
whynowy, line 26: Can we use `switch kind {` and use the constants in https://github.com/numaproj/numaflow/blob/main/pkg/apis/numaflow/v1alpha1/register.go for `case`?

dpadhiar: There is a type mismatch here with conflicting packages `metav1` and `schema`:

```
invalid case dfv1.ISBGroupVersionKind in switch on kind (mismatched types schema.GroupVersionKind and "k8s.io/apimachinery/pkg/apis/meta/v1".GroupVersionKind)
```

dpadhiar: If we still use `kind.Kind` for switch, we can check each `case` like this by comparing just the string `Kind`:
```
	switch kind.Kind {
	case dfv1.ISBGroupVersionKind.Kind:
```

whynowy: > 

This is good.

whynowy:
@dpadhiar - one last thing, could you add the webhook deployment to our local dev env (`make start`)? This can be done by updating `./test/manifests/kustomization.yaml`

test/manifests/kustomization.yaml
```diff
@@ -3,7 +3,7 @@ kind: Kustomization
 
 resources:
   - ../../config/cluster-install
-
+  - ../../config/extensions/webhook
```
whynowy, line 6: Is only adding this gonna make it work?

dpadhiar: Need to include this as well, adding in next commit:
```
  - patch: |-
      - op: add
        path: /spec/template/spec/containers/0/imagePullPolicy
        value: IfNotPresent
    target:
      kind: Deployment
      name: numaflow-webhook
```

config/extensions/webhook/numaflow-webhook-deployment.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: numaflow-webhook
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: numaflow-webhook
```
whynowy, line 9: Update the labels to:

```
      app.kubernetes.io/part-of: numaflow
      app.kubernetes.io/component: numaflow-webhook
```

whynowy: Don't forget also update the selector in the Service.

whynowy (approved):
LGTM

PR #838, opened 2023-07-09
fix: pod tracker logic for calculating processing rate

yhl25:
What did we fix?
    Bug in pod tracking logic, assume we have a vertex with 10 active pods and it scales down to 5, but the active pod list will still have 9 pods (only the pod with index 5 will be removed from the list which is wrong we should remove all the pods from index 5 to 9).

How did we fix?
   We delete all the pods one by one if it's inactive. We never skip deletion.

### DIFF
--- a/pkg/daemon/server/service/rater/pod_tracker.go
+++ b/pkg/daemon/server/service/rater/pod_tracker.go
@@ -99,12 +99,8 @@ func (pt *PodTracker) Start(ctx context.Context) error {
 						if pt.isActive(v.Name, podName) {
 							pt.activePods.PushBack(podKey)
 						} else {
-							// if a pod is not active, we can assume all the following pods are not active as well.
+							// if the pod is not active, remove it from the active pod list
 							pt.activePods.Remove(podKey)
-							// we assume all the pods are ordered with continuous indices, hence as we keep increasing the index, if we don't find one, we can stop looking.
-							// the assumption holds because when we scale down, we always scale down from the last pod.
-							// there can be a case when a pod in the middle crashes, causing us missing counting the following pods.
-							// such case is rare and if it happens, it can lead to lower rate then the real one. It is acceptable because it will recover when the crashed pod is restarted.
 						}
 					}
 				}
--- a/pkg/daemon/server/service/rater/pod_tracker.go
+++ b/pkg/daemon/server/service/rater/pod_tracker.go
@@ -95,12 +95,12 @@ func (pt *PodTracker) Start(ctx context.Context) error {
 					}
 					for i := 0; i < int(v.Scale.GetMaxReplicas()); i++ {
 						podName := fmt.Sprintf("%s-%s-%d", pt.pipeline.Name, v.Name, i)
-						// podKey is used as a unique identifier for the pod, it is used by worker to determine the count of processed messages of the pod.
-						podKey := strings.Join([]string{pt.pipeline.Name, v.Name, fmt.Sprintf("%d", i), vType}, PodInfoSeparator)
+						podKey := pt.getPodKey(i, v.Name, vType)
 						if pt.isActive(v.Name, podName) {
 							pt.activePods.PushBack(podKey)
 						} else {
-							pt.activePods.Remove(podKey)
+							// if a pod is not active, we can assume all the following pods are not active as well.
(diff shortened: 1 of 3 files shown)


pkg/daemon/server/service/rater/pod_tracker.go
```diff
@@ -95,12 +95,12 @@ func (pt *PodTracker) Start(ctx context.Context) error {
...
 					for i := 0; i < int(v.Scale.GetMaxReplicas()); i++ {
 						podName := fmt.Sprintf("%s-%s-%d", pt.pipeline.Name, v.Name, i)
-						// podKey is used as a unique identifier for the pod, it is used by worker to determine the count of processed messages of the pod.
-						podKey := strings.Join([]string{pt.pipeline.Name, v.Name, fmt.Sprintf("%d", i), vType}, PodInfoSeparator)
+						podKey := pt.getPodKey(i, v.Name, vType)
 						if pt.isActive(v.Name, podName) {
 							pt.activePods.PushBack(podKey)
 						} else {
-							pt.activePods.Remove(podKey)
+							// if a pod is not active, we can assume all the following pods are not active as well.
```
whynowy, line 102: Why? A pod could be inaccessible because of node upgrade.

yhl25: https://github.com/numaproj/numaflow/blob/b660b6d99118c945a6a466f3387fbade4a1984e0/pkg/daemon/server/service/rater/pod_tracker.go#L107
this is the existing behaviour we break if we find an inactive pod, to decrease the number of HTTP calls. But there is a bug in that logic, assume we have a vertex with 10 active pods and it scales down to 5, but the active pod list will still have 9 pods (only the pod with index 5 will be removed from the list which is wrong we should remove all the pods from index 5 to 9).

yhl25: I agree with a pod going down because of node upgrade, but in the next tracker refresh the active pod list will be updated with the correct count.

whynowy: > 

I was expecting there's a logic at the place of getting metrics call, if there's an error, remove that pod from the active pod list, but it looks like it's not there.

whynowy: Adding active pods is handled in the pod tracker, removing inactive pods could be handled in both places.

yhl25: I think we should remove inactive pods only inside the pod tracker

whynowy: Then a way to correctly figure out which pods need to be removed is needed.

yhl25: current logic 
```
every 30s
for vertex in vertices:
   for replica range vertex.maxReplica:
        ping the pod
        if pod is active: 
             add to active pod list
       else
            remove all the pods from the list with index > replica index
            break
```
Do you see any issues with this logic?

yhl25: I understand a pod going down for some reason will result in the deletion of all the pods with an index greater than the pod's index. But in the next refresh it will be updated, only for duration of 30s we might end up with a low processing rate.

vigith: doesn't look right, it is not a likely path to be optimized in the first place. let's do the below?

```diff
every 30s
for vertex in vertices:
   for replica range vertex.maxReplica:
        ping the pod
        if pod is active: 
             add to active pod list
       else
-            remove all the pods from the list with index > replica index
+            remove the pod from the list
-            break
```

yhl25: +1
yeah, I thought about it, optimisation is not very useful here and also in case if pod is in a crash loop state our processing rate calculation will not be accurate.

whynowy: This is better.

My original suggestion to @KeranYang was, if there are consecutive x pods not active, then we break. Otherwise the default value of `GetMaxReplicas())` is 50.

yhl25: since we refresh every 30 seconds, it should be ok

whynowy (approved):

PR #842, opened 2023-07-10
chore: this is no longer an error messages now that there are multiple goroutines publishing to Offset Timeline

juliev0:
It used to be the case that a given OffsetTimeline could be expected to reliably be published to the K/V store in increasing order of offset. This is no longer the case now that multiple goroutines (each reading from a different partition in a Map vertex) can publish to that OffsetTimeline. For example, take two goroutines G1 and G2 both publishing to the same OffsetTimeline:  

- G1 determines watermark x and publishes to offsets 100-101
- G2 determines watermark y and publishes to offsets 102-103
- G2 publishes watermark y for offset 103
- G1 publishes watermark x for offset 101

This PR addresses the case where watermark x = y. The receiver of the OffsetTimeline in this case will see offset 103 prior to offset 101. This should not be considered an error.

Note this error is seen in the log occasionally if [this](https://github.com/numaproj/numaflow/blob/main/test/reduce-e2e/testdata/simple-keyed-reduce-pipeline.yaml) e2e test pipeline is modified to set 2 partitions for `atoi` vertex (i.e. 2 goroutines publishing).

### DIFF
--- a/pkg/watermark/timeline/offset_timeline.go
+++ b/pkg/watermark/timeline/offset_timeline.go
@@ -85,6 +85,12 @@ func (t *OffsetTimeline) Put(node wmb.WMB) {
 				}
 				return
 			} else {
+				// This can happen if the publisher of the watermark is a Map Vertex reading from >1 partition. In that case, there is one goroutine per
+				// partition and the following example can happen some of the time for two goroutines G1 and G2 both publishing to the same OffsetTimeline:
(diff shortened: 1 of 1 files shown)


vigith (commented):
could you please add a comment? you can copy paste your description :)

juliev0:
> could you please add a comment? you can copy paste your description :)

good idea, will do

vigith (approved):

PR #847, opened 2023-07-12
chore: clean up for recent rater changes

### LINKED ISSUE #846: Fix potential incorrect processing rate calculation
**Describe the bug**
Described in https://github.com/numaproj/numaflow/pull/810#discussion_r1261203309

---

KeranYang:
This PR does following cleanups.
1. Remove unused constants
2. Add more comments
3. Remove window-related codes
4. Remove some unit tests

Regarding 4, the reason of removing is that unit tests should focus on testing **what** instead of "how". Using `helper.go` as an example, within `helper.go`, `UpdateCount` and `CalculateRate` are **what** it does, while `getDeltaBetweenTimestampedCounts`, `calculatePartitionDelta` and `findStartIndex` are **how** it does. I removed all the unit tests that cover **how**.

Also closes #846 by removing the slope approach and keep using the delta aggregation approach to calculate rate.

### DIFF
--- a/pkg/daemon/server/service/rater/helper.go
+++ b/pkg/daemon/server/service/rater/helper.go
@@ -64,14 +64,6 @@ func CalculateRate(q *sharedqueue.OverflowQueue[*TimestampedCounts], lookbackSec
 		return 0
 	}
 
-	if rate := getDeltaBetweenTimestampedCounts(counts[startIndex], counts[endIndex], partitionName) / float64(timeDiff); rate > 0 {
-		// positive slope, meaning there was no restart in the last lookback seconds
-		// TODO - FIX IT - the statement above doesn't always hold true.
-		// see https://github.com/numaproj/numaflow/pull/810#discussion_r1261203309
-		return rate
-	}
-
-	// maybe there was a restart, we need to iterate through the queue to compute the rate.
 	delta := float64(0)
 	for i := startIndex; i < endIndex; i++ {
 		delta += calculatePartitionDelta(counts[i], counts[i+1], partitionName)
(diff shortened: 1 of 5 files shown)


whynowy:
Is `const CountNotAvailable = -1` being used?

KeranYang:
> Is `const CountNotAvailable = -1` being used?

No. I think all the usages are removed from our previous PRs.

whynowy (approved):

pkg/daemon/server/service/rater/helper.go
```diff
@@ -56,47 +56,30 @@ func CalculateRate(q *sharedqueue.OverflowQueue[*TimestampedCounts], lookbackSec
...
 func calculatePartitionDelta(tc1, tc2 *TimestampedCounts, partitionName string) float64 {
-	prevPodReadCount := tc1.PodReadCountSnapshot()
-	currPodReadCount := tc2.PodReadCountSnapshot()
-
 	delta := float64(0)
+	if tc1 == nil || tc2 == nil {
+		// we calculate delta only when both input timestamped counts are non-nil
+		return delta
+	}
+	prevPodReadCount := tc1.PodPartitionCountSnapshot()
```
yhl25, line 81: can this ever be nil? If I remember correctly if index is not found we return early

pkg/daemon/server/service/rater/helper_test.go
```diff
@@ -329,158 +289,45 @@ func TestCalculateRate(t *testing.T) {
...
 		assert.Equal(t, 5.0, CalculateRate(q, 35, "partition4"))
 		assert.Equal(t, 5.0, CalculateRate(q, 100, "partition4"))
-		assert.Equal(t, 100.0, getDeltaBetweenTimestampedCounts(qItems[0], qItems[2], "partition4"))
 
 		// partition100 rate
 		assert.Equal(t, 0.0, CalculateRate(q, 5, "partition100"))
 		assert.Equal(t, 0.0, CalculateRate(q, 15, "partition100"))
 		assert.Equal(t, 0.0, CalculateRate(q, 25, "partition100"))
 		assert.Equal(t, 0.0, CalculateRate(q, 35, "partition100"))
 		assert.Equal(t, 0.0, CalculateRate(q, 100, "partition100"))
```
yhl25, line 331: I remember @veds-g added these unit tests, to cover the edge cases since we are using binary search to fetch the start index. Please don't remove unit tests, it's always good to have more unit tests. The purpose of unit tests is to ensure each unit of the code performs as expected, so that we get confidence when we run it as a whole. If these edge cases are covered in other tests, I am ok with deleting these.

pkg/daemon/server/service/rater/helper.go
```diff
@@ -56,47 +56,30 @@ func CalculateRate(q *sharedqueue.OverflowQueue[*TimestampedCounts], lookbackSec
...
 
-	delta := float64(0)
 	// time diff in seconds.
 	timeDiff := counts[endIndex].timestamp - counts[startIndex].timestamp
 	if timeDiff == 0 {
 		// if the time difference is 0, we return 0 to avoid division by 0
 		// this should not happen in practice because we are using a 10s interval
 		return 0
 	}
-	rate := getDeltaBetweenTimestampedCounts(counts[startIndex], counts[endIndex], partitionName) / float64(timeDiff)
```
yhl25, line 67: I did consider the case where even when the slope is positive there could be a restart, ignored it because it’s very rare and also we will end up showing a low processing rate which is fine I guess. I am ok with iterating and calculating the delta for all the cases since there will not be a lot of elements in the array.

pkg/daemon/server/service/rater/helper_test.go
KeranYang: > If these edge cases are covered in other tests, I am ok with deleting these.

+1. I will see if I can cover these edge cases in other tests.

> Please don't remove unit tests, it's always good to have more unit tests.

I have a mixed feeling about this statement. As I mentioned in PR description, unit tests should focus on testing what instead of how. I would say it's always good to have more unit tests to test what the function does. When we have a well-written unit test set to cover all the functionalities, any performance optimization code changes should not require any changes to unit tests because it doesn't change **what**.

Using our code as example, changing how we find start index from linear search to binary search is an optimization which shouldn't require unit test update. If it introduces bugs that are not caught by exiting unit tests, then we were not doing a good job testing.

KeranYang: I just checked through the test cases. I confirm that ALL test cases that I deleted (`TestCalculatePartitionDelta` and `TestFindStartIndex`) are already covered in `TestCalculateRate`.

PR #869, opened 2023-07-18
fix: highlight edge when buffer is full

veds-g:
when buffer is full for any partition of the edge, the edge becomes highlighted

image


whynowy (approved):

vigith (approved):
how did we retrogress?

veds-g:
> how did we retrogress?

while fixing the UI after moving buffer to vertex, I missed setting the isFull variable.

Issue #740, opened 2023-05-16
Separate the forwarder for source and udf


whynowy:
# Summary

As we are adding more features, the forwarder logic for Map is getting more complex, it would be better to separate it to two files, one for source, another one for map udf.

```[tasklist]
### Tasks
- [x] Create a new forwarder dedicated for source
- [x] Update Map/Sink forwarder to remove source-related code
```

vigith:
#562

PR #874, opened 2023-07-18
refactor: create a new dedicated data forwarder for source

KeranYang:
This is the first step of separating forward from existing one to source.

In this PR, I copied the existing data forwarder from package forward to package sources and then did the following cleanups:
1. removed streaming since it's not supported for source.
2. removed all the `isSource` check and if branches that were for non-source vertices.
3. removed vertexType that was a required parameter of forwarder.
4. made source watermark publisher a required parameter instead of optional.
5. updated unit tests accordingly.

TODO - clean up the map data forwarder to remove source related code.

Closes first sub-task of #740

### DIFF
--- a/pkg/forward/metrics.go
+++ b/pkg/forward/metrics.go
@@ -23,123 +23,123 @@ import (
 	"github.com/numaproj/numaflow/pkg/metrics"
 )
 
-// readMessagesCount is used to indicate the number of messages read
-var readMessagesCount = promauto.NewCounterVec(prometheus.CounterOpts{
+// ReadMessagesCount is used to indicate the number of messages read
+var ReadMessagesCount = promauto.NewCounterVec(prometheus.CounterOpts{
 	Subsystem: "forwarder",
 	Name:      "read_total",
 	Help:      "Total number of Messages Read",
 }, []string{metrics.LabelVertex, metrics.LabelPipeline, metrics.LabelPartitionName})
 
-// readBytesCount is to indicate the number of bytes read
-var readBytesCount = promauto.NewCounterVec(prometheus.CounterOpts{
+// ReadBytesCount is to indicate the number of bytes read
+var ReadBytesCount = promauto.NewCounterVec(prometheus.CounterOpts{
 	Subsystem: "forwarder",
 	Name:      "read_bytes_total",
 	Help:      "Total number of bytes read",
 }, []string{metrics.LabelVertex, metrics.LabelPipeline, metrics.LabelPartitionName})
 
-// readMessagesError is used to indicate the number of errors messages read
-var readMessagesError = promauto.NewCounterVec(prometheus.CounterOpts{
+// ReadMessagesError is used to indicate the number of errors messages read
+var ReadMessagesError = promauto.NewCounterVec(prometheus.CounterOpts{
 	Subsystem: "forwarder",
 	Name:      "read_error_total",
 	Help:      "Total number of Read Errors",
 }, []string{metrics.LabelVertex, metrics.LabelPipeline, metrics.LabelPartitionName})
 
-// writeMessagesCount is used to indicate the number of messages written
-var writeMessagesCount = promauto.NewCounterVec(prometheus.CounterOpts{
+// WriteMessagesCount is used to indicate the number of messages written
+var WriteMessagesCount = promauto.NewCounterVec(prometheus.CounterOpts{
 	Subsystem: "forwarder",
 	Name:      "write_total",
 	Help:      "Total number of Messages Written",
 }, []string{metrics.LabelVertex, metrics.LabelPipeline, metrics.LabelPartitionName})
 
-// writeBytesCount is to indicate the number of bytes written
-var writeBytesCount = promauto.NewCounterVec(prometheus.CounterOpts{
+// WriteBytesCount is to indicate the number of bytes written
+var WriteBytesCount = promauto.NewCounterVec(prometheus.CounterOpts{
 	Subsystem: "forwarder",
 	Name:      "write_bytes_total",
 	Help:      "Total number of bytes written",
 }, []string{metrics.LabelVertex, metrics.LabelPipeline, metrics.LabelPartitionName})
 
-// writeMessagesError is used to indicate the number of errors messages written
-var writeMessagesError = promauto.NewCounterVec(prometheus.CounterOpts{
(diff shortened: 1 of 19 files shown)


pkg/sources/forward/data_forward.go
```diff
@@ -0,0 +1,621 @@
...
+		}
+		return
+	}
+
+	var dataMessages = make([]*isb.ReadMessage, 0, len(readMessages))
+	// store the offsets of the messages we read from source
+	var readOffsets = make([]isb.Offset, len(readMessages))
+	for idx, m := range readMessages {
+		readOffsets[idx] = m.ReadOffset
+		// for source forwarder, below if statement should always hold true, since we never publish idle watermark to source reader buffer
```
KeranYang, line 236: @jy4096 Hey Juanlu, I am not quite familiar with idle watermark. Could you help identify if I could remove more code blocks regarding idle watermark for this source forwarder? Right now I am thinking maybe I can remove the concept of `dataMessages` from forwarder and just use `readMessages` instead. Because we never send ctrl messages to source.

pkg/forward/forward.go
```diff
@@ -211,9 +211,9 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
 	readMessages, err := isdf.fromBufferPartition.Read(ctx, isdf.opts.readBatchSize)
 	if err != nil {
 		isdf.opts.logger.Warnw("failed to read fromBufferPartition", zap.Error(err))
-		readMessagesError.With(map[string]string{metrics.LabelVertex: isdf.vertexName, metrics.LabelPipeline: isdf.pipelineName, metrics.LabelPartitionName: isdf.fromBufferPartition.GetName()}).Inc()
+		ReadMessagesError.With(map[string]string{metrics.LabelVertex: isdf.vertexName, metrics.LabelPipeline: isdf.pipelineName, metrics.LabelPartitionName: isdf.fromBufferPartition.GetName()}).Inc()
```
whynowy, line 214: Don't see any code in this file gets removed?

KeranYang: Yes. I plan to have a separate PR to clean up this file.

pkg/forward/metrics.go
```diff
@@ -23,123 +23,123 @@ import (
 	"github.com/numaproj/numaflow/pkg/metrics"
 )
```
vigith, line 24: Why are we exporting the metric names? If source and forwarder metrics are going to be different, we can just make a copy and delete it later. IMHO, they should be different, easy to search by metric name rather than by tags.

vigith (approved):

yhl25 (approved):
I noticed that there is a significant amount of duplicate code present in both forwarders. However, I think there is scope to refactor the code by generalizing some of the common functions.

pkg/forward/metrics.go
vigith: then just copy over the metrics file for the time being without exporting.

KeranYang:
I am merging this PR and address comments in a follow-up PR, for ease of review.

pkg/forward/metrics.go
KeranYang: It seems not allowed to register duplicate metrics
```
/Users/kyang5/sdk/go1.19.2/bin/go tool test2json -t /Users/kyang5/Library/Caches/JetBrains/GoLand2023.1/tmp/GoLand/___data_forward_test_go.test -test.v -test.paniconexit0 -test.run ^\QTestNewDataForward\E|\QTestNewDataForwardIdleWatermark\E|\QTestNewDataForwardIdleWatermark_Reset\E|\QTestDataForwardSinglePartition\E|\QTestDataForwardMultiplePartition\E|\QTestWriteToBuffer\E$
panic: duplicate metrics collector registration attempted

goroutine 1 [running]:
github.com/prometheus/client_golang/prometheus.(*Registry).MustRegister(0xc000195680?, {0xc0001bb0e0?, 0x1, 0x0?})
	/Users/kyang5/go/pkg/mod/github.com/prometheus/client_golang@v1.12.1/prometheus/registry.go:403 +0x7f
github.com/prometheus/client_golang/prometheus/promauto.Factory.NewCounterVec({{0x1eaabb0?, 0xc000041a90?}}, {{0x0, 0x0}, {0x1ce19e9, 0x9}, {0x1ce2efb, 0xa}, {0x1cf8656, 0x1d}, ...}, ...)
	/Users/kyang5/go/pkg/mod/github.com/prometheus/client_golang@v1.12.1/prometheus/promauto/auto.go:276 +0x133
github.com/prometheus/client_golang/prometheus/promauto.NewCounterVec(...)
	/Users/kyang5/go/pkg/mod/github.com/prometheus/client_golang@v1.12.1/prometheus/promauto/auto.go:176
github.com/numaproj/numaflow/pkg/sources/forward.init()
	/Users/kyang5/Desktop/development/numaflow/numaflow/pkg/sources/forward/metrics.go:27 +0x152
```

KeranYang: Should we consider using `source_data_forwarder` instead of `forwarder` as new metric name? It's as you mentioned above easier to search but backward incompatible.

vigith: i am okay with not being backward incompatible. i want to clean out the code before we hit 1.0, which is by the end of Sept. WDYT?

KeranYang: Agree. I don't know how many customers are actively using the existing source metric to build their dashboards. It should be straight-forward to update the dashboard accordingly after upgrading numaflow to 1.0, we just need to communicate it to customers.

pkg/sources/forward/data_forward.go
```diff
@@ -0,0 +1,621 @@
...
+	// toVertexWMPublishers stores the toVertex to publisher mapping.
+	toVertexWMPublishers map[string]publish.Publisher
+	// srcWMPublisher is used to publish source watermark.
+	srcWMPublisher isb.SourceWatermarkPublisher
+	opts           options
+	vertexName     string
+	pipelineName   string
+	// idleManager manages the idle watermark status.
+	idleManager *wmb.IdleManager
+	// wmbChecker checks if the idle watermark is valid.
```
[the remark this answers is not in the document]
KeranYang: I think idleManager is still used by source forwarder because after transformation we can have conditional forwarding to toVertices. Correct me if I am wrong. cc: @jy4096

PR #875, opened 2023-07-18
feat: Join Vertex

### LINKED ISSUE #719: JOIN Vertex to join data from more than one Vertices
Numaflow should have the ability to JOIN during the following scenarios. 

## JOIN by Key by Window

JOIN by Reduce and using Watermark to close the book. The output will be Keys with Value as an iterator. Please note that the JOIN could be from more than Reduce vertices or consists of one or more Map and Reduce vertices.

image1

## JOIN on Map/Sink

A pure non-join write to Sink or Map without really joining by any key.

image2

## Multiple Sources

This could be based on the previous two use-cases.

image4

## Cyclic Write

We can write back to itself or a previous vertex, but Watermark cannot retrogress, so this can only happen on a pipeline with only Map, and the data written back will implicitly be delayed.

image3

# Tasks

- [x] A design doc which identifies all the changes needed and challenges
- [x] Controller changes to be able to provision join vertex topologies
- [x] Watermark changes to support multi-bucket [Map/reduce]
- [x] #934
- [x] #935
- [x] #936
- [x] #946

```[tasklist]
### Tasks
```

juliev0:
Fixes #719 

Enables Join Vertex (multiple Vertices publish to a single Vertex). This includes both cases where the Join Vertex is a map type, sink type, or Reduce type.

The implementation of this is based on [this](https://docs.google.com/document/d/1bLtYvUlbCdetXnwZBCMQFYtmTumk4_nhh6ZsG3XdOHI/edit#heading=h.ug6nq2un2wsc) design document. (Sorry it's an internal Intuit link, for those who are external.)

**Changes**

The Design document goes over the need for an `EdgeFetcherSet`, which is the set of `EdgeFetchers` that a Vertex consumes Watermarks from. `EdgeFetcherSet` traverses the `EdgeFetchers` for each incoming Edge.

`EdgeFetcherSet` and `EdgeFetcher` both belong to the same `Fetcher` interface. To that end, I needed to make some small changes to the functions in that interface.

In addition to that, some other changes made include:

- setting the ID of a message to avoid duplicates, thereby incorporating the publishing vertex name into it
- enabling all publishing Vertices to use the same partitioning algorithm seed when publishing to a Reduce Vertex to ensure they all partition the same

**Remaining Work**

Cycles: Plan is to enable Cycles in the pipeline as well as long as there is no Reduce Vertex at the point of the cycle or to the right (since Reduce vertex relies on the Watermark, and Cycles create a scenario in which late data arrives after Close-of-book). As of yet, Cycles have not been tested, and also validation needs to be added to prevent the Cycles mentioned above.

A few minor todos in the code, such as a potential improvement in the e2e test to verify that all lines contain certain values rather than that just one line does. 

**Testing**
The e2e tests include tests for:
- Joined Sink Vertex
- Joined Reduce Vertex
- Joined Map Vertex prior to a Reduce Vertex (the Reduce vertex helps verify that the watermark is propagating correctly)
Also, I have visually verified the UI and also done an Idle test to verify that the Idle watermark is propagating.
Please let me know if you think anything else should be tested.

Screenshot 2023-07-19 at 12 17 29 PM

Screenshot 2023-07-24 at 9 45 46 AM


pkg/watermark/fetch/edge_fetcher.go
```diff
@@ -205,16 +210,13 @@ func (e *edgeFetcher) GetHeadWMB(fromPartitionIdx int32) wmb.WMB {
 	defer e.Unlock()
 	// update the last processed watermark for the partition
 	e.lastProcessedWm[fromPartitionIdx] = headWMB.Watermark
-	// we only consider idle watermark if it is smaller than or equal to min of all the last processed watermarks.
```
whynowy, line 208: Why is this removed?

pkg/watermark/timeline/offset_timeline.go
```diff
@@ -42,13 +42,13 @@ type OffsetTimeline struct {
 }
 
 // NewOffsetTimeline returns OffsetTimeline.
-func NewOffsetTimeline(ctx context.Context, c int) *OffsetTimeline {
+func NewOffsetTimeline(ctx context.Context, c int, bucket string) *OffsetTimeline {
```
whynowy, line 45: Is bucket only for logging purpose?

pkg/watermark/fetch/edge_fetcher.go
juliev0: this check is moved into EdgeFetcherSet::GetHeadWMB(). The check needs to be made across all of the Edges.

pkg/watermark/timeline/offset_timeline.go
juliev0: right

pkg/watermark/generic/jetstream/generic.go
```diff
@@ -51,35 +51,12 @@ func BuildWatermarkProgressors(ctx context.Context, vertexInstance *v1alpha1.Ver
...
-	// create a fetcher that fetches watermark.
-	if vertexInstance.Vertex.IsASource() {
-		fetchWatermark = fetch.NewSourceFetcher(ctx, fromBucket, store.BuildWatermarkStoreWatcher(hbWatch, otWatch), processManager)
-	} else if vertexInstance.Vertex.IsReduceUDF() {
-		fetchWatermark = fetch.NewEdgeFetcher(ctx, fromBucket, storeWatcher, processManager, 1)
-	} else {
-		fetchWatermark = fetch.NewEdgeFetcher(ctx, fromBucket, storeWatcher, processManager, vertexInstance.Vertex.Spec.GetPartitionCount())
+		return nil, nil, err
 	}
+	//TODO: create separate function buildPublishers() and call that
```
juliev0, line 59: (for parity with buildFetcher() function)

pkg/watermark/fetch/edge_fetcher_test.go
```diff
@@ -552,7 +552,7 @@ func Test_edgeFetcher_GetHeadWMB(t *testing.T) {
 		},
 		{
 			name:             "all pods empty timeline",
-			processorManager: processorManager3,
+			processorManager: processorManager4,
```
juliev0, line 555: I think this was a typo?

juliev0:
Note that `TestBuiltinEventTimeExtractor` is the test that's failing, but it's also failing in `main` branch.

pkg/watermark/fetch/edge_fetcher_set.go
```diff
@@ -0,0 +1,145 @@
...
+		logging.FromContext(ctx),
+	}
+}
+
+// ProcessOffsetGetWatermark processes the offset on the partition indicated and returns the overall Watermark
+// from all Partitions
+func (efs *edgeFetcherSet) ProcessOffsetGetWatermark(inputOffset isb.Offset, fromPartitionIdx int32) wmb.Watermark {
+	_ = efs.ProcessOffset(inputOffset, fromPartitionIdx) // even if it errored, we'll keep going
+	return efs.GetWatermark()
+}
```
juliev0, line 50: **Rationale:**
This is the equivalent of the original `EdgeFetcher::GetWatermark()` but for multiple EdgeFetchers.

EdgeFetcherSet is simply the addition of more "publishing processors", so any of the original logic that worked between Processors of a single EdgeFetcher can be extended between EdgeFetchers: in this case, find the minimum Watermark time.

pkg/watermark/fetch/edge_fetcher_set.go
```diff
@@ -0,0 +1,145 @@
...
+		if wm == wmb.InitialWatermark { // unset
+			continue
+		}
+		efs.log.Debugf("Got Edge Head Watermark from vertex=%q while processing partition %d: %v", fromVertex, fromPartitionIdx, wm.UnixMilli())
+		if wm.BeforeWatermark(overallWatermark) {
+			overallWatermark = wm
+		}
+	}
+	return overallWatermark
+}
```
juliev0, line 82: **Rationale:**
EdgeFetcherSet is simply the addition of more "publishing processors", so any of the original logic that worked between Processors of a single EdgeFetcher can be extended between EdgeFetchers: in this case, find the minimum head watermark. 

Also, if `EdgeFetcher::GetHeadWatermark()` returns `wmb.InitialWatermark`, it means there was no Watermark found there. Just like `EdgeFetcher::GetHeadWatermark()` allows an individual Processor to not have a Watermark set, `EdgeFetcherSet::GetHeadWatermark()` allows for an entire EdgeFetcher not to have one set.

pkg/watermark/fetch/edge_fetcher_set.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	// we only consider idle watermark if it is smaller than or equal to min of all the last processed watermarks.
+
+	if overallHeadWMB.Watermark > efs.GetWatermark().UnixMilli() {
+		return wmb.WMB{}
+	}
+	return overallHeadWMB
+
+}
```
juliev0, line 120: **Rationale:**
EdgeFetcherSet is simply the addition of more "publishing processors", so any of the original logic that worked between Processors of a single EdgeFetcher can be extended between EdgeFetchers: in this case, find the minimum Head WMB. 

Thus, the check "we only consider idle watermark if it is smaller than or equal to min of all the last processed watermarks" that was originally in `EdgeFetcher::GetHeadWMB()` was moved to `EdgeFetcherSet::GetHeadWMB()`. 
Also, `EdgeFetcher::GetHeadWMB()` returns `wmb.WMB{}` if any Processor is not Idle. `EdgeFetcherSet::GetHeadWMB()` does the same.

juliev0:
@jy4096 I have added rationale for each of the methods of EdgeFetcherSet(), if that helps to review at all.

jy4096 (commented):
First round review... to address a few small comments first

pkg/isb/testutils/rw.go
```diff
@@ -82,10 +83,13 @@ func BuildTestReadMessagesIntOffset(count int64, startTime time.Time) []isb.Read
 	var readMessages = make([]isb.ReadMessage, count)
 
 	for idx, writeMessage := range writeMessages {
-		offset, _ := strconv.Atoi(writeMessage.Header.ID)
+		splitStr := strings.Split(writeMessage.Header.ID, "-")
+		offset, _ := strconv.Atoi(splitStr[0])
 		readMessages[idx] = isb.ReadMessage{
-			Message:    writeMessage,
-			ReadOffset: isb.SimpleIntOffset(func() int64 { return int64(offset) }),
+			Message: writeMessage,
```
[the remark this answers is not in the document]
juliev0: sorry, what do you mean? I know the VSCODE editor typically auto-corrects any indentation that's wrong.

pkg/reduce/pbq/store/wal/segment_test.go
```diff
@@ -261,7 +261,7 @@ func Test_batchSyncWithMaxBatchSize(t *testing.T) {
 	assert.NoError(t, err)
 	err = wal.Write(&message)
 	assert.NoError(t, err)
-	assert.Equal(t, int64(222), tempWAL.prevSyncedWOffset)
+	assert.Equal(t, int64(252), tempWAL.prevSyncedWOffset)
```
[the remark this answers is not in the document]
juliev0: right

pkg/sources/source.go
```diff
@@ -137,7 +137,7 @@ func (sp *SourceProcessor) Start(ctx context.Context) error {
 	shuffleFuncMap := make(map[string]*shuffle.Shuffle)
 	for _, edge := range sp.VertexInstance.Vertex.Spec.ToEdges {
 		if edge.ToVertexType == dfv1.VertexTypeReduceUDF && edge.GetToVertexPartitionCount() > 1 {
-			s := shuffle.NewShuffle(sp.VertexInstance.Vertex.GetName(), edge.GetToVertexPartitionCount())
+			s := shuffle.NewShuffle(edge.To, edge.GetToVertexPartitionCount())
```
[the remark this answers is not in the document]
juliev0: We need to replace it so that all Vertices that write to a given Reduce Vertex use the same Partitioning Seed to ensure they all partition the same. So, the `edge.To` Vertex enables them all to use the same seed.

pkg/watermark/fetch/edge_fetcher.go
```diff
@@ -39,8 +39,8 @@ import (
 	"github.com/numaproj/numaflow/pkg/watermark/wmb"
 )
 
-// edgeFetcher is a fetcher between two vertices.
-type edgeFetcher struct {
+// EdgeFetcher is a fetcher between two vertices.
+type EdgeFetcher struct {
```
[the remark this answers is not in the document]
juliev0: Oh, at one point with what I was doing I had to do that, but then I changed that and meant to change this back too and forgot. Will do.

pkg/watermark/fetch/edge_fetcher.go
```diff
@@ -223,13 +225,13 @@ func (e *edgeFetcher) Close() error {
 	return nil
 }
 
-// getMinFromLastProcessed returns the smallest watermark among all the last processed watermarks.
-func (e *edgeFetcher) getMinFromLastProcessed(watermark int64) int64 {
-	minWm := watermark
+// GetWatermark returns the smallest watermark among all the last processed watermarks.
+func (e *EdgeFetcher) GetWatermark() wmb.Watermark {
+	minWm := int64(0)
```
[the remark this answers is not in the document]
juliev0: So, line 232 checks for the case of `minWm == int64(0)`. I suppose if `e.lastProcessedWm` were empty then this would get set to 0. Is that the concern? or are you just thinking it's more elegant not to have to do the `minWm == int64(0)` check?

pkg/watermark/fetch/edge_fetcher_set.go
```diff
@@ -0,0 +1,145 @@
...
+	return &edgeFetcherSet{
+		edgeFetchers,
+		logging.FromContext(ctx),
+	}
+}
+
+// ProcessOffsetGetWatermark processes the offset on the partition indicated and returns the overall Watermark
+// from all Partitions
+func (efs *edgeFetcherSet) ProcessOffsetGetWatermark(inputOffset isb.Offset, fromPartitionIdx int32) wmb.Watermark {
+	_ = efs.ProcessOffset(inputOffset, fromPartitionIdx) // even if it errored, we'll keep going
```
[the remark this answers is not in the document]
juliev0: I didn't log it here since it's being logged in the underlying function 
As far as why we keep going, I guess we don't need to, but I figured it was better to try if for some reason something happens. ?

pkg/watermark/fetch/edge_fetcher_set.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+	}
+	return overallWatermark
+}
+
+func (efs *edgeFetcherSet) Close() error {
+	for _, fetcher := range efs.edgeFetchers {
+		err := fetcher.Close()
+		if err != nil {
+			return err
```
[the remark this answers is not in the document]
juliev0: good idea

pkg/watermark/fetch/source_fetcher.go
```diff
@@ -56,19 +56,32 @@ func NewSourceFetcher(ctx context.Context, sourceBufferName string, storeWatcher
...
 // it ignores the input Offset.
-func (e *sourceFetcher) GetWatermark(offset isb.Offset, fromPartitionIdx int32) wmb.Watermark {
+func (e *sourceFetcher) GetWatermark() wmb.Watermark {
 	var epoch int64 = math.MaxInt64
 	var debugString strings.Builder
 
 	for _, p := range e.processorManager.GetAllProcessors() {
+
+		if len(p.GetOffsetTimelines()) != 1 {
+			e.log.Fatalf("sourceFetcher %+v has %d offset timelines??, expected 1", e, len(p.GetOffsetTimelines()))
```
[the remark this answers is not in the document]
juliev0: how do you mean? are you referring to the double question mark?

pkg/isb/testutils/rw.go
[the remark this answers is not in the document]
jy4096: I tested in GoLand, and it will auto reformat these two lines to align `writeMessage` and `isb.xxx`

pkg/watermark/fetch/edge_fetcher.go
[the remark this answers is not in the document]
jy4096: I prefer to assign minWM to MaxInt as the initial value instead of doing a zero check later

pkg/watermark/fetch/source_fetcher.go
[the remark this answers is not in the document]
jy4096: `you referring to the double question mark` yes... why do we have `??` in the log?....

pkg/watermark/fetch/edge_fetcher_set.go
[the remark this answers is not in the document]
jy4096: I just checked underlying code, looks like the only error will be from `var offset, err = inputOffset.Sequence()` and this is not a recoverable error...

we can discuss with Vigith and Derek on how to proceed

pkg/watermark/fetch/source_fetcher.go
```diff
@@ -56,19 +56,32 @@ func NewSourceFetcher(ctx context.Context, sourceBufferName string, storeWatcher
...
+		}
+		offsetTimeline := p.GetOffsetTimelines()[0]
```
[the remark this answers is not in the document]
juliev0: Yes, [here](https://github.com/numaproj/numaflow/blob/main/pkg/watermark/processor/processor_manager.go#L83) if it's a source then `Vertex.fromBufferPartitionCount=1`.
And [here](https://github.com/numaproj/numaflow/blob/main/pkg/watermark/processor/processor_to_fetch.go#L87) len(offsetTimelines) = `Vertex.fromBufferPartitionCount`.

pkg/watermark/processor/processor_manager.go
```diff
@@ -194,7 +196,7 @@ func (v *ProcessorManager) startHeatBeatWatcher() {
 					var entity = NewProcessorEntity(value.Key())
 					// if the processor is a reduce or source processor, then we only need one fromProcessor
 					// because the reduce or source will read from only one partition.
-					fromProcessor := NewProcessorToFetch(v.ctx, entity, 10, v.fromBufferPartitionCount)
+					fromProcessor := NewProcessorToFetch(v.ctx, entity, v.bucket, 10, v.fromBufferPartitionCount)
```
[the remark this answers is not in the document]
juliev0: yes, but it is helpful for logging

pkg/watermark/fetch/edge_fetcher.go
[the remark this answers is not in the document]
juliev0: I agree

pkg/isb/testutils/rw.go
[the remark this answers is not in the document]
juliev0: I am not sure why that's changing it, but the linter we're using - golangcli-lint - which hosts all of these [linters](https://golangci-lint.run/usage/linters) isn't flagging it.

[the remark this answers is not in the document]
jy4096: I would prefer to have the format style consistently
e.g. lint also doesn't detect the ordering of import package but yet we still have a preference.. we fix the format when we see anything inconsistent.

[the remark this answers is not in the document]
jy4096: I just figured that when the next line is multi-line the IDE won't align ... you may ignore this

pkg/watermark/fetch/edge_fetcher.go
```diff
@@ -67,19 +67,27 @@ func NewEdgeFetcher(ctx context.Context, bucketName string, storeWatcher store.W
...
 }
 
-// GetWatermark gets the largest possible watermark from all the active processors of all partitions for the given offset and partition.
-// We calculate the watermark for the given offset and partition, update the lastProcessedWatermark of the given partition to the watermark we just calculate.
-// Then, we compare the lastProcessedWatermark from all partitions and return the minimum as the edge watermark.
-// deletes the processor if it's not active.
-func (e *edgeFetcher) GetWatermark(inputOffset isb.Offset, fromPartitionIdx int32) wmb.Watermark {
+// ProcessOffsetGetWatermark processes the offset on the partition indicated and returns the overall Watermark
+// from all Partitions
+func (e *edgeFetcher) ProcessOffsetGetWatermark(offset isb.Offset, fromPartitionIdx int32) wmb.Watermark {
```
vigith, line 76:
```suggestion
func (e *edgeFetcher) UpdateAndGetWatermark(offset isb.Offset, fromPartitionIdx int32) wmb.Watermark {
```

I think `ProcessOff` can be confusing since we use p

pkg/watermark/fetch/interface.go
```diff
@@ -26,8 +26,13 @@ import (
 // Fetcher fetches watermark data from Vn-1 vertex.
 type Fetcher interface {
 	io.Closer
-	// GetWatermark returns the inorder monotonically increasing watermark of the edge connected to Vn-1.
-	GetWatermark(offset isb.Offset, fromPartitionIdx int32) wmb.Watermark
+	// ProcessOffset updates state according to the incoming offset on the given partition
+	ProcessOffset(offset isb.Offset, fromPartitionIdx int32) error
+	// GetWatermark returns the current watermark based on what has been processed
+	GetWatermark() wmb.Watermark
```
vigith, line 32: do we need to expose these two?

vigith (commented):
partitial review

pkg/watermark/fetch/edge_fetcher.go
jy4096: Do we only call `EdgeFetcher::GetHeadWMB()` in `EdgeFetcherSet::GetHeadWMB()`?? Removing this part will result in the `EdgeFetcher::GetHeadWMB()` returning wrong value.

juliev0: yes. For the purpose of `EdgeFetcherSet::GetHeadWMB()` we can't call it here. If you have a concern about somebody calling this directly in the future as far as this check being removed let me know.

xdevxy (commented):
IIUC, cycle is supported out of box if there's no Reduce Vertex in the pipeline?

juliev0:
> IIUC, cycle is supported out of box if there's no Reduce Vertex in the pipeline?

It will be prohibited at the Join or downstream of it. A subsequent PR will take care of validation for that as well as add an e2e test.

test/reduce-e2e/testdata/joined-map-vertex-pipeline.yaml
```diff
@@ -0,0 +1,74 @@
...
+spec:
+  limits:
+    readBatchSize: 50
+  vertices:
+    - name: in-0
+      scale:
+        min: 1
+      containerTemplate:
+        env:
+          - name: NUMAFLOW_DEBUG
```
xdevxy, line 14: Why do we need to enable debug in the e2e test?

test/reduce-e2e/testdata/joined-map-vertex-pipeline.yaml
```diff
@@ -0,0 +1,74 @@
+apiVersion: numaflow.numaproj.io/v1alpha1
+kind: Pipeline
+metadata:
+  name: joined-map-vertex
```
xdevxy, line 4: Should this be named to 'join-on-map' to match the naming pattern of the above reduce join vertex pipeline? So is the file name.

test/reduce-e2e/testdata/join-reduce-pipeline.yaml
```diff
@@ -0,0 +1,78 @@
+apiVersion: numaflow.numaproj.io/v1alpha1
+kind: Pipeline
+metadata:
+  name: join-on-reduce
+spec:
+  limits:
+    readBatchSize: 50
```
xdevxy, line 7: Any reason we cannot have the default `readBatchSize`?

pkg/watermark/wmb/watermark.go
```diff
@@ -39,6 +39,14 @@ func (w Watermark) After(t time.Time) bool {
 	return time.Time(w).After(t)
 }
 
+func (w Watermark) AfterWatermark(compare Watermark) bool {
```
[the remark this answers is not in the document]
xdevxy: Any places we would not need `AfterWatermark` but just `After`? If not, then we can remove it?

pkg/watermark/generic/jetstream/generic.go
```diff
@@ -51,25 +51,71 @@ func BuildWatermarkProgressors(ctx context.Context, vertexInstance *v1alpha1.Ver
...
+
+	publishWatermark, err := buildPublishers(ctx, pipelineName, vertexInstance)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return fetchWatermark, publishWatermark, nil
+}
+
+func buildFetcher(ctx context.Context, vertexInstance *v1alpha1.VertexInstance) (fetch.Fetcher, error) {
```
xdevxy, line 68: Can you add a comment for this function?

juliev0: sure, can add comments for the various functions I added here

test/reduce-e2e/testdata/join-reduce-pipeline.yaml
juliev0: no, I must have been imitating another test which set it

juliev0: I see the other tests in this directory have it set - should they set it and not this one?

test/reduce-e2e/testdata/joined-map-vertex-pipeline.yaml
juliev0: meant to remove all of that - was using it for my own debugging - will do

pkg/sources/source.go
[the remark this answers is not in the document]
KeranYang: +1. Might be worth adding a comment to explicitly explain the rationale behind the scene, for future reference.

pkg/watermark/fetch/edge_fetcher_set.go
```diff
@@ -0,0 +1,145 @@
+/*
+Copyright 2023 The Numaproj Authors.
```
KeranYang, line 2: nit: shall we update all Copyright to 2023? Derek has a script under the `hack` folder to do so.

pkg/watermark/fetch/interface.go
```diff
@@ -26,8 +26,11 @@ import (
 // Fetcher fetches watermark data from Vn-1 vertex.
 type Fetcher interface {
 	io.Closer
-	// GetWatermark returns the inorder monotonically increasing watermark of the edge connected to Vn-1.
-	GetWatermark(offset isb.Offset, fromPartitionIdx int32) wmb.Watermark
+	// GetWatermark returns the current watermark based on what has been processed
+	GetWatermark() wmb.Watermark
+	// UpdateAndFetchWatermark processes the offset on the partition indicated and returns the overall Watermark
+	// from all Partitions
+	UpdateAndFetchWatermark(offset isb.Offset, fromPartitionIdx int32) wmb.Watermark
```
KeranYang, line 33: For `UpdateAndFetchWatermark` it seems we use one interface to do two things, both writing and reading. For writing, we are using the input parameters but for reading, it doesn't seem care about the input. Interface users could by mistake think the method is fetching from a certain offset/partitionIdx instead of overall watermark.

Would it be better if we separate them? To have `UpdateWatermark(offset, fromPartitionIdx)` and use the existing `GetWatermark()` for fetching.

pkg/watermark/publish/publisher.go
```diff
@@ -148,17 +148,17 @@ func (p *publish) PublishWatermark(wm wmb.Watermark, offset isb.Offset, toVertex
 
 	value, err := otValue.EncodeToBytes()
 	if err != nil {
-		p.log.Errorw("Unable to publish watermark", zap.String("HB", p.heartbeatStore.GetStoreName()), zap.String("OT", p.otStore.GetStoreName()), zap.String("key", key), zap.Error(err))
+		p.log.Errorw("Unable to publish watermark", zap.Int32("partition", toVertexPartitionIdx), zap.String("HB", p.heartbeatStore.GetStoreName()), zap.String("OT", p.otStore.GetStoreName()), zap.String("key", key), zap.Error(err))
```
KeranYang, line 151: zap.Int32("toVertexPartitionIdx", toVertexPartitionIdx)

pkg/watermark/wmb/watermark.go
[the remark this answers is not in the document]
KeranYang: +1

test/reduce-e2e/testdata/join-on-map-pipeline.yaml
```diff
@@ -0,0 +1,58 @@
+apiVersion: numaflow.numaproj.io/v1alpha1
+kind: Pipeline
+metadata:
+  name: join-on-map
```
KeranYang, line 4: Do we need a reduce vertex for join-on-map test? If no, should we remove the reduce vertex to make the test simpler and consider moving it out of the reduce test suite?

juliev0: We do need the reduce vertex for this test. Here I was explicitly trying to verify the case of the Joined Map vertex as far as the watermark, since the Reduce Vertex downstream helps to verify that the Watermark was passed successfully.

pkg/watermark/fetch/interface.go
juliev0: _Interface users could by mistake think the method is fetching from a certain offset/partitionIdx instead of overall watermark_

That was in fact my motivation for replacing the original `GetWatermark(offset, fromPartitionIdx)`.

I also wanted to have two separate methods like you're saying but @jy4096 expressed a desire to still have a single method to do both like we did before since they're generally done together.

I'll make it whatever the team wants in the end.

juliev0: cc @vigith

pkg/watermark/fetch/edge_fetcher.go
jy4096: Could you add a TODO or comment (regarding `this check is moved into EdgeFetcherSet::GetHeadWMB() The check needs to be made across all of the Edges` and `Removing this part will result in the EdgeFetcher::GetHeadWMB() returning a wrong value`)  to inform that this could have a potential problem, and also explain that we have the check in EdgeFetcherSet?

juliev0: Conceivably the note could say that it's the responsibility of the caller to verify it?

juliev0: Or do you feel like the contract of the method says that it will definitively do this?

jy4096: > the contract of the method says that it will definitively do this

We have the `GetHeadWMB()` in the interface. The two datatypes that implement this interface have different behavior: EdgeFetcher doesn't compare the WMB with lastProcessedWMs while EdgeFetcherSet does. This is inconsistent and may confuse developers.

It's up to us how to define what a method does. I think either having the comparison or not having it works as long as the datatypes that implement this method in a way what the method is designed to be.

juliev0: I agree that consistency is good.

pkg/isb/stores/jetstream/writer.go
```diff
@@ -75,7 +75,7 @@ func NewJetStreamBufferWriter(ctx context.Context, client jsclient.JetStreamClie
 		js:           js,
 		opts:         o,
 		isFull:       atomic.NewBool(true),
-		log:          logging.FromContext(ctx).With("bufferWriter", name).With("stream", stream).With("subject", subject),
+		log:          logging.FromContext(ctx).With("bufferWriter", name).With("stream", stream).With("subject", subject).With("partition", partitionIdx),
```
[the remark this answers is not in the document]
juliev0: oh, do we always use the exact variable name here?

pkg/forward/forward_test.go
```diff
@@ -1242,7 +1250,7 @@ func TestInterStepDataForwardSinglePartition(t *testing.T) {
 	assert.NoError(t, err, "expected no error")
 	assert.Len(t, readMessages, int(count))
 	assert.Equal(t, []interface{}{writeMessages[0].Header.Keys, writeMessages[1].Header.Keys}, []interface{}{readMessages[0].Header.Keys, readMessages[1].Header.Keys})
-	assert.Equal(t, []interface{}{writeMessages[0].Header.ID + "-0-0", writeMessages[1].Header.ID + "-0-0"}, []interface{}{readMessages[0].Header.ID, readMessages[1].Header.ID})
+	assert.Equal(t, []interface{}{writeMessages[0].Header.ID, writeMessages[1].Header.ID}, []interface{}{readMessages[0].Header.ID, readMessages[1].Header.ID})
```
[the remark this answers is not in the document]
juliev0: this is a really good catch

pkg/watermark/fetch/interface.go
jy4096: How about we update 
```
	// GetWatermark returns the current watermark based on what has been processed
	GetWatermark() wmb.Watermark
	// UpdateAndFetchWatermark processes the offset on the partition indicated and returns the overall Watermark
	// from all Partitions
	UpdateAndFetchWatermark(offset isb.Offset, fromPartitionIdx int32) wmb.Watermark
```
to
```
	// GetSmallestLastProcessedWatermark returns the overall smallest last processed watermark
	GetSmallestLastProcessedWatermark() wmb.Watermark
	// ComputeWatermark uses the given offset and partition to compute and return the current watermark
	ComputeWatermark(offset isb.Offset, fromPartitionIdx int32) wmb.Watermark
```
And then in edgeFetcher write a godoc like this
```
        // ComputeWatermark uses the given offset and partition to calculate the watermark, 
        // update the lastProcessedWatermark of the given partition to the watermark we just calculate, 
        // and return the minimum lastProcessedWatermark from all partitions as the watermark. 
        // Finally, it deletes the processor if it's not active.
```

jy4096: The original `GetWatermark(offset)` before we introduced multi partition is also more like "compute and return the watermark for the given offset".

jy4096: Before multi partition, we had a method `GetWatermark(offset) watermark`. This function used the offset from `Read` to calculate and return the watermark for the current readBatch.
With the introduction of multi partition, we updated the method to be `GetWatermark(offset, partition) watermark`. This method had two steps: calculate the watermark for the given offset and partition provided by `readABatch`; compare the last processed watermark among all the partitions and return the smallest value.

In this PR, at first, we extracted the above two steps into two functions: `ProcessOffset(offset, partition) watermark` and `GetWatermark()`.

I think that, when we read a batch and need to assign a watermark to this batch, the operation `ProcessOffset` and `GetWatermark` are strictly bonded together. Therefore, we need to have a new function `Something` (we have been discussing the function name for a long time so I will just use `Something` here).
```
func Something(offset, partition) watermark {
   ProcessOffset(offset, partition) watermark
   GetWatermark()
}
```
The `Read` (which imports the watermark) doesn't need to know that it has to `ProcessOffset to update the last processed watermark` and only then it can `Get the correct watermark for the batch`.

Above is why I prefer to have this `Something` function.

pkg/watermark/fetch/edge_fetcher.go
juliev0: I think I'm going to move the check into the callers and this method will make no promises as far as that.

pkg/watermark/fetch/interface.go
KeranYang: @jy4096 thanks for catching me up on the history. +1 on adding more comments which provides ALL important informations about the interface and reduces developer cognitive loads.

I agree that having the `Something` function will simplify the `BatchRead` in a way that we don't need to call two methods in sequence. However, from the view of `Read` (which imports the watermark), I think it needs to know that it has to `UpdateOffset(offset, partition)` first and then `GetWatermark()`, which is exactly why we name `Something` `UpdateAndFetchWatermark` at first place, to explicitly describe what the method does. In general, an interface should provide ALL important informations that the caller needs to know, otherwise we introduce unknowns.

Personally I still prefer separating `Update` and `Get` as it's clearer for the interface. But I am also ok with combining if 1. They are called in sequence very frequently and 2. We put in good descriptions for the `Something` interface, as you did above.

whynowy: It looks right to me to have 2 functions:

1. `GetWatermark()` - the watermark displayed in the UI should be this, right?
2. `Something(offset, partition) watermark`, which includes 2 steps (`ProcessOffset` and `GetWatermark`).

Isn't this true?
->  _xxxx think the method is fetching from a certain offset/partitionIdx instead of overall watermark_

KeranYang: another data point to help us make the decision is how often we will call `Update` without following by a `Get`.

KeranYang: > Isn't this true?
-> xxxx think the method is fetching from a certain offset/partitionIdx instead of overall watermark

By adding interface descriptions we can avoid such confusion.

juliev0: @whynowy this is not true: "the method is fetching from a certain offset/partitionIdx instead of overall watermark". It is fetching overall watermark. Therefore, do you want to change your mind?

jy4096: > another data point to help us make the decision is how often we will call `Update` without following by a `Get`.

we don't have such use case ATM

juliev0: I would still prefer to have `UpdateWatermark(offset, fromPartitionIdx)` and `GetWatermark()` but can be flexible on that.

But if we keep as is, I'm not sure what's unclear about the current comment "UpdateAndFetchWatermark processes the offset on the partition indicated and returns the **overall Watermark from all Partitions**"

pkg/watermark/fetch/edge_fetcher.go
juliev0: Do you consider it leaking if the caller does a sanity check to verify that the Idle Watermark is not more advanced than `GetWatermark()`?

pkg/isb/stores/jetstream/writer.go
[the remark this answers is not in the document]
juliev0: got it, makes sense

pkg/watermark/fetch/interface.go
jy4096: > But if we keep as is, I'm not sure what's unclear about the current comment "UpdateAndFetchWatermark processes the offset on the partition indicated and returns the **overall Watermark from all Partitions**"

https://github.com/numaproj/numaflow/pull/875#discussion_r1274179798
How about this proposal?

pkg/watermark/fetch/edge_fetcher.go
juliev0: Perhaps I should also revisit whether this definitely needs to be yanked out here. At the time, I decided it did but maybe that decision needs revisiting.

pkg/watermark/fetch/edge_fetcher.go
```diff
@@ -205,10 +210,7 @@ func (e *edgeFetcher) GetHeadWMB(fromPartitionIdx int32) wmb.WMB {
 	defer e.Unlock()
 	// update the last processed watermark for the partition
 	e.lastProcessedWm[fromPartitionIdx] = headWMB.Watermark
-	// we only consider idle watermark if it is smaller than or equal to min of all the last processed watermarks.
-	if headWMB.Watermark > e.getMinFromLastProcessed(headWMB.Watermark) {
-		return wmb.WMB{}
-	}
```
jy4096, line 211: Add back the check

pkg/watermark/fetch/edge_fetcher_set.go
```diff
@@ -0,0 +1,145 @@
...
+				overallHeadWMB = watermarkBuffer
+			}
+		}
+	}
+
+	// we only consider idle watermark if it is smaller than or equal to min of all the last processed watermarks.
+
+	if overallHeadWMB.Watermark > efs.GetWatermark().UnixMilli() {
+		return wmb.WMB{}
+	}
```
jy4096, line 113: Remove this check

pkg/watermark/fetch/interface.go
```diff
@@ -26,8 +26,11 @@ import (
 // Fetcher fetches watermark data from Vn-1 vertex.
 type Fetcher interface {
 	io.Closer
-	// GetWatermark returns the inorder monotonically increasing watermark of the edge connected to Vn-1.
-	GetWatermark(offset isb.Offset, fromPartitionIdx int32) wmb.Watermark
+	// GetWatermark returns the current watermark based on what has been processed
+	GetWatermark() wmb.Watermark
```
jy4096, line 30: Unexpose `GetWatermark()`

pkg/watermark/fetch/interface.go
```diff
@@ -26,8 +26,11 @@ import (
...
+	// UpdateAndFetchWatermark processes the offset on the partition indicated and returns the overall Watermark
+	// from all Partitions
```
jy4096, line 32: `ComputeWatermark`

jy4096 (commented):
Proposal

TODO: fetcher interface refactor (open an issue)

pkg/watermark/fetch/interface.go
xdevxy: +1

pkg/watermark/fetch/edge_fetcher.go
juliev0: Per our discussion I have temporarily added the check back and removed it from EdgeFetcherSet until the interfaces for the two structs can be split

pkg/watermark/fetch/interface.go
whynowy: > @whynowy this is not true: "the method is fetching from a certain offset/partitionIdx instead of overall watermark". It is fetching overall watermark. Therefore, do you want to change your mind?

@juliev0 - could you help me understand this?

juliev0: @whynowy As you can see in the implementation of `EdgeFetcher` for example, [this](https://github.com/juliev0/numaflow/blob/join-vertex-new/pkg/watermark/fetch/edge_fetcher.go#L81) method which has been renamed to `ComputeWatermark(offset, fromPartitionIdx)` gets the smallest watermark from all partitions. This both sets the current watermark for this partition and then gets the overall watermark from all partitions. I thought dividing it into `UpdateWatermark(offset, fromPartitionIdx)` and `GetWatermark()` might make sense, but I think Juanlu would like to expose just a single method in the interface. Vigith, Juanlu, Siddhant, Keran, and I met and decided in the end that with the interface method being sufficiently commented, that satisfies everyone enough.

pkg/watermark/fetch/edge_fetcher_set.go
```diff
@@ -0,0 +1,135 @@
+/*
+Copyright 202 The Numaproj Authors.
```
jy4096, line 2:
```suggestion
Copyright 2022 The Numaproj Authors.
```

pkg/watermark/fetch/edge_fetcher_test.go
```diff
@@ -159,8 +159,8 @@ func TestBuffer_GetWatermarkWithOnePartition(t *testing.T) {
 				log:              zaptest.NewLogger(t).Sugar(),
 				lastProcessedWm:  lastProcessed,
 			}
-			if got := b.GetWatermark(isb.SimpleStringOffset(func() string { return strconv.FormatInt(tt.args.offset, 10) }), 0); time.Time(got).In(location) != time.UnixMilli(tt.want).In(location) {
-				t.Errorf("GetWatermark() = %v, want %v", got, wmb.Watermark(time.UnixMilli(tt.want)))
+			if got := b.ComputeWatermark(isb.SimpleStringOffset(func() string { return strconv.FormatInt(tt.args.offset, 10) }), 0); time.Time(got).In(location) != time.UnixMilli(tt.want).In(location) {
+				t.Errorf("ProcessOffsetGetWatermark() = %v, want %v", got, wmb.Watermark(time.UnixMilli(tt.want)))
```
jy4096, line 163:
```suggestion
				t.Errorf("ComputeWatermark() = %v, want %v", got, wmb.Watermark(time.UnixMilli(tt.want)))
```

pkg/watermark/fetch/edge_fetcher_test.go
```diff
@@ -40,19 +40,19 @@ import (
 	"github.com/numaproj/numaflow/pkg/watermark/store"
 )
 
-func TestBuffer_GetWatermarkWithOnePartition(t *testing.T) {
+func TestBuffer_ProcessOffsetGetWatermarkWithOnePartition(t *testing.T) {
```
jy4096, line 43:
```suggestion
func TestBuffer_ComputeWatermarkWithOnePartition(t *testing.T) {
```

pkg/watermark/fetch/edge_fetcher_set_test.go
```diff
@@ -0,0 +1,337 @@
...
+	"github.com/numaproj/numaflow/pkg/isb"
+	"github.com/numaproj/numaflow/pkg/watermark/processor"
+	"github.com/numaproj/numaflow/pkg/watermark/store"
+	"github.com/numaproj/numaflow/pkg/watermark/store/noop"
+	"github.com/numaproj/numaflow/pkg/watermark/wmb"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zaptest"
+)
+
+func Test_EdgeFetcherSet_ProcessOffsetGetWatermark(t *testing.T) {
```
jy4096, line 37:
```suggestion
func Test_EdgeFetcherSet_ComputeWatermark(t *testing.T) {
```

pkg/watermark/fetch/edge_fetcher_test.go
```diff
@@ -169,7 +169,7 @@ func TestBuffer_GetWatermarkWithOnePartition(t *testing.T) {
 	}
 }
 
-func TestBuffer_GetWatermarkWithMultiplePartition(t *testing.T) {
+func TestBuffer_ProcessOffsetGetWatermarkWithMultiplePartition(t *testing.T) {
```
jy4096, line 172:
```suggestion
func TestBuffer_ComputeGetWatermarkWithMultiplePartition(t *testing.T) {
```

pkg/watermark/fetch/edge_fetcher_test.go
```diff
@@ -323,7 +323,7 @@ func TestBuffer_GetWatermarkWithMultiplePartition(t *testing.T) {
 				log:              zaptest.NewLogger(t).Sugar(),
 				lastProcessedWm:  tt.lastProcessedWm,
 			}
-			if got := b.GetWatermark(isb.SimpleStringOffset(func() string { return strconv.FormatInt(tt.args.offset, 10) }), tt.partitionIdx); time.Time(got).In(location) != time.UnixMilli(tt.want).In(location) {
+			if got := b.ComputeWatermark(isb.SimpleStringOffset(func() string { return strconv.FormatInt(tt.args.offset, 10) }), tt.partitionIdx); time.Time(got).In(location) != time.UnixMilli(tt.want).In(location) {
 				t.Errorf("GetWatermark() = %v, want %v", got, wmb.Watermark(time.UnixMilli(tt.want)))
```
jy4096, line 327:
```suggestion
				t.Errorf("ComputeWatermark() = %v, want %v", got, wmb.Watermark(time.UnixMilli(tt.want)))
```

pkg/watermark/fetch/edge_fetcher.go
```diff
@@ -67,19 +67,27 @@ func NewEdgeFetcher(ctx context.Context, bucketName string, storeWatcher store.W
...
-// deletes the processor if it's not active.
-func (e *edgeFetcher) GetWatermark(inputOffset isb.Offset, fromPartitionIdx int32) wmb.Watermark {
+// ComputeWatermark processes the offset on the partition indicated and returns the overall Watermark
+// from all Partitions
+func (e *edgeFetcher) ComputeWatermark(offset isb.Offset, fromPartitionIdx int32) wmb.Watermark {
+	err := e.updateWatermark(offset, fromPartitionIdx)
+	if err != nil {
+		return wmb.InitialWatermark
+	}
+	return e.getWatermark()
```
jy4096, line 81: Do we need a read lock to access lastProcessedWatermark in `getWatermark`?

juliev0: oh, good catch. It looks like that was being locked by the caller before but then the code was rearranged

pkg/watermark/fetch/edge_fetcher.go
```diff
@@ -202,14 +207,22 @@ func (e *edgeFetcher) GetHeadWMB(fromPartitionIdx int32) wmb.WMB {
 		return wmb.WMB{}
 	}
 
-	e.Lock()
-	defer e.Unlock()
 	// update the last processed watermark for the partition
+	e.Lock()
 	e.lastProcessedWm[fromPartitionIdx] = headWMB.Watermark
+	e.Unlock()
+
```
juliev0, line 214: I like putting a lock around the smallest section of code that requires it

jy4096: how about using https://pkg.go.dev/sync#RWMutex

juliev0: that would be more optimal

jy4096 (approved):

PR #866, opened 2023-07-16
feat: controller changes for Side Inputs support

whynowy:
Be able to orchestrate a pipeline spec like following:

```yaml
apiVersion: numaflow.numaproj.io/v1alpha1
kind: Pipeline
metadata:
  name: simple-pipeline
spec:
  sideInputs:
    - name: s1
      container:
        image: my-s1-image
      trigger:
        interval: 300s
    - name: s2
      container:
        image: my-s2-image
      trigger:
        interval: 600s
  vertices:
    - name: in
      source:
        http: {}
    - name: my-udf
       sideInputs:
         - s1
         - s2
      udf:
        container:
          image: my-image
    - name: out
      sink:
        # A simple log printing sink
        log: {}
  edges:
    - from: in
      to: my-udf
    - from: my-udf
      to: out
```

For a pipeline like above, following new resources will be created:

- Side Inputs data store in JetStream.
- Two Side Input Manager Deployments for `s1` and `s2`, each of them contains 1 init container to check if the store is ready, and 2 containers.
- A new init container on the `my-udf` vertex pod, which is supposed to load the data from data store.
- A sidecar container to watch the data store changes.

PR #880, opened 2023-07-19
chore: some initial code for side inputs manager

whynowy:
Explain what this PR does.

### DIFF
--- a/pkg/sideinputs/manager/manager_test.go
+++ b/pkg/sideinputs/manager/manager_test.go
@@ -1,0 +1,62 @@
+/*
+Copyright 2022 The Numaproj Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


vigith (approved):

pkg/sideinputs/manager/manager_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+	dfv1 "github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
+)
+
+func Test_resolveCron(t *testing.T) {
+	f := func() { fmt.Println("hello") }
+
+	t.Run("valid cron schedule", func(t *testing.T) {
+		trigger := &dfv1.SideInputTrigger{
+			Schedule: pointer.String("*/5 * * * *"),
```
jy4096, line 36: https://pkg.go.dev/github.com/robfig/cron#hdr-CRON_Expression_Format
`using 6 space-separated fields` I think the package needs [6 fields](https://github.com/robfig/cron/blob/v1.2.0/parser_test.go#L126) instead of [5 fields](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#schedule-syntax)

I see tests in https://github.com/numaproj/numaflow/pull/866 also uses 5 fields..

jy4096: we can merge this in and I will update the tests in the next PR

vigith: perhaps we can merge to a different branch and fix tests in that branch and then merge to main?

jy4096: ignore above :-) I was looking at v1.2.0 while we use v3.

PR #884, opened 2023-07-19
chore: remove source related code from inter-step forwarder

KeranYang:
Explain what this PR does.


pkg/forward/forward.go
```diff
@@ -110,19 +111,14 @@ func NewInterStepDataForward(vertex *dfv1.Vertex,
...
-		if isdf.opts.enableMapUdfStream {
-			return nil, fmt.Errorf("stream is not supported for source data transformer")
-		}
-	}
-
 	if isdf.opts.enableMapUdfStream && isdf.opts.readBatchSize != 1 {
 		return nil, fmt.Errorf("batch size is not 1 with UDF streaming")
 	}
 
+	if isdf.opts.vertexType == dfv1.VertexTypeSource {
```
vigith, line 118: will this happen? we are not exposing this to users so these kinds of accidents won't happen, right?

vigith (approved):

pkg/forward/forward.go
KeranYang: Right, this won't happen unless a developer passes in a source vertex type. My personal preference is to always add extra check, to be safe.

PR #888, opened 2023-07-21
chore: create separate source forwarder metrics instead of exposing existing ones

KeranYang:
Also update the rater such that for source vertex, rater uses the source metric to get read total count. Tested by creating a new pipeline locally and verified rate shows up at source vertex.


pkg/daemon/server/service/rater/pod_tracker.go
```diff
@@ -90,8 +90,10 @@ func (pt *PodTracker) Start(ctx context.Context) error {
 					var vType string
 					if v.IsReduceUDF() {
 						vType = "reduce"
+					} else if v.IsASource() {
+						vType = "source"
 					} else {
-						vType = "non_reduce"
+						vType = "other"
```
vigith, line 96: i think `non_reduce` is a good word, as it clears highlights the exclusion list without any ambiguity :)

vigith (approved):

pkg/daemon/server/service/rater/pod_tracker.go
KeranYang: I thought about rename it to `non_reduce_non_source` but it seems a bit too long to me. :)

PR #891, opened 2023-07-24
updated default version of Redis used for e2e

juliev0:
This was an oversight from back when the Redis version was updated that this was not updated in turn.


juliev0:
I have this in the Join Vertex PR but should take it out of there.

vigith (approved):

PR #901, opened 2023-07-28
feat: emit k8s events for controller messages. Fixes #856

### LINKED ISSUE #856: Emit Kubernetes events for important controller messages
# Summary

Today, when there's any issue with a pipeline reconciliation in the controller, we log it and also set the error messages in the pipeline or vertex status. This is good but not enough, sometimes ppl miss messages in the status, and hard for them to debug. 

It would be better to emit k8s messages as well when there's an issue as well as some important steps. For example:

- Fail to reconcile a pipeline due to any reason
- Fail to create a pod (due to problems like out of quota)
- Succeed to scale up/down a vertex
- xxxx

This kind of events will be helpful for users, and they will be displayed in tools like Argo CD.

---

dpadhiar:
Explain what this PR does.

Closes #856 

Emits Kubernetes events for pipeline reconciliation issues as well as for important steps. This includes scaling vertices, failing to create pods or failing to reconcile a pipeline for any situation. These events will be viewable on Argo CD for the user to be more clear than messages logged and set in the status of a pipeline/vertex.

### DIFF
--- a/pkg/reconciler/pipeline/controller.go
+++ b/pkg/reconciler/pipeline/controller.go
@@ -44,6 +44,7 @@ import (
 	"github.com/numaproj/numaflow/pkg/reconciler"
 	"github.com/numaproj/numaflow/pkg/shared/logging"
 	sharedutil "github.com/numaproj/numaflow/pkg/shared/util"
+	"k8s.io/client-go/tools/record"
 )
 
 const (
@@ -55,13 +56,14 @@ type pipelineReconciler struct {
 	client client.Client
 	scheme *runtime.Scheme
 
-	config *reconciler.GlobalConfig
-	image  string
-	logger *zap.SugaredLogger
+	config   *reconciler.GlobalConfig
+	image    string
+	logger   *zap.SugaredLogger
+	recorder record.EventRecorder
 }
 
-func NewReconciler(client client.Client, scheme *runtime.Scheme, config *reconciler.GlobalConfig, image string, logger *zap.SugaredLogger) reconcile.Reconciler {
-	return &pipelineReconciler{client: client, scheme: scheme, config: config, image: image, logger: logger}
+func NewReconciler(client client.Client, scheme *runtime.Scheme, config *reconciler.GlobalConfig, image string, logger *zap.SugaredLogger, recorder record.EventRecorder) reconcile.Reconciler {
+	return &pipelineReconciler{client: client, scheme: scheme, config: config, image: image, logger: logger, recorder: recorder}
 }
 
 func (r *pipelineReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
@@ -79,6 +81,7 @@ func (r *pipelineReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 	result, reconcileErr := r.reconcile(ctx, plCopy)
 	if reconcileErr != nil {
 		log.Errorw("Reconcile error", zap.Error(reconcileErr))
+		r.recorder.Event(pl, corev1.EventTypeWarning, "ReconcilePipelineFailed", reconcileErr.Error())
 	}
 	plCopy.Status.LastUpdated = metav1.Now()
 	if needsUpdate(pl, plCopy) {
@@ -777,6 +780,7 @@ func (r *pipelineReconciler) scaleVertex(ctx context.Context, pl *dfv1.Pipeline,
 				return false, err
 			}
 			log.Infow("Scaled vertex", zap.Int32("from", origin), zap.Int32("to", replicas), zap.String("vertex", vertex.Name))
+			r.recorder.Event(pl, corev1.EventTypeNormal, "ScalingVertex", fmt.Sprintf("Scaled vertex %s to %d", vertex.Name, replicas))
 			isVertexPatched = true
 		}
 	}
--- a/pkg/reconciler/pipeline/controller.go
+++ b/pkg/reconciler/pipeline/controller.go
@@ -81,7 +81,6 @@ func (r *pipelineReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 	result, reconcileErr := r.reconcile(ctx, plCopy)
 	if reconcileErr != nil {
 		log.Errorw("Reconcile error", zap.Error(reconcileErr))
-		r.recorder.Event(pl, corev1.EventTypeWarning, "ReconcilePipelineFailed", reconcileErr.Error())
 	}
 	plCopy.Status.LastUpdated = metav1.Now()
 	if needsUpdate(pl, plCopy) {
@@ -106,6 +105,7 @@ func (r *pipelineReconciler) reconcile(ctx context.Context, pl *dfv1.Pipeline) (
 				safeToDelete, err := r.safeToDelete(ctx, pl)
 				if err != nil {
 					log.Errorw("Failed to check if it's safe to delete the pipeline", zap.Error(err))
+					r.recorder.Event(pl, corev1.EventTypeWarning, "ReconcilePipelineFailed", "Failed to check if it's safe to delete the pipeline")
 					return ctrl.Result{}, err
 				}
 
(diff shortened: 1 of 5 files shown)


pkg/reconciler/pipeline/controller_test.go
```diff
@@ -144,6 +146,60 @@ func Test_reconcile(t *testing.T) {
...
+			logger:   zaptest.NewLogger(t).Sugar(),
+			recorder: record.NewFakeRecorder(64),
+		}
+		testObj := testPipeline.DeepCopy()
+		_, err = r.reconcile(ctx, testObj)
+		assert.NoError(t, err)
+		testObj.Spec.Vertices = append(testObj.Spec.Vertices, dfv1.AbstractVertex{Name: "input", Source: &dfv1.Source{}})
+		_, err = r.reconcile(ctx, testObj)
+		assert.Error(t, err)
+		events := getEvents(r, 1)
```
KeranYang, line 198: Can we also add a unit test that emit multiple events within on reconciliation? e.g. a normal event followed by a warning event?

pkg/reconciler/pipeline/controller.go
```diff
@@ -144,7 +149,11 @@ func (r *pipelineReconciler) reconcile(ctx context.Context, pl *dfv1.Pipeline) (
 	}
```
KeranYang, line 150: L143, I think it might also worth a normal event. `log.Infow("Updated pipeline phase", zap.String("originalPhase", string(oldPhase)), zap.String("currentPhase", string(pl.Status.Phase)))`

pkg/reconciler/pipeline/controller.go
```diff
@@ -144,7 +149,11 @@ func (r *pipelineReconciler) reconcile(ctx context.Context, pl *dfv1.Pipeline) (
...
 
 	// Regular pipeline update
-	return r.reconcileNonLifecycleChanges(ctx, pl)
+	result, err := r.reconcileNonLifecycleChanges(ctx, pl)
```
KeranYang, line 152: There is another occurrence of `reconcileNonLifecycleChanges` at L132

pkg/reconciler/vertex/controller.go
```diff
@@ -228,6 +230,7 @@ func (r *vertexReconciler) reconcile(ctx context.Context, vertex *dfv1.Vertex) (
 			if err := r.client.Create(ctx, pod); err != nil {
 				log.Errorw("Failed to create pod", zap.String("pod", pod.Name), zap.Error(err))
 				vertex.Status.MarkPhaseFailed("CreatePodFailed", err.Error())
+				r.recorder.Event(pod, corev1.EventTypeWarning, "CreatePodFailed", err.Error())
```
KeranYang, line 233: For vertex controller, I think for ALL occurrences of `vertex.Status.MarkPhaseFailed`, we can send a event.

It seems `log.Errorw`, `vertex.Status.MarkPhaseFailed`, `r.recorder.Event` and `return ctrl.Result{}, err` are a common sequence of operations, consider extract them into a single helper function.

pkg/reconciler/vertex/controller_test.go
```diff
@@ -169,7 +170,7 @@ func init() {
 
 func Test_NewReconciler(t *testing.T) {
```
KeranYang, line 171: Vertex controller unit tests and Pipeline controller are independent. We should also add unit tests for event here even though we have similar ones for pipeline.

pkg/reconciler/pipeline/controller.go
dpadhiar: Should this also include oldPhase to currentPhase in the message?
```
r.recorder.Event(pl, corev1.EventTypeNormal, "UpdatePipelinePhase", "Updated pipeline phase")
```

KeranYang: I think yes. It makes the event more detailed and easy for debugging.

pkg/reconciler/pipeline/controller.go
```diff
@@ -103,6 +105,7 @@ func (r *pipelineReconciler) reconcile(ctx context.Context, pl *dfv1.Pipeline) (
 				safeToDelete, err := r.safeToDelete(ctx, pl)
 				if err != nil {
 					log.Errorw("Failed to check if it's safe to delete the pipeline", zap.Error(err))
+					r.recorder.Event(pl, corev1.EventTypeWarning, "ReconcilePipelineFailed", "Failed to check if it's safe to delete the pipeline")
 					return ctrl.Result{}, err
```
juliev0, line 109: I see the underlying errors aren't being included in the message - maybe it would be helpful to include them? (unless there's some reason we don't want the message to be too verbose) If I glance at the use of Events in the Argo Workflows code, at least in certain places I see we are including a full error.

dpadhiar: This is true, there is a mix of it in Argo Workflows. There is a fear that the events would be too verbose - these are supposed to be readable.

@KeranYang thoughts?

juliev0: for the sake of eliminating redundancy, you could consider doing an `fmt.Sprintf()` to format the string which becomes both the error to log and the event message to emit

dpadhiar: There is also the event function that does formatting:
```
// example
r.recorder.Eventf(pl, corev1.EventTypeWarning, "ReconcilePipelineFailed", "Failed to check if it's safe to delete the pipeline: %v", err)
```

pkg/reconciler/pipeline/controller_test.go
dpadhiar: Currently, there would need to be multiple reconciliation calls now to get a normal event followed by a warning event as when the NormalEvent for updating a pipeline phase occurs, there is a return statement. Otherwise, we do not hit the check for any nonLifeCycleChanges that would cause a warning.

pkg/reconciler/vertex/controller.go
dpadhiar: Helper function:
```
func (r *vertexReconciler) markPhaseLogEvent(vertex *dfv1.Vertex, log *zap.SugaredLogger, reason, message, logMsg string, logWith ...interface{}) {
	log.Errorw(logMsg, logWith)
	vertex.Status.MarkPhaseFailed(reason, message)
	r.recorder.Event(vertex, corev1.EventTypeWarning, reason, message)
}
```

Condensed logic vs with helper:
```
r.markPhaseLogEvent(vertex, log, "FindISBSvcFailed", err.Error(), "Failed to get ISB Service", zap.String("isbsvc", isbSvcName), zap.Error(err))
// log.Errorw("Failed to get ISB Service", zap.String("isbsvc", isbSvcName), zap.Error(err))
// vertex.Status.MarkPhaseFailed("FindISBSvcFailed", err.Error())
// r.recorder.Event(vertex, corev1.EventTypeWarning, "FindISBSvcFailed", err.Error())
```

pkg/reconciler/pipeline/controller.go
KeranYang: I am ok with including the full error.

pkg/reconciler/pipeline/controller.go
```diff
@@ -44,6 +44,7 @@ import (
 	"github.com/numaproj/numaflow/pkg/reconciler"
 	"github.com/numaproj/numaflow/pkg/shared/logging"
 	sharedutil "github.com/numaproj/numaflow/pkg/shared/util"
+	"k8s.io/client-go/tools/record"
```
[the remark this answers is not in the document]
KeranYang: Please sort the imports properly.

KeranYang (approved):

pkg/reconciler/pipeline/controller.go
juliev0: I was thinking we could remove the redundancy of writing out the error message twice? Like:
`fmt.Sprintf("Failed to check if it's safe to delete the pipeline: %v", err.Error())` and then apply it to both the log and the event?

KeranYang: +1

KeranYang (approved):

PR #910, opened 2023-08-01
feat: watermark delay in tooltip

### LINKED ISSUE #876: UX to display relative time of watermark to current time
It is very difficult with the timestamp in UTC to figure out how old is the watermark propagation. It will be great if, on mouse over, we can also see the relative time too (e.g., 60sec, 2min, 5h2min, etc.). It is implicit that the timestamp of the watermark will be behind `now()`, so we do not have to make it signed `+` or `-`.

veds-g:
Fixes #876 

- Shows delay of watermark wrt to `time.now()` when hovering on the edge label
- Shows difference in precision from years (~ 5yr 4mo) up to milliseconds (~ 549ms)

image
image
image

### DIFF
--- a/ui/src/components/pages/Pipeline/partials/Graph/partials/CustomEdge/index.tsx
+++ b/ui/src/components/pages/Pipeline/partials/Graph/partials/CustomEdge/index.tsx
@@ -50,6 +50,51 @@ const CustomEdge: FC<EdgeProps> = ({
       labelY = targetY + 2;
     }
   }
+  const getMinWM = () => {
+    if (data?.edgeWatermark?.watermarks) {
+      return Math.min.apply(null, data?.edgeWatermark?.watermarks);
+    } else {
+      return -1;
+    }
+  };
+
+  const getDelay = () => {
+    const str = "Delay - ";
+    const fetchWMTime = data?.edgeWatermark?.WMFetchTime;
+    const minWM = getMinWM();
+    let difference =
+      (fetchWMTime ? fetchWMTime : Date.now()) - (minWM == -1 ? 0 : minWM);
+
+    const years = Math.floor(difference / (1000 * 60 * 60 * 24 * 365));
+    difference -= years * (1000 * 60 * 60 * 24 * 365);
+    const months = Math.floor(difference / (1000 * 60 * 60 * 24 * 30));
+    difference -= months * (1000 * 60 * 60 * 24 * 30);
+    const days = Math.floor(difference / (1000 * 60 * 60 * 24));
+    difference -= days * (1000 * 60 * 60 * 24);
+    const hours = Math.floor(difference / (1000 * 60 * 60));
+    difference -= hours * (1000 * 60 * 60);
+    const minutes = Math.floor(difference / (1000 * 60));
+    difference -= minutes * (1000 * 60);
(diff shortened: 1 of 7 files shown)


yhl25 (approved):

ui/src/components/pages/Pipeline/partials/Graph/partials/CustomEdge/index.tsx
```diff
@@ -50,6 +50,51 @@ const CustomEdge: FC<EdgeProps> = ({
...
+    const str = "Delay - ";
+    const fetchWMTime = data?.edgeWatermark?.WMFetchTime;
+    const minWM = getMinWM();
+    let difference =
+      (fetchWMTime ? fetchWMTime : Date.now()) - (minWM == -1 ? 0 : minWM);
+
+    const years = Math.floor(difference / (1000 * 60 * 60 * 24 * 365));
+    difference -= years * (1000 * 60 * 60 * 24 * 365);
+    const months = Math.floor(difference / (1000 * 60 * 60 * 24 * 30));
+    difference -= months * (1000 * 60 * 60 * 24 * 30);
```
KeranYang, line 71: Using 30 days as a month is not always precise. Consider using a third-party library like `moment` to calculate time. Something like below:

```
const startTime = minWM === -1 ? 0 : minWM;
const endTime = fetchWMTime || Date.now();

const duration = moment.duration(moment(endTime).diff(moment(startTime)));
const years = duration.years();
const months = duration.months();
...
```

KeranYang (changes requested):

ui/src/components/pages/Pipeline/partials/Graph/partials/CustomEdge/index.tsx
veds-g: It's not precise but third-party libraries also use a constant for number of days in a year, ...
for example here using moment js, the difference between same dates over `leap-year and non-leap` & `two non-leaps years` is not constant ~ (library considers 366 days / yr)
image

should we use the current implementation or switch to this library instead?

vigith:
Should we use the word `old` rather than `delay`? I am concerned whether it will confuse the users. The problem with `old` and `delay` is that it is relative and there is a reference point, and hence open to interpretation. 

Any suggestions?

KeranYang:
> Should we use the word old rather than delay? I am concerned whether it will confuse the users. The problem with old and delay is that it is relative and there is a reference point, and hence open to interpretation. Any suggestions?

`555ms behind now`?

ui/src/components/pages/Pipeline/partials/Graph/partials/CustomEdge/index.tsx
```diff
@@ -52,47 +53,43 @@ const CustomEdge: FC<EdgeProps> = ({
   }
   const getMinWM = () => {
     if (data?.edgeWatermark?.watermarks) {
+      return Date.now() - 678;
```
KeranYang, line 56: Is this line for test? can we remove it?

veds-g: yes, missed cleaning that.

KeranYang (approved):

mshakira (approved):

PR #920, opened 2023-08-05
Pipelines with Cycles: e2e testing, and pipeline validation

juliev0:
Fixes: #934 

We will allow Cycles in Pipelines as long as there's no Reduce Vertex at the point of the Cycle or to the right of it (to avoid issues of late data after close-of-book).

This PR includes:
- e2e tests of valid Cycles
- Validation to prevent invalid Cycles

As a side note, I have observed how the cycle looks in the UI, and I believe this does need modification (i.e. if Vertex A sends to Vertex B, but Vertex B cycles back to A, there should probably be a second line (edge); and if Vertex B cycles to itself, that probably needs a line (edge))


juliev0:
I think this is failing because maybe the image for [this](https://github.com/numaproj/numaflow-go/tree/main/pkg/function/examples/retry) hasn't been published to quay.io?

test/e2e-suite-2/testdata/cycle-backward.yaml
```diff
@@ -0,0 +1,43 @@
...
+  vertices:
+    - name: in
+      source:
+        http: {}
+    - name: cat
+      udf:
+        builtin:
+          name: cat
+    - name: retry
+      replicas: 1
```
juliev0, line 15: actually, based on conversation with Derek I need to remove the `replicas` field as it's not part of this spec. My intention was to turn autoscaling off so need to do that the right way

examples/11-cycle-to-prev.yaml
vigith, line 1: can you rename to 10-cycle-to-prev? so both 10s are for cycle?

pkg/reconciler/pipeline/validate.go
```diff
@@ -308,3 +310,145 @@ func isReservedContainerName(name string) bool {
...
+		}
+	}
+
+	return nil
+}
+
+// getCycles locates the vertices where there's a Cycle, if any
+// eg. if A->B->A, then return A
+// Since there are multiple Sources, and since each Source produces a Tree, then we can return multiple Cycles
+func getCycles(pipelineSpec *dfv1.PipelineSpec) (map[string]struct{}, error) {
```
vigith, line 350: NIT: what will be the error for this func? is it only "no vertex found", isn't it already covered in 320?

examples/11-cycle-to-prev.yaml
juliev0: done. I didn't realize how the numbering worked before

vigith (approved):

pkg/reconciler/pipeline/validate.go
juliev0: Yes. Well, I suppose there's nothing to prevent getCycles() from being called from somewhere else, so for completeness I think I'll leave it in.

PR #937, opened 2023-08-08
feat: add vertex template to pipeline spec

### LINKED ISSUE #878: Be able to add label to every vertex in a pipeline by specifying it in a single place
# Summary

It would be good to be able to add a label in a single place for it to be picked up as a label in all the vertices within the pipeline.

# Use Cases

We have a label for splunk index, also another label which we use as a selector for prometheus metrics that is required in all vertices, currently we add it as a label individually to every vertex, but it would be good to be able to add the label in a single place like the pipeline labels (with an optional tag or something similar) that would make the label a part of every vertex.

---

dpadhiar:
Explain what this PR does.

Fixes #878


pkg/reconciler/vertex/controller.go
```diff
@@ -186,6 +186,24 @@ func (r *vertexReconciler) reconcile(ctx context.Context, vertex *dfv1.Vertex) (
 		if needToCreate {
 			labels := map[string]string{}
 			annotations := map[string]string{}
+			if pipeline.Spec.Templates != nil && pipeline.Spec.Templates.VertexTemplate != nil {
+				apt := pipeline.Spec.Templates.VertexTemplate.AbstractPodTemplate
+				apt.ApplyToPodSpec(podSpec)
+				if apt.Metadata != nil && len(apt.Metadata.Labels) > 0 {
+					for k, v := range apt.Metadata.Labels {
+						if _, ok := labels[k]; !ok {
```
whynowy, line 194: The existence check is not needed.

pkg/reconciler/vertex/controller.go
```diff
@@ -186,6 +186,24 @@ func (r *vertexReconciler) reconcile(ctx context.Context, vertex *dfv1.Vertex) (
...
+							labels[k] = v
+						}
+					}
+				}
+				if apt.Metadata != nil && len(apt.Metadata.Annotations) > 0 {
+					for k, v := range apt.Metadata.Annotations {
+						if _, ok := annotations[k]; !ok {
```
whynowy, line 201: Same here.

pkg/reconciler/vertex/controller.go
```diff
@@ -328,6 +346,14 @@ func (r *vertexReconciler) buildPodSpec(vertex *dfv1.Vertex, pl *dfv1.Pipeline,
 		return nil, fmt.Errorf("failed to generate pod spec, error: %w", err)
 	}
 
+	if pl.Spec.Templates != nil && pl.Spec.Templates.VertexTemplate.ContainerTemplate != nil {
```
whynowy, line 349: `VertexTemplate` nil check is missing.

pkg/reconciler/vertex/controller.go
```diff
@@ -328,6 +346,14 @@ func (r *vertexReconciler) buildPodSpec(vertex *dfv1.Vertex, pl *dfv1.Pipeline,
...
+		pl.Spec.Templates.VertexTemplate.ContainerTemplate.ApplyToNumaflowContainers(podSpec.Containers)
+	}
+
+	if pl.Spec.Templates != nil && pl.Spec.Templates.VertexTemplate.InitContainerTemplate != nil {
```
whynowy, line 353: Same here.

whynowy:
Duplicated and raised another one https://github.com/numaproj/numaflow/pull/947.

PR #939, opened 2023-08-09
fix: retry logic for fetching last updated kv time

yhl25:
if there are no keys in the store we should not retry while trying to calculate the last updated kv time.


KeranYang (approved):

pkg/shared/kvs/jetstream/kv_watch.go
```diff
@@ -205,15 +207,37 @@ func (jsw *jetStreamWatch) newWatcher(ctx context.Context) nats.KeyWatcher {
 
 // lastUpdateKVTime returns the last update time of the kv store
```
vigith, line 208:
```suggestion
// lastUpdateKVTime returns the last update time of the kv store. if the key is missing, it will return time.Zero.
```

vigith (approved):

PR #947, opened 2023-08-10
feat: add vertex template to pipeline spec

### LINKED ISSUE #878: Be able to add label to every vertex in a pipeline by specifying it in a single place

whynowy:
Explain what this PR does.

Fixes #878.

This PR introduces `vertex` for `pipeline.spec.templates` which allows a user to define `VertexTemplate` from the pipeline spec. This includes `AbstractPodTemplate` to include annotations and labels in the metadata and `Container` and `InitContainer` templates as well.

This change will allow users the option to add labels to all vertices of a pipeline for selectors. Note that templates on a vertex will override this spec as expected.

Example of vertex template:
```
  templates:
    vertex:
        metadata:
          labels:
            numaflow.numaproj.io/prometheus: "test"
```


vigith:
Should we add user doc for this?

whynowy:
> Should we add user doc for this?

just added.

docs/user-guide/reference/configuration/pipeline-customization.md
```diff
@@ -116,3 +116,67 @@ spec:
           limits:
             memory: 500Mi
 ```
+
+## Vertices
+
+The following example shows how to configure the all the vertex pods owned by a pipeline with all currently supported fields. Be aware these configurations applied to all vertex pods can be overridden by the vertex configuration.
```
vigith, line 122: can we provide a hyperlink for `vertex configuration`?

vigith (approved):

PR #949, opened 2023-08-11
chore: use partition name while publishing watermark from source

### LINKED ISSUE #890: watermark stops progressing when one of the input pods fails to read messages.
If an input pod stops reading messages, the watermark stops progressing, which results in the windows not closing.

yhl25:
- use partition instead of pod name as processing entity while publishing watermark inside source
- add partition information to offset
- introduced new structs `kafkaOffset`, `redisOffset`, `simplePartitionStringOffset` and `simplePartitionIntOffset` to avoid slicing the string to get offset and partition
- we don't have to explicitly add partition information to msg-id for dedup, we can achieve it using offset.String() method, since it will already have the partition information
- source data forward should close the publishers since its created by them (TODO as part of wm refactor we might have to introduce a separate interface for source wm publishers and fetchers, will tackle it in upcoming refactor prs)

fixes #890


pkg/sources/generator/tickgen.go
```diff
@@ -389,18 +389,19 @@ func (mg *memgen) generator(ctx context.Context, rate int, timeunit time.Duratio
 }
 
 func (mg *memgen) newReadMessage(key string, payload []byte, offset int64) *isb.ReadMessage {
+	readOffset := util.NewSimpleIntPartitionOffset(offset, mg.vertexInstance.Replica)
```
KeranYang, line 392: It seems when a source doesn't naturally support partitioning, we use the `vertexInstance.Replica` as the partition index. Is the Replica mapped to a pod in the vertex? I thought we are moving from using the pod as a processing unit to using the partition.

Or the Replica isn't really a partition index and we are just using it here to build the unique ID? If so, would it be better if we use the following statement?

```
readOffset := util.NewSimpleIntPartitionOffset(strconv.FormatInt(offset, 10) + "-" + strconv.FormatInt(int64(mg.vertexInstance.Replica), 10), 0) // 0 is the partition index
```

pkg/sources/source.go
```diff
@@ -67,9 +66,14 @@ func (sp *SourceProcessor) Start(ctx context.Context) error {
 	defer natsClientPool.CloseAll()
 	// watermark variables no-op initialization
 	// publishWatermark is a map representing a progressor per edge, we are initializing them to a no-op progressor
```
KeranYang, line 68: nit: please update comments

KeranYang (commented):
I took a peek at this draft PR since it's quite related to the UDSource and left some early comments. Please take a look. Thanks. :)

pkg/sources/forward/data_forward.go
```diff
@@ -53,8 +55,9 @@ type DataForward struct {
 	toWhichStepDecider forward.ToWhichStepDecider
 	transformer        applier.MapApplier
 	wmFetcher          fetch.Fetcher
+	toVertexWMStores   map[string]store.WatermarkStore
```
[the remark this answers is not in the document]
yhl25: we create the publisher based on the data, it's not created when we create the forward.

pkg/sources/forward/data_forward.go
```diff
@@ -269,6 +294,7 @@ func (isdf *DataForward) forwardAChunk(ctx context.Context) {
 		}
 	}
 
+	var sourcePartitions = make([]int32, 0)
```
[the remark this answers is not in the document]
yhl25: done

pkg/sources/forward/data_forward.go
```diff
@@ -557,6 +602,18 @@ func (isdf *DataForward) whereToStep(writeMessage *isb.WriteMessage, messageToSt
 	return nil
 }
 
+// createToVertexWatermarkPublisher creates a watermark publisher for the given toVertexName and partition
+func (isdf *DataForward) createToVertexWatermarkPublisher(toVertexName string, partition int32) publish.Publisher {
+
+	wmStore := isdf.toVertexWMStores[toVertexName]
```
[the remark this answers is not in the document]
yhl25: same comment

pkg/sources/generator/tickgen.go
vigith: we need either the source partition or the replica id. either of the ones is required for the correctness of the watermark.

pkg/shared/simpleoffset/simpleoffset.go
```diff
@@ -0,0 +1 @@
+package simpleoffset
```
whynowy, line 1: What is this used for?

pkg/isb/interfaces.go
```diff
@@ -27,6 +27,8 @@ import (
 	"io"
 	"math"
 	"strconv"
+
+	"github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
```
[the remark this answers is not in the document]
yhl25: only constants are stored in that package, so should be fine.

pkg/shared/util/poffset.go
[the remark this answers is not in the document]
yhl25: I moved it to isb

pkg/shared/clients/redis/redis_reader.go
```diff
@@ -140,10 +141,12 @@ func (br *RedisStreamsRead) Ack(_ context.Context, offsets []isb.Offset) []error
 	dedupOffsets := make(map[string]struct{}) // essentially a Set
 	strOffsets := []string{}
 	for _, o := range offsets {
-		_, found := dedupOffsets[o.String()]
+		// for redis, we don't consider partition id to ack
+		ofs := strings.Split(o.String(), "-")[0]
```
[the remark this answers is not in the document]
yhl25: `.String()` has the partition information as well, but for redis we only need seq to acknowledge the message

pkg/sources/forward/data_forward.go
```diff
@@ -289,6 +315,8 @@ func (isdf *DataForward) forwardAChunk(ctx context.Context) {
 				return
 			}
 		}
+		// get the list of source partitions for which we have read messages, we will use this to publish watermarks to toVertices
+		sourcePartitionsIndices = append(sourcePartitionsIndices, m.readMessage.ReadOffset.PartitionIdx())
```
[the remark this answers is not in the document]
yhl25: yes it should be a set, fixed it.

pkg/apis/numaflow/v1alpha1/const.go
```diff
@@ -164,6 +164,9 @@ const (
 
 	// UDF map streaming
 	MapUdfStreamKey = "numaflow.numaproj.io/map-stream"
+
+	// DefaultPartitionIdx Default partition index
+	DefaultPartitionIdx = 0
```
whynowy, line 169: Move it to `pkg/isb/interfaces.go`?

pkg/shared/simpleoffset/simpleoffset.go
yhl25: I forgot to delete it, removed.

pkg/apis/numaflow/v1alpha1/const.go
yhl25: done

pkg/sources/generator/tickgen.go
KeranYang: > we need either the source partition or the replica id.

Yes, I agree, but I thought this PR is to choose partition over replica explicitly.

Sorry, I have two questions:

1. Is this PR to move away from using the pod name to create a processing entity? If so, does it imply that the pod name is not needed in the read offset, hence Replica is not needed?

2. If we translate the `NewSimpleIntPartitionOffset` statement to native Golang initialization syntax, it will look like below:

```
&simpleIntPartitionOffset{
		seq:          offset,
		partitionIdx: mg.vertexInstance.Replica,
	}
```

Does `mg.vertexInstance.Replica` represent the partition index? e.g. for HTTP source, should partitionIdx always be 0?

vigith: > Yes, I agree, but I thought this PR is to choose partition over replica explicitly.

This PR is to make sure we use the same "processing unit" we use for the first publish between the source and the source vertex, between the source vertex and the next vertex. Earlier, we were publishing the source vertex's offset timeline only based on pod even though the source supported a different sharding/partitioning scheme.


> Sorry, I have two questions:
> 
> 1. Is this PR to move away from using the pod name to create a processing entity? If so, does it imply that the pod name is not needed in the read offset, hence Replica is not needed?

Making sure the source vertex publishes the offset using the same sharding/partitioning semantics that the source provides.

> 2. If we translate the `NewSimpleIntPartitionOffset` statement to native Golang initialization syntax, it will look like below:
> 
> ```
> &simpleIntPartitionOffset{
> 		seq:          offset,
> 		partitionIdx: mg.vertexInstance.Replica,
> 	}
> ```
> 
> Does `mg.vertexInstance.Replica` represent the partition index? e.g. for HTTP source, should partitionIdx always be 0?

Yes.

pkg/forward/forward.go
```diff
@@ -447,9 +447,8 @@ func (isdf *InterStepDataForward) streamMessage(
 		// to send the result to. Then update the toBuffer(s) with writeMessage.
 		msgIndex := 0
 		for writeMessage := range writeMessageCh {
-			// add partition to the ID, to make sure that the ID is unique across partitions
 			// also add vertex name to the ID, since multiple vertices can publish to the same vertex and we need uniqueness across them
```
juliev0, line 450: this comment now sounds a little weird starting with "also" given that the previous comment was removed - maybe the comments here can say what each of the components are? (same issue on line 672 below)

pkg/sources/forward/data_forward.go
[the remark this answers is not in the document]
vigith: we need to think about it from an ownership point of view. we will have to tackle it in the near future.

pkg/sources/kafka/reader.go
```diff
@@ -445,37 +485,30 @@ func (r *KafkaSource) startConsumer() {
...
-	tokens := strings.Split(offset, ":")
-	if len(tokens) < 3 {
-		return "", 0, 0, errors.New("malformed offset")
-	}
-	var poffset, partition int64
-	var err error
-	poffset, err = strconv.ParseInt(tokens[2], 10, 64)
-	if err == nil {
-		partition, err = strconv.ParseInt(tokens[1], 10, 32)
+func extractTopicFromOffset(offset isb.Offset) (string, error) {
```
vigith, line 508: can't we just get it from kafka.Offset? just by type casting? why do we need string manipulation?

pkg/watermark/publish/publisher_inmem_test.go
```diff
@@ -23,13 +23,13 @@ import (
 	"testing"
 	"time"
 
+	"github.com/numaproj/numaflow/pkg/isb"
```
vigith, line 26: fix sorting?

pkg/sources/generator/tickgen.go
KeranYang: I see. Thanks for the clarification. Resolving this comment.

pkg/forward/forward.go
yhl25: done

pkg/watermark/publish/publisher_inmem_test.go
yhl25: done

pkg/sources/kafka/reader.go
vigith: delete this func then `extractTopicFromOffset`?

pkg/watermark/publish/publisher_test.go
```diff
@@ -23,12 +23,12 @@ import (
 	"time"
 
 	"github.com/nats-io/nats.go"
+	"github.com/numaproj/numaflow/pkg/isb"
```
vigith, line 26: wrong sorting

vigith (approved):

PR #950, opened 2023-08-11
feat: Fetcher refactor

jy4096:
Draft PR still WIP

### DIFF
--- a/pkg/daemon/server/daemon_server.go
+++ b/pkg/daemon/server/daemon_server.go
@@ -146,7 +146,7 @@ func (ds *daemonServer) Run(ctx context.Context) error {
 
 func (ds *daemonServer) newGRPCServer(
 	isbSvcClient isbsvc.ISBService,
-	wmFetchers map[v1alpha1.Edge][]fetch.Fetcher,
+	wmFetchers map[v1alpha1.Edge][]fetch.UXFetcher,
 	rater server.Ratable) (*grpc.Server, error) {
 	// "Prometheus histograms are a great way to measure latency distributions of your RPCs.
 	// However, since it is a bad practice to have metrics of high cardinality the latency monitoring metrics are disabled by default.
--- a/pkg/isbsvc/jetstream_service.go
+++ b/pkg/isbsvc/jetstream_service.go
@@ -24,14 +24,15 @@ import (
 	"time"
 
 	"github.com/nats-io/nats.go"
+	"github.com/spf13/viper"
+	"go.uber.org/zap"
+
 	jsclient "github.com/numaproj/numaflow/pkg/shared/clients/nats"
 	"github.com/numaproj/numaflow/pkg/shared/kvs/jetstream"
 	"github.com/numaproj/numaflow/pkg/shared/logging"
 	"github.com/numaproj/numaflow/pkg/watermark/fetch"
 	"github.com/numaproj/numaflow/pkg/watermark/processor"
 	"github.com/numaproj/numaflow/pkg/watermark/store"
-	"github.com/spf13/viper"
-	"go.uber.org/zap"
 )
 
 type jetStreamSvc struct {
@@ -140,7 +141,7 @@ func (jss *jetStreamSvc) CreateBuffersAndBuckets(ctx context.Context, buffers, b
 			}
 			log.Infow("Succeeded to create a consumer for a stream", zap.String("stream", streamName), zap.String("consumer", streamName))
 		}
-		//TODO: remove sleep and use a better way to wait for the stream to be ready
+		// TODO: remove sleep and use a better way to wait for the stream to be ready
 		time.Sleep(3 * time.Second)
 	}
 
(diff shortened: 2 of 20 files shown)


pkg/reduce/data_forward_test.go
```diff
@@ -348,7 +348,7 @@ func TestDataForward_StartWithNoOpWM(t *testing.T) {
 
 // ReadMessage size = 0
 func TestReduceDataForward_IdleWM(t *testing.T) {
-	//FIXME: fix this test
+	// FIXME: fix this test
```
jy4096, line 351: @yhl25 Do you remember what the bug is?... I ran this locally but didn't get any error..

pkg/isbsvc/jetstream_service.go
```diff
@@ -315,8 +316,8 @@ func (jss *jetStreamSvc) GetBufferInfo(ctx context.Context, buffer string) (*Buf
 	return bufferInfo, nil
 }
```
vigith, line 318:
```suggestion
// CreateUXWatermarkFetcher is used to create watermark fetcher for the given bucket
```

pkg/watermark/fetch/edge_fetcher.go
```diff
@@ -126,18 +112,18 @@ func (e *edgeFetcher) updateWatermark(inputOffset isb.Offset, fromPartitionIdx i
...
+// ComputeHeadWatermark returns the smallest head watermark among all the processors for given partition
 // This can be used in showing the watermark
 // progression for a vertex when not consuming the messages directly (eg. UX, tests)
 // NOTE
 //   - We don't use this function in the regular pods in the vertex.
-//   - UX only uses GetHeadWatermark, so the `p.IsDeleted()` check in updateWatermark() never happens.
+//   - UX only uses ComputeHeadWatermark, so the `p.IsDeleted()` check in updateWatermark() never happens.
 //     Meaning, in the UX (daemon service) we never delete any processor.
-func (e *edgeFetcher) GetHeadWatermark(fromPartitionIdx int32) wmb.Watermark {
+func (e *edgeFetcher) ComputeHeadWatermark(fromPartitionIdx int32) wmb.Watermark {
```
vigith, line 126:
```suggestion
func (e *edgeFetcher) ComputeHeadWatermarkForUX(fromPartitionIdx int32) wmb.Watermark {
```

since this is used only by UX?

pkg/watermark/fetch/edge_fetcher_set.go
```diff
@@ -22,20 +22,37 @@ import (
...
 
-func NewEdgeFetcherSet(ctx context.Context, edgeFetchers map[string]Fetcher) Fetcher {
+// NewEdgeFetcherSet creates a new edgeFetcherSet object which implements the Fetcher interface.
+func NewEdgeFetcherSet(ctx context.Context, vertexInstance *dfv1.VertexInstance, processorManagers map[string]*processor.ProcessorManager) Fetcher {
+	var edgeFetchers = make(map[string]*edgeFetcher)
+	for key, processorManager := range processorManagers {
+		var fetchWatermark *edgeFetcher
+		// create a fetcher that fetches watermark.
+		if vertexInstance.Vertex.IsASource() {
+			// skip: source vertex is handled using new source fetcher
```
vigith, line 48: panic because this shouldn't happen?

vigith (approved):

pkg/watermark/fetch/edge_fetcher.go
jy4096: hmmmm the function itself is a method of the UXFetcher interface... so I didn't add the UX to the function name
```
// UXFetcher computes the watermark for Vn.
type UXFetcher interface {
	...
	// ComputeHeadWatermark computes a valid head watermark for the given partition
	ComputeHeadWatermark(fromPartitionIdx int32) wmb.Watermark
}
```

pkg/daemon/server/daemon_server.go
```diff
@@ -86,7 +86,7 @@ func (ds *daemonServer) Run(ctx context.Context) error {
 	default:
 		return fmt.Errorf("unsupported isbsvc buffer type %q", ds.isbSvcType)
 	}
-	wmFetchers, err := service.GetEdgeWatermarkFetchers(ctx, ds.pipeline, isbSvcClient)
+	wmFetchers, err := service.GetUXEdgeWatermarkFetchers(ctx, ds.pipeline, isbSvcClient)
```
whynowy, line 89: Can we get `UX` out of the function name? It limits the usage of the function only for UX purpose.

whynowy: What I meant was, use a different word from `UX`, otherwise it will be tricky if somehow some other services need to use it in the future.

vigith: should we be exposing it for other use-cases? as of now only UX uses it and ideally nobody should be using it.

whynowy: For example, we write a CLI someday, it displays the status of a pipeline, which includes the watermark of each edge, then it most probably uses the same interface/function.

vigith: we can rename it then? the only concern I have is that people might misuse it without knowing the side effects.

jy4096: Maybe we can rename later if we are to add new use cases?... Can't think of a proper name for UXFetcher and related methods/functions............. `Fetcher` and `SimpleFetcher` ?..

PR #956, opened 2023-08-16
chore: define vertex spec for user define source

KeranYang:
Also re-arrange our e2e test suites to avoid CI workflow time out.


pkg/reconciler/pipeline/validate.go
```diff
@@ -89,6 +89,17 @@ func ValidatePipeline(pl *dfv1.Pipeline) error {
 		return fmt.Errorf("pipeline has no sink, at least one vertex with 'sink' defined is required")
 	}
 
+	for k, s := range sources {
+		if s.IsUDSource() {
+			if s.Source.UDSource.Container == nil || s.Source.UDSource.Container.Image == "" {
+				return fmt.Errorf("invalid user-defined source vertex %q, a customized image is required", k)
+			}
+			if s.Source.HTTP != nil || s.Source.Kafka != nil || s.Source.Nats != nil || s.Source.RedisStreams != nil || s.Source.Generator != nil {
```
vigith, line 97: can you add a TODO to move `redis`, `nats`, to user-defined-source, and not in the core?

vigith (approved):

pkg/apis/numaflow/v1alpha1/source.go
```diff
@@ -108,3 +113,39 @@ func (s Source) getUDTransformerContainer(mainContainerReq getContainerReq) core
 	}
 	return container
 }
+
+func (s Source) getUDSourceContainer(mainContainerReq getContainerReq) corev1.Container {
+	c := containerBuilder{}.
+		name(CtrUdsource).
+		imagePullPolicy(mainContainerReq.imagePullPolicy). // Use the same image pull policy as the main container
+		appendVolumeMounts(mainContainerReq.volumeMounts...)
```
whynowy, line 121: Should revisit this.

pkg/apis/numaflow/v1alpha1/source.go
```diff
@@ -108,3 +113,39 @@ func (s Source) getUDTransformerContainer(mainContainerReq getContainerReq) core
...
+		c = c.appendEnv(x.Env...).appendVolumeMounts(x.VolumeMounts...).resources(x.Resources).securityContext(x.SecurityContext).appendEnvFrom(x.EnvFrom...)
+		if x.ImagePullPolicy != nil {
+			c = c.imagePullPolicy(*x.ImagePullPolicy)
+		}
+	}
+	container := c.build()
+	container.LivenessProbe = &corev1.Probe{
+		ProbeHandler: corev1.ProbeHandler{
+			HTTPGet: &corev1.HTTPGetAction{
+				Path:   "/sidecar-livez",
```
whynowy, line 141: Do we have the logic change in `/sidecar-livez`?

pkg/reconciler/pipeline/validate.go
KeranYang: Could you elaborate more on why?

vigith: i want to move those away from our core numaflow code; our core code is getting bloated. i would like to do that with UDSource release (0.11).

pkg/apis/numaflow/v1alpha1/source.go
KeranYang: synced offline, I will address this comment in follow-up PRs.

whynowy (approved):

PR #964, opened 2023-08-18
chore: upgrade k8s api to v0.26

whynowy:
Explain what this PR does.


vigith (approved):

config/base/crds/full/numaflow.numaproj.io_interstepbufferservices.yaml
```diff
@@ -489,6 +489,18 @@ spec:
                         type: string
                       resources:
                         properties:
+                          claims:
```
vigith, line 492: what is this claims thingy?

whynowy: New optional property under `resource`.

```
	// Claims lists the names of resources, defined in spec.resourceClaims,
	// that are used by this container.
	//
	// This is an alpha field and requires enabling the
	// DynamicResourceAllocation feature gate.
	//
	// This field is immutable. It can only be set for containers.
	//
	// +listType=map
	// +listMapKey=name
	// +featureGate=DynamicResourceAllocation
	// +optional
```

PR #979, opened 2023-08-23
fix: send keys for udsink

### LINKED ISSUE #978: sinksdk Datum.Keys() is returning empty string list
**Describe the bug**
I have two sinks, built-in log sink and one udsink. When I log Datum.Keys() in udsink it is an empty list, but the previous vertex is sending output with Keys. Also, I can see Keys in the log sink output. 
So what is the reason for Datum.Keys to be empty for udsink

**To Reproduce**

1. Implement a UD Sink with previous vertex sending data with Keys

**Expected behavior**
Keys should have the expected keys list passed on from previous vertex 

**Screenshots**

Using numaflow-go sdk v0.4.5

- Kubernetes: [e.g. v1.18.6]
- Numaflow: [e.g. v0.5.1]

**Additional context**

---

yhl25:
fixes #978

### DIFF
--- a/pkg/sinks/udsink/sink.go
+++ b/pkg/sinks/udsink/sink.go
@@ -111,6 +111,7 @@ func (s *UserDefinedSink) Write(ctx context.Context, messages []isb.Message) ([]
 			// NOTE: key is not used anywhere ATM
 			Id:        m.ID,
 			Value:     m.Payload,
+			Keys:      m.Keys,
 			EventTime: timestamppb.New(m.EventTime),
 			// Watermark is only available in readmessage....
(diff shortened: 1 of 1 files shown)


pkg/sinks/udsink/sink.go
```diff
@@ -111,6 +111,7 @@ func (s *UserDefinedSink) Write(ctx context.Context, messages []isb.Message) ([]
 			// NOTE: key is not used anywhere ATM
```
KeranYang, line 111: Does this statement still hold true? 🤔

KeranYang (approved):

pkg/sinks/udsink/sink.go
yhl25: removed

PR #980, opened 2023-08-23
feat: implement user-defined source

### LINKED ISSUE #928: UDSource - Implement UDSource READ, ACK and Pending under numaflow repo

KeranYang:
### Description
* Implement gRPC on the platform side to enable calling user-defined source APIs.
* Enable the simple E2E test which verifies that the simple source image can generate data and the data reaches sink.
This PR completes the functionality of the user-defined source.

closes #928 

### TODO
* I think we can remove the retry strategy for ack in the source forwarder. I will do that using a separate PR.


pkg/sources/udsource/grpc_udsource.go
```diff
@@ -0,0 +1,161 @@
...
+func (u *GRPCBasedUDSource) WaitUntilReady(ctx context.Context) error {
+	for {
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("failed on readiness check: %w", ctx.Err())
+		default:
+			if _, err := u.client.IsReady(ctx, &emptypb.Empty{}); err == nil {
+				return nil
+			}
+			time.Sleep(1 * time.Second)
```
vigith, line 65: I am wondering whether we should log the error and sleep so users will know what is happening, esp. if someone has attached the wrong container to the vertex. other wise, it is an infinite wait.

pkg/sources/udsource/grpc_udsource.go
```diff
@@ -0,0 +1,161 @@
...
+	var r = &sourcepb.AckRequest{
+		Request: &sourcepb.AckRequest_Request{
+			Offsets: rOffsets,
+		},
+	}
+	_, err := u.client.AckFn(ctx, r)
+	return err
+}
+
+func constructMessageID(r *sourcepb.ReadResponse_Result) string {
```
vigith, line 158: make it a stringer?

test/udsource-e2e/udsource_test.go
```diff
@@ -44,10 +44,10 @@ func (s *UserDefinedSourceSuite) TestSimpleSource() {
 
 	// the user-defined simple source sends the read index of the message as the message content
 	// verify the sink get the first batch of data
-	w.Expect().SinkContains("out", "0")
-	w.Expect().SinkContains("out", "1")
+	w.Expect().VertexPodLogContains("out", "0")
+	w.Expect().VertexPodLogContains("out", "1")
 	// verify the sink get the second batch of data
-	w.Expect().SinkContains("out", "2")
+	w.Expect().VertexPodLogContains("out", "2")
```
vigith, line 50: we used redis to do e2e because `VertexPodLogContains` was too flaky.

pkg/sources/udsource/grpc_udsource.go
KeranYang: The proto generated `sourcepb.ReadResponse_Result` does have a stringer which uses the following method to translate the message to a string.

```
// MessageStringOf returns the message value as a string,
// which is the message serialized in the protobuf text format.
func (Export) MessageStringOf(m protoreflect.ProtoMessage) string {
	return prototext.MarshalOptions{Multiline: false}.Format(m)
}
```

I don't feel comfortable using it because it's not within our control. It might not enforce uniqueness and it the output can be quite verbose, probably the ID generated by it is longer than my current approach.

test/udsource-e2e/udsource_test.go
vigith: please make a comment why we are not using Redis sink for posterity.

vigith (approved):

yhl25 (approved):

PR #983, opened 2023-08-24
chore: remove metadata from message

yhl25:


vigith (commented):
if we remove this how will users know whether this is a non-processable message that is  just looping?

vigith:
let's close this for now. we will have to rethink.

PR #986, opened 2023-08-24
chore: add some small refactors mainly for source

KeranYang:
1. Remove retry on Ack.
For most of our existing sources, the Ack() functions return `return make([]error, len(offsets))`. The only exception is the RedisStream source, it uses `err := br.Client.XAck()` to batch acknowledge and populates the error array with the same err. Hence, for ALL sources, it's either ALL success or ALL failure and we don't need to retry acking.
2. Replace `err ==` with `errors.Is`.
3. Fix some typos and nit grammars as IntelliJ is giving me a lot of red lines :)
4. Remove some gRPC unit tests because I don't think they are necessary. The whole point of mocking a dependency is so that we don't worry about how the dependency constructs the returned value. Some of our gRPC unit tests are implementing the dependency and verifying the functionality of the implementation, which doesn't make sense to me. e.g. in 
`TestHGRPCBasedUDF_ApplyWithMockClient`, we implement `multiplyBy2` and verify the result is `multipliedBy2`.

### DIFF
--- a/pkg/sources/forward/data_forward.go
+++ b/pkg/sources/forward/data_forward.go
@@ -372,7 +372,7 @@ func (isdf *DataForward) forwardAChunk(ctx context.Context) {
 
 	// when we apply transformer, we don't handle partial errors (it's either non or all, non will return early),
 	// so we should be able to ack all the readOffsets including data messages and control messages
-	err = isdf.reader.Ack(ctx, readOffsets)[0]
+	err = isdf.ackFromSource(ctx, readOffsets)
 	// implicit return for posterity :-)
 	if err != nil {
 		isdf.opts.logger.Errorw("failed to ack from source", zap.Error(err))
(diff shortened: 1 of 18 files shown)


pkg/sources/forward/data_forward.go
```diff
@@ -374,7 +372,7 @@ func (isdf *DataForward) forwardAChunk(ctx context.Context) {
 
 	// when we apply transformer, we don't handle partial errors (it's either non or all, non will return early),
 	// so we should be able to ack all the readOffsets including data messages and control messages
-	err = isdf.ackFromBuffer(ctx, readOffsets)
+	err = isdf.reader.Ack(ctx, readOffsets)[0]
```
whynowy, line 375: Even this theory is correct since we do all success or all failure, this line (only checking the first item) is confusing.

If we want to do that, it would be better do it in the separate function (`ackFromBuffer`), and add some comments.

whynowy (approved):
Assuming your theory is true :)

Issue #333, opened 2022-11-09
Add Pipeline pausing docs


dseapy:
# Summary

Just learned you can scale down an entire pipeline with paused when looking through repo, which is super useful.  don't think it's in the docs yet, might be good to add

# Use Cases

user wanting to pause a pipeline might not know it's a supported feature.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

whynowy:
That's right, hidden features :-)

PR #989, opened 2023-08-25
docs: how to pause/resume a pipeline

whynowy:
Closes: #333

### DIFF
--- a/docs/user-guide/reference/configuration/pipeline-operations.md
+++ b/docs/user-guide/reference/configuration/pipeline-operations.md
@@ -14,7 +14,7 @@ To summarize, if there are unprocessed messages in the pipeline, and the new pip
 
 ## Pause a Pipeline
 
-To pause a pipeline, use the command below, it will bring the pipeline to `Paused` status, and terminate all running vertex pods.
+To pause a pipeline, use the command below, it will bring the pipeline to `Paused` status, and terminate all the running vertex pods.
 
 ```bash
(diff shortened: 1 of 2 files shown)


vigith (approved):
should we write when to pause a pipeline v/s delete and recreate? and what is the difference? call out the caveat that when it resumes it will start from the last known state and can read a lot of data from source?

vigith:
also call this out #990 :)

whynowy:
> also call this out #990 :)

We should fix those bugs instead.

whynowy:
> > also call this out #990 :)
> 
> We should fix those bugs instead.

You are right, call out the impact of the WAL.

docs/user-guide/reference/configuration/pipeline-operations.md
```diff
@@ -0,0 +1,39 @@
...
+
+To pause a pipeline, use the command below, it will bring the pipeline to `Paused` status, and terminate all the running vertex pods.
+
+```bash
+  kubectl patch pl my-pipeline --type=merge --patch '{"spec": {"lifecycle": {"desiredPhase": "Paused"}}}'
+```
+
+Pausing a pipeline will not cause data loss. It does not clean up the unprocessed data in the pipeline, but just terminates the running pods. When the pipeline is resumed, the pods will be restarted and continue processing the unprocessed data.
+
+When pausing a pipeline, it will shutdown the source vertex pods first, and then wait for the other vertices to finish the backlog before terminating them.
```
whynowy, line 25: There's not a timeout (grace period) for pausing, need to add it.

vigith (approved):

PR #999, opened 2023-08-29
fix: avoid unwanted watcher creation and reduce being stuck with udf is restarted

### LINKED ISSUE #1001: reduce pod gets stuck when the udf container is restarted.

yhl25:
fixes #1001

### DIFF
--- a/pkg/udf/rpc/grpc_reduce.go
+++ b/pkg/udf/rpc/grpc_reduce.go
@@ -110,13 +110,18 @@ func (u *GRPCBasedReduce) ApplyReduce(ctx context.Context, partitionID *partitio
 
 	// create datum from isbMessage and send it to datumCh channel for reduceFn
 	go func() {
-	readLoop:
+		// after reading all the messages from the messageStream or if ctx was canceled close the datumCh channel
+		defer close(datumCh)
 		for {
 			select {
 			case msg, ok := <-messageStream:
-				// if the messageStream is closed or if the message is nil, break the loop
-				if !ok || msg == nil {
-					break readLoop
+				// if the messageStream is closed, break the loop
+				if !ok {
+					return
+				}
+				// if the message is nil, break the loop
+				if msg == nil {
+					return
 				}
 				d := createDatum(msg)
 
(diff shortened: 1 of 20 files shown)


vigith (approved):

pkg/udf/rpc/grpc_reduce.go
```diff
@@ -75,111 +74,104 @@ func (u *GRPCBasedReduce) WaitUntilReady(ctx context.Context) error {
...
+	)
 
 	// pass key and window information inside the context
 	mdMap := map[string]string{
 		shared.WinStartTime: strconv.FormatInt(partitionID.Start.UnixMilli(), 10),
 		shared.WinEndTime:   strconv.FormatInt(partitionID.End.UnixMilli(), 10),
 	}
 
-	ctx = metadata.NewOutgoingContext(ctx, metadata.New(mdMap))
+	cctx, cancel := context.WithCancel(ctx)
```
vigith, line 91: why do you need a new context?

pkg/udf/rpc/grpc_reduce.go
```diff
@@ -75,111 +74,104 @@ func (u *GRPCBasedReduce) WaitUntilReady(ctx context.Context) error {
...
+		close(responseCh)
 	}()
 
-readLoop:
-	for {
-		select {
-		case msg, ok := <-messageStream:
-			if msg != nil {
+	// create datum from isbMessage and send it to datumCh channel for reduceFn
+	go func() {
```
whynowy, line 112:
```golang
go func() {
  defer close(datumCh)
  for {
    select {
      if xxx {
        return
      }
      xxxx
    }
  }
}()
```

pkg/udf/rpc/grpc_reduce.go
yhl25: to exit from readloop go routine when we return from ApplyReduce.

pkg/udf/rpc/grpc_reduce.go
yhl25: done

pkg/udf/rpc/grpc_reduce.go
```diff
@@ -75,111 +74,103 @@ func (u *GRPCBasedReduce) WaitUntilReady(ctx context.Context) error {
...
+			select {
+			case msg, ok := <-messageStream:
+				// if the messageStream is closed, break the loop
+				if !ok {
+					return
+				}
+				// if the message is nil, break the loop
+				if msg == nil {
+					return
+				}
```
vigith, line 125: reverted this change?

yhl25: fixed

pkg/udf/rpc/grpc_reduce.go
whynowy: It should be okay to just do `ctx, cancel := context.WithCancel(ctx)`, do not need to give a new `cctx`.

This is a common patten to make sure everyone in this function watches this `ctx` will properly exit if the function returns, no matter if it's expectedly or unexpectedly.

whynowy (approved):

PR #1009, opened 2023-08-30
docs: put square vertices in start guide diagrams

jasonzeshengchen:
Replaced the even-odd diagram with a new one containing square vertices, to be more consistent with the UI.


docs/quick-start.md
```diff
@@ -94,7 +94,7 @@ kubectl delete -f https://raw.githubusercontent.com/numaproj/numaflow/stable/exa
 
 Now we will walk you through creating an advanced pipeline. In our example, this is called the `even-odd` pipeline, illustrated by the following diagram:
 
-![Pipeline Diagram](assets/even-odd.png)
+![Pipeline Diagram](assets/even-odd-square.png)
```
whynowy, line 97: Name does not match, PNG vs. png.

If we use the new image, remove the old one from assets.

jasonzeshengchen: should be good now. thanks!

whynowy (approved):

Issue #1006, opened 2023-08-30
Pipeline shows 0 processing rate when it fails scaling up due to exceeding quota


KeranYang:
**Describe the bug**
When a numaflow pipeline can't scale up because of exceeding resource quota. The vertex shows a 0 processing rate.

The following pipeline is retrying scaling up with error `exceeded quota: resource-quota, requested: pods=1, used: pods=50, limited: pods=50`, while the sink vertex shows 0 processing rate.

![Screenshot 2023-08-30 at 9 19 45 AM](https://github.com/numaproj/numaflow/assets/13165977/fed6272a-7760-4db3-88e1-4a86d22b5cb6)

**To Reproduce**
Find an iksm cluster with N pod count quota, deploy a pipeline with N+1 pods. Monitor the UI.

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**

- Kubernetes: [e.g. v1.18.6]
- Numaflow: [e.g. v0.5.1]

**Additional context**

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

PR #1014, opened 2023-09-01
fix: reconcile headless services before pods

KeranYang:
The problem: when we start a pipeline and create vertex A with 10 pods and the resource quota is 8, the vertex reconciler will keep creating new pods and start to fail reconciliation when trying to create the 9th pod. Currently in our vertex reconciler, the pod reconciliation happens before service hence the vertex reconciler will NEVER create the headless service for the vertex. 

The headless service is used for exposing metrics like processing rate, even if we have 8 pods which don't meet our specification, it shouldn't stop us from emitting metrics for the vertex.

This change moves the service creation to before pod creation to mitigate the issue.

Fixes: #1006

### DIFF
--- a/docs/development/debugging.md
+++ b/docs/development/debugging.md
@@ -47,18 +47,6 @@ spec:
         env:
           - name: NUMAFLOW_DEBUG
             value: "true" # DO NOT forget the double quotes!!!
-  vertices:
-    - name: in
-      source:
-        generator:
-          rpu: 100
-          duration: 1s
-    - name: out
-      sink:
-        log: {}
-  edges:
-    - from: in
-      to: out
 ```
 
 ## Profiling
--- a/docs/development/debugging.md
+++ b/docs/development/debugging.md
@@ -32,6 +32,35 @@ spec:
       to: out
 ```
 
+To enable debug logs in the daemon pod, set environment variable `NUMAFLOW_DEBUG` to `true` for the daemon pod.
+For example:
+
+```yaml
+apiVersion: numaflow.numaproj.io/v1alpha1
+kind: Pipeline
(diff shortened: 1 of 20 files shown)


docs/development/debugging.md
```diff
@@ -32,6 +32,35 @@ spec:
...
+      source:
+        generator:
+          rpu: 100
+          duration: 1s
+    - name: out
+      sink:
+        log: {}
+  edges:
+    - from: in
+      to: out
```
vigith, line 61: you can skip this section to make the README smaller

whynowy (approved):

pkg/daemon/server/service/rater/rater.go
```diff
@@ -281,3 +285,11 @@ func (r *Rater) buildLookbackSecondsMap(vertexName string) map[string]int64 {
 	}
 	return lookbackSecondsMap
 }
+
+func stringify(q *sharedqueue.OverflowQueue[*TimestampedCounts]) string {
```
vigith, line 289: write Stringer on `sharedqueue.OverflowQueue` type?

pkg/reconciler/vertex/controller.go
```diff
@@ -160,6 +160,56 @@ func (r *vertexReconciler) reconcile(ctx context.Context, vertex *dfv1.Vertex) (
 		return ctrl.Result{}, err
 	}
 
+	// Create services
```
vigith, line 163: Can you please add a comment on why this "move" is important? A summary of the PR description is good enough.

Any constants should be applied before the runtime variables can come into play.

vigith (approved):

PR #1019, opened 2023-09-06
chore: update sdk version

yhl25:
- updated sdk version to v0.5.0
- keeping a copy of sdk constants in numaflow repo
- added unit tests for sdk clients
- fixed inmem kv-store, which was causing unit tests to fail


pkg/apis/numaflow/v1alpha1/const.go
```diff
@@ -161,7 +161,6 @@ const (
 	DefaultKeyForNonKeyedData = "NON_KEYED_STREAM"
 
 	// Default gRPC max message size
```
whynowy, line 163: Delete it.

whynowy (approved):

PR #1036, opened 2023-09-11
refactor: re-arrange some of the rater implementations

KeranYang:
* Add one more unit test to cover the case when a single pod is serving multiple partitions.
* Refactor the vertex type key usage by making them const and shared between the pod tracker and rater.
* Fix a TODO to make userSpecifiedLookbackSecond static instead of being figured out dynamically every time getRates gets called.
* Break early when looping through the metric label list to find the partition name.

### DIFF
--- a/pkg/daemon/server/service/rater/helper.go
+++ b/pkg/daemon/server/service/rater/helper.go
@@ -23,18 +23,18 @@ import (
 )
 
 const (
-	IndexNotFound = -1
+	indexNotFound = -1
 
 	// The following string set is used to identify the vertex type of pod
 
-	// KeyVertexTypeReduce is the vertex type string for reduce vertex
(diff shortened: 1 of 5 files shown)


pkg/daemon/server/service/rater/helper.go
```diff
@@ -22,7 +22,20 @@ import (
...
 )
 
-const IndexNotFound = -1
+const (
+	IndexNotFound = -1
+
+	// The following string set is used to identify the vertex type of pod
+
+	// KeyVertexTypeReduce is the vertex type string for reduce vertex
+	KeyVertexTypeReduce = "reduce"
```
whynowy, line 31: Make it private?

whynowy (approved):

PR #1040, opened 2023-09-12
fix: allow udsource pending api to return negative count to indicate PendingNotAvailable

KeranYang:
Such that we distinguish the PendingNotAvailable scenario from there are not pending messages. Hence auto-scaler knows whether to skip scaling or to scale down.

### DIFF
--- a/pkg/sources/udsource/grpc_udsource.go
+++ b/pkg/sources/udsource/grpc_udsource.go
@@ -73,7 +73,10 @@ func (u *GRPCBasedUDSource) WaitUntilReady(ctx context.Context) error {
 // ApplyPendingFn returns the number of pending messages in the source.
 func (u *GRPCBasedUDSource) ApplyPendingFn(ctx context.Context) (int64, error) {
 	if resp, err := u.client.PendingFn(ctx, &emptypb.Empty{}); err == nil {
-		return int64(resp.Result.Count), nil
+		if resp.Result.Count < 0 {
+			return isb.PendingNotAvailable, nil
+		}
+		return resp.Result.Count, nil
 	} else {
 		return isb.PendingNotAvailable, err
 	}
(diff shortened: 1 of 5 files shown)


whynowy (approved):

pkg/sources/udsource/grpc_udsource.go
```diff
@@ -73,7 +73,10 @@ func (u *GRPCBasedUDSource) WaitUntilReady(ctx context.Context) error {
...
 func (u *GRPCBasedUDSource) ApplyPendingFn(ctx context.Context) (int64, error) {
 	if resp, err := u.client.PendingFn(ctx, &emptypb.Empty{}); err == nil {
-		return int64(resp.Result.Count), nil
+		if resp.Result.Count < 0 {
+			return isb.PendingNotAvailable, nil
+		}
+		return resp.Result.Count, nil
 	} else {
 		return isb.PendingNotAvailable, err
 	}
```
juliev0, line 82: I'm not sure if the caller is logging the `isb.PendingNotAvailable` but even if it is, it might still be helpful to have a log line for line 77 and line 81, in which the log lines are different to distinguish what happened.

KeranYang: Below is how the caller uses the API.

```
if pending, err := lagReader.Pending(ctx); err != nil {
        } else {
        if pending != isb.PendingNotAvailable {
        ts := timestampedPending{pending: pending, timestamp: time.Now().Unix()}
        ms.partitionPendingInfo[partitionName].Append(ts)
        }
}
```

If the err is non-nil, we log the err. If PendingNotAvailable, we skip adding the pending information.

I would prefer not to log from the grpc_udsource.go side because 1. the caller has a way to distinguish between L77 and L81, 2. currently grpc_udsource.go doesn't have a logger and it's using the returned error to include logs, expecting the caller to dive deep. Would like to keep this pattern across APIs. and 3. pending gets called frequently so would like to reduce amount of logs.

Issue #1002, opened 2023-08-29
complete pipeline validations


vigith:
We need to do an exhaustive validation on incompatible/unsupported pipeline states

Validate the following:
- [x] forests
- [x] dangling vertices
- [x] #974 
- [x] disallow if sink is not the last vertex

juliev0:
I could potentially do this along with [this](https://github.com/numaproj/numaflow/issues/974) one that's related. Or if anyone else works on it, just make sure that if you're traversing vertices that you don't get into an infinite loop now that we can have cycles. Note [this](https://github.com/numaproj/numaflow/blob/main/pkg/reconciler/pipeline/validate.go#L413) method may or may not be useful.

vigith:
I have added #974 to the list

juliev0:
One way to look at the forest problem: If you start with any Vertex and from there you traverse "from" and "to" and you stop whenever you hit something that you already visited, you can compile a unique Set. Then you can verify that the length of that Set is equal to the number of Vertices. If it's less than the number of Vertices then you have a forest.

PR #1043, opened 2023-09-13
feat: implement pipeline validation for unsupported states

dpadhiar:
Partially completes #1002 

Adds pipeline validation to prevent pipelines to be submitted that have:

- the last vertex not be a sink
- duplicate edges
- dangling vertices


pkg/reconciler/pipeline/validate.go
```diff
@@ -89,6 +89,17 @@ func ValidatePipeline(pl *dfv1.Pipeline) error {
 		return fmt.Errorf("pipeline has no sink, at least one vertex with 'sink' defined is required")
 	}
 
+	for _, v := range pl.Spec.Vertices {
```
whynowy, line 92: Shall we do something like below:

1. Sources should have `to`, but no `from`,
2. Sinks should have `from`, but no`to`;
3. UDFs should have both.

And this can be done together with the for loop above.

whynowy: If this logic is in place, maybe we can remove the logic in line 165-170.

pkg/reconciler/pipeline/validate.go
```diff
@@ -89,6 +89,17 @@ func ValidatePipeline(pl *dfv1.Pipeline) error {
...
+		if len(pl.GetToEdges(v.Name)) == 0 {
+			if len(pl.GetFromEdges(v.Name)) == 0 {
+				return fmt.Errorf("not all the vertex names are defined in edges")
```
KeranYang, line 95: Is this error message right?

pkg/reconciler/pipeline/validate.go
KeranYang: +1.

pkg/reconciler/pipeline/validate.go
dpadhiar: This is the error message on lines 181-183 that handles dangling vertices as well.

I think a more appropriate message might be that the vertex is not connected to the pipeline.

KeranYang: yes, would be better if we embed the vertex name into the err msg.

dpadhiar: This line may not be necessary now with this implementation. Since we are checking that each UDF/Source/Sink has the appropriate number of edges, those error messages will inform us which vertex is dangling.

pkg/reconciler/pipeline/validate.go
```diff
@@ -67,12 +70,18 @@ func ValidatePipeline(pl *dfv1.Pipeline) error {
 			if v.Source != nil || v.UDF != nil {
 				return fmt.Errorf("invalid vertex %q, only one of 'source', 'sink' and 'udf' can be specified", v.Name)
 			}
+			if len(pl.GetFromEdges(v.Name)) == 0 || len(pl.GetToEdges(v.Name)) > 0 {
+				return fmt.Errorf("invalid vertex %q, sink must have at least 1 from edge and 0 from edges", v.Name)
```
KeranYang, line 74:
```suggestion
				return fmt.Errorf("invalid vertex %q, sink must have at least 1 from edge and 0 to edges", v.Name)
```

dpadhiar: Good catch!

pkg/reconciler/pipeline/validate.go
```diff
@@ -58,6 +58,9 @@ func ValidatePipeline(pl *dfv1.Pipeline) error {
 			if v.Sink != nil || v.UDF != nil {
 				return fmt.Errorf("invalid vertex %q, only one of 'source', 'sink' and 'udf' can be specified", v.Name)
 			}
+			if len(pl.GetToEdges(v.Name)) == 0 || len(pl.GetFromEdges(v.Name)) > 0 {
+				return fmt.Errorf("invalid vertex %q, source must have at least 1 to edge and 0 from edges", v.Name)
```
KeranYang, line 62: This error msg seems ambiguous. One can interpret it as source must have at least 1 to edge and any number of from edges. Consider changing to something like `must have 0 from edges and at least 1 to edge`

KeranYang (approved):

whynowy (approved):

PR #1046, opened 2023-09-14
feat: added toolbar and updated graph background

veds-g:
Modified the graph background wrt to the new design
Added toolbar with the following interactions
- Toggle panning on the graph area
- Toggle zooming with scroll/pinch
- Trigger fitView / fullscreen
- Toggle side panel (yet to add logic)
- Zoom in/out with percentage info

Updated the logo with transparent background

pipeline


ui/src/components/pages/Pipeline/partials/Graph/index.tsx
```diff
@@ -74,6 +85,101 @@ const getLayoutedElements = (
...
+          onClick={() => setIsScrollLocked((prevState) => !prevState)}
+        >
+          <img src={scrollToggle} alt={"scrollLock"} />
+        </IconButton>
+        <div className={"divider"} />
+        <IconButton onClick={() => fitView()}>
+          <img src={fullscreen} alt={"fullscreen"} />
+        </IconButton>
+        <IconButton
+          onClick={() => {
```
bbehnke, line 134: Would be good to move all these in-line callback method definitions to useCallback methods. It will result in less overall re-rendering of children.

veds-g: sure, let me do that

bbehnke (approved):

PR #1063, opened 2023-09-18
feat: add forest validation for pipelines. Fixes #1002

### LINKED ISSUE #1002: complete pipeline validations

dpadhiar:
Explain what this PR does.

Fixes #1002 

Adds validation to prevent users from submitting pipeline specs that are forests.


pkg/reconciler/pipeline/validate_test.go
```diff
@@ -273,6 +324,42 @@ func TestValidatePipeline(t *testing.T) {
...
+
+	t.Run("forest - pipelines have cycles", func(t *testing.T) {
+		testObj := testForestPipeline.DeepCopy()
+		testObj.Spec.Edges = append(testObj.Spec.Edges, dfv1.Edge{From: "p1", To: "p1"})
+		testObj.Spec.Edges = append(testObj.Spec.Edges, dfv1.Edge{From: "p2", To: "p2"})
+		err := ValidatePipeline(testObj)
+		assert.Error(t, err)
+		assert.Contains(t, err.Error(), "invalid pipeline")
+	})
+
```
juliev0, line 362: looks like some good test cases for invalid pipelines. I wonder if we should add one for an "interesting" valid pipeline - something with multiple sources and multiple sinks?

pkg/reconciler/pipeline/validate.go
```diff
@@ -472,3 +476,37 @@ func toVerticesMappedByFrom(edges []dfv1.Edge, verticesByName map[string]*dfv1.A
...
 }
+
+func findForests(vtx dfv1.AbstractVertex, pl *dfv1.Pipeline) bool {
+
+	visited := map[string]struct{}{}
+	findForestHelper(vtx.Name, visited, pl)
+
+	// true if forest is found
+	return len(visited) != len(pl.Spec.Vertices)
+
```
juliev0, line 487: might be useful to have a comment for this function to go over how it works

pkg/reconciler/pipeline/validate.go
```diff
@@ -472,3 +476,37 @@ func toVerticesMappedByFrom(edges []dfv1.Edge, verticesByName map[string]*dfv1.A
...
+}
+
+func findForestHelper(vtxName string, visited map[string]struct{}, pl *dfv1.Pipeline) {
+
+	visited[vtxName] = struct{}{}
+
+	// construct list all to and from vertices
```
juliev0, line 494: "construct list of all"...

pkg/reconciler/pipeline/validate.go
```diff
@@ -107,6 +107,10 @@ func ValidatePipeline(pl *dfv1.Pipeline) error {
 				return fmt.Errorf("invalid user-defined source vertex %q, only one of 'http', 'kafka', 'nats', 'redisStreams', 'generator' and 'udSource' can be specified", k)
 			}
 		}
+
+		if findForests(s, pl) {
```
KeranYang, line 111: Do we need to `findForests` for all the sources? Can we define the findForests as below?

```
func findForests(pl *dfv1.Pipeline) bool
```

Inside the function, we randomly pick a vertex and DFS. If the pipeline is a forest, then ANY of the vertices cannot visit ALL vertices. Hence if the eventual lengths don't match, the pipeline is a forest.

pkg/reconciler/pipeline/validate.go
```diff
@@ -472,3 +476,37 @@ func toVerticesMappedByFrom(edges []dfv1.Edge, verticesByName map[string]*dfv1.A
 	}
 	return mappedEdges, nil
 }
+
+func findForests(vtx dfv1.AbstractVertex, pl *dfv1.Pipeline) bool {
+
+	visited := map[string]struct{}{}
+	findForestHelper(vtx.Name, visited, pl)
```
KeranYang, line 483: nit - naming: we have findForest**s** and findForestHelper, do we need the extra s? Also consider naming it `isAForest`

pkg/reconciler/pipeline/validate.go
juliev0: Correct. I didn't notice we were doing it for all sources. We should only do it for one vertex.

dpadhiar: In this case, how would we randomly pick the vertex? I'm in favor of just calling it on `pl.Spec.Vertices[0]` each time as well.

juliev0: that's fine

vigith (approved):

juliev0 (approved):

PR #1066, opened 2023-09-19
feat: cluster summary page, namespace summary page base work, side bar base work

bbehnke:
Functional 1.0 cluster summary page
Start of namespace cluster summary page
Start of sidebar


ui/src/utils/fetchWrappers/namespaceSummaryFetch.ts
```diff
@@ -0,0 +1,259 @@
...
+    isbsActiveCount,
+    isbsInactiveCount,
+    isbsHealthyCount,
+    isbsWarningCount,
+    isbsCriticalCount,
+    pipelineSummaries,
+  };
+};
+
+export interface NamespacePipelineSummary {
```
veds-g, line 131: can we move the interfaces to the top or define them inside `types/declarations` for better readability

ui/src/components/pages/Namespace/index.tsx
```diff
@@ -1,219 +1,111 @@
...
-    if (namespaceListError && systemInfo && systemInfo?.namespaced === false) {
-      notifyError([
-        {
-          error: "Failed to fetch the available namespaces",
-          options: { toastId: "ns-cluster", autoClose: false },
-        },
-      ]);
-    }
-  }, [namespaceListError, systemInfo]);
+  console.log("DATA", data);
```
veds-g, line 21: cleanup

ui/src/components/pages/Namespace/index.tsx
```diff
@@ -1,219 +1,111 @@
...
-import { GetStore } from "../../../localStore/GetStore";
+import CircularProgress from "@mui/material/CircularProgress";
+import { useNamespaceSummaryFetch } from "../../../utils/fetchWrappers/namespaceSummaryFetch";
+import { SummaryPageLayout, SummarySection, SummarySectionType } from "../../common/SummaryPageLayout";
 
 import "./style.css";
 
-const NS_Curr_Store = "curr_namespace";
-const NS_List_Store = "namespace_list";
+const DATA_REFRESH_INTERVAL = 15000; // ms
```
veds-g, line 10: for other APIs for graph we refresh at 30sec and 1min, should we set a standard refresh interval for all?

vigith: Let's not have a single refresh interval for all. The pipeline view will be the most aggressive one, while others can be a little slower. Refresh at a cluster level is the most resource-intensive one with the least ROI and hence can be slower.

ui/src/utils/fetchWrappers/namespaceSummaryFetch.ts
bbehnke: Updated.

mshakira (approved):

PR #1085, opened 2023-09-25
feat: set correct number of replices for some types of vertices after resuming pipeline

### LINKED ISSUE #991: Resuming a paused pipeline sets incorrect replicas for some type of vertices
**Describe the bug**

If a pipeline with an HTTP source is paused, when resuming it, it only brings the HTTP source replica back to 1, even if its `spec.scale.min` is set to a number > 1.

When we resume a pipeline, we always set the replica number to 1, and let the autoscaling kick in to bring the number back to normal, but it does not work for sources like HTTP, because they do not support autoscaling.

---

dpadhiar:
Fixes #991 

Currently if a user resumes a paused pipeline with a vertex that does not support autoscaling, the replica count for the vertex will always be 1 even if `spec.scale.min` is set greater than 1. This is because the work for scaling the vertex is expected to be done by an autoscaler. This is fixed by checking that a vertex is not scalable and has min set when resuming the pipeline to assure the correct replica count is set.


vigith (approved):

KeranYang (commented):
LGTM

PR #1106, opened 2023-09-28
fix: update route endpoints to v1.1 for graph, added error side panel

veds-g:
- Updated api endpoints to v1_1 for pipeline component
- Added error sidebar to appear when clicking icon to top right corner of the page
- Added generator spec to sidebar
image
image


ui/src/components/common/SlidingSidebar/index.tsx
```diff
@@ -43,9 +54,13 @@ export function SlidingSidebar({
   vertexDetailsProps,
   edgeDetailsProps,
   pipelineSpecsProps,
+  generatorDetailsProps,
+  errorsProps,
   onClose,
 }: SlidingSidebarProps) {
-  const [width, setWidth] = useState<number>(pageWidth ? pageWidth / 2 : 0);
+  const [width, setWidth] = useState<number>(
+    errorsProps ? 350 : pageWidth ? pageWidth / 2 : 0
```
bbehnke, line 62: Use MIN_WIDTH_BY_TYPE map value here instead of hardcoded 350

ui/src/components/common/SlidingSidebar/index.tsx
```diff
@@ -119,22 +144,24 @@ export function SlidingSidebar({
...
-      >
-        <img
-          onMouseDown={dragHandler}
-          src={slider}
-          alt="slider"
-          className={"sidebar-drag-icon"}
-          draggable={false}
-        />
-      </Box>
+      {!errorsProps && (
```
bbehnke, line 147: Can we add "slide" boolean prop to top level SidebarProps w/ a default of true for this? 

Then when you call setSidebarProps, just add "slide: false". Just like type in the unnested props.

This will make any other future cases where slider isn't needed much easier.

ui/src/utils/fetcherHooks/pipelineViewFetch.ts
```diff
@@ -167,79 +195,88 @@ export const usePipelineViewFetch = (
...
+        fetch(`${BASE_API}/vertices/metrics`)
+          .then((response) => {
+            if (response.ok) {
+              return response.json();
+            } else {
+              return Promise.reject(response);
+            }
+          })
+          .then((json) => {
+            if (json?.data) {
```
bbehnke, line 208: Maybe methodize some of this if possible. Or add some comments.

ui/src/components/common/SlidingSidebar/partials/GeneratorDetails/style.css
```diff
@@ -0,0 +1,29 @@
+.vertex-details-header-text {
```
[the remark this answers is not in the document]
veds-g: its essentially same as vertex details sidebar instead of 4 it has only 1 tab, should I still update the classname?

ui/src/utils/fetcherHooks/pipelineViewFetch.ts
veds-g: added comments

ui/src/components/common/SlidingSidebar/index.tsx
veds-g: done

ui/src/components/common/SlidingSidebar/index.tsx
veds-g: done

bbehnke (approved):

PR #1117, opened 2023-09-29
feat: add buffer info in vertex side panel

veds-g:
- updated remaining api endpoints to v1_1
- added buffer info for non-source vertices
image


ui/src/utils/fetchWrappers/systemInfoFetch.ts
```diff
@@ -23,12 +20,17 @@ export const useSystemInfoFetch = () => {
       setLoading(false);
       return;
     }
+    if (data?.errMsg) {
```
bbehnke, line 23: In the other fetchwrappers I just combine error and errMsg, into just error in the result. Can we do that here too? The components using it will need to do less "error" checks this way.

bbehnke (approved):

PR #1135, opened 2023-10-02
chore: Version update v1 api

jy4096:
Replace current v1 API with v1_1 and keep the v1 version.


vigith (approved):

vigith:
i do see some dead code, are we planning to delete those?

jy4096:
> i do see some dead code, are we planning to delete those?

👌

PR #1136, opened 2023-10-02
fix: PR review actions

jy4096:
1. bug fix
previously, to get the pipeline lag we updated the proto file to have from and to in the watermark response
but we only wet these two fields when watermark is enabled
when watermark is disabled, we also need to set the from and to in the query response
2. bug fix
UpdateISBSVC API doesn't use the new request body to udpate
3. update isbsvc validate to have min 3 replicas
3. add TODO
4. use c.BindJSON
5. update kube rest config fetch logic in NewHandler


vigith (approved):
do we still need the proto changes?

jy4096:
> do we still need the proto changes?

yes.. for calculate pipeline lag

PR #1138, opened 2023-10-02
feat: add timeout for pausing pipeline. Fixes #992

### LINKED ISSUE #992: No grace period (timeout) for pausing a pipeline
# Summary

Should be able to set it in `spec.lifecycle`, and respect it in the controller logic.

---

dpadhiar:
Explain what this PR does.

Fixes #992 

Allows user to add timeout (grace period) for pausing a pipeline. Default is 30 seconds and can be configured in `spec.Lifecycle.PauseGracePeriodSeconds`.


pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -430,6 +442,7 @@ type PipelineSpec struct {
 	Edges []Edge `json:"edges,omitempty" protobuf:"bytes,3,rep,name=edges"`
 	// Lifecycle define the Lifecycle properties
 	// +kubebuilder:default={"deleteGracePeriodSeconds": 30, "desiredPhase": Running}
+	// +kubebuilder:default={"deleteGracePeriodSeconds": 30, "desiredPhase": Running, "pauseGracePeriodSeconds": 30}
```
whynowy, line 445: Merge line 445 and 444.

pkg/reconciler/pipeline/controller.go
```diff
@@ -753,7 +770,14 @@ func (r *pipelineReconciler) pausePipeline(ctx context.Context, pl *dfv1.Pipelin
 	if err != nil {
 		return true, err
 	}
-	if drainCompleted {
+
+	pauseTimestamp, err := time.Parse(time.RFC3339, pl.GetAnnotations()["pauseTimestamp"])
+	if err != nil {
+		return true, err
```
whynowy, line 776: If error happens during time parsing, do not requeue  the request as it won't help.

pkg/reconciler/pipeline/controller.go
```diff
@@ -725,6 +725,15 @@ func (r *pipelineReconciler) updateDesiredState(ctx context.Context, pl *dfv1.Pi
 }
 
 func (r *pipelineReconciler) resumePipeline(ctx context.Context, pl *dfv1.Pipeline) (bool, error) {
+
+	// reset pause timestamp
+	if pl.GetAnnotations()["pauseTimestamp"] != "" {
```
whynowy, line 730: Expose the annotation as a constant.

`KeyPauseTimestamp    = "numaflow.numaproj.io/pause-timestamp"`

pkg/reconciler/pipeline/controller.go
```diff
@@ -725,6 +725,15 @@ func (r *pipelineReconciler) updateDesiredState(ctx context.Context, pl *dfv1.Pi
...
+		delete(pl.GetAnnotations(), "pauseTimestamp")
+		if err := r.client.Update(ctx, pl.DeepCopy()); err != nil {
```
whynowy, line 732: Do not update the pipeline here, instead, return the pipeline object with modified data, and let the reconciler do it. https://github.com/numaproj/numaflow/blob/main/pkg/reconciler/pipeline/controller.go#L88

You also need to change the logic of `needsUpdate()`

pkg/reconciler/pipeline/controller.go
```diff
@@ -734,6 +743,14 @@ func (r *pipelineReconciler) resumePipeline(ctx context.Context, pl *dfv1.Pipeli
 }
 
 func (r *pipelineReconciler) pausePipeline(ctx context.Context, pl *dfv1.Pipeline) (bool, error) {
+
+	if pl.GetAnnotations() == nil || pl.GetAnnotations()["pauseTimestamp"] == "" {
+		pl.SetAnnotations(map[string]string{"pauseTimestamp": time.Now().Format(time.RFC3339)})
+		if err := r.client.Update(ctx, pl.DeepCopy()); err != nil {
```
whynowy, line 749: same here.

pkg/reconciler/pipeline/controller.go
```diff
@@ -534,6 +534,9 @@ func needsUpdate(old, new *dfv1.Pipeline) bool {
 	if !equality.Semantic.DeepEqual(old.Finalizers, new.Finalizers) {
 		return true
 	}
+	if !equality.Semantic.DeepEqual(old.GetAnnotations(), new.GetAnnotations()) {
```
whynowy, line 537: Have a concern of comparing all the annotations might cause unexpected behaviors, maybe just check if the keys starting with `numaflow.numaproj.io/` has any change?

pkg/reconciler/pipeline/controller.go
```diff
@@ -534,6 +534,21 @@ func needsUpdate(old, new *dfv1.Pipeline) bool {
...
+	oldAnnotations := old.GetAnnotations()
+	newAnnotations := new.GetAnnotations()
+	if oldAnnotations != nil && newAnnotations == nil {
+		return true
+	}
+	if oldAnnotations == nil && newAnnotations != nil {
+		return true
+	}
+
+	for k, v := range oldAnnotations {
```
whynowy, line 547: The logic we want is to compare the annotations with key starting with `numaflow.numaproj.io/`, if there's difference, return true. Difference could be number of key/value pairs are different, same key has different value, etc.

pkg/reconciler/pipeline/controller.go
```diff
@@ -534,7 +534,23 @@ func needsUpdate(old, new *dfv1.Pipeline) bool {
...
+	newNumaAnnotations := annotSlice("numaflow.numaproj.io/", new.GetAnnotations())
+
+	return !equality.Semantic.DeepEqual(oldNumaAnnotations, newNumaAnnotations)
+}
+
+func annotSlice(label string, annotations map[string]string) map[string]string {
+
+	slice := make(map[string]string)
+	for k, v := range annotations {
+		if strings.Contains(k, label) {
```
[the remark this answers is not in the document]
dpadhiar: changed to `HasPrefix`

whynowy:
@dpadhiar - I tested the change, the annotation causes one extra reconciliation. Can you look into it?

dpadhiar:
> @dpadhiar - I tested the change, the annotation causes one extra reconciliation. Can you look into it?

With a log, I am seeing that using `update` causes the reconciliation to happen twice when resuming the pipeline. When pausing the pipeline, reconciliation occurs twice to change the Status from Running to Pausing to Paused which is expected. 

If we use `patch` instead of `update`, reconciliation only occurs once when resuming.

```
if pl.GetAnnotations() == nil || pl.GetAnnotations()[dfv1.KeyPauseTimestamp] == "" {
		pl.SetAnnotations(map[string]string{dfv1.KeyPauseTimestamp: time.Now().Format(time.RFC3339)})
		body, err := json.Marshal(pl)
		if err != nil {
			return false, err
		}
		err = r.client.Patch(ctx, pl, client.RawPatch(types.MergePatchType, body))
		if err != nil && !apierrors.IsNotFound(err) {
			return false, err
		}
	}
```

pkg/reconciler/pipeline/controller.go
```diff
@@ -725,6 +726,20 @@ func (r *pipelineReconciler) updateDesiredState(ctx context.Context, pl *dfv1.Pi
 }
 
 func (r *pipelineReconciler) resumePipeline(ctx context.Context, pl *dfv1.Pipeline) (bool, error) {
+
+	// reset pause timestamp
+	if pl.GetAnnotations()[dfv1.KeyPauseTimestamp] != "" {
+		pl.GetAnnotations()[dfv1.KeyPauseTimestamp] = ""
```
whynowy, line 732: This means the annotation is not nil, why don't use `del()`?

pkg/reconciler/pipeline/controller.go
```diff
@@ -725,6 +726,20 @@ func (r *pipelineReconciler) updateDesiredState(ctx context.Context, pl *dfv1.Pi
...
+		body, err := json.Marshal(pl)
+		if err != nil {
+			return false, err
+		}
+		err = r.client.Patch(ctx, pl, client.RawPatch(types.MergePatchType, body))
+		if err != nil && !apierrors.IsNotFound(err) {
+			return false, err
```
whynowy, line 739: In this case, we should let it re-queue, right?

pkg/reconciler/pipeline/controller.go
```diff
@@ -734,6 +749,19 @@ func (r *pipelineReconciler) resumePipeline(ctx context.Context, pl *dfv1.Pipeli
 }
 
 func (r *pipelineReconciler) pausePipeline(ctx context.Context, pl *dfv1.Pipeline) (bool, error) {
+
+	if pl.GetAnnotations() == nil || pl.GetAnnotations()[dfv1.KeyPauseTimestamp] == "" {
```
whynowy, line 753: No else?

pkg/reconciler/pipeline/controller.go
```diff
@@ -750,10 +778,17 @@ func (r *pipelineReconciler) pausePipeline(ctx context.Context, pl *dfv1.Pipelin
 		_ = daemonClient.Close()
 	}()
 	drainCompleted, err := daemonClient.IsDrained(ctx, pl.Name)
+	if err != nil {
+		return false, err
```
whynowy, line 782: For this error, we should let it requeue.

pkg/reconciler/pipeline/controller.go
```diff
@@ -750,10 +778,17 @@ func (r *pipelineReconciler) pausePipeline(ctx context.Context, pl *dfv1.Pipelin
...
+	}
+
+	pauseTimestamp, err := time.Parse(time.RFC3339, pl.GetAnnotations()[dfv1.KeyPauseTimestamp])
 	if err != nil {
```
whynowy, line 786: For this error, should not requeue.

pkg/reconciler/pipeline/controller.go
dpadhiar: For this else, we can just return `false` since it means there is a timestamp set.

pkg/reconciler/pipeline/controller.go
dpadhiar: MergePatchType only merges the patch with the current object, it doesn't replace the `annotations` map. We would need to use `StrategicMergePatch` which isn't supported for this.

dpadhiar: Found that you can delete this annotation using this JSON patch: 
`[{"op": "remove", "path": "/metadata/annotations/numaflow.numaproj.io~1pause-timestamp"}]`

Implemented and added patch string to `const.go`

whynowy: This is nice!

pkg/reconciler/pipeline/controller.go
```diff
@@ -725,6 +726,15 @@ func (r *pipelineReconciler) updateDesiredState(ctx context.Context, pl *dfv1.Pi
 }
 
 func (r *pipelineReconciler) resumePipeline(ctx context.Context, pl *dfv1.Pipeline) (bool, error) {
+
+	// reset pause timestamp
+	if pl.GetAnnotations()[dfv1.KeyPauseTimestamp] != "" {
+		err := r.client.Patch(ctx, pl, client.RawPatch(types.JSONPatchType, []byte(dfv1.RemovePauseTimestampPatch)))
+		if err != nil && !apierrors.IsNotFound(err) {
+			return true, err
+		}
```
whynowy, line 735: Just a tiny issue here:
if err != nill && IsNotFound, return false, nil

Somehow the pipeline can not be found, and we skip it.

whynowy (approved):
All the rest looks good to me, except the one commented.

whynowy (changes requested):
.

pkg/reconciler/pipeline/controller.go
dpadhiar: Updated to `if err != nil && apierrors.IsNotFound(err) {
			return false, nil
		}`

dpadhiar:
@whynowy Not sure why this jetstream e2e test continues to fail, otherwise should be done.

pkg/reconciler/pipeline/controller.go
dpadhiar: For this case, yes. We should requeue the pipeline if there is a patching error that is not IsNotFound
```
		if err != nil {
			// if pipeline cannot be found, do not requeue
			if apierrors.IsNotFound(err) {
				return false, nil
		        // otherwise, requeue pipeline
			} else {
				return true, err
			}
		}
```

pkg/reconciler/pipeline/controller.go
dpadhiar:
```
if err != nil {
  if apierrors.IsNotFound(err) {
    return false, nil // skip pl if we can't find it
  } else {
      return false, err // otherwise requeue resume
  }
}
```

whynowy (approved):

PR #1139, opened 2023-10-02
chore: unify all sources to be scalable

KeranYang:
For existing built-in sources that cannot provide pending information, we update them to implement Pending and return `isb.PendingNotAvailable`. The auto-scaler treats them as scalable but excludes them from being auto-scaled.


vigith (approved):
can you please add `LagReader` also to this interface https://github.com/numaproj/numaflow/blob/main/pkg/sources/sourcer.go#L26-L28?

PR #1159, opened 2023-10-04
fix: get pl kind apiversion

jy4096:
because of https://github.com/kubernetes/kubernetes/issues/80609 
get pipeline doesn't return "kind" and "apiVersion"

since we know the value for these two fields, set the fields in handler manually

test result:
image

### DIFF
--- a/pkg/client/clientset/versioned/typed/numaflow/v1alpha1/pipeline.go
+++ b/pkg/client/clientset/versioned/typed/numaflow/v1alpha1/pipeline.go
@@ -20,6 +20,7 @@ package v1alpha1
 
 import (
 	"context"
+	"encoding/json"
 	"time"
 
 	v1alpha1 "github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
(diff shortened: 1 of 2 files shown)


pkg/client/clientset/versioned/typed/numaflow/v1alpha1/pipeline.go
```diff
@@ -65,15 +66,32 @@ func newPipelines(c *NumaflowV1alpha1Client, namespace string) *pipelines {
...
+	pipeline = &v1alpha1.Pipeline{}
+	result := c.client.Get().
 		Namespace(c.ns).
 		Resource("pipelines").
 		Name(name).
 		VersionedParams(&options, scheme.ParameterCodec).
-		Do(ctx).
-		Into(result)
+		Do(ctx)
+	raw, err := result.Raw()
```
whynowy, line 77: Why doing all of there? You can set the pipeline object filed value directly right?

pkg/client/clientset/versioned/typed/numaflow/v1alpha1/pipeline.go
```diff
@@ -65,15 +66,32 @@ func newPipelines(c *NumaflowV1alpha1Client, namespace string) *pipelines {
 }
 
 // Get takes name of the pipeline, and returns the corresponding pipeline object, and an error if there is any.
-func (c *pipelines) Get(ctx context.Context, name string, options v1.GetOptions) (result *v1alpha1.Pipeline, err error) {
-	result = &v1alpha1.Pipeline{}
-	err = c.client.Get().
+func (c *pipelines) Get(ctx context.Context, name string, options v1.GetOptions) (pipeline *v1alpha1.Pipeline, err error) {
```
whynowy, line 69: Hold on - this file is codgen generated, you should not change it.

server/apis/v1/handler.go
```diff
@@ -209,6 +209,9 @@ func (h *handler) GetPipeline(c *gin.Context) {
 		h.respondWithError(c, fmt.Sprintf("Failed to fetch pipeline %q namespace %q, %s", pipeline, ns, err.Error()))
 		return
 	}
+	// set pl kind and apiVersion
+	pl.Kind = "Pipeline"
```
whynowy, line 213: Don't hard code it.

pkg/client/clientset/versioned/typed/numaflow/v1alpha1/pipeline.go
whynowy, line 1: Revert the change on this file, or the ci will not pass.

whynowy (approved):

PR #1171, opened 2023-10-06
chore: update broken links

KeranYang:
* Also add a link checker to the CI.
* I chose to check the test folder because the developers uses the link, docs numaflow users.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -249,7 +249,7 @@ docs-serve: docs
 
 .PHONY: docs-linkcheck
 docs-linkcheck: /usr/local/bin/lychee
-	lychee --exclude-path=CHANGELOG.md --exclude-mail *.md
+	lychee --exclude-path=CHANGELOG.md --exclude-mail *.md --include "https://github.com/numaproj/*" ./test/**/* ./docs/**/*.md
 
 # pre-push checks
 
--- a/Makefile
+++ b/Makefile
@@ -228,6 +228,7 @@ endif
 
 /usr/local/bin/mkdocs:
 	$(PYTHON) -m pip install mkdocs==1.3.0 mkdocs_material==8.3.9 mkdocs-embed-external-markdown==2.3.0
+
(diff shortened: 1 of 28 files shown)


Makefile
```diff
@@ -249,7 +250,7 @@ docs-serve: docs
 
 .PHONY: docs-linkcheck
 docs-linkcheck: /usr/local/bin/lychee
-	lychee --exclude-path=CHANGELOG.md --exclude-mail *.md
+	lychee --exclude-path=CHANGELOG.md --exclude-mail *.md --include "https://github.com/numaproj/*" $(shell find ./test -type f) $(wildcard ./docs/*.md)
```
whynowy, line 253: Does this make sense? The CI is checking remote files instead of local files (with changes)?

KeranYang: Are you referring to the --include specification?

```
--include <INCLUDE>
          URLs to check (supports regex).
```

It basically specifies that lychee is going to scan all files under ./test and ./docs, and look for every url that starts with "https://github.com/numaproj/". Then lychee will ping the URLs to check that the link is working and return errors if not.

KeranYang: The PR is mainly to make sure when we update SDK packages with path change, we update the example links in numaflow accordingly.

vigith (approved):
awesome

Makefile
whynowy: I see. But is it limited to only check the links starting with "https://github.com/numaproj/"? which means ignore that links even they are broken?

KeranYang: Yes. Mainly because if I don't add this prefix, it will scan URLs like "localhost:8080" which is naturally broken and also other links like "https://v1-18.docs.kubernetes.io/docs/reference/generated/*", "https://kubernetes.io/docs/*". The number of links can be very high and I noticed lychee calls being throttled. So I decided to just check numaproj links, for now.

PR #1174, opened 2023-10-09
DRAFT feat: Jetstream support for replica of 1 Fixes #944

### LINKED ISSUE #944: Jetstream Isbsvc Replicas
# Summary

What change needs making?

I've noticed in the documentation and in the source code that if the number of replicas of the jetsream isbsvc is <3, it will still create 3 pods and replicas.  I would like like to have a single isbsvc pod with no extra replicas.  Are there any fundamental limitations that would prevent this?

# Use Cases

When would you use this?

I am running on an embedded device and trying to cut down on overhead as much as possible.  The device is frequently powered on and off as well.  On startup, we sometimes have issues with the isbsvc pods struggling to communicate with each other (e.g. establish a raft leader - Some of these issues have been addressed in recent jetstream releases, but we still have issues sometimes).  I think that life would be a lot easier if we just had a single pod for the buffer rather than 3.  Since we are running on an embedded device, I don't see any advantage to having 3 replicas and we are paying extra costs for it.

---

joelcomp1:
Fixes #944 

This allows for Jetstream event buses to have a replica set 1 instead of the minimum required 3.  Modifications taken mostly from the same change done to Argo Events.  

Tested this locally by modifying the numaflow install to set the bufferConfig to have replica of 1 for each (instead of the default 3) and then set my ISBS to explicitly have a replica of 1.   Then verified I could run the example simple pipeline on the website.  Verified in the ISBS that if replicas are not set that it defaults to 3


whynowy:
@joelcomp1 - something is wrong with you git...

joelcomp1:
I know I was trying to fix my signoff and I made a mess :( I'm going to try to clean out my commits and try again. Sorry for the distraction

PR #1199, opened 2023-10-12
feat: get current status of ISB service

dpadhiar:
Explain what this PR does.

Implement `getIsbServiceStatus` method in `handler.go` to get current status of an ISB service.


server/apis/v1/handler.go
```diff
@@ -824,6 +831,19 @@ func getPipelineStatus(pipeline *dfv1.Pipeline) (string, error) {
 	return retStatus, nil
 }
 
+// GetIsbServiceStatus is used to provide the status of a given InterStepBufferService
+func getIsbServiceStatus(isbsvc *dfv1.InterStepBufferService) (string, error) {
+	retStatus := ISBServiceStatusHealthy
```
whynowy, line 836: We will figure out the correct way of determining if pipeline/isbsvc/vertex is healthy later on, can you please add a TODO for that? But this is good to clean up those TODOs to abstract to this function.

whynowy:
@dpadhiar - please look at the test failures.

dpadhiar:
> @dpadhiar - please look at the test failures.

Made changes to test case to check that ISBSvc is `inactive`.  Logic currently is that if the phase of the ISB is unknown, we mark it as `inactive`, if pending or running it is `healthy` and if failed it is `critical`.

test/api-e2e/api_test.go
```diff
@@ -60,7 +60,7 @@ func (s *APISuite) TestISBSVC() {
 		Expect().
 		Status(200).Body().Raw()
 	assert.Contains(s.T(), getISBSVCBody, fmt.Sprintf(`"name":"%s"`, testISBSVCName))
-	assert.Contains(s.T(), getISBSVCBody, `"status":"healthy"`)
+	assert.Contains(s.T(), getISBSVCBody, `"status":"inactive"`)
```
whynowy, line 63: Is it because it's too fast, that controller hasn't get a chance to reconcile?

whynowy (approved):

test/api-e2e/api_test.go
whynowy: It would be better to have a try with timeout until it's `healthy`.

dpadhiar: I'll make that change, should I open a smaller PR or revert this one?

vigith: please open a new one.

PR #1203, opened 2023-10-13
pods component error fix

mshakira:
-  handle errors when pods not found


bbehnke (changes requested):
Move pods length check to error check if statement as discussed offline.

bbehnke (approved):

PR #1225, opened 2023-10-17
feat: user info context and account menu

bbehnke:
- base work for user context in app
- account menu added
Screenshot 2023-10-18 at 11 54 57 AM

### DIFF
--- a/ui/src/components/pages/Login/index.tsx
+++ b/ui/src/components/pages/Login/index.tsx
@@ -11,12 +11,11 @@ export function Login() {
 
   const handleLoginClick = useCallback(async () => {
     try {
-      const response = await fetch(`/api/v1/login?returnUrl=${returnURL}`, {
-        method: "GET",
-        headers: {
-          "Content-Type": "application/json",
-        },
-      });
+      const response = await fetch(`/api/v1/login?returnUrl=${returnURL}`);
+      if (response.ok) {
+        window.location.replace(response.url);
+        return;
+      }
       const error = await getAPIResponseError(response);
       if (error) {
         setError(error);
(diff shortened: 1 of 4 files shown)


KeranYang (approved):

server/apis/v1/handler.go
```diff
@@ -88,8 +88,8 @@ func (h *handler) Login(c *gin.Context) {
 	// TODO - send a request to Dex to get the real user identity token.
 	token := "dummy-token"
 	c.SetCookie("user-identity-token", token, 3600, "/", "", true, true)
-	returnUrl := c.DefaultQuery("returnUrl", "/cluster-summary")
-	c.Redirect(http.StatusOK, returnUrl)
+	returnUrl := c.DefaultQuery("returnUrl", "/")
+	c.Redirect(http.StatusTemporaryRedirect, returnUrl)
```
qhuai, line 92: 1. Should this redirect take the browser to dex server (not the return url)?
2. the http code may be a minor one. the dex doc uses http.StatusFound. https://dexidp.io/docs/using-dex/

ui/src/components/pages/Login/index.tsx
```diff
@@ -10,21 +10,20 @@ export function Login() {
...
-        setError(error);
-      }
-    } catch (e) {
-      setError(e);
-    }
-  }, []);
+    // try {
+    //   const response = await fetch(`/api/v1/login?returnUrl=${returnURL}`);
+    //   if (response.ok) {
+    window.location.replace(`/api/v1/login?returnUrl=${returnURL}`);
```
qhuai, line 16: Should the returnURL be determined by the dex server? The app (Numaflow api server) should have a config in place with the dex server so that the dex server knows which redirect url to use after a successful authentication.

bbehnke: This PR is not final work. It is just some shared starting code for mocking. 

This returnUrl is the url we want the user to eventually come back to once they are logged in. Once dex is integrated there will be other redirects before we come back to returnUrl.

darshansimha (approved):

PR #1228, opened 2023-10-18
feat: Dex Initial Intergration

jy4096:
This is an initial version for Dex Authentication integration. 
- Github oauth app is set for POC purpose only (user need to be added into the org to get authenticated)
- The Dex server is deployed manually.
- Dex reverse proxy

Flow:
- The "/login" returns the authCodeURL to frontend.
- The frontend will make a call to the authCodeURL, then return back to homepage/login with returned code and state params in the url.
- Then the frontend will call the "/callback" to set up the cookies

TODO (Will open an issue after GED to track the progress) :
- TLS
- Dex reverse proxy (improvement) 
- make Dex configurable 
- "state" use case
- add root cmd to start the Dex server instead of manually deploy it
- improvement: revisit set/verify cookie, avoid attack
- improvement: redirect to the previous page before login (instead of always to the homepage).
- improvement: refresh token/session, logout
- Tests
- Official Github oauth app for default setup


KeranYang (approved):

vigith:
just curious, without reverse-proxy how do you try it?

server/apis/v1/dexauth.go
```diff
@@ -0,0 +1,177 @@
...
+	issuerURL := "http://numaflow-dex-server:5556/dex"
+	provider, err := oidc.NewProvider(ctx, issuerURL)
+	if err != nil {
+		log.Fatalf("failed to query provider %q: %v", issuerURL, err)
+	}
+	verifier := provider.Verifier(&oidc.Config{ClientID: clientID})
+
+	return &DexObject{
+		clientID:       clientID,
+		clientSecret:   "ZXhhbXBsZS1hcHAtc2VjcmV0",
```
KeranYang, line 57: where is this secret coming from?

KeranYang (approved):

server/apis/v1/dexauth.go
jy4096: I copied and pasted from the Dex example app. It's hardcoded. I don't think we need this. Will clean up later as part of `make Dex configurable` TODO.

jy4096:
Update: add initial version of Dex Reverse Proxy.

PR #1229, opened 2023-10-18
feat: Styling the login page

darshansimha:
Screenshot 2023-10-18 at 5 06 51 PM


ui/src/components/pages/Login/index.tsx
```diff
@@ -26,23 +29,117 @@ export function Login() {
...
             style={{
               color: "#fff",
-              width: "82%",
               fontSize: "32px",
             }}
+            className="flex row"
           >
-            Unlock the power of data streaming with Numaflow!
-          </span>
+            Unlock the power of
```
bbehnke, line 44: Can you update this to be one div? Just need to set a max-width on it to get the stacking you want. 

Also, convert px to rem

bbehnke (changes requested):

bbehnke (approved):

PR #1230, opened 2023-10-18
chore: add app.kubernetes.io/name label

whynowy:
This is to make istio be able to pick a proper name to expose metrics.


pkg/reconciler/vertex/controller.go
```diff
@@ -265,6 +265,7 @@ func (r *vertexReconciler) reconcile(ctx context.Context, vertex *dfv1.Vertex) (
 			labels[dfv1.KeyPartOf] = dfv1.Project
 			labels[dfv1.KeyManagedBy] = dfv1.ControllerVertex
 			labels[dfv1.KeyComponent] = dfv1.ComponentVertex
+			labels[dfv1.KeyAppName] = vertex.Spec.PipelineName + "-" + vertex.Name
```
vigith, line 268: why does the `KeyAppName` have `vertex.Name`?

vigith (approved):

pkg/reconciler/vertex/controller.go
whynowy: Different vertex should have different name.

PR #1251, opened 2023-10-23
feat: allow specifying disable-auth when creating the server

KeranYang:
User can update the numaflow deployment to turn on and off AuthN/AuthZ by passing in an argument.

```
...
containers:
        - name: main
          image: quay.io/numaproj/numaflow:latest
          args:
            - server
            - '--disable-auth=true'
...
```

Tested by turning on and off the auth and verifying that after turning it off, my read-only role can perform write operations like CreatePipeline.


server/routes/routes.go
```diff
@@ -30,6 +30,8 @@ type SystemInfo struct {
 	ManagedNamespace string `json:"managedNamespace"`
 	Namespaced       bool   `json:"namespaced"`
 	Version          string `json:"version"`
+	DisableAuth      bool   `json:"disableAuth"`
```
whynowy, line 33: This `SystemInfo` is exposed to the client, do not add those information to this struct.

KeranYang: +1. I am going to create a separate struct AuthInfo to hold the `DisableAuth` and `DexServerAddr`, hence the `/systemInfo` API call won't expose auth configurations.

vigith:
isn't it better to disable auth by default, and enable only on-demand. which experience will be better for users testing it out?

config/base/numaflow-server/numaflow-server-deployment.yaml
```diff
@@ -38,6 +38,8 @@ spec:
           image: quay.io/numaproj/numaflow:latest
           args:
           - "server"
+          # By default, turn off authentication and authorization.
```
KeranYang, line 41: I made this change, thinking if we by default enable auth, it could add complexity to first-time numaflow users. @jy4096 also mentioned that we want to disable auth for e2e API testing.

KeranYang: cc: @vigith This also replies to your comment above. :)

cmd/commands/server.go
```diff
@@ -64,7 +64,7 @@ func NewServerCommand() *cobra.Command {
 	command.Flags().BoolVar(&namespaced, "namespaced", false, "Whether to run in namespaced scope, defaults to false.")
 	command.Flags().StringVar(&managedNamespace, "managed-namespace", sharedutil.LookupEnvStringOr("NAMESPACE", "numaflow-system"), "The namespace that the server watches when \"--namespaced\" is \"true\".")
 	command.Flags().StringVar(&baseHref, "base-href", "/", "Base href for Numaflow server, defaults to '/'.")
-	command.Flags().BoolVar(&disableAuth, "disable-auth", false, "Whether to disable authentication, defaults to false.")
+	command.Flags().BoolVar(&disableAuth, "disable-auth", true, "Whether to disable authentication and authorization, defaults to true for easy on-boarding.")
```
whynowy, line 67: It would be better to set the default to `false` here, but add the arg `true` in the manifests.

KeranYang: sounds good.

whynowy (approved):

PR #1259, opened 2023-10-24
chore: authenticate user before each api call

KeranYang:
* Add user authN to the middleware such that authN runs before each authZ.
* rename pkg auth to authz

### Testing
* Deployed locally and verified can login and logout.
* Also verified when invoking APIs, can reach the authorization part of the middleware.


server/routes/routes.go
```diff
@@ -165,6 +167,43 @@ func authMiddleware(enforcer *casbin.Enforcer) gin.HandlerFunc {
...
+func authenticate(c *gin.Context) (v1.CallbackResponse, error) {
+	// Validate the cookie exists
+	userIdentityTokenStr, err := c.Cookie("user-identity-token")
+	if err != nil {
+		return v1.CallbackResponse{}, err
+	}
+	userIdentityToken := v1.GetUserIdentityToken(userIdentityTokenStr)
+
+	clientID := "example-app"
+	issuerURL := "https://numaflow-server:8443/dex"
```
vigith, line 179: we will have to get these from configs?

whynowy:
@jy4096 - review?

whynowy (approved):

PR #1261, opened 2023-10-25
chore: cleanup and test fix

jy4096:


server/apis/v1/noauthhandler.go
```diff
@@ -14,8 +14,12 @@ type noAuthHandler struct {
 
 // NewNoAuthHandler is used to provide a new instance of the handler type
 func NewNoAuthHandler(serverAddr string, proxyAddr string) (*noAuthHandler, error) {
+	dexObj, err := NewDexObject(serverAddr, proxyAddr)
```
vigith, line 17: Why does NoAuth need dex?

vigith (approved):

server/apis/v1/noauthhandler.go
jy4096: "no auth" means the three pages (/login, /logout, /callback) that can access even if the user don't have auth...

PR #1263, opened 2023-10-25
fix: fixed the ns-summary page to allow creation of pipeline when no …

darshansimha:
Screenshot 2023-10-25 at 12 02 40 PM

### DIFF
--- a/ui/src/components/pages/Namespace/index.tsx
+++ b/ui/src/components/pages/Namespace/index.tsx
@@ -13,8 +13,27 @@ import { AppContextProps } from "../../../types/declarations/app";
 import { SidebarType } from "../../common/SlidingSidebar";
 import { NamespacePipelineListing } from "./partials/NamespacePipelineListing";
 import { ErrorDisplay } from "../../common/ErrorDisplay";
+import { NamespaceSummaryData } from "../../../types/declarations/namespace";
 
 import "./style.css";
+
+const defaultNamespaceSummaryData: NamespaceSummaryData = {
+  pipelinesCount: 0,
+  pipelinesActiveCount: 0,
+  pipelinesInactiveCount: 0,
+  pipelinesHealthyCount: 0,
+  pipelinesWarningCount: 0,
+  pipelinesCriticalCount: 0,
+  isbsCount: 0,
+  isbsActiveCount: 0,
+  isbsInactiveCount: 0,
+  isbsHealthyCount: 0,
+  isbsWarningCount: 0,
+  isbsCriticalCount: 0,
+  pipelineSummaries: [],
+  pipelineRawData: null,
+  isbRawData: null,
+};
 
 export function Namespaces() {
   const { namespaceId } = useParams();
@@ -25,7 +44,6 @@ export function Namespaces() {
       loadOnRefresh: false,
       addError,
     });
-
   const handleK8sEventsClick = useCallback(() => {
     if (!namespaceId || !setSidebarProps) {
       return;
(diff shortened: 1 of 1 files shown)


ui/src/components/pages/Namespace/index.tsx
```diff
@@ -15,6 +15,7 @@ import { NamespacePipelineListing } from "./partials/NamespacePipelineListing";
 import { ErrorDisplay } from "../../common/ErrorDisplay";
 
 import "./style.css";
+import { NamespaceSummaryData } from "../../../types/declarations/namespace";
```
bbehnke, line 18: move import above style import

ui/src/components/pages/Namespace/index.tsx
```diff
@@ -157,18 +218,10 @@ export function Namespaces() {
...
-          title="Error loading namespace pipelines"
-          message="No resources found."
-        />
-      );
-    }
     return (
       <NamespacePipelineListing
         namespace={namespaceId || ""}
-        data={data}
+        data={data ? data : ({} as NamespaceSummaryData)}
```
bbehnke, line 224: Please provide a proper NamespaceSummaryData object here when data is not available. Just set all values to 0 and empty arrays.

there is some data references in NamespacePipelineListing, want to avoid any undefined usage errors such as here: https://github.com/numaproj/numaflow/blob/f5ea9c1b8fcd22165e0a001c5dbf1ab9411d684f/ui/src/components/pages/Namespace/partials/NamespacePipelineListing/index.tsx#L98

ui/src/components/pages/Namespace/index.tsx
```diff
@@ -60,7 +61,67 @@ export function Namespaces() {
       ];
     }
     if (!data) {
-      return [];
+      return [
```
bbehnke, line 64: Instead of duplicating the whole array. Just use a "default" version of data in the existing array that is built. 

i.e  
let dataToUse = data;
if (!dataToUse) {
 dataToUse = defaultData; // 0s and empty arrays
}
return //existing array but using "dataToUse" instead of the original data object.

bbehnke: This "default" data can be stored as a const up top and also used down on line 224

bbehnke (changes requested):

ui/src/components/pages/Namespace/index.tsx
darshansimha: Done!

ui/src/components/pages/Namespace/index.tsx
darshansimha: Done

ui/src/components/pages/Namespace/index.tsx
darshansimha: Done!

ui/src/components/pages/Namespace/index.tsx
```diff
@@ -35,6 +35,81 @@ export function Namespaces() {
       k8sEventsProps: { namespaceId },
     });
   }, [namespaceId, setSidebarProps]);
+  const defaultPipelinesData = [
```
bbehnke, line 38: Move these up outside of the component, with names like DEFAULT_PIPELINES_DATA. 

In current state, it recreates them on each render which may cause unneeded re-renders

bbehnke (changes requested):

ui/src/components/pages/Namespace/index.tsx
darshansimha: The default data needs handleK8sEventsClick method, I can move this inside a useMemo, that would be better.

darshansimha: Done!

bbehnke (approved):

PR #1286, opened 2023-10-27
refactor: create interfaces for AuthN and AuthZ

KeranYang:
This change defines interfaces for AuthN and AuthZ, which helps modularize the auth components in a way that each package takes care of its own business and has minimal knowledge of the implementation details of dependencies. The goal is to hide complexity and reduce engineer cognitive load.

### DIFF
--- a/docs/user-guide/user-defined-functions/map/examples.md
+++ b/docs/user-guide/user-defined-functions/map/examples.md
@@ -1,0 +1,119 @@
+# Map Examples
+
+Please read [map](./map.md) to get the best out of these examples.
+
+## Prerequisites
+
+### Inter-Step Buffer Service (ISB Service)
+
+#### What is ISB Service?
+An Inter-Step Buffer Service is described by a [Custom Resource](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/), which is used to pass data between vertices of a numaflow pipeline.
+Please refer to the doc [Intern-Step Buffer Service](../../../core-concepts/inter-step-buffer.md) for more information on ISB.
+
+
+#### How to install the ISB Service
+
+```shell
+kubectl apply -f https://raw.githubusercontent.com/numaproj/numaflow/stable/examples/0-isbsvc-jetstream.yaml
+```
+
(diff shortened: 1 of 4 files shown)


server/authz/route_map.go
```diff
@@ -20,20 +20,20 @@ package authz
...
-		RequiresAuth: requiresAuth,
+		Object:        object,
+		RequiresAuthZ: requiresAuthZ,
 	}
 }
 
-// RouteMap is a map of routes to their corresponding RouteInfo objects. This map is used to certain
-// information about the route.
+// RouteMap is a map of routes to their corresponding RouteInfo objects.
+// This map is used to save certain information about the route.
```
vigith, line 36: lets be explicit about what information :)

vigith (approved):
LGTM

server/apis/v1/dexauthn.go
```diff
@@ -98,8 +98,24 @@ func (d *DexObject) oauth2Config(scopes []string) (*oauth2.Config, error) {
 	}, nil
 }
 
-// Verify is used to validate the user ID token.
-func (d *DexObject) Verify(ctx context.Context, rawIDToken string) (*oidc.IDToken, error) {
+func (d *DexObject) Authenticate(c *gin.Context) (authn.UserIdInfo, error) {
```
whynowy, line 101: Call it `UserInfo`?

server/apis/v1/dexauthn.go
```diff
@@ -98,8 +98,8 @@ func (d *DexObject) oauth2Config(scopes []string) (*oauth2.Config, error) {
 	}, nil
 }
 
-func (d *DexObject) Authenticate(c *gin.Context) (authn.UserIdInfo, error) {
-	userInfo := authn.UserIdInfo{}
+func (d *DexObject) Authenticate(c *gin.Context) (authn.UserInfo, error) {
```
whynowy, line 101: BTW - the common practice is returning `(pointer, error)`, which makes more sense.

KeranYang: +1.

whynowy (approved):

PR #1288, opened 2023-10-27
feat: add scopes to authorization

kohlisid:


config/base/numaflow-server/numaflow-server-rbac-config.yaml
```diff
@@ -9,6 +9,13 @@ data:
     p, role:readonly, *, *, GET
     # Groups go here
     # g, admin, role:admin
+    g, sidhant.kohli@gmail.com, role:admin
     # g, my-github-org:my-github-team, role:readonly
   rbac.conf: |
     policy.default: role:readonly
+    // The scopes field controls which authentication scopes to examine during rbac enforcement.
```
whynowy, line 16: Does `#` work for the comments? Having both `#` and `//` in the same config map for comments is very confusing.

server/authz/consts.go
```diff
@@ -20,6 +20,10 @@ const (
 	// PolicyMapPath is the path to the policy map.
 	policyMapPath = "/etc/numaflow/rbac-policy.csv"
 
+	// RbacPropertiesPath is the path to the rbac properties file. It includes configuraion for authorization like
+	// scope, default policy etc.
+	RbacPropertiesPath = "/etc/numaflow/rbacConf.yaml"
```
vigith, line 25: should it be exported?

server/authz/rbac.go
```diff
@@ -20,49 +20,71 @@ import (
...
+	"github.com/spf13/viper"
+
+	"github.com/numaproj/numaflow/pkg/shared/logging"
+	"github.com/numaproj/numaflow/server/authn"
 )
 
 var (
 	//go:embed rbac-model.conf
 	rbacModel string
+	logger    = logging.NewLogger().Named("server")
```
vigith, line 38: why is the logger in the module scope? also, why not use https://github.com/numaproj/numaflow/blob/main/pkg/shared/logging/log.go#L28 ?

server/authz/rbac.go
```diff
@@ -153,3 +175,82 @@ func extractObject(c *gin.Context) string {
...
+}
+
+// GetRbacScopes returns the scopes from the rbac properties file. If no scopes are provided, it returns Group as the
+// default scope. The scopes are used to determine the user identity token to be used for authorization.
+// If the scope is group, the user identity token will be the groups assigned to the user from the authentication
+// system. If the scope is email, the user identity token will be the email assigned to
+// the user from the authentication.
+// The scopes are provided as a comma separated list in the rbac properties file.
+// Example: policy.scopes=groups,email
+func GetRbacScopes(config *viper.Viper) []string {
```
vigith, line 205:
```suggestion
func GetRBACScopes(config *viper.Viper) []string {
```

[initialism](https://zchee.github.io/golang-wiki/CodeReviewComments/#initialisms)

server/authz/consts.go
kohlisid: Not using it outside scope after change, removing

server/authz/interface.go
```diff
@@ -1,12 +1,28 @@
...
+	// Authorize also returns the policy count of the user. The policy count is used to check if there are any policies
+	// defined for the given user, if not we will allocate a default policy for the user.
+	Authorize(c *gin.Context, userIdentityToken *authn.UserInfo, scope string) (bool, int)
+	// GetConfig returns the config file of the authorizer. We use a config file to store the policy params
+	// so that we can change the policy without restarting the server. The config file is in the format of yaml.
+	// The config file is read by viper.
+	GetConfig() *viper.Viper
+	// GetScopes returns the scopes of the authorizer. The scopes are used to check the authentication params to check
+	// if the user is authorized to access the resource.
+	GetScopes() []string
```
KeranYang, line 27: Keep the interface simple. Please see if we can keep only one function `Authorize` for this interface. Try to move GetConfig and GetScopes to under Authorize such that the interface is deep. At routes.go, we should be able to use this interface as simply as follow:

```
if this request needs to be authorized
    authorizer.authorize(gin, userInfo)
    (next or abort)
else
   c.next
```

The goal is to completely hide the authorization implementation details from the caller(routes.go). Current implementation expose such knowledge as Config and Scope to routes.go

vigith: +1

server/authz/rbac.go
```diff
@@ -153,3 +216,98 @@ func extractObject(c *gin.Context) string {
...
+// enforceCheck checks if the user has permission based on the Casbin model and policy.
+// It returns true if the user is authorized, otherwise false.
+func enforceCheck(enforcer *casbin.Enforcer, user, resource, object, action string) bool {
+	ok, _ := enforcer.Enforce(user, resource, object, action)
+	return ok
+}
+
+// ConfigFileReload is used to reload the config file when it is changed. This is used to reload the policy without
+// restarting the server. The config file is in the format of yaml. The config file is read by viper.
+func ConfigFileReload(e fsnotify.Event) {
```
KeranYang, line 261: `configFileReload`

server/authz/rbac.go
```diff
@@ -20,16 +20,28 @@ import (
...
 
 var (
 	//go:embed rbac-model.conf
-	rbacModel string
+	rbacModel     string
+	logger        = logging.NewLogger()
+	userPermCount map[string]int
+	currentScopes []string
+	policyDefault string
+	configReader  *viper.Viper
```
KeranYang, line 44: Can we make them instance variables belonging to the casbin struct instead of global? They are data that are specific to the object and are used across multiple methods inside the struct. Having too many global variables can make code harder to maintain and understand.

server/authz/rbac.go
```diff
@@ -45,24 +57,75 @@ func NewCasbinObject() (*CasbinObject, error) {
...
 	return &CasbinObject{
 		enforcer: enforcer,
 	}, nil
 }
 
-func (cas *CasbinObject) Authorize(c *gin.Context, groups []string) (bool, error) {
+// Authorize checks if a user is authorized to access the resource.
+// It returns true if the user is authorized, otherwise false.
+// It also returns the policy count of the user. The policy count is used to check if there are any policies defined
+// for the given user, if not we will allocate a default policy for the user.
```
KeranYang, line 84: Please update the description.

server/authz/rbac.go
```diff
@@ -45,24 +57,75 @@ func NewCasbinObject() (*CasbinObject, error) {
...
+func (cas *CasbinObject) Authorize(c *gin.Context, userIdentityToken *authn.UserInfo) bool {
```
KeranYang, line 85:
```suggestion
func (cas *CasbinObject) Authorize(c *gin.Context, userInfo *authn.UserInfo) bool {
```

server/authz/rbac.go
```diff
@@ -45,24 +57,75 @@ func NewCasbinObject() (*CasbinObject, error) {
...
-			return true, nil
+	userHasPolicies := false
+	// Check for the given scoped list if the user is authorized using any of the subjects in the list.
+	for _, scopedSubject := range scopedList {
+		// Check if the user has permissions in the policy for the given scoped subject.
+		userHasPolicies = userHasPolicies || getPermissionCount(cas.enforcer, scopedSubject)
+		ok := enforceCheck(cas.enforcer, scopedSubject, resource, object, action)
+		if ok {
+			return ok
 		}
```
KeranYang, line 100: nit, same for below
```suggestion
		if ok := enforceCheck(cas.enforcer, scopedSubject, resource, object, action); ok {
			return ok
		}
```

server/authz/rbac.go
```diff
@@ -153,3 +216,98 @@ func extractObject(c *gin.Context) string {
...
+	defaultPolicy := getRbacProperty(RbacPropertyDefaultPolicy, config)
+	if defaultPolicy == emptyString {
+		return emptyString
+	}
+	return defaultPolicy.(string)
+}
+
+// getPermissionCount returns if there are permissions defined for a user in the RBAC policy.
+// If the user has permissions in the policy, it returns true, otherwise false.
+func getPermissionCount(enforcer *casbin.Enforcer, user string) bool {
```
KeranYang, line 291: This func performs both read and write. While the name of it indicates readonly. It looks we are updating the user permission cache, can we move the update to somewhere else?

The func also returns a boolean, could you update the name to reflect it? usually a function returning a boolean is called isABC or hasABC etc.

server/authz/rbac.go
```diff
@@ -153,3 +216,98 @@ func extractObject(c *gin.Context) string {
...
+// getDefaultPolicy returns the default policy from the rbac properties file. The default policy is used when the
+// requested resource is not present in the policy.
+// The default policy is provided in the rbac properties file in the format "policy.default: value"
+// Example: policy.default: deny
+func getDefaultPolicy(config *viper.Viper) string {
+	defaultPolicy := getRbacProperty(RbacPropertyDefaultPolicy, config)
+	if defaultPolicy == emptyString {
+		return emptyString
+	}
+	return defaultPolicy.(string)
```
KeranYang, line 286: would it work?

```suggestion
	return getRbacProperty(RbacPropertyDefaultPolicy, config)
```

KeranYang (commented):
Thanks for making the change Sidhant, I made some comments, please take a look. I had a hard time reviewing the rbac file, probably due to my limited knowledge on casbin. As we keep adding logics to our RBAC component, when we have time, let's look into adding unit tests to help ensure accuracy and readability.

server/authz/rbac.go
```diff
@@ -45,24 +57,75 @@ func NewCasbinObject() (*CasbinObject, error) {
...
+	var scopedList []string
+	// If the scope is group, fetch the groups from the user identity token.
+	if slices.Contains(scopes, ScopeGroup) {
+		scopedList = append(scopedList, userIdentityToken.IDTokenClaims.Groups...)
+	}
+	// If the scope is email, fetch the email from the user identity token.
+	if slices.Contains(scopes, ScopeEmail) {
+		scopedList = append(scopedList, userIdentityToken.IDTokenClaims.Email)
+	}
+	return scopedList
```
KeranYang, line 128: What if a group is named "abc@gmail.com"?

server/authz/rbac.go
kohlisid: We return an interface from the `getRbacProperty `, would want to cast that to string before sending. 
Or can it be done without that?

server/authz/rbac.go
kohlisid: If there is a policy added for `"[abc@gmail.com](mailto:abc@gmail.com)"` then it can work.
So the difference from the authorization side is which param to read from the authentication token. We do not add any type enforcing

server/authz/rbac.go
KeranYang: sorry I meant would `return getRbacProperty(RbacPropertyDefaultPolicy, config).(string)` work? meaning if it's empty string, it also fits into this single statement.

kohlisid: I have removed the extra check for emptyString

server/authz/rbac.go
kohlisid: Moved to Casbin Struct

server/authz/interface.go
```diff
@@ -1,12 +1,17 @@
...
+import (
+	"github.com/gin-gonic/gin"
+
+	"github.com/numaproj/numaflow/server/authn"
+)
 
 type Authorizer interface {
 	// Authorize checks if a user is authorized to access the resource.
 	// c is the gin context.
 	// g is the list of groups the user belongs to.
```
KeranYang, line 12: we can also remove these 2 lines

server/authz/rbac.go
```diff
@@ -20,49 +20,91 @@ import (
...
+
+	"github.com/numaproj/numaflow/pkg/shared/logging"
+	"github.com/numaproj/numaflow/server/authn"
 )
 
 var (
 	//go:embed rbac-model.conf
-	rbacModel string
+	rbacModel     string
+	logger        = logging.NewLogger()
```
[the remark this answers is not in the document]
kohlisid: Will check in refactor later

KeranYang (approved):

PR #1290, opened 2023-10-29
refactor: unified metrics names for forwarders

whynowy:
Metrics names for source, sink, map and reduce forwarders are different, which makes it difficult to set up unified dashboard for the metrics of each vertex. For example, there's no way to do same prometheus query on the processing rate of each vertex by using the `read_total`.

This PR makes each forwarder use the same names for common metrics, such as `read_total`, `write_total`, `ack_total`, etc.


pkg/daemon/server/service/rater/rater.go
```diff
@@ -217,15 +218,9 @@ func sleep(ctx context.Context, duration time.Duration) {
...
-		keyVertexTypeSource: "source_forwarder_read_total",
-		keyVertexTypeSink:   "sink_forwarder_data_read",
-	}
-
-	readTotalMetricName, ok := metricNames[vertexType]
-	if !ok {
-		readTotalMetricName = "forwarder_data_read"
+	readTotalMetricName := "forwarder_data_read_total"
+	if keyVertexTypeSource == vertexType {
+		readTotalMetricName = "forwarder_read_total"
```
KeranYang, line 223: For source, we use `forwarder_read_total` instead of `forwarder_data_read_total`. Is it because currently source only processes real data messages hence the numbers are the same between these two metrics? If so I wonder if we can also use `forwarder_data_read_total` for source, since we plan to look into source idling and could introduce other non-data messages.

pkg/reduce/data_forward.go
```diff
@@ -464,12 +464,13 @@ messagesLoop:
...
-		metrics.LabelPipeline:           df.pipelineName,
-		metrics.LabelVertexReplicaIndex: strconv.Itoa(int(df.vertexReplica)),
-	}).Observe(float64(time.Since(startTime).Milliseconds()))
+	defer func(t time.Time) {
+		metrics.PBQWriteTime.With(map[string]string{
+			metrics.LabelVertex:             df.vertexName,
+			metrics.LabelPipeline:           df.pipelineName,
+			metrics.LabelVertexReplicaIndex: strconv.Itoa(int(df.vertexReplica)),
+		}).Observe(float64(time.Since(t).Milliseconds()))
+	}(time.Now())
```
KeranYang, line 473: nice, I was under an impression that `time.Now()` is evaluated when defer function executes. It turns out it's evaluated immediately when encountering the defer statement. Learnt something new.

pkg/reduce/pnf/processandforward.go
```diff
@@ -92,12 +92,13 @@ func newProcessAndForward(ctx context.Context,
...
-		metrics.LabelVertex:             p.vertexName,
-		metrics.LabelPipeline:           p.pipelineName,
-		metrics.LabelVertexReplicaIndex: strconv.Itoa(int(p.vertexReplica)),
-	}).Observe(float64(time.Since(startTime).Milliseconds()))
+	defer func(t time.Time) {
+		metrics.ReduceProcessTime.With(map[string]string{
+			metrics.LabelVertex:             p.vertexName,
+			metrics.LabelPipeline:           p.pipelineName,
+			metrics.LabelVertexReplicaIndex: strconv.Itoa(int(p.vertexReplica)),
+		}).Observe(float64(time.Since(t).Milliseconds()))
```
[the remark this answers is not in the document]
KeranYang: Was it wrong because there could be tiny time diffs between `startTime := time.Now()` and `defer`?

pkg/daemon/server/service/rater/rater.go
whynowy: > For source, we use `forwarder_read_total` instead of `forwarder_data_read_total`. Is it because currently source only processes real data messages hence the numbers are the same between these two metrics? If so I wonder if we can also use `forwarder_data_read_total` for source, since we plan to look into source idling and could introduce other non-data messages.

Wan't it you added this logic? 

There's no `data_read` metric exposed in the source.

pkg/metrics/metrics_server.go
```diff
@@ -0,0 +1,291 @@
+/*
```
whynowy, line 1: No change, but file rename.

pkg/reduce/pnf/processandforward.go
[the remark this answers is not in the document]
whynowy: > Was it wrong because there could be tiny time diffs between `startTime := time.Now()` and `defer`?

Assume you are good with it now?

pkg/daemon/server/service/rater/rater.go
KeranYang: Yes, I added this logic. Now that we are unifying the metrics, I think we can change the source forwarder to also use `data_read`. There are two benefits: 1. it potentially paves the road for future source idling implementations and 2. with all vertices sharing the same metric name, we can completely remove the `vertexType` part of the raptor podkey, simplifying the code.

yhl25:
I support this change this will be more convenient for the users when creating dashboards. Should we consider adding a vertexType tag to the metrics?

vigith (approved):

KeranYang (approved):

whynowy:
> I support this change. With this change, it will be more convenient for the users when creating dashboards. Should we consider adding a vertexType tag to the metrics?



> I support this change. With this change, it will be more convenient for the users when creating dashboards. Should we consider adding a vertexType tag to the metrics?



> I support this change. With this change, it will be more convenient for the users when creating dashboards. Should we consider adding a vertexType tag to the metrics?

Added.

PR #1291, opened 2023-10-30
chore: set stream.replicas=1 when in non-cluster mode

whynowy:
When jetstream is not in cluster mode, automatically sets the stream/KV replica to 1. @joelcomp1 

```golang
	if r.isbSvc.Spec.JetStream.GetReplicas() < 3 {
		// Replica can not > 1 with non-cluster mode
		v.Set("otbucket.replicas", 1)
		v.Set("procbucket.replicas", 1)
		v.Set("stream.replicas", 1)
	}
```

### DIFF
--- a/pkg/reconciler/isbsvc/controller.go
+++ b/pkg/reconciler/isbsvc/controller.go
@@ -52,60 +52,60 @@ func NewReconciler(client client.Client, kubeClient kubernetes.Interface, scheme
 }
 
 func (r *interStepBufferServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	isbs := &dfv1.InterStepBufferService{}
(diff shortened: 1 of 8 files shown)


joelcomp1:
So is that KV store configuration different then the one here: https://github.com/numaproj/numaflow/blob/86df4a84083613cf3912f35ada400434ce1b200c/pkg/isbsvc/jetstream_service.go#L100

must be?

vigith (approved):

whynowy:
> So is that KV store configuration different then the one here:
> 
> https://github.com/numaproj/numaflow/blob/86df4a84083613cf3912f35ada400434ce1b200c/pkg/isbsvc/jetstream_service.go#L100
> 
> must be?

We can use the stream.replicas for sideinput KV as well for now, can always revisit later one.

PR #1292, opened 2023-10-30
feat: added separate colors for sideInput and dynamic legend

veds-g:
- Added separate color for each generator nodes as well as for the corresponding input to the vertices
image

- Added dynamic legend (only shows the nodes present in graph)
image

- Fix graph height to occupy 100% area, when summary is collapsed / uncollapsed
image
image

- Added tooltip for graph interaction buttons
image

- Styled the collapsed summary on pipeline view and adjusted the play/pause buttons, error panel and graph interaction panel
image


ui/src/components/common/SummaryPageLayout/index.tsx
```diff
@@ -301,6 +305,7 @@ export function SummaryPageLayout({
   ]);
 
   const contentMargin = useMemo(() => {
+    if (collapsable) return 0;
```
bbehnke, line 308: Confused about this change. We want "contentMargin" to be the number of px to push the main content down so the summary isn't overlaying it. 

Wouldn't this always return 0 for the pipeline's usage, given it is "collapsable"? 

Looking further into the changes, I see what you are doing. I think it would be best to have SummaryPageLayout still apply the marginTop as it did before. If you want it to not apply any margin when it collapses, just pass `offsetOnCollapse` as false. 

Adding it to context to adjust your other items (buttons and etc.) is fine, but i think the main layout margin should be controlled here.

bbehnke (changes requested):

ui/src/components/pages/Pipeline/partials/Graph/partials/CustomNode/index.tsx
```diff
@@ -45,6 +52,21 @@ const CustomNode: FC<NodeProps> = ({
     sideInputEdges,
   } = useContext<HighlightContextProps>(HighlightContext);
 
+  const generatorToColorMap: Map<string, string> = useContext(
+    GeneratorColorContext
+  );
+
+  const getSideInputColor = (nodeName: string) => {
```
bbehnke, line 59: useCallback needed

bbehnke, line 59: useCallback needed

bbehnke: Or just put outside of the component so it is not in render loop.

ui/src/components/pages/Pipeline/partials/Graph/partials/CustomNode/index.tsx
```diff
@@ -98,11 +120,18 @@ const CustomNode: FC<NodeProps> = ({
     return {
       border: `${isSelected(
         highlightValues[data?.name] && highlightValues[text]
-      )} ${getBorderColor(data?.type)}`,
+      )} ${getSideInputColor(data?.name)}`,
     };
   };
 
   if (data?.type === "sideInput") {
+    const generatorImage = {
```
bbehnke, line 128: +1 can be moved outside of component (const at top of file)

ui/src/components/pages/Pipeline/partials/Graph/partials/CustomNode/index.tsx
```diff
@@ -329,22 +358,30 @@ const CustomNode: FC<NodeProps> = ({
...
-                left: `${50 - idx * 10}%`,
+                left: `${50 - idx * 9}%`,
               }}
             />
           );
         })}
       </div>
-      {data?.nodeInfo?.sideInputs?.map((_, idx) => {
+      {data?.nodeInfo?.sideInputs?.map((input, idx) => {
+        const inputImage = {
```
bbehnke, line 368: +1 outside of component

bbehnke (changes requested):

bbehnke (approved):

PR #1296, opened 2023-10-30
docs: need metrics server

vigith:


docs/quick-start.md
```diff
@@ -22,6 +22,13 @@ Then use one of the following options to create a local Kubernete Cluster:
 
 You will also need `kubectl` to manage the cluster. [Follow these steps to install kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/). In case you need a refresher, all the `kubectl` commands used in this quick start guide can be found in the [kubectl Cheat Sheet](https://kubernetes.io/docs/reference/kubectl/cheatsheet/).
 
+### Metrics Server
```
whynowy, line 25: For quick start, shall we move this to the place before talking about the UI? (Too many installation steps scale ppl)

Also use some soft words like `You might need to install the metrics server xxxxx`, and  give the explicit commands in this doc instead of asking ppl to switch to a different page.

whynowy (approved):

PR #1298, opened 2023-10-31
chore: remove unwanted locks

yhl25:


whynowy (approved):
Lock is not needed?

yhl25:
> Lock is not needed?

No, we don't require a lock here. Underneath, NATS KV is essentially a stream and since the NATS connection is thread-safe, we should be in a good. Additionally, the lock was necessary for the timer, but since it's specific to the watcher, I've moved it inside the `Watch()` method.

PR #1302, opened 2023-10-31
chore: update rater package name

KeranYang:
* rename the package from server to rater.
* add doc.go
* let pod tracker own parsing the pod key to pod information, the rater doesn't need to know how to parse a key using separator

### DIFF
--- a/pkg/daemon/server/service/rater/pod_tracker.go
+++ b/pkg/daemon/server/service/rater/pod_tracker.go
@@ -131,22 +131,22 @@ func (pt *PodTracker) GetActivePodsCount() int {
 	return pt.activePods.Length()
 }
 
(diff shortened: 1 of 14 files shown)


pkg/daemon/server/service/rater/pod_tracker.go
```diff
@@ -131,9 +131,28 @@ func (pt *PodTracker) GetActivePodsCount() int {
 	return pt.activePods.Length()
 }
 
+// PodInfo represents the information of a pod that is used for tracking the processing rate
+type PodInfo struct {
```
whynowy, line 135: `podInfo`?

vigith (approved):

PR #1303, opened 2023-10-31
fix: fixed the timer not clearing issue

darshansimha:
This fixes the timer not clearing issue

### DIFF
--- a/ui/src/components/pages/Pipeline/partials/Graph/index.tsx
+++ b/ui/src/components/pages/Pipeline/partials/Graph/index.tsx
@@ -174,10 +174,22 @@ const Flow = (props: FlowProps) => {
     undefined
   );
   const [statusPayload, setStatusPayload] = useState<any>(undefined);
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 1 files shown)


ui/src/components/pages/Pipeline/partials/Graph/index.tsx
```diff
@@ -176,9 +177,12 @@ const Flow = (props: FlowProps) => {
   const [statusPayload, setStatusPayload] = useState<any>(undefined);
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const [timerDateStamp, setTimerDateStamp] = useState<any>(undefined);
-  const [timer, setTimer] = useState<any>(undefined);
+  const timer = useRef<any>(undefined);
```
bbehnke, line 180: I think the state you had before was fine. Would be type `number | undefined`.

darshansimha: No, state was the actual problem, the state was changing before the react did batch updates. THat's the reason I went with ref.

bbehnke (approved):

PR #1305, opened 2023-10-31
fix: Split cookie to meet the cookie length requirement

jy4096:
- Split cookie to meet the cookie length requirement
- Remove dex-proxy-addr

### DIFF
--- a/server/routes/routes.go
+++ b/server/routes/routes.go
@@ -37,7 +37,6 @@ type SystemInfo struct {
 type AuthInfo struct {
 	DisableAuth   bool   `json:"disableAuth"`
 	DexServerAddr string `json:"dexServerAddr"`
-	DexProxyAddr  string `json:"dexProxyAddr"`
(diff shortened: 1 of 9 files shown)


server/routes/routes.go
```diff
@@ -66,7 +66,7 @@ func Routes(r *gin.Engine, sysInfo SystemInfo, authInfo AuthInfo, baseHref strin
 		// Add the AuthN/AuthZ middleware to the group.
 		r1Group.Use(authMiddleware(authorizer, dexObj))
 	}
-	v1Routes(r1Group)
+	v1Routes(r1Group, dexObj)
```
whynowy, line 69: Curious, when auth is disabled, what will be the dexObj? will it have any impact?

whynowy (approved):

server/routes/routes.go
jy4096: updated....

whynowy (approved):

PR #1310, opened 2023-11-01
fix: logout fix

jy4096:


server/common/cookie.go
```diff
@@ -90,7 +91,7 @@ func JoinCookies(key string, cookieList []*http.Cookie) (string, error) {
 		}
 		sb.WriteString(strings.Join(parts[1:], ":"))
 	} else {
-		return "", fmt.Errorf("invalid cookie for key %s", key)
+		return "", fmt.Errorf("%s:%s", InvalidCookieError, key)
```
whynowy, line 94: Use `fmt.Errorf("xxxxx %w", InvalidCookieError)`, so that `errors.As()` can be used for error checking.

jy4096: I want to include the "key" also........

server/apis/v1/noauthhandler.go
```diff
@@ -51,15 +51,17 @@ func (h *noAuthHandler) Callback(c *gin.Context) {
 func (h *noAuthHandler) Logout(c *gin.Context) {
 	cookies := c.Request.Cookies()
 	tokenString, err := common.JoinCookies(common.UserIdentityCookieName, cookies)
-	if err != nil {
+	if strings.HasPrefix(err.Error(), common.InvalidCookieError) {
```
whynowy, line 54: Use `errors.As/Is` to check, and no else?

whynowy (approved):

PR #1327, opened 2023-11-02
chore: fix update vertex validation

kohlisid:
Fix validation for single vertex update, does not update pipeline when validation requested.
Add mutex to permCount map


vigith (approved):
this `dryrun` flag is required because we have one end point for validate and update, correct?

kohlisid:
@vigith Yes correct! 
This check was missing in updateVertex thus we were updating the pipeline even while validating

jy4096 (approved):

PR #1328, opened 2023-11-02
fix: allow user to update isbsvc metadata labels

jy4096:


server/apis/v1/handler.go
```diff
@@ -547,6 +547,7 @@ func (h *handler) UpdateInterStepBufferService(c *gin.Context) {
 		return
 	}
 	isbSVC.Spec = updatedSpec.Spec
+	isbSVC.ObjectMeta.Labels = updatedSpec.ObjectMeta.Labels
```
whynowy, line 550: Hold this on. Let's revisit it later.

jy4096: 👌

PR #1329, opened 2023-11-02
RC2.0 UI fixes

mshakira:
- fixing sorting bug when new ISB is created no last updated time is available and it leads to crashing
- grey out status images when count is 0.
- change Active/Non-active to Running/Not-Running


bbehnke (approved):
LGTM

PR #1335, opened 2023-11-03
chore: start args in a configmap

whynowy:
With more starting args introduced, updating the deployment args becomes inconvenient. Change to support both configmap and args (recommend to use configmap, but args take precedence).

Use a single optional configmap for both controller and UX server.


whynowy:
Do not merge, need to think through.

vigith (approved):

whynowy:
This is good to review. @numaproj/numaflow-dev

config/base/shared-config/numaflow-cmd-params-config.yaml
```diff
@@ -0,0 +1,31 @@
...
+metadata:
+  name: numaflow-cmd-params-config
+data:
+  ### Whether to run the controller and the UX server in namespaced scope, defaults to false.
+  # namespaced: "false"
+  #
+  ### The namespace that the controller and the UX server watch when "namespaced" is true.
+  # managed.namespace: numaflow-system
+  #
+  ### Where to disable leader election for the controller, defaults to false
```
KeranYang, line 12:
```suggestion
  ### Whether to disable leader election for the controller, defaults to false
```

docs/operations/installation.md
```diff
@@ -74,11 +100,16 @@ To do managed namespace installation, besides `--namespaced`, add `--managed-nam
...
-```
-      name: NUMAFLOW_LEADER_ELECTION_DISABLED
-      value: "true"
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: numaflow-cmd-params-config
+data:
+  # Where to disable leader election for the controller, defaults to false
```
KeranYang, line 111:
```suggestion
  # Whether to disable leader election for the controller, defaults to false
```

KeranYang (approved):

vigith (approved):

PR #1339, opened 2023-11-03
fix: removed latest from isb create

darshansimha:


ui/src/components/common/SlidingSidebar/partials/ISBCreate/index.tsx
```diff
@@ -22,7 +22,7 @@ metadata:
   name: default
 spec:
   jetstream:
-    version: latest # Do NOT use "latest" in real deployment, check "numaflow-controller-config" ConfigMap to get available versions.
+    version: 2.9.22 #check "numaflow-controller-config" ConfigMap to get available versions.
```
whynowy, line 25: Do `2.10.3` for now.

PR #1345, opened 2023-11-03
Upstreammain

mshakira:
- fix isb service data access from spec when it is available
- add isb image
Screenshot 2023-11-03 at 1 53 59 PM


vigith (commented):
i think you can remove the previous ISBSvc image?

vigith:
> i think you can remove the previous ISBSvc image?

nm, we were using the pipeline image

jy4096 (approved):

darshansimha (approved):

PR #1354, opened 2023-11-09
chore: updated testing libraries + unit test for graph component

veds-g:
updating testing libraries
- testing-library/jest-dom
- testing-library/react
- testing-library/user-event

image


ui/src/components/pages/Pipeline/partials/Graph/index.tsx
```diff
@@ -175,7 +175,7 @@ const Flow = (props: FlowProps) => {
   );
   const [statusPayload, setStatusPayload] = useState<any>(undefined);
   const [timerDateStamp, setTimerDateStamp] = useState<any>(undefined);
-  const [timer, setTimer] = useState<number | undefined>(undefined);
```
darshansimha, line 178: Why is this any, because timer will return number and when it is cleared it will be undefined

veds-g: had set it as any because, we assign a pauseTimer variable of type `NodeJS.Timer` inside `handleTimer` function

veds-g: so either we can add `NodeJS.Timer` in addition to the above two or keep it as `any`

darshansimha (approved):
LGTM

PR #1358, opened 2023-11-14
fix: add more checks to isbsvc validation

KeranYang:
1. Use IsDNS1053Label to validate the ISBSVC name.
2. Replace gin ShouldBindJson with the json decoder to decode the request body to the spec object. json decoder supports disallowing unknown fields. This catches typos such as `replica`.
3. Disallow updating ISBSVC name and namespace from UI.

### DIFF
--- a/server/apis/v1/handler.go
+++ b/server/apis/v1/handler.go
@@ -470,7 +470,7 @@ func (h *handler) CreateInterStepBufferService(c *gin.Context) {
 		return
 	}
 
-	isValid := validateISBSVCSpec(h, nil, &isbsvcSpec, ValidTypeCreate)
+	isValid := validateISBSVCSpec(nil, &isbsvcSpec, ValidTypeCreate)
 	if isValid != nil {
 		h.respondWithError(c, fmt.Sprintf("Failed to create interstepbuffer service spec, %s", isValid.Error()))
 		return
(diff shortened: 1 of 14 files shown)


server/apis/v1/handler.go
```diff
@@ -969,11 +952,20 @@ func validatePipelineSpec(h *handler, oldPipeline *dfv1.Pipeline, newPipeline *d
...
 // validateISBSVCSpec is used to validate the ISB service spec
-func validateISBSVCSpec(h *handler, prevSpec *dfv1.InterStepBufferService,
+func validateISBSVCSpec(prevSpec *dfv1.InterStepBufferService,
 	newSpec *dfv1.InterStepBufferService, validType string) error {
-	ns := newSpec.Namespace
-	isbClient := h.numaflowClient.InterStepBufferServices(ns)
-	valid := validator.NewISBServiceValidator(h.kubeClient, isbClient, prevSpec, newSpec)
+	// UI-specific validations: updating namespace and name from UI is not allowed
+	if validType == ValidTypeUpdate && prevSpec != nil {
+		if prevSpec.Namespace != newSpec.Namespace {
```
whynowy, line 959: What happens if `namespace` is not specified in the new spec?

KeranYang: Thanks for catching this edge case. If a user doesn't specify the namespace, it will be an empty string. K8S treats empty namespace as the same as "default". I updated the PR to reflect this.

whynowy: This is not right. 

When you create an object in a namespace, if namespace filed is missing, it should use the specified namespace.

KeranYang: Yes exactly. When creating, I implemented it to use the specified namespace.

KeranYang: To summarize, if a user creates a brand-new spec and the namespace field is missing, we use the namespace from the URL parameter. If a user updates an existing spec, if the namespace field is missing, it means the user purposely removed the namespace field, in this case, we assume the user wants to use the "default" namespace and validate it against the previous namespace.

Let me know if this logic needs to be revised. @whynowy

whynowy: I think the logic for updating should be:

If namespace is missing, we always use the namespace in the URL - which means we do not need to validate it. This is the same behavior as we using kubectl to update something.

whynowy (approved):

PR #1359, opened 2023-11-14
chore: add unit tests for authZ

kohlisid:


server/authz/testData/test-policy-map.csv
```diff
@@ -0,0 +1,9 @@
+# Policies go here
```
whynowy, line 1: To be consistent with the test data folder in ./test/, use `testdata` as the folder name.

server/authz/rbac_test.go
```diff
@@ -0,0 +1,298 @@
...
+	authorizer, err := NewCasbinObject(WithPolicyMap(testPolicyMapPath), WithPropertyFile(testPropertyFilePath))
+	assert.NoError(t, err)
+
+	// Test that the authorizer is not nil
+	assert.NotNil(t, authorizer)
+
+	// Check the default policy, which is set to "role:readonly" in the test data
+	defaultPolicy := authorizer.getDefaultPolicy()
+	assert.Equal(t, "role:readonly", defaultPolicy)
+
```
KeranYang, line 94: nit: I think we can consider removing this check as it's indirectly covered by the `Authorize` test right below. `getDefaultPolicy` is the implementation details of `Authorize` and we should aim to test what(behaviour) instead of how(implementation).

server/authz/rbac_test.go
```diff
@@ -0,0 +1,298 @@
...
+	// GET requests are for the test user as part of the read only group
+	isAuth = authorizer.Authorize(getRequest, userInfo)
+	assert.True(t, isAuth)
+
+	// POST requests are allowed for the test user email
+	postRequest := mockHttpRequest("POST", "")
+	isAuth = authorizer.Authorize(postRequest, userInfo)
+	assert.True(t, isAuth)
+
+	// PATCH requests are allowed for the test user email
```
KeranYang, line 150:
```suggestion
	// PATCH requests are NOT allowed for the test user email
```

server/authz/rbac_test.go
KeranYang: Same for other test cases. We don't necessarily need to call `get` to verify policy or scope.

server/authz/rbac_test.go
```diff
@@ -0,0 +1,298 @@
...
+
+	// get a test gin context for GET request
+	// The default policy allows only GET requests
+	// So, the GET request should be authorized
+	getRequest := mockHttpRequest("GET", "")
+	isAuth := authorizer.Authorize(getRequest, userInfo)
+	assert.True(t, isAuth)
+
+	// Change the scopes to "email"
+	authorizer.currentScopes = []string{"email"}
```
KeranYang, line 136: use setCurrentScope

server/authz/rbac_test.go
```diff
@@ -0,0 +1,298 @@
...
+	// Request for a namespace that is authorized for the user
+	getRequest = mockHttpRequest("GET", "test_ns")
+	isAuth = authorizer.Authorize(getRequest, userInfo)
+	assert.True(t, isAuth)
+
+}
+
+// TestConfigFileReload is a test implementation of the ConfigFileReload functionality.
+// It tests that the RBAC properties are reloaded correctly when the config file is changed.
+func TestConfigFileReload(t *testing.T) {
```
KeranYang, line 181: A better approach could be to verify the `Authorize` behaviour change after the config file reload instead of checking scopes.

vigith (commented):
good job on writing a well-commented code 😄

server/authz/rbac_test.go
kohlisid: So once the config file is reloaded it will update the scopes and the authorize flow would be the same as the TestScopes where we check auth based on scope change, so I thought that we are testing that through the TestScopes test case. 
What do you think?

server/authz/rbac_test.go
kohlisid: Moved them to TestCreateAuthorizer

server/authz/rbac_test.go
KeranYang: I see. The real scenario would be there is a file reload which changes the scope and hence changes the authorization result. Hence I think a single unit test which reloads file and verifies the authorization result change will cover both file reload and scope change.

KeranYang: My main point across most of the comments is to hide implementation details from unit tests. And I think things like getScope and getPolicy are all implementation details. We should test against Authorize and Authorize only.

KeranYang: I am also okay with the current structure. It already covers all the cases. Feel free to disagree and merge.

KeranYang (approved):

server/authz/rbac_test.go
kohlisid: Thanks for this Keran!

I was considering the path where we would try to test the smallest units/functions using individual UTs to see if they are working as expected. 
Considering the testing of Authorize only would probably at an interface level where we just want to verify if the Auth works as expected. So maybe at an E2E level would make more sense or some UTs  at a package where we do not care about which type of Authorization implementation is used. (Like today we are just using 1 kind, we might add more impls later)
Here we might want to verify if our overall RBAC implementation is covered, like in the file reload we could have scenarios where the property is given an un-parsable value, so our code should be covering and tested for such cases using UTs.

This was my understanding for this, would love to discuss further to understand more and probably even come up with a more streamlined approach for our UTs as we go forward in our development. Given that we're 1.0 already, we definitely need to make our testing framework more robust

whynowy (approved):

server/authz/rbac_test.go
KeranYang: Sounds good Sidhant! We can discuss this more offline. Let's make sure we have good UT coverage and at the same time, make the UTs simple.

PR #1360, opened 2023-11-15
chore: update java e2e test images

KeranYang:
Explain what this PR does.


whynowy:
Maybe we should give a tag `stable` for the image, so that we do not need to update it in the test cases.

vigith (approved):

vigith:
> Maybe we should give a tag `stable` for the image, so that we do not need to update it in the test cases.

+1

PR #1376, opened 2023-11-17
fix: disallow updating an existing isbsvc's persistence strategy

KeranYang:
Don't allow updating an existing isbsvc's persistence strategy. This includes removing, adding and modifying.

### DIFF
--- a/webhook/validator/isbsvc.go
+++ b/webhook/validator/isbsvc.go
@@ -63,7 +63,7 @@ func (v *isbsvcValidator) ValidateUpdate(_ context.Context) *admissionv1.Admissi
 			return DeniedResponse("can not change ISB Service type from Redis to Jetstream")
 		}
 		// check the persistence of ISB Service is not changed
-		if !equality.Semantic.DeepEqual(v.oldISBService.Spec.Redis.Native.Persistence, v.newISBService.Spec.Redis.Native.Persistence) {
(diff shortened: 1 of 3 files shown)


webhook/validator/isbsvc.go
```diff
@@ -42,17 +43,28 @@ func (v *isbsvcValidator) ValidateCreate(_ context.Context) *admissionv1.Admissi
...
+			return DeniedResponse("can not change persistence of Jetstream ISB Service")
 		}
 	case v.oldISBService.Spec.Redis != nil:
+		// check the type of ISB Service is not changed
 		if v.newISBService.Spec.JetStream != nil {
-			return DeniedResponse("Can not change ISB Service type from Redis to Jetstream")
+			return DeniedResponse("can not change ISB Service type from Redis to Jetstream")
+		}
+		// check the persistence of ISB Service is not changed
+		if !equality.Semantic.DeepEqual(v.oldISBService.Spec.Redis.Native.Persistence, v.newISBService.Spec.Redis.Native.Persistence) {
```
whynowy, line 66: Need to check if `native` is nil.

KeranYang: Thanks! I am adding a nil check before deepequal. Do we need to handle `native` changing from nil to non-nil and vice versa?

whynowy: It's needed.

KeranYang: done.

whynowy (approved):

PR #1385, opened 2023-11-23
chore!: Generate Idle Watermark if the source is idling. Fixes #633

### LINKED ISSUE #633: Idle Source

chandankumar4:
Explain what this PR does.
- Generate the idle Watermark if the source is idling

Fixes #633

Example Pipeline to test the changes:
```yaml
kind: Pipeline
metadata:
  name: even-odd-sum
spec:
  watermark:
    idleSource:
      maxWait: 5s
      minIncrement: 5s
  vertices:
    - name: in
      scale:
        min: 2
      source:
        http: {}
    - name: atoi
      scale:
        min: 1
      udf:
        container:
          # Tell the input number is even or odd, see https://github.com/numaproj/numaflow-go/tree/main/pkg/mapper/examples/even_odd
          image: quay.io/numaio/numaflow-go/map-even-odd:v0.5.0
    - name: compute-sum
      udf:
        container:
          # compute the sum
          image: quay.io/numaio/numaflow-go/reduce-sum:v0.5.0
        groupBy:
          window:
            fixed:
              length: 60s
          keyed: true
          storage:
            persistentVolumeClaim:
              volumeSize: 10Gi
              accessMode: ReadWriteOnce
      partitions: 2
    - name: sink-log
      scale:
        min: 1
      sink:
        log: {}
  edges:
    - from: in
      to: atoi
    - from: atoi
      to: compute-sum
    - from: compute-sum
      to: sink-log

```

Note:
This will not work in below case:
- In case of Kafka with 12 partition and 6 partition is assigned to one pod and rest 6 to another, now If only 3 of them is receiving the data out of 6 partition.


pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -499,6 +499,19 @@ type Watermark struct {
 	// +kubebuilder:default="0s"
 	// +optional
 	MaxDelay *metav1.Duration `json:"maxDelay,omitempty" protobuf:"bytes,2,opt,name=maxDelay"`
+	// IdleSource defines the idle watermark properties, it could be configured in case source is idling.
+	// +kubebuilder:default={"maxWait": "-1", "minIncrement": "-1"}
+	// +optional
+	IdleSource *IdleSource `json:"idleSource,omitempty" protobuf:"bytes,3,opt,name=idleSource"`
+}
+
+type IdleSource struct {
```
vigith, line 508: we will have to add one more wait, which is wait after detecting idle source

pkg/sources/forward/data_forward.go
```diff
@@ -204,6 +211,31 @@ func (isdf *DataForward) forwardAChunk(ctx context.Context) {
...
+			// Get the head watermark value then add the minIncrement to it and publish the idle watermark with updated value.
+			headWatermark := isdf.wmFetcher.ComputeHeadWatermark(0)
+			updatedWM := headWatermark.Add(minIncrement)
+
+			isdf.srcWMPublisher.PublishIdleWatermarks(updatedWM)
+			// set the time for last updated watermark which will be used for calculating maximum wait time is passed
+			// or not in case of idle watermark.
+			isdf.lastUpdateWM = time.Now()
+			// currentWatermark is used to calculate when watermark is idling
+			isdf.currentWatermark = isdf.wmFetcher.ComputeWatermark(nil, 0)
```
yhl25, line 233: add a comment about `isWatermarkIdle`, also move this inside the if check, in line 241

pkg/sources/udsource/user_defined_source.go
```diff
@@ -194,6 +194,10 @@ func (u *userDefinedSource) PublishSourceWatermarks(msgs []*isb.ReadMessage) {
 	}
 }
 
+func (u *userDefinedSource) PublishIdleWatermarks(wm time.Time) {
+	// won't be used
+}
+
```
yhl25, line 200: iterate over the publishers and publish the idle watermark

pkg/watermark/fetch/edge_fetcher_set.go
```diff
@@ -59,6 +59,11 @@ func NewEdgeFetcherSet(ctx context.Context, vertexInstance *dfv1.VertexInstance,
 	}
 }
 
+func (efs *edgeFetcherSet) ComputeHeadWatermark(fromPartitionIdx int32) wmb.Watermark {
+	// won't be used
+	return wmb.Watermark{}
+}
```
yhl25, line 65: to avoid this, lets introduce you separate interface for source fetcher 

```
// SourceFetcher fetches watermark data for source vertex.
type SourceFetcher interface {
	// ComputeWatermark computes a valid watermark for the given offset on the given partition
	ComputeWatermark(offset isb.Offset, fromPartitionIdx int32) wmb.Watermark
	// ComputeHeadIdleWMB computes a valid head idle WMB for the given partition.
	ComputeHeadIdleWMB(fromPartitionIdx int32) wmb.WMB
	// ComputeHeadWatermark computes a valid head watermark for the given partition
	ComputeHeadWatermark(fromPartitionIdx int32) wmb.Watermark
}
```

pkg/sources/udsource/user_defined_source.go
chandankumar4: Publisher is not an array to iterate. I think it should be like below?
```
	publisher := u.loadSourceWatermarkPublisher(0)
	publisher.PublishWatermark(wmb.Watermark(t), nil, 0) // Source publisher does not care about the offset
```

yhl25: https://github.com/numaproj/numaflow/blob/112bc3b67ddc7433ac50988a16725853d91d32f2/pkg/sources/udsource/user_defined_source.go#L70
It is a map, we can iterate over the  map and publish the watermarks

pkg/sources/forward/data_forward.go
```diff
@@ -108,14 +111,18 @@ func NewDataForward(
 		vertexName:           vertexInstance.Vertex.Spec.Name,
 		pipelineName:         vertexInstance.Vertex.Spec.PipelineName,
 		vertexReplica:        vertexInstance.Replica,
+		Watermark:            vertexInstance.Vertex.Spec.Watermark,
```
KeranYang, line 114:
```suggestion
		watermarkConfig:            vertexInstance.Vertex.Spec.Watermark,
```

pkg/sources/forward/data_forward.go
```diff
@@ -63,8 +63,11 @@ type DataForward struct {
 	vertexName     string
 	pipelineName   string
 	vertexReplica  int32
+	Watermark      dfv1.Watermark
 	// idleManager manages the idle watermark status.
-	idleManager wmb.IdleManager
+	idleManager      wmb.IdleManager
+	lastUpdateWM     time.Time
+	currentWatermark wmb.Watermark
```
KeranYang, line 70: We have been adding attributes to the data forwarder and it's now taking care of too many things, especially watermark-related logics. I am wondering if we can move these attributes to the watermark publisher itself. That way we delegate all the watermark publishing logics to the publisher hence data forwarder doesn't need to know about how an idle source watermark is published.

Such delegation makes our system more modularized and potentially easier to unit test. 

Ideally, it would be great if we could have the following statement:

```
if len(readMesages) == 0 {
  isdf.srcWMPublisher.PublishIdleWatermarks()
}
```

and let srcWMPublisher take care of tracking updatedWM etc.

pkg/sources/forward/data_forward_test.go
```diff
@@ -845,6 +849,10 @@ func (f mySourceForwardTest) ApplyTransform(ctx context.Context, message *isb.Re
 type TestSourceWatermarkPublisher struct {
 }
 
+func (p TestSourceWatermarkPublisher) PublishIdleWatermarks(t time.Time) {
+	// PublishIdleWatermarks is not tested in data_forwarder_test.go
```
KeranYang, line 853: can we add unit tests and one e2e test for source idling? can be in a separate PR.

pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -499,6 +499,19 @@ type Watermark struct {
...
+	// IdleSource defines the idle watermark properties, it could be configured in case source is idling.
+	// +kubebuilder:default={"maxWait": "-1s", "minIncrement": "-1s"}
+	// +optional
+	IdleSource *IdleSource `json:"idleSource,omitempty" protobuf:"bytes,3,opt,name=idleSource"`
+}
+
+type IdleSource struct {
+	// MaxWait is the wait duration before progressing the watermark in idle case.
+	// +kubebuilder:default="-1s"
+	MaxWait *metav1.Duration `json:"maxWait,omitempty" protobuf:"bytes,1,opt,name=maxWait"`
```
whynowy, line 511: protobuf index is not right.

pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -499,6 +499,19 @@ type Watermark struct {
 	// +kubebuilder:default="0s"
 	// +optional
 	MaxDelay *metav1.Duration `json:"maxDelay,omitempty" protobuf:"bytes,2,opt,name=maxDelay"`
+	// IdleSource defines the idle watermark properties, it could be configured in case source is idling.
+	// +kubebuilder:default={"maxWait": "-1s", "minIncrement": "-1s"}
```
whynowy, line 503: What is the reason to use as `-1s` as the default?

pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -509,6 +522,30 @@ func (wm Watermark) GetMaxDelay() time.Duration {
...
+
+	return nil
+}
+
+func (is IdleSource) GetMaxWait() time.Duration {
+	if is.MaxWait != nil {
+		return is.MaxWait.Duration
+	}
+
+	return time.Duration(-1)
```
whynowy, line 538: The default value is different from `-1s` defined above.

pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -509,6 +522,30 @@ func (wm Watermark) GetMaxDelay() time.Duration {
...
+}
+
+func (is IdleSource) GetMinIncrement() time.Duration {
+	if is.MinIncrement != nil {
+		return is.MinIncrement.Duration
+	}
+
+	return time.Duration(-1)
```
whynowy, line 546: same here.

pkg/apis/numaflow/v1alpha1/pipeline_types.go
chandankumar4: I'll update it, `-1` as a thought to put initially as a default value but I don't see any benefits for it.

pkg/apis/numaflow/v1alpha1/pipeline_types.go
chandankumar4: ohh, missed to update it while moving field arounds

pkg/sources/forward/data_forward_test.go
chandankumar4: yaa, I'm working on e2e test

pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -499,6 +499,19 @@ type Watermark struct {
...
+	// +kubebuilder:default={"maxWait": "-1s", "minIncrement": "-1s"}
+	// +optional
+	IdleSource *IdleSource `json:"idleSource,omitempty" protobuf:"bytes,3,opt,name=idleSource"`
+}
+
+type IdleSource struct {
+	// MaxWait is the wait duration before progressing the watermark in idle case.
+	// +kubebuilder:default="-1s"
+	MaxWait *metav1.Duration `json:"maxWait,omitempty" protobuf:"bytes,1,opt,name=maxWait"`
+	// MinIncrement is the time duration for increment the
```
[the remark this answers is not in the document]
chandankumar4: updated

pkg/sources/forward/data_forward.go
```diff
@@ -207,6 +214,30 @@ func (isdf *DataForward) forwardAChunk(ctx context.Context) {
...
+			if isWMIdle {
+				minIncrement := isdf.Watermark.IdleSource.GetMinIncrement()
+				// Get the head watermark value then add the minIncrement to it and publish the idle watermark with updated value.
+				headWatermark := isdf.wmFetcher.ComputeHeadWatermark(0)
+				updatedWM := headWatermark.Add(minIncrement)
+
+				isdf.srcWMPublisher.PublishIdleWatermarks(updatedWM)
+				// set the time for last updated watermark which will be used for calculating maximum wait time is passed
+				// or not in case of idle watermark.
+				isdf.lastUpdateWM = time.Now()
```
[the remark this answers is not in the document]
chandankumar4: probably yes, I'll check

pkg/sources/forward/data_forward.go
```diff
@@ -390,6 +433,31 @@ func (isdf *DataForward) forwardAChunk(ctx context.Context) {
...
+		maxWait := isdf.WatermarkConfig.IdleSource.GetMaxWait()
+		if maxWait != nil && time.Since(isdf.lastUpdateWM) > *maxWait {
+			computedWM := isdf.wmFetcher.ComputeWatermark(nil, 0)
+			if isdf.currentWatermark == nil {
+				isdf.currentWatermark = &computedWM
+			}
+			if computedWM == *isdf.currentWatermark {
+				// set idleWMFound time if not already set and watermark is found to be idle
+				if isdf.idleWMFound.IsZero() {
+					isdf.idleWMFound = time.Now()
```
KeranYang, line 451: isWatermarkIdle based on the name, is a READ(non-write) method which simply tells whether the source is idling. We have some write operations inside isWatermarkIdle which modify(WRITE) the states of the forwarder(L446, L451). Can we move them to the outside?

KeranYang (commented):
I put some comments. Please consider addressing https://github.com/numaproj/numaflow/pull/1385#discussion_r1405506612 . Or create a separate issue to track it. I believe there could be a better way of modularizing the idle source logics.

pkg/watermark/fetch/interface.go
```diff
@@ -29,6 +29,13 @@ type Fetcher interface {
...
 
+// SourceFetcher fetches watermark data for source vertex.
+type SourceFetcher interface {
+	Fetcher
+	// ComputeHeadWatermark computes a valid head watermark for the given partition
+	ComputeHeadWatermark(fromPartitionIdx int32) wmb.Watermark
+}
+
 // UXFetcher computes the watermark for Vn.
 type UXFetcher interface {
```
vigith, line 40:
```suggestion
// SourceFetcher fetches watermark data for source vertex.
type SourceFetcher interface {
	Fetcher
	HeadFetcher
}

// UXFetcher computes the watermark for Vn.
type HeadFetcher interface {
```

test/idle-source-e2e/testdata/idle-source-reduce-pipeline.yaml
```diff
@@ -0,0 +1,55 @@
+apiVersion: numaflow.numaproj.io/v1alpha1
+kind: Pipeline
+metadata:
+  name: idle-source
+spec:
+  watermark:
+    idleSource:
+      maxWait: 5s
+      minIncrement: 5s
+      maxDelay: 5s
```
vigith, line 10: fix me with new names

pkg/sources/udsource/user_defined_source.go
```diff
@@ -208,3 +208,10 @@ func (u *userDefinedSource) loadSourceWatermarkPublisher(partitionID int32) publ
 	u.srcWMPublishers[partitionID] = sourcePublishWM
 	return sourcePublishWM
 }
+
+// PublishIdleWatermarks will publish the watermark when source is idling
+func (u *userDefinedSource) PublishIdleWatermarks(t time.Time) {
+	for partitionID, publisher := range u.srcWMPublishers {
+		publisher.PublishWatermark(wmb.Watermark(t), nil, partitionID)
```
vigith, line 215:
```suggestion
		publisher.PublishIdleWatermark(wmb.Watermark(t), nil, partitionID)
```

vigith (commented):
please generalize PublishIdleWatermarks across native and UDSource.

chandankumar4:
> I put some comments. Please consider addressing [#1385 (comment)](https://github.com/numaproj/numaflow/pull/1385#discussion_r1405506612) . Or create a separate issue to track it. I believe there could be a better way of modularizing the idle source logics.

Created an issue to track it: https://github.com/numaproj/numaflow/issues/1396

pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -499,6 +499,21 @@ type Watermark struct {
 	// +kubebuilder:default="0s"
 	// +optional
 	MaxDelay *metav1.Duration `json:"maxDelay,omitempty" protobuf:"bytes,2,opt,name=maxDelay"`
+	// IdleSource defines the idle watermark properties, it could be configured in case source is idling.
+	// +optional
+	IdleSource *IdleSource `json:"idleSource,omitempty" protobuf:"bytes,3,opt,name=idleSource"`
+}
+
+type IdleSource struct {
+	// Threshold is the duration in seconds after which a source is marked a Idle due to lack of data.
```
whynowy, line 508: Not `in seconds`.

pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -499,6 +499,21 @@ type Watermark struct {
...
+	// Ex: If watermark found to be idle after the Threshold duration then the watermark is progressed by `IncrementBy`.
+	Threshold *metav1.Duration `json:"threshold,omitempty" protobuf:"bytes,1,opt,name=threshold"`
+	// StepInterval is the duration in seconds between the subsequent increment of the watermark as long the source remains Idle.
```
whynowy, line 511: Not `in seconds`.

pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -499,6 +499,21 @@ type Watermark struct {
...
+	// +kubebuilder:default="0s"
+	// +optional
+	StepInterval *metav1.Duration `json:"stepInterval,omitempty" protobuf:"bytes,2,opt,name=stepInterval"`
+	// IncrementBy is the duration in seconds to be added to the current watermark to progress the watermark when source is idling.
```
whynowy, line 515: Not `in seconds`.

pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -509,6 +524,37 @@ func (wm Watermark) GetMaxDelay() time.Duration {
 	return time.Duration(0)
 }
 
+func (wm Watermark) GetIdleSource() *IdleSource {
```
whynowy, line 527: What is the point of this function? Isn't it the same as `return wm.IdleSource`?

test/idle-source-e2e/testdata/idle-source-reduce-pipeline.yaml
whynowy: Wrong indents for `maxDelay`.

pkg/apis/numaflow/v1alpha1/pipeline_types.go
chandankumar4: yaa, it's same as `wm.IdleSource` also it's not used anywhere, I'll remove it

pkg/sources/forward/data_forward.go
```diff
@@ -207,6 +216,32 @@ func (isdf *DataForward) forwardAChunk(ctx context.Context) {
...
 	// There is a natural looping here if there is an internal error while reading, and we are not able to proceed.
 	if len(readMessages) == 0 {
+		// calculate that watermark is idling or not for the given maxWait duration
+		// If no any message is coming and watermark is idling then start increasing the watermark with "minIncrement" config
+		// value after waiting for each "maxWait" duration.
+		if isdf.isWatermarkIdle() {
+			// wait for maxWait duration before publishing the idle watermark with MinIncrement value.
+			if time.Since(isdf.idleWMFound) < isdf.WatermarkConfig.IdleSource.GetMaxDelay() {
+				return
+			}
```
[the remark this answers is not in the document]
chandankumar4: isWatermarkIdle() is only check for idling and this piece of code is for "StepInterval" after idling found that's why I have put it outside

pkg/sources/forward/data_forward.go
```diff
@@ -283,6 +318,7 @@ func (isdf *DataForward) forwardAChunk(ctx context.Context) {
 	}
 	// publish source watermark
 	isdf.srcWMPublisher.PublishSourceWatermarks(transformedReadMessages)
+	isdf.lastUpdateWM = time.Now()
```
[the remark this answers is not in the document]
chandankumar4: I'll move it all to a separate fn

pkg/sources/forward/data_forward.go
```diff
@@ -207,6 +216,32 @@ func (isdf *DataForward) forwardAChunk(ctx context.Context) {
...
+			updatedWM := headWatermark.Add(*minIncrement)
+			// check if updated watermark is greater than the current time then reset the updated watermark with current time.
+			if updatedWM.After(time.Now()) {
+				updatedWM = time.Now()
+			}
+			isdf.srcWMPublisher.PublishIdleWatermarks(updatedWM)
+			// set the time for last updated watermark which will be used for calculating maximum wait time is passed?
+			isdf.lastUpdateWM = time.Now()
+			// reset the currentWatermark to nil which is used to calculate when watermark is idling.
+			isdf.currentWatermark = nil
```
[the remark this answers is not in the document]
chandankumar4: I can't fetch the WM value and set it here because it's containing the older valuer after publishing the new WM also, In the next iteration only I'm able to fetch updated value of WM

I'm not sure what's the reason behind it but that's what I have encountered!!!

pkg/sources/forward/data_forward.go
chandankumar4: I can move it to a different fn and call if after idle found

vigith (changes requested):
we need to refactor the source watermark interface a bit

pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -499,6 +499,21 @@ type Watermark struct {
...
+type IdleSource struct {
+	// Threshold is the duration after which a source is marked a Idle due to lack of data.
+	// Ex: If watermark found to be idle after the Threshold duration then the watermark is progressed by `IncrementBy`.
+	Threshold *metav1.Duration `json:"threshold,omitempty" protobuf:"bytes,1,opt,name=threshold"`
+	// StepInterval is the duration between the subsequent increment of the watermark as long the source remains Idle.
+	// +kubebuilder:default="0s"
+	// +optional
+	StepInterval *metav1.Duration `json:"stepInterval,omitempty" protobuf:"bytes,2,opt,name=stepInterval"`
+	// IncrementBy is the duration to be added to the current watermark to progress the watermark when source is idling.
+	IncrementBy *metav1.Duration `json:"incrementBy,omitempty" protobuf:"bytes,3,opt,name=incrementBy"`
```
whynowy, line 516: Any validation on these new configs?

chandankumar4: Added

pkg/sources/generator/tickgen.go
```diff
@@ -217,42 +197,35 @@ func NewMemGen(
...
 	return mg.vertexName
 }
 
-// GetPartitionIdx returns the partition number for the source vertex buffer
-// Source is like a buffer with only one partition. So, we always return 0
-func (mg *memgen) GetPartitionIdx() int32 {
-	return 0
+// Partitions returns the partitions for the source.
+func (mg *memGen) Partitions() []int32 {
+	return []int32{mg.vertexInstance.Replica}
```
KeranYang, line 221: For single partition sources, we are using Replica as default index, would it be better if we just use `0`? It's better in a way that it explicitly shows `I am the first partition`.

pkg/sources/generator/tickgen.go
```diff
@@ -217,42 +197,35 @@ func NewMemGen(
 		}
 	}
 
-	// attach a source publisher so the source can assign the watermarks.
-	genSrc.sourcePublishWM = genSrc.buildSourceWatermarkPublisher(publishWMStores)
+	// create a source watermark publisher
+	sourceWmPublisher := publish.NewSourcePublish(ctx, genSrc.pipelineName, genSrc.vertexName, publishWMStores,
```
KeranYang, line 201: Same question as below, can we use `0` instead of Replica?

pkg/sources/kafka/handler.go
```diff
@@ -46,8 +46,15 @@ func newConsumerHandler(readChanSize int) *consumerHandler {
 
 // Setup is run at the beginning of a new session, before ConsumeClaim
 func (consumer *consumerHandler) Setup(sess sarama.ConsumerGroupSession) error {
+	consumer.logger.Info("setup was invoked for a new session")
+	for topic, partitions := range sess.Claims() {
+		println("topic: ", topic)
+		for _, partition := range partitions {
+			println("partition: ", partition)
+		}
+	}
```
KeranYang, line 55: nit: is this for debugging purpose? can we instead of using println, use logger? Same for below

pkg/sources/sourcer/sourcer.go
```diff
@@ -0,0 +1,51 @@
...
+	// Read reads a chunk of messages and returns at the first occurrence of an error. Error does not indicate that the
+	// array of result is empty, the callee should process all the elements in the array even if the error is set. Read
+	// will not mark the message in the buffer as "READ" if the read for that index is erring.
+	// There is a chance that we have read the message and the container got forcefully terminated before processing. To provide
+	// at-least-once semantics for reading, during the restart we will have to reprocess all unacknowledged messages.
+	Read(context.Context, int64) ([]*isb.ReadMessage, error)
+	// Ack acknowledges an array of offset.
+	Ack(context.Context, []isb.Offset) []error
+	// NoAck cancels acknowledgement of an array of offset.
+	NoAck(context.Context, []isb.Offset)
```
KeranYang, line 40: Do we NoAck for source?

pkg/sources/udsource/user_defined_source.go
```diff
@@ -130,15 +121,28 @@ func (u *userDefinedSource) GetName() string {
...
 func (u *userDefinedSource) Read(ctx context.Context, count int64) ([]*isb.ReadMessage, error) {
-	return u.sourceApplier.ApplyReadFn(ctx, count, u.readTimeout)
+	readMessages, err := u.sourceApplier.ApplyReadFn(ctx, count, u.readTimeout)
+	if err != nil {
+		return nil, err
+	}
+	for _, msg := range readMessages {
+		if _, ok := u.partitions[msg.ReadOffset.PartitionIdx()]; !ok {
+			u.partitions[msg.ReadOffset.PartitionIdx()] = true
+		}
```
KeranYang, line 143: nit: would it be simpler if we 
```suggestion
		u.partitions[msg.ReadOffset.PartitionIdx()] = true
```

pkg/watermark/publish/src_publisher.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/numaproj/numaflow/pkg/watermark/store"
+	"github.com/numaproj/numaflow/pkg/watermark/wmb"
+)
+
+// SourcePublisher publishes source watermarks based on a list of isb.ReadMessage
+// and also publishes idle watermarks. It internally calls the main WM Publisher to publish
+// the watermark. The only difference is that in source we do not care about offsets in the first
+// Publish within the source itself (huh? :-D). Also, when the source boots up, it has to load
+// the watermark information so it can know about the global source WM state.
+type SourcePublisher interface {
```
KeranYang, line 35: Nice abstraction!

pkg/shared/idlehandler/src_idlehandler.go
```diff
@@ -0,0 +1,83 @@
...
+	return true
+}
+
+// hasStepIntervalPassed verifies if the step interval has passed.
+func (iw *SrcIdleHandler) hasStepIntervalPassed() bool {
+	return time.Since(iw.lastIdleWmPublishedTime) >= iw.config.IdleSource.GetStepInterval()
+}
+
+// PublishSrcIdleWatermark publishes an idle watermark if the conditions are met for source.
+func (iw *SrcIdleHandler) PublishSrcIdleWatermark(partitions []int32) bool {
```
KeranYang, line 55: Nice abstraction. Closes #1396

pkg/sources/sourcer/sourcer.go
```diff
@@ -0,0 +1,51 @@
...
+	// Partitions returns the partitions of the source.
+	Partitions() []int32
```
KeranYang, line 42: Can we add more descriptions to this method? I have some questions about it. 1. could you help me confirm my understanding that calling this method multiple times can return different results (source partition list can dynamically change)? and 2. with this change, all udsources need to implement it too, what's the benefit of introducing this method?

pkg/sources/udsource/user_defined_source.go
yhl25: why perform a write/update operation if the key is already present?

pkg/sources/sourcer/sourcer.go
yhl25: 1. Yes it can result in different results for sources like Kafka which involves partition rebalancing when a pod comes up or gets deleted.
2. Yes, it should not be a complex operation since we already expect the users to provide partition information in the offset. The reason is when we are publishing idle watermarks we cannot publish them to all the partitions instead we should only publish idle watermarks to the partitions which are assigned/managed by this pod.

pkg/sources/sourcer/sourcer.go
yhl25: NoAck is not currently used or exposed for user-defined sources, it seems sensible to remove it for now. We can reintroduce it if needed in the future.

pkg/sources/generator/tickgen.go
yhl25: Using '0' for sources like HTTP and Tickgen doesn't work, as we utilize the pod replica as the partition ID to publish the watermark. This approach helps us maintain coordination between the pods for the progression of watermarks.

pkg/sources/generator/tickgen.go
yhl25: https://github.com/numaproj/numaflow/pull/1385#discussion_r1414901009

pkg/sinks/sink.go
```diff
@@ -191,8 +191,14 @@ func (u *SinkProcessor) Start(ctx context.Context) error {
...
 	}
+
+	// create lag readers from buffer readers
+	var lagReaders []isb.LagReader
+	for _, reader := range readers {
+		lagReaders = append(lagReaders, reader)
+	}
 	// start metrics server and pass the sinkHandler to it, so that it can be used to check the readiness of the sink
-	metricsOpts := metrics.NewMetricsOptions(ctx, u.VertexInstance.Vertex, []metrics.HealthChecker{sinkHandler}, readers)
+	metricsOpts := metrics.NewMetricsOptions(ctx, u.VertexInstance.Vertex, []metrics.HealthChecker{sinkHandler}, lagReaders)
```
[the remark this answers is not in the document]
yhl25: we cannot pass `[]isb.LagReader` in place of `[]isb.BufferReader`

pkg/reconciler/pipeline/validate.go
```diff
@@ -199,6 +199,28 @@ func ValidatePipeline(pl *dfv1.Pipeline) error {
...
+// if threshold is specified, incrementBy should be specified
+func validateIdleSource(pl dfv1.Pipeline) error {
+	if pl.Spec.Watermark.IdleSource != nil {
+		if pl.Spec.Watermark.IdleSource.Threshold != nil {
+			if pl.Spec.Watermark.IdleSource.IncrementBy == nil {
+				return fmt.Errorf("invalid idle source watermark config, if threshold is specified, incrementBy should be specified")
+			}
+		}
+
+		if pl.Spec.Watermark.IdleSource.Threshold.Duration < pl.Spec.Watermark.IdleSource.IncrementBy.Duration {
```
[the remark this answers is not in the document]
chandankumar4: no, I'm adding those validation check as well

pkg/reconciler/pipeline/validate.go
```diff
@@ -199,6 +199,28 @@ func ValidatePipeline(pl *dfv1.Pipeline) error {
...
+		return err
+	}
+
+	return nil
+}
+
+// validateIdleSource validates the idle source watermark config.
+// The threshold should be greater than or equal to incrementBy.
+// if threshold is specified, incrementBy should be specified
+func validateIdleSource(pl dfv1.Pipeline) error {
```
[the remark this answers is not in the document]
chandankumar4: Added

pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -509,6 +524,29 @@ func (wm Watermark) GetMaxDelay() time.Duration {
 	return time.Duration(0)
 }
 
+func (is IdleSource) GetThreshold() time.Duration {
+	if is.Threshold != nil {
+		return is.Threshold.Duration
+	}
+
+	return time.Duration(0)
+}
```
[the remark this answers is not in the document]
yhl25: yes only len of read messages is zero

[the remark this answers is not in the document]
KeranYang: Got it. If threshold is not specified, every time source forwarder receives an empty batch, we publish idle watermark. Might worth documenting it.

pkg/sources/sourcer/sourcer.go
KeranYang: +1. Thanks for the explanation!

pkg/apis/numaflow/v1alpha1/pipeline_types.go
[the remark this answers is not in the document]
vigith: default value should not be 0, let's make sure of that in the validator. user has to specify a value.

pkg/sources/udsource/user_defined_source.go
yhl25:
```
func BenchmarkMapUpdate(b *testing.B) {
	tempMap := make(map[string]struct{})

	for i := 0; i < b.N; i++ {
		tempMap["test"] = struct{}{}
	}
	b.ReportAllocs()
}

func BenchmarkMapReadAndUpdate(b *testing.B) {
	tempMap := make(map[string]struct{})

	for i := 0; i < b.N; i++ {
		if _, ok := tempMap["test"]; !ok {
			tempMap["test"] = struct{}{}
		}
	}
	b.ReportAllocs()
}

//goos: darwin
//goarch: arm64
//pkg: github.com/numaproj/numaflow/pkg/window/strategy/sliding
//BenchmarkMapUpdate
//BenchmarkMapUpdate-10    	154240488	         7.747 ns/op	       0 B/op	       0 allocs/op
//PASS

//goos: darwin
//goarch: arm64
//pkg: github.com/numaproj/numaflow/pkg/window/strategy/sliding
//BenchmarkMapReadAndUpdate
//BenchmarkMapReadAndUpdate-10    	442961821	         2.655 ns/op	       0 B/op	       0 allocs/op
//PASS
```

KeranYang: Thanks for running the benchmark. This proves that `checking before adding` have a performance advantage in the case of a large map with many key collisions. I still think for our case(small map with many key collision), `directly adding` is equally efficient(or less efficient but neglect-able) and preferred for simplicity and clarity in code.

Feel free to disagree and proceed with current approach. No strong opinion.

test/idle-source-e2e/testdata/idle-source-reduce-pipeline.yaml
```diff
@@ -0,0 +1,55 @@
+apiVersion: numaflow.numaproj.io/v1alpha1
+kind: Pipeline
+metadata:
+  name: idle-source
+spec:
+  watermark:
+    idleSource:
+      threshold: 5s
+      incrementBy: 3s
+      stepInterval: 2s
```
KeranYang, line 10: Can we add some comments on these numbers on how they are used to test source idling? The comment should make it clear why removing these configurations will break the test.

pkg/shared/idlehandler/src_idlehandler.go
vigith, line 1: can we add a unit test of idle handler?

pkg/sources/generator/tickgen.go
vigith: yes

pkg/sources/udsource/user_defined_source.go
```diff
@@ -130,15 +121,28 @@ func (u *userDefinedSource) GetName() string {
...
-	return u.sourceApplier.ApplyReadFn(ctx, count, u.readTimeout)
+	readMessages, err := u.sourceApplier.ApplyReadFn(ctx, count, u.readTimeout)
+	if err != nil {
+		return nil, err
+	}
+	for _, msg := range readMessages {
+		if _, ok := u.partitions[msg.ReadOffset.PartitionIdx()]; !ok {
+			u.partitions[msg.ReadOffset.PartitionIdx()] = struct{}{}
+		}
+	}
```
vigith, line 144: this is not right, we only append by never remove. we will have to ask users to add `Partitions()` to ud-source interface. 

@KeranYang WDYT?

pkg/watermark/publish/src_publisher.go
```diff
@@ -0,0 +1,123 @@
...
+	// will be used by http, nats and tickgen source whose partitionIdx is same
+	// as the vertex replica index
+	if sp.opts.defaultPartitionIdx != -1 {
+		entityName := fmt.Sprintf("%s-%s-%d", sp.pipelineName, sp.vertexName, sp.opts.defaultPartitionIdx)
+		processorEntity := entity.NewProcessorEntity(entityName)
+		// toVertexPartitionCount is 1 because we publish watermarks within the source itself.
+		sourcePublishWM := NewPublish(sp.ctx, processorEntity, sp.srcPublishWMStores, 1, IsSource(), WithDelay(sp.opts.delay))
+		sp.sourcePublishWMs[sp.opts.defaultPartitionIdx] = sourcePublishWM
+	}
+
```
vigith, line 80: please document `else` branch

vigith (commented):
need few changes as discussed.

pkg/sources/udsource/user_defined_source.go
vigith: #1405

KeranYang: I agree this is not right. With this change we now have two sources of truth on how many partitions a source has. One is the read batch and the other is the source Partitions() interface.

The removing part can be tricky because if we remove a certain partition, we also need to remove the corresponding watermark KV store etc. We can look into details as part of issue 1405 that you created.

pkg/sinks/sink.go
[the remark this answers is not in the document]
KeranYang: Resolving my comment here as in Go, slices of different interface types are NOT directly assignable.

test/idle-source-e2e/testdata/idle-source-reduce-pipeline.yaml
KeranYang: I am un-resolving this one. We configured the idle source, updated the simple keyed reduce test with a new start time and new wait time before sending next batch.

I have a hard time understanding how this test works in terms of verifying idle source works. Can we write some comments on why this test is testing the source idling functionality? How these numbers `5s`, `3s`, `2s`, `startTime := 100, `time.Sleep(10 * time.Millisecond)` and `WithTimeout(120*time.Second)` work together to achieve idle source test?

KeranYang (approved):
I have only [one minor comment](https://github.com/numaproj/numaflow/pull/1385#discussion_r1417990391) open. Approving.

vigith (approved):

test/idle-source-e2e/testdata/idle-source-reduce-pipeline.yaml
vigith: #1409

Issue #1357, opened 2023-11-14
Source transformer should progress the watermark even if the data is filtered by source-transformer


vigith:
# Summary

We should publish and progress via WMB with idle=true even if the data is dropped by the source-transformer.

KeranYang:
### Reproducing the problem

Trying to reproduce using the following pipeline spec:

```
spec:
  vertices:
    - name: in
      source:
        generator:
          rpu: 5
          duration: 1s
          msgSize: 8
        transformer:
          builtin:
            name: filter
            kwargs:
              expression: 1 == 0
      scale: {}
    - name: cat
      udf:
        container: null
        builtin:
          name: cat
    - name: out
      sink:
        log: {}
  edges:
    - from: in
      to: cat
    - from: cat
      to: out
```

I observed that although all the messages are filtered out by the source transformer, we are still publishing idle watermarks.

![Screenshot 2023-11-19 at 9 51 11 PM](https://github.com/numaproj/numaflow/assets/13165977/f9514ecb-003c-4483-bd85-afb2f23bcc4f)

Looking into the source forwarder, I see that 1. When we [publish source watermarks](https://github.com/numaproj/numaflow/blob/main/pkg/sources/forward/data_forward.go#L285), the messages dropped by source transformers are also counted, and 2. When we [publish an idle watermark](https://github.com/numaproj/numaflow/blob/main/pkg/sources/forward/data_forward.go#L371) for an idle out edge, we publish the source watermark of the read batch. Hence, we are progressing the watermark even if the data is filtered by source transformer.



Did I miss anything? (I assumed source idling is NOT part of this problem)

whynowy:
@KeranYang - thanks for confirmation! If we are already doing it right, this can be closed.

KeranYang:
@whynowy cool, I am closing it.

vigith:
I am not sure how this is working, especially now that #1386 has been opened.

KeranYang:
Thanks for re-opening the issue. @vigith @yhl25 

This is working for 2 reasons: 

1. The numaflow main branch depends on an old version of numaflow-go (v0.5.1), which when creating MessagesToDrop, assigns `time.Time{}` to event time. I was using the built-in filter transformer hence v0.5.1 was used.

[...]

PR #1386, opened 2023-11-25
fix: dropped messages should not be considered for watermark propagation

yhl25:
* dropped messages event time will be set to -1, so we should exclude them from watermark propagation


vigith:
Should dropped messages still sent event-time so we can propagate the watermark?

vigith:
i am not sure how #1357 is working.

yhl25:
> Should dropped messages still sent event-time so we can propagate the watermark?

What should be the event time for dropped messages? should we ask the users to provide event time for dropped messages as well?

yhl25:
> i am not sure how #1357 is working.

+1

KeranYang:
Thanks for looking into this issue. I put a comment on #1357 on why it's working. We can discuss how to assign event time to dropped messages.

whynowy:
1. `MessageToDrop()` uses default event time;
2. Add a new `MessageToDropWithEventTime(et time.Time)` to give event time?

vigith (approved):
we need to revisit this.

PR #1387, opened 2023-11-26
fix: update numaflow-go version

KeranYang:
* update the numaflow-go version hence the built-in source transformer uses the latest SDK.
* for messages dropped by the source transformer, use the original event time to calculate the source watermark.

The rationale is that a dropped message is also a processed message, it should be included in the watermark calculation.

To be determined: if a user chooses to use the source transformer to assign event time and filter, should we require assigning event time to dropped messages? (I think we should because it gives user the full control of event time. However it will be a breaking change for existing pipelines.)

### DIFF
--- a/pkg/sources/forward/data_forward.go
+++ b/pkg/sources/forward/data_forward.go
@@ -24,9 +24,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/numaproj/numaflow-go/pkg/sourcetransformer"
 	"go.uber.org/zap"
-	"k8s.io/utils/strings/slices"
 
 	dfv1 "github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
 	"github.com/numaproj/numaflow/pkg/forwarder"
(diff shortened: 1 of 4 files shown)


pkg/sources/forward/data_forward.go
```diff
@@ -518,10 +520,12 @@ func (isdf *DataForward) applyTransformer(ctx context.Context, readMessage *isb.
...
 			continue
 		} else {
-			// if we do not get a time from Transformer, we set it to the time from (N-1)th vertex
 			for index, m := range writeMessages {
 				m.ID = fmt.Sprintf("%s-%s-%d", readMessage.ReadOffset.String(), isdf.vertexName, index)
-				if m.EventTime.IsZero() {
+				// if the message is dropped by the source transformer, we assign the original event time to the message.
+				// the original event time is the event time of the message before the source transformer.
+				// this way, even when the message is dropped, the event time is still preserved and be used for watermark calculation.
+				if slices.Contains(m.Tags, sourcetransformer.DROP) {
```
vigith, line 528: I think we might have to check for IsZero condition, too, or else the watermark might unintendedly progress far ahead.

KeranYang: Thanks, Vigith. Do you mean that zero event time might be unintentionally assigned to the transformed messages? if that's the case, do you mean far behind(in the past) or far ahead(in the future) ?

vigith: also, will checking for `slices.Contains` in the hot code bath be detrimental to performance?

pkg/udf/forward/forward.go
```diff
@@ -664,10 +664,10 @@ func (isdf *InterStepDataForward) applyUDF(ctx context.Context, readMessage *isb
 			}
 			continue
 		} else {
-			// if we do not get a time from map UDF, we set it to the time from (N-1)th vertex
 			for index, m := range writeMessages {
 				// add vertex name to the ID, since multiple vertices can publish to the same vertex and we need uniqueness across them
 				m.ID = fmt.Sprintf("%s-%s-%d", readMessage.ReadOffset.String(), isdf.vertexName, index)
+				// if we do not get a time from map UDF, we set it to the time from (N-1)th vertex
```
vigith, line 670: we do not ever accept time in any other vertex other than source-vertex, right?

vigith (approved):

PR #1403, opened 2023-12-04
fix: access path for auth endpoints

veds-g:
- added access path for `auth/v1` api endpoints

### DIFF
--- a/server/authz/route_map.go
+++ b/server/authz/route_map.go
@@ -18,8 +18,9 @@ package authz
 
 import (
 	"fmt"
-
 	"github.com/gin-gonic/gin"
+	sharedutil "github.com/numaproj/numaflow/pkg/shared/util"
+	"strings"
 )
 
 // RouteInfo is a struct which contains the route information with the object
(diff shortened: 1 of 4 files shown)


server/authz/route_map.go
```diff
@@ -18,8 +18,9 @@ package authz
 
 import (
 	"fmt"
-
 	"github.com/gin-gonic/gin"
+	sharedutil "github.com/numaproj/numaflow/pkg/shared/util"
+	"strings"
```
whynowy, line 23: Minor - regroup the imports.

server/authz/route_map.go
```diff
@@ -38,34 +39,45 @@ func newRouteInfo(object string, requiresAuthZ bool) *RouteInfo {
...
-	"DELETE:/api/v1/namespaces/:namespace/isb-services/:isb-service":              newRouteInfo(ObjectISBSvc, true),
-	"GET:/api/v1/namespaces/:namespace/pipelines/:pipeline/isbs":                  newRouteInfo(ObjectPipeline, true),
-	"GET:/api/v1/namespaces/:namespace/pipelines/:pipeline/watermarks":            newRouteInfo(ObjectPipeline, true),
-	"PUT:/api/v1/namespaces/:namespace/pipelines/:pipeline/vertices/:vertex":      newRouteInfo(ObjectPipeline, true),
-	"GET:/api/v1/namespaces/:namespace/pipelines/:pipeline/vertices/metrics":      newRouteInfo(ObjectPipeline, true),
-	"GET:/api/v1/namespaces/:namespace/pipelines/:pipeline/vertices/:vertex/pods": newRouteInfo(ObjectPipeline, true),
-	"GET:/api/v1/metrics/namespaces/:namespace/pods":                              newRouteInfo(ObjectPipeline, true),
-	"GET:/api/v1/namespaces/:namespace/pods/:pod/logs":                            newRouteInfo(ObjectPipeline, true),
-	"GET:/api/v1/namespaces/:namespace/events":                                    newRouteInfo(ObjectEvents, true),
+	"GET:" + baseHref + "api/v1/sysinfo":                                                         newRouteInfo(ObjectPipeline, false),
```
vigith, line 57:
```suggestion
	"GET:" + GetBaseHref() + "api/v1/sysinfo":                                                         newRouteInfo(ObjectPipeline, false),
```

this way you can avoid a module scoped level var

server/authz/route_map.go
vigith, line 23:
```suggestion
	"github.com/gin-gonic/gin"
	"strings"
	
	sharedutil "github.com/numaproj/numaflow/pkg/shared/util"
```

server/authz/route_map.go
```diff
@@ -38,34 +39,45 @@ func newRouteInfo(object string, requiresAuthZ bool) *RouteInfo {
 	}
 }
 
+// GetBaseHref returns the base href of the server with a trailing slash.
+func GetBaseHref() string {
+	baseHref := sharedutil.LookupEnvStringOr("NUMAFLOW_SERVER_BASE_HREF", "/")
```
whynowy, line 44: This is not right. Even though we can use the the configmap to get the baseHref config, but it's passed through the arguments. We need to refactor the code to use the passed in argument.

veds-g: updated

server/authz/route_map.go
```diff
@@ -38,41 +38,66 @@ func newRouteInfo(object string, requiresAuthZ bool) *RouteInfo {
...
-	"DELETE:/api/v1/namespaces/:namespace/isb-services/:isb-service":              newRouteInfo(ObjectISBSvc, true),
-	"GET:/api/v1/namespaces/:namespace/pipelines/:pipeline/isbs":                  newRouteInfo(ObjectPipeline, true),
-	"GET:/api/v1/namespaces/:namespace/pipelines/:pipeline/watermarks":            newRouteInfo(ObjectPipeline, true),
-	"PUT:/api/v1/namespaces/:namespace/pipelines/:pipeline/vertices/:vertex":      newRouteInfo(ObjectPipeline, true),
-	"GET:/api/v1/namespaces/:namespace/pipelines/:pipeline/vertices/metrics":      newRouteInfo(ObjectPipeline, true),
-	"GET:/api/v1/namespaces/:namespace/pipelines/:pipeline/vertices/:vertex/pods": newRouteInfo(ObjectPipeline, true),
-	"GET:/api/v1/metrics/namespaces/:namespace/pods":                              newRouteInfo(ObjectPipeline, true),
-	"GET:/api/v1/namespaces/:namespace/pods/:pod/logs":                            newRouteInfo(ObjectPipeline, true),
-	"GET:/api/v1/namespaces/:namespace/events":                                    newRouteInfo(ObjectEvents, true),
+var GlobalRouteMap *routeMap
```
whynowy, line 46: `AuthRouteMap`? The word `Global` is too big.

whynowy: And there's no need to use pointer.

whynowy (approved):

server/authz/route_map.go
```diff
@@ -38,41 +38,61 @@ func newRouteInfo(object string, requiresAuthZ bool) *RouteInfo {
...
-	"DELETE:/api/v1/namespaces/:namespace/isb-services/:isb-service":              newRouteInfo(ObjectISBSvc, true),
-	"GET:/api/v1/namespaces/:namespace/pipelines/:pipeline/isbs":                  newRouteInfo(ObjectPipeline, true),
-	"GET:/api/v1/namespaces/:namespace/pipelines/:pipeline/watermarks":            newRouteInfo(ObjectPipeline, true),
-	"PUT:/api/v1/namespaces/:namespace/pipelines/:pipeline/vertices/:vertex":      newRouteInfo(ObjectPipeline, true),
-	"GET:/api/v1/namespaces/:namespace/pipelines/:pipeline/vertices/metrics":      newRouteInfo(ObjectPipeline, true),
-	"GET:/api/v1/namespaces/:namespace/pipelines/:pipeline/vertices/:vertex/pods": newRouteInfo(ObjectPipeline, true),
-	"GET:/api/v1/metrics/namespaces/:namespace/pods":                              newRouteInfo(ObjectPipeline, true),
-	"GET:/api/v1/namespaces/:namespace/pods/:pod/logs":                            newRouteInfo(ObjectPipeline, true),
-	"GET:/api/v1/namespaces/:namespace/events":                                    newRouteInfo(ObjectEvents, true),
+var AuthRouteMap routeMap
```
vigith, line 46: global vars that are mutable could lead to race/concurrency issues.

veds-g: right, today we only update it once in the beginning, later we only perform read operations. Do you think this works as is now or should we refactor it?

whynowy: @veds-g - let's make `routeMap` an argument to the `Routes()` function, and pass to the casbin object.

vigith (approved):

PR #1404, opened 2023-12-05
fix: include dropped messages in source watermark calculation

### LINKED ISSUE #1357: Source transformer should progress the watermark even if the data is filtered by source-transformer

KeranYang:
closes #1357

### DIFF
--- a/pkg/sources/transformer/builtin/time_extraction_filter/time_extraction_filter.go
+++ b/pkg/sources/transformer/builtin/time_extraction_filter/time_extraction_filter.go
@@ -71,7 +71,7 @@ func New(args map[string]string) (sourcetransformer.SourceTransformFunc, error) 
 func (e expressions) apply(et time.Time, payload []byte) (sourcetransformer.Message, error) {
 	result, err := expr.EvalBool(e.filterExpr, payload)
 	if err != nil {
-		return sourcetransformer.MessageToDrop(), err
+		return sourcetransformer.MessageToDrop(et), err
 	}
 	if result {
 		timeStr, err := expr.EvalStr(e.eventTimeExpr, payload)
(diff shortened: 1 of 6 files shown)


vigith (approved):

pkg/sources/transformer/builtin/time_extraction_filter/time_extraction_filter.go
```diff
@@ -71,7 +71,7 @@ func New(args map[string]string) (sourcetransformer.SourceTransformFunc, error)
 func (e expressions) apply(et time.Time, payload []byte) (sourcetransformer.Message, error) {
 	result, err := expr.EvalBool(e.filterExpr, payload)
 	if err != nil {
-		return sourcetransformer.MessageToDrop(), err
+		return sourcetransformer.MessageToDrop(et), err
```
whynowy, line 74: This means the dropped messages use the default event time coming from the original datum?

KeranYang: Yes. Thanks for raising this question. It seems using the extracted event time should be more accurate. WDYT?

vigith: `et` is "almost" accurate because we subtract `maxDelay` before publishing the watermark.

One could also argue that, since this is an "error", perhaps we should stick to `-1` watermark. 

Let's change this in the future based on the feedback we receive.

whynowy: If we use default et (or -1), then why don't we use the same in the original `MessageToDrop()`?

KeranYang: This is just for the builtin extractEventTimeFilter, where we choose to assign the original event time to dropped messages. For other transformers, users have to provide a new event time.

For extractEventTimeFilter, I think we should also assign the new event time if it has been successfully extracted from the payload. But I am also ok with making the change in the future after feedbacks.

PR #1416, opened 2023-12-15
feat: local user support for Numaflow

### LINKED ISSUE #1367: Local user support for Numaflow UI
# Summary

Today, there are 2 authentication options to use the UI:

- No auth
- Auth through Dex, to integrated with OAuth solutions

For users can not use OAuth solutions, there's no other options for them besides no-auth. It would be better to provide a local user authentication solution for these users. To be specifically:

- First time installation gets a `admin` user name with randomly generated password stored in a k8s secret.
- `admin` user should have the admin privileges by default.
- The user (someone who did the installation) should be able to add more users (could be stored in a configmap), and passwords (stored in a secret).
- local users should be able to be disabled, including `admin`.
- On the login page, there should be an option to allow local user to login, and after login, the authz follows the same process as using OAuth solution.

# Use Cases

When would you use this?

---

veds-g:
Fixes #1367 

- Added local user support for Numaflow UI
- Added new endpoints for login/logout under - `auth/local/v1`
- Enables authn using `JWT(JSON Web Tokens)`
- Added proxy middleware for both `Dex` and `LocalUser`
- default `admin` user will be available for login - with a random generated password, stored in `numaflow-server-secrets` with the key: `admin.initial-password`
- to add users, update the configMap: `numaflow-server-config` like `data: {your_username}.enabled: "true"`
- likewise, a hashed password needs to be generated for the user using -> `htpasswd -bnBC 10 "" {your_password} | tr -d ':\n'`
- password should be added in the secret: `numaflow-server-secrets` like `stringData: {your_username}.password: {hashed_value_generated_above}`
- update the admin password like  `stringData: admin.password: {hashed_value_generated_above}`
- users can be enabled/disabled by updating above configMap like `data: {your_username}.enabled: "false"`
- use {your_username}, {your_password} to login in the UI

Config Map -
image

Secret - 
image

Disable Users -
image

- configs can be updated during runtime without the need to restart the application

image


config/base/numaflow-server/numaflow-server-cm.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: numaflow-server-cm
```
whynowy, line 4: `numaflow-server-config`

config/base/numaflow-server/numaflow-server-deployment.yaml
```diff
@@ -42,6 +42,11 @@ spec:
         volumeMounts:
         - mountPath: /opt/numaflow
           name: env-volume
+      - name: secret-init
+        image: quay.io/numaproj/numaflow:latest
+        args:
+        - "secret-init"
```
whynowy, line 48: `server-secrets-init`

config/base/numaflow-server/numaflow-server-secret.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: numaflow-server-secret
```
whynowy, line 4: `-secrets`

same for file name.

config/cluster-install/rbac/numaflow-server/numaflow-server-cluster-role.yaml
```diff
@@ -37,13 +37,15 @@ rules:
       - pods/log
       - pods/exec
       - configmaps
+      - secrets
```
whynowy, line 40: For this, use a separated `Role` and `RoleBinding` - the UX server does not need to have secret access to cluster scope.

server/cmd/serversecretinit/start.go
```diff
@@ -0,0 +1,102 @@
...
+			return fmt.Errorf("failed to update k8s secret with admin password and secretKey, %w", err)
+		}
+	}
+
+	return nil
+}
+
+// generateRandomBytes generates a random byte slice of the specified length
+// and base64 encodes it
+func generateRandomBytes(length int) (string, error) {
```
whynowy, line 91: Use pkg/shared/utils/RandomString

whynowy:
Can we do

`accounts.{your_username}.disabled: "true"`? defaults to `false`, which means, when adding new local users, there's no need to configure it, it's only needed when disabling a user.

server/apis/v1/noauthlocalhandler.go
```diff
@@ -0,0 +1,126 @@
...
+	"fmt"
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+
+	"github.com/numaproj/numaflow/server/authn"
+	"github.com/numaproj/numaflow/server/common"
+)
+
+type noAuthLocalHandler struct {
```
whynowy, line 29: The name `noAuthxxx` is very confusing, can we use a better one?

server/authn/accounts.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+// Account holds local account information
+type Account struct {
+	InitialPasswordHash string
+	PasswordHash        string
+	Enabled             bool
+}
+
+var logger = logging.NewLogger().Named("accounts")
```
whynowy, line 39: Are we able to use the same logger? If it's not easy to refactor, leave a TODO.

server/authn/accounts.go
```diff
@@ -0,0 +1,142 @@
...
+	account, ok := accounts[name]
+	if !ok {
+		return nil, fmt.Errorf("account: %v, does not exist", name)
+	}
+	return &account, nil
+}
+
+// GetAccounts returns list of configured accounts
+func GetAccounts(ctx context.Context, kubeClient kubernetes.Interface) (map[string]Account, error) {
+	secret, err := kubeClient.CoreV1().Secrets(common.NumaflowAccountsNamespace).Get(ctx, common.NumaflowAccountsSecret, metav1.GetOptions{})
```
whynowy, line 56: Use the namespace from ENV.

server/authn/accounts.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return nil, err
+	}
+	cm, err := kubeClient.CoreV1().ConfigMaps(common.NumaflowAccountsNamespace).Get(ctx, common.NumaflowAccountsConfigMap, metav1.GetOptions{})
```
whynowy, line 60: same here.

whynowy (approved):

PR #1421, opened 2023-12-18
chore: file rename and default config change

whynowy:
Explain what this PR does.


config/advanced-install/namespaced-controller-wo-crds.yaml
```diff
@@ -125,7 +125,7 @@ subjects:
 apiVersion: v1
 data:
   namespaced: "true"
-  server.disable.auth: "true"
+  server.disable.auth: "false"
```
vigith, line 128: let's keep as true

whynowy: done.

vigith (approved):

PR #1429, opened 2023-12-21
feat: update tcp client connections

kohlisid:
Updates the grpc client to use multiple sockets for TCP connections instead of a single port with reuse

### DIFF
--- a/pkg/sdkclient/grpc_resolver/client_resolver.go
+++ b/pkg/sdkclient/grpc_resolver/client_resolver.go
@@ -81,9 +81,9 @@ func (*multiProcResolver) Resolve(target resolver.Target)          {}
 
 // buildConnAddrs Populate the connection list for the clients
 // Format (serverAddr, serverIdx) : (0.0.0.0:5551, 1), (0.0.0.0:5552, 2)
-func buildConnAddrs(numCpu int, servPorts []string) []string {
-	var conn = make([]string, numCpu)
-	for i := 0; i < numCpu; i++ {
+func buildConnAddrs(servPorts []string) []string {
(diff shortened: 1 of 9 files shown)


pkg/sdkclient/grpc_resolver/client_resolver.go
```diff
@@ -81,12 +79,15 @@ func (*multiProcResolver) ResolveNow(o resolver.ResolveNowOptions) {}
 func (*multiProcResolver) Close()                                  {}
 func (*multiProcResolver) Resolve(target resolver.Target)          {}
 
-// BuildConnAddrs Populate the connection list for the clients
-// Format (serverAddr, serverIdx) : (0.0.0.0:5551, 1)
-func BuildConnAddrs(numCpu int) []string {
+// buildConnAddrs Populate the connection list for the clients
+// Format (serverAddr, serverIdx) : (0.0.0.0:5551, 1), (0.0.0.0:5552, 2)
+func buildConnAddrs(numCpu int, servPorts []string) []string {
```
whynowy, line 84: `numCpu` could be potentially not equal to `len(servPorts)`, so the logic below is risky.

kohlisid: From the SDK side they will both be the same value as numCpu is written to the server info as the number of server instances required in multiproc. 
But I can actually do away with the variable itself and just use servPorts and len(servPorts), I guess would be easier for understanding

whynowy (approved):

vigith (approved):
We could have used UDS (multiple ones) to achieve the same? We used TCP because of lack of support for multiplexing in UDS.

PR #1433, opened 2024-01-02
fix: server-secrets-init container restart 

veds-g:
- added env variable for namespace for accessing secrets and configMap for the `server-secrets-init` container
- enabled access to UI by completing initialization for the server


yhl25 (approved):
fix codegen

vigith (approved):
How is this env var being used?

whynowy (approved):

veds-g:
> How is this env var being used?

this var is the same namespace for which we will update k8s secret with data(secretKey and initial admin password) for local user authn

PR #1435, opened 2024-01-03
chore: add the unit for vertex processing rates on ui

KeranYang:
Quentin Faidide from numaflow slack channel:

> The units are not shown in the Processing Rates panel. It's not the end of the world since it's easy to check against the 1m one with the unit in the graph, but it's free lunch to add it and have users figuring out instantly what the numbers means

## Before this PR

Screenshot 2024-01-03 at 5 01 50 PM

## After

Screenshot 2024-01-03 at 5 03 36 PM

### DIFF
--- a/ui/src/components/common/SlidingSidebar/partials/VertexDetails/partials/ProcessingRates/index.tsx
+++ b/ui/src/components/common/SlidingSidebar/partials/VertexDetails/partials/ProcessingRates/index.tsx
@@ -82,9 +82,9 @@ export function ProcessingRates({
               foundRates.map((metric) => (
                 <TableRow key={metric.partition}>
                   <TableCell>{metric.partition}</TableCell>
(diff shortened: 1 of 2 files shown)


ui/src/components/common/SlidingSidebar/partials/VertexDetails/partials/ProcessingRates/index.tsx
```diff
@@ -39,14 +39,14 @@ export function ProcessingRates({
       rates.push({
         partition: index,
         oneM: item.processingRates["1m"]
-          ? item.processingRates["1m"].toFixed(2)
```
darshansimha, line 42: Instead of changing the rates and their types, can you add the "/sec" text to the JSX on line number : 85.
That way we can keep the data types as it is.

veds-g (approved):

PR #1437, opened 2024-01-04
chore: update some comments for reduce

KeranYang:
Make some small updates on comments and fix some typos while I learn how reduce works.

### DIFF
--- a/pkg/window/strategy/fixed/fixed.go
+++ b/pkg/window/strategy/fixed/fixed.go
@@ -52,6 +52,8 @@ func NewFixedWindow(length time.Duration, message *isb.ReadMessage) window.Timed
 	}
 }
 
+var _ window.TimedWindow = (*fixedWindow)(nil)
+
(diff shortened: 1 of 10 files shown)


pkg/window/strategy/fixed/fixed.go
```diff
@@ -147,12 +148,13 @@ func (w *Windower) InsertWindow(tw window.TimedWindow) {
 // returns a list of TimedWindowRequests, each request contains the window operation and the window
 // which needs to be closed.
 func (w *Windower) CloseWindows(time time.Time) []*window.TimedWindowRequest {
+	// TODO - we are updating both active and closed windows. do we need lock? same question for other windower methods.
```
vigith, line 151:
```suggestion
	// FIXME - we are updating both active and closed windows. do we need lock? same question for other windower methods.
```

data races are severe.

vigith (approved):

PR #1445, opened 2024-01-10
fix: memory leak inside session windower

yhl25:
* fixes memory leak inside session windower
* metrics to track number of active and closed windows
* enhance tickgen to produce out of order messages


pkg/sources/kafka/reader.go
```diff
@@ -142,20 +143,33 @@ func (r *kafkaSource) Partitions(context.Context) []int32 {
 }
 
 func (r *kafkaSource) Read(_ context.Context, count int64) ([]*isb.ReadMessage, error) {
+	latestEtMap := make(map[int32]int64)
```
whynowy, line 146: Is this for debug purpose?

pkg/window/strategy/sliding/sliding.go
```diff
@@ -183,6 +192,19 @@ func (w *Windower) CloseWindows(time time.Time) []*window.TimedWindowRequest {
 		windowOperations = append(windowOperations, operation)
 		w.closedWindows.InsertBack(win)
 	}
+
+	metrics.ActiveWindowsCount.With(map[string]string{
```
KeranYang, line 196: Just out of curiosity, why do we choose to emit metrics when we call CloseWindows, as opposed to other methods that are also updating the active&closed windows?

pkg/sources/forward/data_forward.go
```diff
@@ -311,16 +311,26 @@ func (df *DataForward) forwardAChunk(ctx context.Context) {
...
+			if latestEt, ok := latestEtMap[m.readMessage.ReadOffset.PartitionIdx()]; !ok || message.EventTime.UnixNano() < latestEt {
+				latestEtMap[m.readMessage.ReadOffset.PartitionIdx()] = message.EventTime.UnixNano()
+			}
 			transformedReadMessages = append(transformedReadMessages, message.ToReadMessage(m.readMessage.ReadOffset, time.UnixMilli(-1)))
 		}
 	}
+
+	// print the latest event time for each partition
+	for partition, latestEt := range latestEtMap {
+		df.opts.logger.Infow("Latest event time for partition - ", zap.Int32("partition", partition), zap.Int64("latestEt", int64(time.Since(time.Unix(0, latestEt)).Minutes())))
```
whynowy, line 332: Debugw?

pkg/sources/forward/data_forward.go
```diff
@@ -311,16 +311,26 @@ func (df *DataForward) forwardAChunk(ctx context.Context) {
...
+	latestEtMap := make(map[int32]int64)
+
 	for _, m := range transformerResults {
 		writeMessages = append(writeMessages, m.writeMessages...)
 		for _, message := range m.writeMessages {
 			// we convert each writeMessage to isb.ReadMessage by providing its parent ReadMessage's ReadOffset.
 			// since we use message event time instead of the watermark to determine and publish source watermarks,
 			// time.UnixMilli(-1) is assigned to the message watermark. transformedReadMessages are immediately
 			// used below for publishing source watermarks.
+			if latestEt, ok := latestEtMap[m.readMessage.ReadOffset.PartitionIdx()]; !ok || message.EventTime.UnixNano() < latestEt {
```
whynowy, line 323: For all of these debugging purpose logic, we should think about using an approach that does not affect the performance. e.g. turn it off by default.

pkg/window/strategy/sliding/sliding.go
```diff
@@ -99,8 +105,11 @@ type Windower struct {
 	closedWindows *window.SortedWindowListByEndTime
 }
 
-func NewWindower(length time.Duration, slide time.Duration) window.TimedWindower {
+func NewWindower(length time.Duration, slide time.Duration, vertexInstance *dfv1.VertexInstance) window.TimedWindower {
```
KeranYang, line 108: Have you thought about adding a new method to the Windower interface `EmitMetrics(vertexInstance *dfv1.VertexInstance)`? That way the windower can emit metrics not only after closing windows but also after other operations, which is up to the caller to decide. It also simplifies the Windower struct itself.

pkg/sources/kafka/reader.go
yhl25: I had added it for debugging, will remove it

pkg/sources/forward/data_forward.go
yhl25: +1

pkg/sources/kafka/reader.go
```diff
@@ -142,20 +143,33 @@ func (r *kafkaSource) Partitions(context.Context) []int32 {
...
 		case <-timeout:
 			// log that timeout has happened and don't return an error
 			r.logger.Debugw("Timed out waiting for messages to read.", zap.Duration("waited", r.readTimeout))
 			break loop
 		}
 	}
+	//// print the latest event time for each partition
+	//for partition, latestEt := range latestEtMap {
+	//	r.logger.Infow("Latest event time for partition - ", zap.Int32("partition", partition), zap.Int64("latestEt", int64(time.Since(time.Unix(0, latestEt)).Minutes())))
+	//}
```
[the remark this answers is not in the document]
yhl25: will remove

pkg/window/strategy/sliding/sliding.go
yhl25: CloseWindows() is invoked once per batch, whereas other functions are invoked for every message. I think it's good to track the windows count once per batch.

pkg/window/strategy/sliding/sliding.go
yhl25: I don't think adding Prometheus metrics related functionality to the interface is ideal, the caller shouldn't be tasked with handling these metrics, since they are not directly related to window functionality.

KeranYang (approved):

Issue #1450, opened 2024-01-11
Refactor - move the ServerInfo check logic out from SDK clients


whynowy:
# Summary

The `ServerInfo` written by the SDK server includes very useful information such as the SDK version, written language, protocol, etc. This information check sits together with the SDK clients, it runs at the time the container starts. It should be abstracted to be standalone module, so that it can be used to support sort of backward compatibility - based on the sdk version to start the correct client version.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #1455, opened 2024-01-12
chore: extract sdk server info

jy4096:
fix: https://github.com/numaproj/numaflow/issues/1450

### DIFF
--- a/pkg/sdkclient/sinker/client.go
+++ b/pkg/sdkclient/sinker/client.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 
 	sinkpb "github.com/numaproj/numaflow-go/pkg/apis/proto/sink/v1"
+	"github.com/numaproj/numaflow-go/pkg/info"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/protobuf/types/known/emptypb"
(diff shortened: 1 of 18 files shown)


pkg/sdkclient/sinker/client.go
```diff
@@ -38,23 +36,13 @@ type client struct {
 
 var _ Client = (*client)(nil)
 
-// New creates a new client object.
+// New creates a new client object. Sinker doesn't require server info to start ATM.
```
whynowy, line 39: Is it true? https://github.com/numaproj/numaflow-go/blob/main/pkg/sinker/server.go#L39

pkg/sdkclient/source/client/client.go
```diff
@@ -39,24 +37,14 @@ type client struct {
 
 var _ Client = (*client)(nil)
 
-// New creates a new client object.
+// New creates a new client object. Source doesn't require server info to start ATM.
```
whynowy, line 40: https://github.com/numaproj/numaflow-go/blob/main/pkg/sourcer/server.go#L41

whynowy (commented):
`source` and `sourceTransformer` use the same server info path, this needs to be fixed.

pkg/sdkclient/sinker/client.go
jy4096: The source and sinker clients don't require serverinfo to start.. fixed the godoc

whynowy (approved):

PR #1460, opened 2024-01-16
feat: pipeline health status for UI

### LINKED ISSUE #1325: reinvent health and status definitions for numaflow UI
reinvent health and status definitions for numaflow UI

veds-g:
Fixes #1325 

- Updated health status for pipeline based on new health API
- Calculated health based on consolidated value for `Resource Health` and `Data Flow Health`
- Updated health status on both Namespace page and Pipeline page
- Added tooltip to show detailed health information on hover
image


pkg/daemon/server/service/pipeline_metrics_query.go
```diff
@@ -55,9 +55,9 @@ type PipelineMetadataQuery struct {
 }
 
 const (
-	PipelineStatusOK       = "OK"
+	PipelineStatusHealthy  = "healthy"
```
whynowy, line 58: What is the reason of this sort of inconsistency (upper case, lower case)?

veds-g: I formatted to match constants as defined in the UI code, uppercase for pipeline running status and lowercase for pipeline health

whynowy: Can we rename the constant names? There's no way to tell at all.

whynowy:
According to the screenshot:

1. Do not capital first letter of every word of the tooltip.
2. `DataFlow` -> `Data Flow`.
3. Instead of using one large tooltip, can we do it separately when hover on the different icons?

whynowy (approved):

Issue #925, opened 2023-08-08
Kafka source to ingest from a specific offset


alexanghh:
# Summary

Allow configuration of Kafka source to ingest from a specific offset

What change needs making?

Kafka source.

# Use Cases

- When creating a new pipeline which requires to ingest from specific point in time/offset.
- When fixing bugs/new features in pipeline and re-ingesting partially from kafka topic.

PS: I can help to implement this but would like to seek advice if this is as simple as adding the offset configuration in the kafka source (e.g. would the kafka source pods get restart, and use the offset in the configuration?)

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

yhl25:
Thank you for creating this issue. While Sarama does not directly support reading from a specific time, it does provide the functionality to read from a specific offset. This can be achieved using the [ResetOffset()](https://github.com/IBM/sarama/blob/23c42869222b1467841a9b6318a7ceb159c6a6fd/consumer_group.go#L644) method. I am currently investigating this feature to understand it better and will update you with any findings or potential solutions.

yhl25:
Following my testing, I have been able to utilize the ResetOffset() method from Sarama to facilitate consumption from a specific offset I think we can offer this feature in our Kafka source. Since the offset is unique to each partition, how do we allow users to specify the offsets in the pipeline spec? any suggestions? @vigith @whynowy @alexanghh

alexanghh:
Hi @yhl25, may I proceed to implement the Consumer.Offsets.Initial first while figuring out the viability of setting specific offset?
At least making it possible to consume from the oldest offset. Current implementation only allows msg to be consumed from the newest.

alexanghh:
> Following my testing, I have been able to utilize the ResetOffset() method from Sarama to facilitate consumption from a specific offset I think we can offer this feature in our Kafka source. Since the offset is unique to each partition, how do we allow users to specify the offsets in the pipeline spec? any suggestions? @vigith @whynowy @alexanghh

May I suggest to use a map to store the offsets per partition? This would be similar to what [Apache Spark](https://spark.apache.org/docs/latest/structured-streaming-kafka-integration.html) (referencing startingOffsetsByTimestamp and startingOffsets) has done.

[...]

PR #1469, opened 2024-01-19
chore: update kafka source docs with start from a specific date time

jy4096:
fix: https://github.com/numaproj/numaflow/issues/925

Added the current solution in the Kafka source docs before we have further plans on implementing the feature in Numaflow.


docs/user-guide/sources/kafka.md
```diff
@@ -57,3 +57,39 @@ spec:
...
+
+GROUP                  TOPIC             PARTITION  CURRENT-OFFSET  LOG-END-OFFSET  LAG             CONSUMER-ID     HOST            CLIENT-ID
+console-consumer-94457 quickstart-events 0          56              56              0               -               -               -
+console-consumer-94457 quickstart-events 1          38              38              0               -               -               -
+console-consumer-94457 quickstart-events 2          4               4               0               -               -               -
+```
+To achieve that, before the pipeline start, we need to first stop the consumers in the consumer group `console-consumer-94457` because offsets can only be reset if the group `console-consumer-94457` is inactive. Then, reset the offsets using the desired date and time. The example command below uses UTC time.
+```shell
+➜  kafka_2.13-3.6.1 bin/kafka-consumer-groups.sh --bootstrap-server localhost:9092 --execute --reset-offsets --group console-consumer-94457 --topic quickstart-events --to-datetime 2024-01-19T19:26:00.000
```
whynowy, line 88: 1. Please verify this is working from anywhere (not just kafka server).
2. https://docs.confluent.io/kafka/operations-tools/use-kafka-tools-ccloud.html#use-kafka-tool-ccloud
3. https://docs.confluent.io/kafka/operations-tools/manage-consumer-groups.html

vigith (commented):
Please document what privileges are required to run these commands in case special privileges are required. Just run it in our e2e and see whether it works.

jy4096:
Tested in our e2e with a configuration file `--consumer.config e2e.properties` and worked.

vigith (approved):

PR #1472, opened 2024-01-20
fix: bug in late message handling for sliding window

yhl25:


whynowy (approved):

vigith:
this has a correctness problem, let's remove late data acceptance for sliding window in this PR and cut 1.1.4

PR #1490, opened 2024-02-02
feat: configure standardResources via controller configmap

### LINKED ISSUE #1487: Better way to configure the memory and CPU resource requests
# Summary

- On our internal deployment of numaflow, we observe a generally a lower utilization. Have a lot of pipelines, pods the requested resource stay underutilized significantly.
- Currently, pipeline vertex request resource (request) is configured at code-level ([refer](https://github.com/numaproj/numaflow/blob/main/pkg/apis/numaflow/v1alpha1/const.go#L188)). 

One of the approaches is to fork, make the change, build our own image and use it. I feel having a cleaner approach can help other community members of Numaflow as well.

We can use [controller deployment env](https://github.com/numaproj/numaflow/blob/main/config/base/controller-manager/controller-manager-deployment.yaml#L29) instead (open to better ways) rather than code-level

---

inishchith:
Better way to configure the memory and CPU resource requests

closes #1487

Sample configmap

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: numaflow-controller-config
data:
  controller-config.yaml: |+
    defaults:
      containerResources: |
        limits:
          memory: "256Mi"
          cpu: "200m"
        requests:
          memory: "128Mi"
          cpu: "100m"
```


vigith:
Is standardResources only for UDF containers?

pkg/reconciler/config.go
```diff
@@ -21,13 +21,21 @@ import (
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 )
 
 // GlobalConfig is the configuration for the controllers, it is
 // supposed to be populated from the configmap attached to the
 // controller manager.
 type GlobalConfig struct {
-	ISBSvc *ISBSvcConfig `json:"isbsvc"`
+	StandardResources *ResourceConfig `json:"standardResources"`
```
whynowy, line 32:
```
data:
  controller-config.yaml: |+
    defaults:
      containerResources:
          limits:
            cpu: xxxx
            memory: xxxx
          requests:
            cpu: xxx
            memory: xxxx
    isbsvc:
      redis:
...
```

This way we may add more `defaults` in the future.

Also, use `*apiv1.ResourceRequirements` directly.

pkg/reconciler/config.go
```diff
@@ -69,6 +77,27 @@ type JetStreamVersion struct {
 	StartCommand         string `json:"startCommand"`
 }
 
+func (g *GlobalConfig) GetStandardResources() corev1.ResourceRequirements {
```
whynowy, line 80: `GetDefaultContainerResources()`, and it always return a new instance, so that the invokers do not need to deepCopy.

pkg/apis/numaflow/v1alpha1/get_spec_req.go
```diff
@@ -21,21 +21,22 @@ import (
...
+	RedisContainerPort        int32                       `protobuf:"bytes,7,opt,name=redisContainerPort"`
+	SentinelContainerPort     int32                       `protobuf:"bytes,8,opt,name=sentinelContainerPort"`
+	RedisMetricsContainerPort int32                       `protobuf:"bytes,9,opt,name=redisMetricsContainerPort"`
+	CredentialSecretName      string                      `protobuf:"bytes,10,opt,name=credentialSecretName"`
+	TLSEnabled                bool                        `protobuf:"bytes,11,opt,name=tlsEnabled"`
+	PvcNameIfNeeded           string                      `protobuf:"bytes,12,opt,name=pvcNameIfNeeded"`
+	ConfConfigMapName         string                      `protobuf:"bytes,13,opt,name=confConfigMapName"`
+	ScriptsConfigMapName      string                      `protobuf:"bytes,14,opt,name=scriptsConfigMapName"`
+	HealthConfigMapName       string                      `protobuf:"bytes,15,opt,name=healthConfigMapName"`
+	StandardResources         corev1.ResourceRequirements `protobuf:"bytes,16,opt,name=standardResources"`
```
whynowy, line 39: Use `DefaultResources`?

pkg/apis/numaflow/v1alpha1/get_spec_req.go
```diff
@@ -45,35 +46,38 @@ type GetRedisServiceSpecReq struct {
...
-	Image               string            `protobuf:"bytes,2,opt,name=image"`
-	PullPolicy          corev1.PullPolicy `protobuf:"bytes,3,opt,name=pullPolicy,casttype=k8s.io/api/core/v1.PullPolicy"`
-	Env                 []corev1.EnvVar   `protobuf:"bytes,4,rep,name=env"`
-	SideInputsStoreName string            `protobuf:"bytes,5,opt,name=sideInputsStoreName"`
+	ISBSvcType          ISBSvcType                  `protobuf:"bytes,1,opt,name=isbSvcType"`
+	Image               string                      `protobuf:"bytes,2,opt,name=image"`
+	PullPolicy          corev1.PullPolicy           `protobuf:"bytes,3,opt,name=pullPolicy,casttype=k8s.io/api/core/v1.PullPolicy"`
+	Env                 []corev1.EnvVar             `protobuf:"bytes,4,rep,name=env"`
+	SideInputsStoreName string                      `protobuf:"bytes,5,opt,name=sideInputsStoreName"`
+	StandardResources   corev1.ResourceRequirements `protobuf:"bytes,6,opt,name=standardResources"`
```
whynowy, line 54: Use `DefaultResources`?

pkg/apis/numaflow/v1alpha1/get_spec_req.go
```diff
@@ -45,35 +46,38 @@ type GetRedisServiceSpecReq struct {
...
 type GetDaemonDeploymentReq struct {
-	ISBSvcType ISBSvcType        `protobuf:"bytes,1,opt,name=isbSvcType"`
-	Image      string            `protobuf:"bytes,2,opt,name=image"`
-	PullPolicy corev1.PullPolicy `protobuf:"bytes,3,opt,name=pullPolicy,casttype=k8s.io/api/core/v1.PullPolicy"`
-	Env        []corev1.EnvVar   `protobuf:"bytes,4,rep,name=env"`
+	ISBSvcType        ISBSvcType                  `protobuf:"bytes,1,opt,name=isbSvcType"`
+	Image             string                      `protobuf:"bytes,2,opt,name=image"`
+	PullPolicy        corev1.PullPolicy           `protobuf:"bytes,3,opt,name=pullPolicy,casttype=k8s.io/api/core/v1.PullPolicy"`
+	Env               []corev1.EnvVar             `protobuf:"bytes,4,rep,name=env"`
+	StandardResources corev1.ResourceRequirements `protobuf:"bytes,5,opt,name=standardResources"`
```
whynowy, line 62: Same here.

pkg/apis/numaflow/v1alpha1/get_spec_req.go
```diff
@@ -45,35 +46,38 @@ type GetRedisServiceSpecReq struct {
...
+	ClusterPort                int32                       `protobuf:"bytes,6,opt,name=clusterPort"`
+	ClientPort                 int32                       `protobuf:"bytes,7,opt,name=clientPort"`
+	MonitorPort                int32                       `protobuf:"bytes,8,opt,name=monitorPort"`
+	MetricsPort                int32                       `protobuf:"bytes,9,opt,name=metricsPort"`
+	ServerAuthSecretName       string                      `protobuf:"bytes,10,opt,name=serverAuthSecretName"`
+	ServerEncryptionSecretName string                      `protobuf:"bytes,11,opt,name=serverEncryptionSecretName"`
+	ConfigMapName              string                      `protobuf:"bytes,12,opt,name=configMapName"`
+	PvcNameIfNeeded            string                      `protobuf:"bytes,13,opt,name=pvcNameIfNeeded"`
+	StartCommand               string                      `protobuf:"bytes,14,opt,name=startCommand"`
+	StandardResources          corev1.ResourceRequirements `protobuf:"bytes,15,opt,name=standardResources"`
```
whynowy, line 80: Same here.

pkg/apis/numaflow/v1alpha1/get_spec_req.go
```diff
@@ -85,8 +89,9 @@ type GetJetStreamServiceSpecReq struct {
...
 type GetSideInputDeploymentReq struct {
-	ISBSvcType ISBSvcType        `protobuf:"bytes,1,opt,name=isbSvcType"`
-	Image      string            `protobuf:"bytes,2,opt,name=image"`
-	PullPolicy corev1.PullPolicy `protobuf:"bytes,3,opt,name=pullPolicy,casttype=k8s.io/api/core/v1.PullPolicy"`
-	Env        []corev1.EnvVar   `protobuf:"bytes,4,rep,name=env"`
+	ISBSvcType        ISBSvcType                  `protobuf:"bytes,1,opt,name=isbSvcType"`
+	Image             string                      `protobuf:"bytes,2,opt,name=image"`
+	PullPolicy        corev1.PullPolicy           `protobuf:"bytes,3,opt,name=pullPolicy,casttype=k8s.io/api/core/v1.PullPolicy"`
+	Env               []corev1.EnvVar             `protobuf:"bytes,4,rep,name=env"`
+	StandardResources corev1.ResourceRequirements `protobuf:"bytes,5,opt,name=standardResources"`
```
whynowy, line 96: And this.

inishchith:
> Is standardResources only for UDF containers?

@vigith - Not just UDF. All init and main containers (vertex, deamon redis_buffer_service,)

pkg/reconciler/config.go
inishchith: Agree with additional defaults in the future. ResourceList is a Map which makes the valid yaml as following, if I were to use `*apiv1.ResourceRequirements` directly.

```yaml
defaults:
  containerResources:
    limits:
      memory:
        value: xxxx
      cpu:
        value: xxxx
    requests:
      memory:
        value: xxxx
      cpu:
        value: xxxx
```

checking if there could be a simpler way to handle this while opening up scope for more defaults.

inishchith: Viper uses the mapstructure package for decoding, which doesn't support resource.Quantity out of the box.

pkg/reconciler/config.go
inishchith: updated

pkg/reconciler/config.go
inishchith: updated to cover defaults

whynowy: If Viper does not support, make it a string, and then we parse it.

whynowy:
```
data:
  controller-config.yaml: |+
    defaults:
      containerResources: |
          limits:
            cpu: xxxx
            memory: xxxx
          requests:
            cpu: xxx
            memory: xxxx
    isbsvc:
      redis:

Comment (whynowy) [reply pkg/reconciler/config.go]: Actually I didn't get it, Viper does not support customized structs? we are already doing that, right?

Comment (inishchith) [reply pkg/reconciler/config.go]: > Viper does not support customized structs?
No, it does not parse `resource.Quantity` i.e the value of the map object (under limits/resource) correctly - assigns them 0. Which is due to the underlying mapstructure.

Comment (inishchith) [reply pkg/reconciler/config.go]: > If Viper does not support, make it a string, and then we parse it.

Better. updated!

**Comment by whynowy** [comment]
@inishchith - It looks good to me, can you please:

1. Add the new config to https://github.com/numaproj/numaflow/blob/main/config/base/controller-manager/numaflow-controller-config.yaml#L7, use the default resource settings;
2. Add some doc to https://github.com/numaproj/numaflow/blob/main/docs/operations/controller-configmap.md, also add the new config to https://github.com/numaproj/numaflow/blob/main/docs/operations/numaflow-controller-config.yaml.
3. Appreciate if you could add a config_test.go to cover the unit tests (unit tests for config.go are missing, you can add the one for the new function you added, but always welcome if you could cover more).

**Comment by inishchith** [comment]
@whynowy Thanks!

I have updated the docs and config.
I have also added minimal tests that covers the newly introduced method. Covering the config file might require refactoring some `LoadConfig`, which I think might diverge from the goal of this change list. Let me know your thoughts.

File: pkg/reconciler/config_test.go [hunk 8]
```diff
@@ -0,0 +1,76 @@
+// add test for config.go
```
whynowy, line 1: Remove these.

pkg/reconciler/config.go
```diff
@@ -69,6 +77,29 @@ type JetStreamVersion struct {
...
+	}
+
+	if g.Defaults == nil || g.Defaults.ContainerResources == "" {
+		return &defaultResources, nil
+	}
+
+	var resourceConfig corev1.ResourceRequirements
+	err := yaml.Unmarshal([]byte(g.Defaults.ContainerResources), &resourceConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed unmarshal configuration file. %w", err)
```
whynowy, line 97: `panic(ftm.Errorf("failed to unmarshal default container resources, %w", err))`

In this way there's no need to return an error but a struct only, that will make the invokers easier - letting the reconciler capture the error and keep retrying does not make sense.

docs/operations/controller-configmap.md
```diff
@@ -15,11 +15,54 @@ metadata:
...
+      containerResources: |
+        limits:
+          memory: "256Mi"
+          cpu: "200m"
+        requests:
+          memory: "128Mi"
+          cpu: "100m"
+```
+
+Similarly, you can set the default container limits resources for steps across all the pipelines:
```
whynowy, line 48: This section has no difference from above, can be removed?

inishchith: it has a difference, it explains not setting request. let me know if need to remove this

whynowy: Why not setting `requests` it will default to limits?

pkg/reconciler/config.go
```diff
@@ -69,6 +77,29 @@ type JetStreamVersion struct {
 	StartCommand         string `json:"startCommand"`
 }
 
+func (g *GlobalConfig) GetDefaultContainerResources() *corev1.ResourceRequirements {
```
whynowy, line 80: Now you can return the struct directly.

docs/operations/controller-configmap.md
inishchith: As the ResourceRequirements are directly sourced to k8s API that we use ([refer](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L2553)).

pkg/reconciler/config.go
inishchith: thanks for pointing out. fixed.

docs/operations/controller-configmap.md
inishchith: removed it!

whynowy (approved):
LGTM, thanks!

Issue #1456, opened 2024-01-13
`source` and `sourceTransformer` sdk use the same serverInfoPath


whynowy:
**Describe the bug**
In the SDKs, `source` and `sourceTransformer` use the same serverInfoPath, this will potentially cause problem.

We should use different path for different sdks maybe.

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**

- Kubernetes: [e.g. v1.18.6]
- Numaflow: [e.g. v0.5.1]

**Additional context**

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

KeranYang:
@a3hadi will pick up this issue.

a3hadi:
@whynowy I just had a quick question. Is this a problem because two separate containers would write to the same location? I.e., would a container containing a python udf and a container containing a java udf both write to /var/run/numaflow/server-info and this could potentially cause a conflict? If this is the case, should we address this just for the sourcer service?

whynowy:
> @whynowy I just had a quick question. Is this a problem because two separate containers would write to the same location? I.e., would a container containing a python udf and a container containing a java udf both write to /var/run/numaflow/server-info and this could potentially cause a conflict? If this is the case, should we address this just for the sourcer service?

Potentially there could be 2 containers write to the same location (e.g. udsource and source transformer).

What we should do is, for different type of UD container, use different path (instead of only addressing source and transformer).

a3hadi:
steps to take:

1. Create new paths for all UD services, and use these new paths in all SDKs
2. Implement the same changes on the client side in numaflow
3. Update all images with new version number so that they use the new paths
4. Add note to documentation/readme that users using new SDK version also update numaflow to new version, otherwise there will be compatibility issues

KeranYang:
Great job @a3hadi on closing this issue. 👍

PR #1494, opened 2024-02-07
add separate server info file paths for services on client side

a3hadi:
Implement #1456 on client side


KeranYang:
@a3hadi , with this change, the platform stops supporting all the old UDF images (hence all e2e fail.). Users have to re-build UDF/Source/Sink images using the latest SDKs. Since server info hasn't been widely used, let's see if we can update on the platform side in a way that when a server info file is not found because of incompatible file path, the platform numa container can decide to ignore it and continue processing.

KeranYang:
@a3hadi , could you also look into why the udsource e2e succeeded? Thanks!

a3hadi:
> @a3hadi , with this change, the platform stops supporting all the old UDF images (hence all e2e fail.). Users have to re-build UDF/Source/Sink images using the latest SDKs. Since server info hasn't been widely used, let's see if we can update on the platform side in a way that when a server info file is not found because of incompatible file path, the platform numa container can decide to ignore it and continue processing.

I think ignoring incompatible server info path, with a message to remind to upgrade sdk is a good idea (since using the new server info paths isn't mission critical to continue processing anyways). I will focus on getting the current tests to pass first and then look into this.

a3hadi:
After discussing with Keran, I think the best course of action for the "error handling" is simply, if the server info file path is not found, set it to the default path of `/var/run/numaflow/server-info`. This is because, the server info file will always be created. However there are two possible paths that it can have. If an old sdk version is being used it will be the aforementioned default path, if the sdk version is up to date it will be using the new isolated paths for each service. Thus falling back to the default path when the server info file path is not found will accommodate older sdk versions.

pkg/sdkclient/mapper/client.go
```diff
@@ -36,7 +37,7 @@ type client struct {
 
 // New creates a new client object.
 func New(serverInfo *info.ServerInfo, inputOptions ...sdkclient.Option) (Client, error) {
-	var opts = sdkclient.DefaultOptions(sdkclient.MapAddr)
+	var opts = sdkclient.DefaultOptions(sdkclient.MapAddr, sdkserverinfo.MapServerInfoFile)
```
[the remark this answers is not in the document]
KeranYang: Please also learn about the Go option pattern and see if we are using it accurately here.

pkg/sdkclient/sideinput/client.go
```diff
@@ -21,8 +21,6 @@ type client struct {
 	grpcClt sideinputpb.SideInputClient
 }
 
-var _ Client = (*client)(nil)
```
KeranYang, line 24: Please keep this line. https://github.com/KeranYang/numaflow-notes/blob/main/GoFundamental.md#2-a-trick-to-ensure-an-object-implements-an-interface

pkg/sdkclient/sinker/client.go
```diff
@@ -35,23 +35,19 @@ type client struct {
 	grpcClt sinkpb.SinkClient
 }
 
-var _ Client = (*client)(nil)
```
KeranYang, line 38: please add back

pkg/sdkclient/source/client.go
```diff
@@ -36,26 +36,20 @@ type client struct {
 	grpcClt sourcepb.SourceClient
 }
 
-var _ Client = (*client)(nil)
```
KeranYang, line 39: please add back

KeranYang (approved):
LGTM. Please add back the interface check and then we can merge this in.

PR #1501, opened 2024-02-12
chore: minor fixes

yhl25:
* fix dedup for aligned reduce (TODO: unaligned)
* increase the nats ping interval from 1 second to 3 seconds to avoid a reconnect storm, when the server goes down.
* fix seg fault when the watermark is disabled in the pipeline
* adding ISB latency metrics for better debugging in case of an ISB issue
* adding a metric to track the number of control messages


whynowy (approved):
LGTM

pkg/udf/rpc/grpc_aligned_reduce.go
```diff
@@ -91,6 +97,7 @@ func (u *GRPCBasedAlignedReduce) ApplyReduce(ctx context.Context, partitionID *p
 
 	grpcCtx := metadata.NewOutgoingContext(ctx, metadata.New(mdMap))
 
+	index := 0
```
whynowy, line 100: Move it into the goroutine?

pkg/udf/rpc/grpc_aligned_reduce.go
```diff
@@ -104,7 +111,9 @@ func (u *GRPCBasedAlignedReduce) ApplyReduce(ctx context.Context, partitionID *p
 					close(responseCh)
 					return
 				}
-				responseCh <- parseReduceResponse(result)
+				// create a unique message id for each response message which will be used for deduplication
+				msgId := fmt.Sprintf("%s-%d-%s-%d", u.vertexName, u.vertexReplica, partitionID.String(), index)
+				responseCh <- parseReduceResponse(result, msgId)
```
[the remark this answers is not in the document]
yhl25: done

pkg/udf/rpc/grpc_aligned_reduce.go
yhl25: done

pkg/shared/idlehandler/idlehandler.go
```diff
@@ -18,17 +18,19 @@ package idlehandler
...
 	dfv1 "github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
 	"github.com/numaproj/numaflow/pkg/isb"
+	"github.com/numaproj/numaflow/pkg/metrics"
 	"github.com/numaproj/numaflow/pkg/watermark/publish"
 	"github.com/numaproj/numaflow/pkg/watermark/wmb"
 )
 
 // PublishIdleWatermark publishes a ctrl message with isb.Kind set to WMB. We only send one ctrl message when
-func PublishIdleWatermark(ctx context.Context, toBufferPartition isb.BufferWriter, wmPublisher publish.Publisher, idleManager wmb.IdleManager, logger *zap.SugaredLogger, vertexType dfv1.VertexType, wm wmb.Watermark) {
+func PublishIdleWatermark(ctx context.Context, toBufferPartition isb.BufferWriter, wmPublisher publish.Publisher, idleManager wmb.IdleManager, logger *zap.SugaredLogger, vertexName string, pipelineName string, vertexType dfv1.VertexType, vertexReplica int32, wm wmb.Watermark) {
```
vigith, line 33: NIT: all these new args just for metrics?

test/http-e2e/testdata/http-source.yaml
```diff
@@ -3,6 +3,8 @@ kind: Pipeline
 metadata:
   name: http-source
 spec:
+  watermark:
```
vigith, line 6: do we have one with watermark on?

vigith (approved):
this PR will increase the total metrics count (PPS, points per second). should we think about having tiered metrics, eg., P0 (must-haves), P1, P2, etc. and ability to turn on on-demand.

vigith:
@yhl25 please describe in the description about all the minor fixes 😄

test/http-e2e/testdata/http-source.yaml
yhl25: In all the other tests watermark is enabled

pkg/shared/idlehandler/idlehandler.go
yhl25: yes

yhl25:
> this PR will increase the total metrics count (PPS, points per second). should we think about having tiered metrics, eg., P0 (must-haves), P1, P2, etc. and ability to turn on on-demand.

yeah, I think we should.

PR #1502, opened 2024-02-12
fix: idle manager refactor

jy4096:
The idleManager is shared across all forwarders for map and sink vertex, but out current implementation doesn't take multiple forwarders into consideration, causing race condition when publishing control messages.

In this PR
1. Refactor the current idle manager interface to include fromBufferPartition.
2. Refactor the current idle manager structure to "singleIdleManager". It is used by source and reduce where one pod has only one forwarder. (TODO: revisit the source section)
3. Add a new "sharedIdleManager". It's used by map and sink where one pod could have multiple forwarders (each fromPartition has its own forwarder).

I originally tried to add the fromBufferPartitions to the current idleManager but found it to be complicated for source and reduce. Thus, I chose this approach: refactor and add a new data type.


pkg/watermark/wmb/shared_idle_manager.go
```diff
@@ -0,0 +1,93 @@
...
+func (im *sharedIdleManager) NeedToSendCtrlMsg(toBufferPartitionName string) bool {
+	im.lock.RLock()
+	defer im.lock.RUnlock()
+	for _, activePartitions := range im.forwarderActiveToPartition {
+		if activePartitions[toBufferPartitionName] {
+			// at least one (forwarder, toBufferPartition) is active
+			// no need to send ctrl msg
+			return false
+		}
+	}
```
vigith, line 65: thinking whether we can avoid this using bitops.

pkg/watermark/wmb/shared_idle_manager.go
```diff
@@ -0,0 +1,96 @@
...
+	// for the given forwarder.
+	forwarderActiveToPartition map[string]int64
+	lock                       sync.RWMutex
+}
+
+// NewSharedIdleManager returns an sharedIdleManager object as the IdleManager interface type to track the watermark idle status.
+func NewSharedIdleManager(numOfFromPartitions int, numOfToPartitions int) (IdleManager, error) {
+	if numOfFromPartitions <= 1 {
+		return nil, fmt.Errorf("failed to create a new shared idle manager: numOfFromPartitions should be > 1")
+	}
```
vigith, line 41:
```suggestion
	if numOfFromPartitions <= 1 || numOfFromPartitions > 64 {
		return nil, fmt.Errorf("failed to create a new shared idle manager: numOfFromPartitions should be > 1 and < 64")
	}
```

jy4096:
opened another PR https://github.com/numaproj/numaflow/pull/1512
closing this one...

PR #1512, opened 2024-02-16
fix: idle manager refactor for multi partitions

jy4096:
The idleManager is shared across all forwarders for map and sink vertex, but out current implementation doesn't take multiple forwarders into consideration, causing race condition when publishing control messages.

In this PR
- Refactor the current idle manager interface to include `fromBufferParitionIndex`.
- Refactor the current idle manager structure to have a map[toBufferParitionName]int64.
```go
// forwarderActivePartition is a map[toPartitionName]uint64 to record if a forwarder is sending to the toPartition.
// for each "toPartition" we have an integer represents in binary, and mark the #n bit as 1 if the #n forwarder
// is sending to the "toPartition"
// example:
//   if we have three forwarders, the initial value in binary format will be {"toPartition": 000} which is 0 in decimal
//   if forwarder0 is sending data to the toPartition, then the value will become {"toPartition": 001} which is 1 in decimal
//   if forwarder1 is sending data to the toPartition, then the value will become {"toPartition": 011} which is 3 in decimal
// when we do the ctrlMsg check, we reply on the value to decide if we need to send a ctrlMsg
// note that we use uint64, therefore there is a limit of maximum 64 partitions
forwarderActiveToPartition map[string]uint64
```

### DIFF
--- a/pkg/reduce/data_forward.go
+++ b/pkg/reduce/data_forward.go
@@ -251,7 +251,7 @@ func (df *DataForward) forwardAChunk(ctx context.Context) {
 			for toVertexName, toVertexBuffer := range df.toBuffers {
 				if publisher, ok := df.wmPublishers[toVertexName]; ok {
 					for _, bufferPartition := range toVertexBuffer {
-						idlehandler.PublishIdleWatermark(ctx, 0, bufferPartition, publisher, df.idleManager, df.log, df.vertexName, df.pipelineName, dfv1.VertexTypeReduceUDF, df.vertexReplica, wmb.Watermark(time.UnixMilli(processorWMB.Watermark)))
+						idlehandler.PublishIdleWatermark(ctx, wmb.PARTITION_0, bufferPartition, publisher, df.idleManager, df.log, df.vertexName, df.pipelineName, dfv1.VertexTypeReduceUDF, df.vertexReplica, wmb.Watermark(time.UnixMilli(processorWMB.Watermark)))
 					}
 				}
 			}
@@ -274,7 +274,7 @@ func (df *DataForward) forwardAChunk(ctx context.Context) {
 				for toVertexName, toVertexBuffer := range df.toBuffers {
 					if publisher, ok := df.wmPublishers[toVertexName]; ok {
 						for _, bufferPartition := range toVertexBuffer {
-							idlehandler.PublishIdleWatermark(ctx, 0, bufferPartition, publisher, df.idleManager, df.log, df.vertexName, df.pipelineName, dfv1.VertexTypeReduceUDF, df.vertexReplica, wmb.Watermark(watermark))
+							idlehandler.PublishIdleWatermark(ctx, wmb.PARTITION_0, bufferPartition, publisher, df.idleManager, df.log, df.vertexName, df.pipelineName, dfv1.VertexTypeReduceUDF, df.vertexReplica, wmb.Watermark(watermark))
 						}
 					}
 				}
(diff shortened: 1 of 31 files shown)


pkg/watermark/wmb/idle_manager.go
```diff
@@ -17,23 +17,38 @@ limitations under the License.
...
 type idleManager struct {
+	// forwarderActivePartition is a map[toPartitionName]uint64 to record if a forwarder is sending to the toPartition.
+	// for each "toPartition" we have an integer represents in binary, and mark the #n bit as 1 if the #n forwarder
+	// is sending to the "toPartition"
+	// example:
+	//   if we have three forwarders, the initial value in binary format will be {"toPartition": 000} which is 0 in decimal
+	//   if forwarder0 is sending data to the toPartition, then the value will become {"toPartition": 001} which is 1 in decimal
+	//   if forwarder1 is sending data to the toPartition, then the value will become {"toPartition": 011} which is 3 in decimal
+	// when we do the ctrlMsg check, we reply on the value to decide if we need to send a ctrlMsg
+	// note that we use uint64, therefore there is a limit of maximum 64 partitions
```
vigith, line 36:
```suggestion
	//   if we have three forwarders, the initial value in binary format will be {"toPartition": 000}, which is 0 in decimal
	//   if forwarder0 is sending data to the toPartition, then the value will become {"toPartition": 001}
	//   if forwarder1 is sending data to the toPartition, then the value will become {"toPartition": 010}
	//   if forwarder2 is sending data to the toPartition, then the value will become {"toPartition": 100}
	// when we do the ctrlMsg check, we rely on the value (0 or non-0) to decide if we need to send a ctrlMsg
	// note that we use uint64, therefore there is a limit of maximum 64 partitions
```

pkg/reduce/data_forward.go
```diff
@@ -251,7 +251,7 @@ func (df *DataForward) forwardAChunk(ctx context.Context) {
 			for toVertexName, toVertexBuffer := range df.toBuffers {
 				if publisher, ok := df.wmPublishers[toVertexName]; ok {
 					for _, bufferPartition := range toVertexBuffer {
-						idlehandler.PublishIdleWatermark(ctx, bufferPartition, publisher, df.idleManager, df.log, df.vertexName, df.pipelineName, dfv1.VertexTypeReduceUDF, df.vertexReplica, wmb.Watermark(time.UnixMilli(processorWMB.Watermark)))
+						idlehandler.PublishIdleWatermark(ctx, 0, bufferPartition, publisher, df.idleManager, df.log, df.vertexName, df.pipelineName, dfv1.VertexTypeReduceUDF, df.vertexReplica, wmb.Watermark(time.UnixMilli(processorWMB.Watermark)))
```
vigith, line 254: should we create a `const` for `0`th index called PARTITION_0? this will give a hint that we only have one partition?

vigith (approved):

pkg/watermark/wmb/idle_manager.go
```diff
@@ -17,29 +17,51 @@ limitations under the License.
 package wmb
 
 import (
+	"fmt"
 	"sync"
 
 	"github.com/numaproj/numaflow/pkg/isb"
 )
 
+const PARTITION_0 = 0
```
vigith, line 26:
```suggestion
// PARTITION_0 is used when it is the only partition of the buffer, and this is used only for those cases
// where we cannot have more than one partition (e.g., Reduce vertex, Source, etc.)
const PARTITION_0 = 0
```

PR #1515, opened 2024-02-20
fix: avoid panic when ctx is canceled

yhl25:


vigith (approved):
how was it working so far?

pkg/udf/rpc/errors.go
```diff
@@ -44,3 +44,8 @@ func (e ApplyUDFErr) IsInternalErr() bool {
 func (e ApplyUDFErr) Error() string {
 	return fmt.Sprint(e.Message)
 }
+
+// Is checks if the error is of the same type
+func (e ApplyUDFErr) Is(target error) bool {
```
whynowy, line 49: Where is the function used?

vigith: i think it is coming from "session" code.

vigith: https://github.com/numaproj/numaflow/blob/main/pkg/reduce/pnf/processandforward.go#L136

vigith: looks like it hit the never in `session` code :)

whynowy (approved):

PR #1516, opened 2024-02-20
fix: podSpec incorrectly configured in case template exist

### LINKED ISSUE #1514: ServiceAccountName incorrectly configured in case vertex template exist
**Describe the bug**

In case the vertex template is configured, the `serviceAccountName` isn't respected and is set to `default`.

**To Reproduce**

A sample pipeline - in this case the `cat` vertice should be configured with serviceAccountName as `test-two` instead of `default`.

```yaml
kind: Pipeline
metadata:
  name: simple-pipeline
spec:
  templates:
    vertex:
      metadata:
        annotations:
          cluster-autoscaler.kubernetes.io/safe-to-evict: "true"
      tolerations:
        - key: "kubernetes.azure.com/scalesetpriority"
          operator: "Equal"
          value: "spot"
          effect: "NoSchedule"
  vertices:
    - name: in
      source:
        generator:
          rpu: 5
          duration: 1s
    - name: cat
      serviceAccountName: "test-two"
      udf:
        builtin:
          name: cat # A built-in UDF which simply cats the message
    - name: out
      sink:
        # A simple log printing sink
        log: {}
  edges:
    - from: in
      to: cat
    - from: cat
      to: out
```

**Expected behavior**
The serviceAccountName at the vertice level should be adhered

**Screenshots**

- Kubernetes: v1.29.2
- Numaflow: v1.1.5

**Additional context**

---

inishchith:
closes #1514

With this change PodSpec fields are assigned with values from template only if they are null/empty.


whynowy (dismissed):
LGTM, thanks!

inishchith:
@whynowy - not entirely sure what's breaking. if you could help identify, happy to fix and get this through

whynowy:
> @whynowy - not entirely sure what's breaking. if you could help identify, happy to fix and get this through

Thanks @inishchith ! I will take a look.

whynowy:
> E2e failed because of our recent e2e test image upgrade. It is now fixed after rebasing. cc: @whynowy

:) Thanks @KeranYang

whynowy (approved):

PR #1518, opened 2024-02-21
feat: add support for SASL SCRAM 256 and 512 for Kafka

Dnomyar:
Hey folks,

First, thanks for this project!

I am proposing to add support for SASL SCRAM 256 and 512 for the Kafka connector.
I only tested SASL SCRAM 512 with a source (not a sink).

There are a few things that are missing:
- tests
- some code is duplicated

This is my first time doing Go so I appreciate any feedback

Cheers


pkg/shared/util/scram_client.go
```diff
@@ -0,0 +1,38 @@
+// copied from https://github.com/IBM/sarama/blob/main/examples/sasl_scram_client/scram_client.go
+package util
```
whynowy, line 2: Maybe combining this file with sasl_config.go, since it's only used for that at this moment?

pkg/apis/numaflow/v1alpha1/sasl.go
```diff
@@ -90,3 +96,11 @@ type SASLPlain struct {
 	PasswordSecret *corev1.SecretKeySelector `json:"passwordSecret" protobuf:"bytes,2,opt,name=passwordSecret"`
 	Handshake      bool                      `json:"handshake" protobuf:"bytes,3,opt,name=handshake"`
 }
+type SASLSCRAMSHA struct {
```
whynowy, line 99: Nit: Leave an empty line before it.

pkg/apis/numaflow/v1alpha1/sasl.go
```diff
@@ -90,3 +96,11 @@ type SASLPlain struct {
...
+	// UserSecret refers to the secret that contains the user
+	UserSecret *corev1.SecretKeySelector `json:"userSecret" protobuf:"bytes,1,opt,name=userSecret"`
```
whynowy, line 101: Use embedding struct `BasicAuth`?

vigith (commented):
LGTM, thank you for the contribution :)

vigith:
could you please update the doc in https://github.com/numaproj/numaflow/blob/main/docs/user-guide/sources/kafka.md and https://github.com/numaproj/numaflow/blob/main/docs/user-guide/sinks/kafka.md to include SCRAM 256 and 512?

Dnomyar:
I have updated the docs and added some tests. 

Could you indicate how should I fix conflicts for the go.sum, please? I am not so familiar with go

vigith:
can you do a git-pull and run `make codegen` ?

pkg/shared/util/volume.go
```diff
@@ -30,9 +30,20 @@ var (
 	configMapKeySelectorType = reflect.TypeOf(&corev1.ConfigMapKeySelector{})
 )
 
+type GetSecrets interface {
```
whynowy, line 33: 1. I assume the purpose of the interface is to be able to mock it for testing? If that's the case, make it and `OsFile`?
2. The name does not match the 2 functions.

pkg/shared/util/sasl_config.go
```diff
@@ -37,6 +40,23 @@ func GetSASL(saslConfig *dfv1.SASL) (*struct {
 	SCRAMClientGeneratorFunc func() sarama.SCRAMClient
 	TokenProvider            sarama.AccessTokenProvider
 	GSSAPI                   sarama.GSSAPIConfig
+}, error) {
+	return GetSASLStrategy(saslConfig, OsFile{})
+}
+
+// GetSASLConfig A utility function to get sarama.Config.Net.SASL
+func GetSASLStrategy[GetSecretsStrategy GetSecrets](saslConfig *dfv1.SASL, strategy GetSecretsStrategy) (*struct {
```
whynowy, line 48: What is the reason you use generics instead of just writing it as:

```
func GetSASLStrategy(saslConfig *dfv1.SASL, strategy GetSecrets) (*struct {
...
```

whynowy: Also, make the function private.

pkg/shared/util/sasl_config.go
```diff
@@ -54,30 +74,60 @@ func GetSASL(saslConfig *dfv1.SASL) (*struct {
...
+			}
+			config.Net.SASL.Handshake = scram.Handshake
+		}
 	default:
 		return nil, fmt.Errorf("SASL mechanism not supported: %s", *saslConfig.Mechanism)
 	}
 	return &config.Net.SASL, nil
 }
 
+func SetUserPassword[GetSecretsStrategy GetSecrets](config *sarama.Config, userSecret *corev1.SecretKeySelector, passwordSecret *corev1.SecretKeySelector, strategy GetSecretsStrategy) error {
```
whynowy, line 111: Make it private.

pkg/shared/util/volume.go
```diff
@@ -30,9 +30,20 @@ var (
...
+type OsFile struct{}
+
+func GetSecretFromVolume(selector *corev1.SecretKeySelector) (string, error) {
+	return OsFile{}.GetSecretFromVolume(selector)
+}
+
 // GetSecretFromVolume retrieves the value of mounted secret volume
 // "/var/numaflow/secrets/${secretRef.name}/${secretRef.key}" is expected to be the file path
-func GetSecretFromVolume(selector *corev1.SecretKeySelector) (string, error) {
+func (_ OsFile) GetSecretFromVolume(selector *corev1.SecretKeySelector) (string, error) {
```
whynowy, line 46: Using underscore will bring warning in some IDEs.

pkg/shared/util/volume.go
Dnomyar: > I assume the purpose of the interface is to be able to mock it for testing?
yes I added that as a comment 

> If that's the case, make it and OsFile?
Sorry, I don't understand, what you mean by that

> The name does not match the 2 functions.
I renamed to `volumeReader`, what do you think?

pkg/shared/util/sasl_config.go
Dnomyar: >What is the reason you use generics instead of just writing it as:
That's a very good point! Changed!

> Also, make the function private.
I went through the code I changed the visibility of things that should be private

pkg/shared/util/volume.go
Dnomyar: >Using underscore will bring warning in some IDEs.
I thought the underscore was required, changed!

Dnomyar:
Thanks for you feedback @whynowy! I think I have fixed most things. Let me know if I forgot anything

whynowy (approved):
LGTM, thanks!

vigith:
thank you @Dnomyar for your contribution. Would you like to add your use-case to https://github.com/numaproj/numaflow/blob/main/USERS.md?

syayi:
Thank you @Dnomyar for the contribution! We're keen to hear about your experience with Numaflow and gather feedback, would you be open for a chat?

PR #1532, opened 2024-03-02
feat: noop persistence store for reduce

### LINKED ISSUE #1523: Support for no-op persistence storage for groupBy
# Summary
For our use case we don't need to persist the message since the data is not critical, currently numaflow enforces us to provide either PVC or emptyDir as persistence storage.

yhl25:
fixes #1523


pkg/apis/numaflow/v1alpha1/udf.go
```diff
@@ -168,8 +168,14 @@ type PBQStorage struct {
 	PersistentVolumeClaim *PersistenceStrategy `json:"persistentVolumeClaim,omitempty" protobuf:"bytes,1,opt,name=persistentVolumeClaim"`
 	// +optional
 	EmptyDir *corev1.EmptyDirVolumeSource `json:"emptyDir,omitempty" protobuf:"bytes,2,opt,name=emptyDir"`
+	// +optional
+	// add a none empty object
+	None *None `json:"none,omitempty" protobuf:"bytes,3,opt,name=none"`
 }
 
+// None is used to indicate no persistence storage is needed for a vertex.
```
vigith, line 176:
```suggestion
// None means there will be no persistence storage and there will be data loss during pod restarts.
// Use this option only if you do not care about correctness (e.g., approx statistics pipeline like sampling rate, etc.).
```

vigith (approved):

PR #1535, opened 2024-03-03
fix: break from retry loop when key is not found

### LINKED ISSUE #1536: Delete KeyValue entry from the watermark bucket when the processor goes down

### LINKED ISSUE #1540: Add unit tests for KeyValue store and watcher

yhl25:
* update the nats sdk
* avoid retrying forever when the key is not found 
* unit tests for kv store 
* update the Watch() function signature to avoid race condition between updates and done channel

Closes #1540 
Closes #1536


pkg/shared/kvs/jetstream/kv_store.go
```diff
@@ -289,22 +290,28 @@ retryLoop:
...
 				time.Sleep(100 * time.Millisecond)
+				value, err = jss.kv.Get(key)
 			}
 		}
 		if value.Created().After(lastUpdate) {
 			lastUpdate = value.Created()
 		}
 	}
+
+	_ = keyLister.Stop()
```
vigith, line 314: will we end up in double stopping because KeyLister closes stop if all is read?

pkg/shared/kvs/jetstream/kv_store.go
```diff
@@ -289,22 +290,28 @@ retryLoop:
 
 	}
 
-	for _, key := range keys {
+	for key := range keyLister.Keys() {
 		value, err = jss.kv.Get(key)
 		for err != nil {
+			// keys can be deleted when the previous vertex pod is deleted/restarted.
+			if errors.Is(err, nats.ErrKeyNotFound) {
```
vigith, line 297: let's add a Info statement here.

pkg/shared/kvs/jetstream/kv_store.go
whynowy: Shall we use `defer`?

test/idle-source-e2e/idle_source_test.go
```diff
@@ -47,15 +47,16 @@ func (is *IdleSourceSuite) TestIdleKeyedReducePipeline() {
 	done := make(chan struct{})
 	go func() {
 		// publish messages to source vertex, with event time starting from 1000
```
vigith, line 49:
```suggestion
		// publish messages to source vertex, with event time starting from 0
```

vigith (approved):

PR #1548, opened 2024-03-07
fix: initialize inflightAcks channel to not nil channel

### LINKED ISSUE #1541: Kafka source panic does not restart the container
**Describe the bug**

In Kafka source implementation, if there's any error returned from kafka server side during message consumption, we panic the service intentionally to let the container restart, so that it could start over to initialize the application. However, there are chances that the panic log is seen but the container does not restart, the log shows as below.

```
"panic: Kafka consumer failed with error:
goroutine 358 [running]:
go.uber.org/zap/zapcore.CheckWriteAction.OnWrite(0x1?, 0x210ca40?, {0x0?, 0x0?, 0xc000e07420?})
        /home/runner/go/pkg/mod/go.uber.org/zap@v1.24.0/zapcore/entry.go:198 +0x65
go.uber.org/zap/zapcore.(*CheckedEntry).Write(0xc002b74ea0, {0xc0046f4500, 0x1, 0x1})
        /home/runner/go/pkg/mod/go.uber.org/zap@v1.24.0/zapcore/entry.go:264 +0x3ec
go.uber.org/zap.(*SugaredLogger).log(0xc000491be0, 0x4, {0x25c4562?, 0x2940610?}, {0x0?, 0xc0003d4eb8?, 0xc00011df88?}, {0xc00a607f28, 0x1, 0x1})
        /home/runner/go/pkg/mod/go.uber.org/zap@v1.24.0/sugar.go:295 +0xee
go.uber.org/zap.(*SugaredLogger).Panicw(...)
        /home/runner/go/pkg/mod/go.uber.org/zap@v1.24.0/sugar.go:235
github.com/numaproj/numaflow/pkg/sources/kafka.(*kafkaSource).startConsumer.func2()
        /home/runner/work/numaflow/numaflow/pkg/sources/kafka/reader.go:414 +0x225
created by github.com/numaproj/numaflow/pkg/sources/kafka.(*kafkaSource).startConsumer
        /home/runner/work/numaflow/numaflow/pkg/sources/kafka/reader.go:406 +0x54a"
```

**Expected behavior**

The container should restart.

**Screenshots**

- Kubernetes: [e.g. v1.18.6]
- Numaflow: [e.g. v0.5.1]

**Additional context**

---

afugazzotto:
Closes #1541

Initialize inflightAcks channel to not nil channel to avoid hanging on cleanup function when there are no messages to acknowledge.

How to Test/Reproduce:
1. setup a Kafka client and a Kafka topic with 10 partitions
2. in a bash script, set Kafka client to send 10 messages per second
3. setup a Numaflow pipeline with source vertex to consume from the Kafka client (disable autoscaling)
4. manually scale source vertex to 12 or more pods (`kubectl scale vtx <vtx_name> --replicas=12`)
5. manually delete pod-0
6. wait for a new pod-0 to restart
7. verify it has no topics assigned to it in the metrics: `kafka_source_read_total`
8. scale down to 1 pod
9. at this point, before the fix, the messages flow to the log sink would stop. After the fix, pod-0 is not stuck in the cleanup state and can resume consuming messages (`kafka_source_read_total` should show for all 10 partitions)

### DIFF
--- a/pkg/sources/kafka/handler.go
+++ b/pkg/sources/kafka/handler.go
@@ -57,14 +57,17 @@ func (consumer *consumerHandler) Setup(sess sarama.ConsumerGroupSession) error {
 	consumer.readyCloser.Do(func() {
 		close(consumer.ready)
 	})
+	consumer.logger.Info("Kafka Consumer Setup complete")
 	return nil
 }
 
 // Cleanup is run at the end of a session, once all ConsumeClaim goroutines have exited
 func (consumer *consumerHandler) Cleanup(sess sarama.ConsumerGroupSession) error {
+	consumer.logger.Info("Kafka Consumer Starting Cleanup routine, waiting for in-flight-acks to complete")
 	// wait for inflight acks to be completed.
 	<-consumer.inflightAcks
 	sess.Commit()
+	consumer.logger.Info("Kafka Consumer Cleanup complete")
 	return nil
 }
 
@@ -73,6 +76,7 @@ func (consumer *consumerHandler) ConsumeClaim(session sarama.ConsumerGroupSessio
 	// The `ConsumeClaim` itself is called within a goroutine, see:
 	// https://github.com/IBM/sarama/blob/main/consumer_group.go#L27-L29
 	for {
+		consumer.logger.Info("Kafka Consumer about to claim Messages from the Kafka broker")
 		select {
 		case msg, ok := <-claim.Messages():
 			if !ok {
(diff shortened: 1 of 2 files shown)


KeranYang (commented):
Why do we need `inflightAcks` at first place? My understanding is Ack is a blocking call hence `inflightAcks` should be either nil or closed. Am I missing anything? cc: @vigith

afugazzotto:
> Why do we need `inflightAcks` at first place? My understanding is Ack is a blocking call hence `inflightAcks` should be either nil or closed. Am I missing anything? cc: @vigith

Based on this question, would it make more sense or be preferred to remove the inflight ack check in the Cleanup function (https://github.com/numaproj/numaflow/blob/main/pkg/sources/kafka/handler.go#L59)?

KeranYang:
> would it make more sense (or be preferred) to remove the inflight ack "check" in the Cleanup function

or completely remove the var `inflightAcks`?

vigith (approved):

KeranYang (approved):

PR #1561, opened 2024-03-14
doc: add section to the dev docs about e2e testing

QuentinFAIDIDE:
This change add a section about end-to-end tests to the documentation. These are things that I found useful and had to figure out by reading the code or Makefile.
I'm not a native English speaker and I just got started playing around with Numaflow, so it's far for perfect, but you are welcomed to suggest enhancements, correct errors or refuse the PR.


KeranYang:
Thank you, @QuentinFAIDIDE for improving the doc. Regarding e2e testing, I agree that we should add the command to the development file. I don't think we need to be detailed since as you mentioned above users can have different environments etc. I will create a PR to address this.

KeranYang:
https://github.com/numaproj/numaflow/pull/1562

PR #1565, opened 2024-03-14
feat: read Kafka header and propagate in the payload

vigith:
Fixes: #1566 

Propagate Kafka RecordHeaders in the Payload. This changes the payload struct though.

## With Header
If header is included it will be of the following format

```json
{
  "body": "... original payload ...",
  "headers": {
    "k1": "v1",
    "k2": "v2"
  }
}
```

## Without Header (default)

The format won't change, the payload will be a `[]byte` seq.


yhl25 (approved):
should we support headers in kafka sink as well?

PR #1574, opened 2024-03-18
chore: fix noack when all the messages fail

### LINKED ISSUE #1563: Flaky e2e test TestSimpleSessionPipelineFailOverUsingWAL
The TestSimpleSessionPipelineFailOverUsingWAL doesn't expect string `1` in the sink but some times we get 2 of it.

example: https://github.com/numaproj/numaflow/actions/runs/8274462313/job/22668313990

yhl25:
* pass ctx to jetstream fetch method(to avoid reading when the context is cancelled).
* no ack function was not invoked for failed messages when there were no successful messages written to PBQ(PBQ writes only fail during sigterm, which can lead to messages being redelivered when the new pod comes up).
* add watermark delay of 60s to e2e test(since we are using HTTP source, data can come out of order).

fixes #1563


KeranYang:
Thanks @yhl25 for fixing it. Will this change also fix #1575 ?

KeranYang (approved):

PR #1577, opened 2024-03-19
chore: increase idle source e2e test timeout

yhl25:


pkg/isb/stores/jetstream/reader.go
```diff
@@ -122,7 +122,7 @@ func (jr *jetStreamReader) Read(ctx context.Context, count int64) ([]*isb.ReadMe
 	rctx, cancel := context.WithTimeout(ctx, jr.opts.readTimeOut)
 	defer cancel()
 	msgs, err := jr.sub.Fetch(int(count), nats.Context(rctx))
-	if err != nil && !errors.Is(err, nats.ErrTimeout) {
+	if err != nil && !errors.Is(err, nats.ErrTimeout) && !errors.Is(err, context.DeadlineExceeded) && !errors.Is(err, context.Canceled) {
 		isbReadErrors.With(map[string]string{"buffer": jr.GetName()}).Inc()
 		return nil, fmt.Errorf("failed to fetch messages from jet stream subject %q, %w", jr.subject, err)
 	}
```
vigith, line 128: do we really need all this check.. why not just `err != nil`?

vigith (approved):

pkg/isb/stores/jetstream/reader.go
yhl25: yes, those are not errors.. we don't want to log those

PR #1578, opened 2024-03-19
feat: support headers for message

### LINKED ISSUE #1568: Sources have metadata associated with payload and we need to propagate those along with the payload
# Summary

Numaflow source should have a way to propagate headers (e.g., Kafka Record Headers, HTTP Headers, etc.) along with the payload. The only way it is possible today is by passing the headers along with the payload after encoding it in JSON. 

---

We could perhaps pass it in ISB.Header and expose it via the Metadata interface. This is not a clean approach because only sources have metadata and could be dropped after the first hop (transformer, Map UDF, etc.).

Another approach is not to include and let users write custom User-defined sources if they need to access that data. The most significant side effect is the need for sharability of commonly used sources, and hence, it is not worthwhile to reimplement them repeatedly.

# Use Cases

Needs to access Header in processing vertex because they contain valuable metadata.

---

### LINKED ISSUE #1572: Pass Headers KV[string]string from source to sink
# Summary

`ISB.Header` should support an additional field which a map of key, value of `String` type. Since headers are part of the data, the `Datum`, we should expose `Datum.GetHeaders()` method that returns `KV[string]string`.

# Use Cases

- Sources like HTTP, Kafka, etc., have headers and UDFs want to access those
- Distributed Tracing
- Ability to pass transaction-id(tx_id), auth_headers, etc.

# Notes

It is easy to propagate for `Map` vertices, but for `Reduce`, we cannot do a merge of many to one, hence it is okay to create an empty header after the `Reduce` operation.

---

```[tasklist]
### Numaflow
- [ ] add KV[string]string to isb.Headers
- [ ] update WALs to take in the headers
- [ ] expose GetHeaders() in SDKs
- [ ] SetHeader in User-Defined Source
- [ ] Reset Header after Reduce (M:1 merge)
```

```[tasklist]
### SDK Tracking
- [ ] https://github.com/numaproj/numaflow/issues/1586
- [ ] https://github.com/numaproj/numaflow/issues/1587
- [ ] https://github.com/numaproj/numaflow/issues/1588
```

yhl25:
closes #1572
closes #1568


pkg/sources/kafka/reader.go
```diff
@@ -454,44 +453,21 @@ func (ks *kafkaSource) toReadMessage(m *sarama.ConsumerMessage) *isb.ReadMessage
 		topic:        m.Topic,
 	}
 
-	var body isb.Body
+	var body = isb.Body{Payload: m.Value}
+	var headers = make(map[string]string, len(m.Headers))
 
 	if ks.includeHeaders {
```
whynowy, line 459: Remove `includeHeaders`?

pkg/sources/http/http.go
```diff
@@ -159,11 +159,18 @@ func New(
 			}
 			eventTime = time.UnixMilli(i)
 		}
+
+		headers := make(map[string]string, len(r.Header))
+		for k, v := range r.Header {
+			headers[k] = v[0]
```
whynowy, line 165: Exclude headers like `Authorization`?

whynowy: Also leave a comment telling we only take the first value of the array.

vigith: Should we pass in Auth headers, too? For example, what if some downstream vertex wants to access an external entity using the same auth headers, e.g., if the auth is BrowserAuth+ and passes the user's token for accessing the store?

whynowy: This is going to be a security concern.

The `Authorization` header is supposed to be only accessed by the client who initializes the request, and the server who receives the request, and it's encrypted during transmission. If we keep it and store it in ISB, that means the credential is leaked to and stored in a 3rd party.

vigith: Yes. We will have to enable encryption at rest and in transit.

vigith: opened #1583

whynowy (approved):

vigith (approved):

PR #1590, opened 2024-03-21
fix: flaky e2e tests

vigith:
Our E2E tests are flaky, and we do not know why. This is an attempt to understand and possibly fix it.


vigith:
idle-source failed again, how come i am not seeing the logs? didn't we enable by default?

yhl25 (approved):

Issue #1413, opened 2023-12-13
Buttons overflow in the UI


veds-g:
While creating Pipeline / ISB in the UI, in case of a long error while validating / submitting,  the buttons overflow the container

<img width="1727" alt="image" src="https://github.com/numaproj/numaflow/assets/49195734/a2001c83-5450-48be-984f-035b6c1836a8">
<img width="1728" alt="image" src="https://github.com/numaproj/numaflow/assets/49195734/dcd96bcb-5c11-4611-b181-30f12ff870ee">

adarsh0728:
Closed with https://github.com/numaproj/numaflow/pull/1591

PR #1591, opened 2024-03-21
fix: error message that cause Buttons overflow in the UI

a49ibrahim:
Explain what this PR does.
 
This PR fixes the text over flow from the error message in the specEditor from the issue here https://github.com/numaproj/numaflow/issues/1413


darshansimha (approved):
LGTM!

a49ibrahim:
Screenshot 2024-03-22 at 9 53 45 AM

For testing purposes. Tried to put a video but video size is too large.

Issue #1363, opened 2023-11-15
CI should create a stable tag for SDKs 


vigith:
Maybe we should give a tag `stable` for the image, so that we do not need to update it in the test cases.

_Originally posted by @whynowy in https://github.com/numaproj/numaflow/issues/1360#issuecomment-1811925843_

KeranYang:
Adding a formal problem statement... 

Every time we make a change to the SDK, we need to manually rebuild the images under the example folder, and update the image tags under the numaflow test/ folder, such that the e2e test suites can run with the latest SDK changes. It is an operational inefficiency. If we can automate the process, we streamline the development process and also free engineering time to work on innovative tasks.

@a3hadi I am assigning this one to you. We can work on it as the next task.

PR #1594, opened 2024-03-22
chore: update example image tags to use stable

ayildirim21:
To be consistent with the recent CI improvements, as per [#1363](https://github.com/numaproj/numaflow/issues/1363)

### DIFF
--- a/test/reduce-two-e2e/testdata/session-reduce/simple-session-sum-pipeline.yaml
+++ b/test/reduce-two-e2e/testdata/session-reduce/simple-session-sum-pipeline.yaml
@@ -15,8 +15,8 @@ spec:
       partitions: 1
       udf:
         container:
-          # see https://github.com/numaproj/numaflow-go/tree/main/pkg/sessionreducer/examples/counter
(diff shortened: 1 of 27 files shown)


vigith (approved):

test/reduce-two-e2e/testdata/session-reduce/simple-session-sum-pipeline.yaml
```diff
@@ -16,7 +16,7 @@ spec:
       udf:
         container:
           # see https://github.com/numaproj/numaflow-go/tree/main/pkg/sessionreducer/examples/counter
-          image: quay.io/numaio/numaflow-go/session-sum:v0.6.1
+          image: quay.io/numaio/numaflow-go/session-counter:stable
```
KeranYang, line 19: Should we use session-sum?

KeranYang: please also update the link.

KeranYang (approved):

PR #1603, opened 2024-03-26
test: remove unnecessary pod ID lookup

whynowy:
Explain what this PR does.

### DIFF
--- a/test/fixtures/when.go
+++ b/test/fixtures/when.go
@@ -42,33 +42,12 @@ type When struct {
 
 	portForwarderStopChannels map[string]chan struct{}
 	streamLogsStopChannels    map[string]chan struct{}
-	// Key: vertex label selector
-	// Value: the ip of, one of the pods matching the label selector
-	vertexToPodIpMapping map[string]string
 }
 
 // SendMessageTo sends msg to one of the pods in http source vertex.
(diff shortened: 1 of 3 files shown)


test/fixtures/when.go
```diff
@@ -42,33 +42,12 @@ type When struct {
...
-		}
-		if len(podList.Items) == 0 {
-			w.t.Fatalf("No running pod found in pipeline %s, vertex: %s", pipelineName, vertexName)
-		}
-		w.vertexToPodIpMapping[labelSelector] = podList.Items[0].Status.PodIP
-	}
-
-	// There could be a rare corner case when a previous added pod gets replaced by a new one, making the mapping entry no longer valid.
-	// Considering current e2e tests are all lightweight, we assume no such case for now.
-	SendMessageTo(w.vertexToPodIpMapping[labelSelector], vertexName, req)
```
KeranYang, line 71: Thanks for making the change Derek. I was not aware of the headless service. nit: if you search for the keyword `podIp` across the code base, you will see that I used it a lot. I think with this change, we can update it to maybe `headlessEndpoint` or another better name.

KeranYang (approved):

vigith (approved):

PR #1613, opened 2024-03-29
fix: dedup in user defined source

### LINKED ISSUE #1611: User-defined source write metrics not matching the read metrics from next vertex
**Describe the bug**

Kafka is showing 50K TPS, while Sink is showing only 4K TPS.

image_udsource-bug

yhl25:
fixes #1611

### DIFF
--- a/pkg/sources/udsource/utils/offset.go
+++ b/pkg/sources/udsource/utils/offset.go
@@ -18,6 +18,7 @@ package utils
 
 import (
 	"encoding/base64"
+	"fmt"
 
 	sourcepb "github.com/numaproj/numaflow-go/pkg/apis/proto/source/v1"
 
@@ -40,7 +41,7 @@ func NewSimpleSourceOffset(offset *sourcepb.Offset) isb.Offset {
 }
 
 func (s *simpleSourceOffset) String() string {
-	return s.offset
+	return fmt.Sprintf("%s-%d", s.offset, s.partitionIdx)
 }
 
 func (s *simpleSourceOffset) PartitionIdx() int32 {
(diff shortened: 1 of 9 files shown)


vigith (approved):

whynowy (approved):

pkg/sources/udsource/utils/offset.go
```diff
@@ -56,12 +60,9 @@ func (s *simpleSourceOffset) NoAck() error {
 }
 
 func ConvertToSourceOffset(offset isb.Offset) *sourcepb.Offset {
+	decoded, _ := base64.StdEncoding.DecodeString(offset.String())
```
qhuai, line 63: Could DecodeString ever fail?

yhl25: With us controlling the encoding, `DecodeString` should never fail.

KeranYang (commented):
Thank you @yhl25  for root causing and fixing this. My understanding is adding jetstream dedup metric helps with debugging the issue and adding partition index to write message id fixes the issue. Do we need the other changes under `udsource` folder and `utils` folder? I agree that base64 encoding is safer than `string()`.

pkg/sources/udsource/utils/offset.go
```diff
@@ -17,21 +17,25 @@ limitations under the License.
...
 
-func NewSimpleSourceOffset(o string, p int32) isb.Offset {
+func NewSimpleSourceOffset(offset *sourcepb.Offset) isb.Offset {
 	return &simpleSourceOffset{
-		offset:       o,
-		partitionIdx: p,
+		offset:       base64.StdEncoding.EncodeToString(offset.GetOffset()),
+		partitionIdx: offset.GetPartitionId(),
 	}
 }
```
KeranYang, line 41: Should we include partitionIdx in the `String()` of offset as you did in your first commit of the PR? Since partitionIdx is now an integral part of simpleSourceOffset.

KeranYang (approved):

vigith (changes requested):
based on our meeting we need a few more changes

vigith (approved):

PR #1615, opened 2024-03-29
Add a company that uses Numaflow to the list of users

QuentinFAIDIDE:
Document a new Numaflow user.


QuentinFAIDIDE:
The change was based off my Numaflow fork main branch which has diverted 2 commits from master. Going to make a new clean one.

PR #1616, opened 2024-03-29
ci: upgrade deprecated actions

whynowy:
Explain what this PR does.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,13 +13,13 @@ jobs:
     timeout-minutes: 6
     steps:
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: Setup Node
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
-          node-version: "16"
+          node-version: "20"
       - name: Setup Node-Cache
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yaml
```diff
@@ -137,32 +137,31 @@ jobs:
     timeout-minutes: 20
     strategy:
       fail-fast: false
-      max-parallel: 13
```
whynowy, line 140: Maximumly use the resources.

vigith (approved):

vigith:
We need to have a different strategy for redis since most of them are Skip. The current setup is very slow.

whynowy:
> We need to have a different strategy for redis since most of them are Skip. The current setup is very slow.

true.

vigith:
> > We need to have a different strategy for redis since most of them are Skip. The current setup is very slow.
> 
> true.

Disable Redis for the time being?

whynowy:
> > > We need to have a different strategy for redis since most of them are Skip. The current setup is very slow.
> > 
> > 
> > true.
> 
> Disable Redis for the time being?



> > > We need to have a different strategy for redis since most of them are Skip. The current setup is very slow.
> > 
> > 
> > true.
> 
> Disable Redis for the time being?

Partial enable.

Issue #1554, opened 2024-03-12
Buffer Usage Calculation and ISB Writing Race Condition: Potential Data Loss in High-Throughput Pipelines


yhl25:
https://github.com/numaproj/numaflow/issues/1551

vigith:
Move to 1.3 because we have a work around.

QuentinFAIDIDE:
I was thinking of something. Can't we modify the discard policy in the buffer creation right here ?
https://github.com/numaproj/numaflow/blob/f1e5ba0eb222edf3e5a5593769efc3626b092c1b/pkg/isbsvc/jetstream_service.go#L117

We hard-code this `DiscardOld` policy, which is the reason we encounter data loss instead of a write error according to the [Jetstream doc](https://docs.nats.io/using-nats/developer/develop_jetstream/model_deep_dive).
```
The Discard Policy sets how messages are discarded when limits set by LimitsPolicy are reached. The DiscardOld option removes old messages making space for new, while DiscardNew refuses any new messages.
```

QuentinFAIDIDE:
Could you take a minute to elaborate on why it can't work ? I got all the test passing in a draft PR and have started torturing Numaflow with this change, and for now no issue yet. Sorry for being persistent, I'm a curious person, I will leave you in peace once I understand :laughing: 
Also, it feels like it would make your lives easier to not be living under a buffer that silently drops data when it's full, but just error out :)

yhl25:
When the stream is configured with the `Limits` policy and `DiscardOld`, Jetstream will automatically delete old messages if either the maxMsgs limit of 100k or the maxAge limit of 72h is met. However, when operating with `DiscardNew` and `Limits`, reaching the maxMsgs of 100k will result in the stream returning an error when attempting to write new messages and old messages will not be deleted from the stream, so the pipeline will get stuck. On the other hand, using `DiscardNew` with `WorkQueue` works fine because Jetstream deletes messages once they are consumed and acked.

PR #1624, opened 2024-04-01
fix: support DiscardNew policy for Jetstream streams

### LINKED ISSUE #1551: Messages dropped in Sinks when buffer is full and sink is failing
**Describe the bug**
I have noticed a few times that messages could disappear in very specific conditions. I accidentally discovered how to reproduce the issue in my pipelines. I apologize for the lengthy description, I have not yet managed to reproduce it in a simple pipeline. Here is a detailed description of the issue:

The pipeline is made of the following vertices:
image

- **block-metadata** is a source that broadcast ordered metadata for transaction blocks. It goes from offset 0 to 800k without gaps. The offset is stored in redis for resilient restarts. I clear the redis before creating the pipeline, and I can confirm it starts at height 0 thanks to the witness writing all first blocks.
- **block-tx-fetch**: uses the block metadata to download all the block transactions, it has a ratio of approximately 3000 transactions (output) for 1 block (input). It outputs for each block the transactions, and a block metadata.
- The upper part with coinjoin and tx-writer is not relevant here.
- The **block-writer** writes block metadata into Scylla/Cassandra.
- The **block-witness** does the same, but unlike **block-writer**, I always set it to a valid keyspace to serve as witness of the lost data in **block-writer**.

If I start the pipeline, and configure `block-writer` on an invalid keyspace that doesn't have the block table, it will set the sink response to failure with the error message from the gocql driver for cassandra: `underfined table block` (UDSink code below).
As mentioned earlier, I set `block-witness` to a valid keyspace `btc_raw_dev_b` that has the block table.
After I let it run for a while, until the buffer for the failing block-writer gets full, I edit the vertex yaml in Numaflow webui to set it to a valid keyspace named `btc_raw_dev`.
I then wait for the vertex restart, and the now healthy **block-writer** sink consumes messages that are in the buffer untill there is nothing left in it.
At the end of the process, I end up with the `btc_raw_dev` block table missing the first blocks, while the `btc_raw_dev_b` one has all of these missing blocks, and show no sign of missing entries:
image
I have reproduced this error consistently, it never fails to happen in the described context. I found missing entries in the past in my output tables, even without the misconfigurations, so I imagine that errors in the Sink are not properly recovered on db failures.
Here is the code of the Sink (I use the go SDK):
```go
// Sink is the sink function that relies on the cqlRawClient's WriteBlock utility
func (l *BlockSink) Sink(ctx context.Context, datumStreamCh <-chan sinksdk.Datum) sinksdk.Responses {
	result := sinksdk.ResponsesBuilder()
	for d := range datumStreamCh {

		blockJSON := d.Value()
		var block bitcoinclients.Block
		err := json.Unmarshal(blockJSON, &block)
		if err != nil {
			result = result.Append(sinksdk.ResponseFailure(d.ID(), err.Error()))
			continue
		}

		dbBlock := blockchaindb.Block{
			Height:        block.Height,
			BlockHash:     block.Hash,
			NoTransaction: block.NoTxs,
			Timestamp:     block.Timestamp,
		}

		err = l.CqlRawClient.WriteBlock(dbBlock)
		if err != nil {
			result = result.Append(sinksdk.ResponseFailure(d.ID(), err.Error()))
			continue
		}

		result = result.Append(sinksdk.ResponseOK(d.ID()))
	}
	return result
}

// [... in another file]

// WriteBlock writes the block in the cassandra database.
func (rbc *RawBlockchainClient) WriteBlock(block Block) error {
	return rbc.cqlSession.Query(
		"INSERT INTO block (height, block_hash, no_transactions, "+
			"timestamp) VALUES (?, ?, ?, ?)",
		block.Height,
		block.BlockHash,
		block.NoTransaction,
		block.Timestamp,
	).Exec()
}
```
And here is the pipeline manifest that I use to reproduce the bug (it starts with the wrong keyspace name in the envars for tx and block writer, but not for witness):
```yaml
kind: Pipeline
metadata:
  name: ingest-error-test-3
  namespace: poseidon-dev
spec:
  vertices:
    - name: block-metadata
      source:
        udsource:
          container:
            image: registry.gitlab.com/mycompany/pipeline/bitcoin-block-metadata-source:v1.61.0
            env:
              - name: REDIS_HOST
                value: redis-block-source-master.poseidon-dev.svc.cluster.com
            resources: {}
      imagePullSecrets:
        - name: registry-credentials
      scale:
        min: 1
        max: 1
        
    - name: block-tx-fetcher
      udf:
        container:
          image: registry.gitlab.com/mycompany/pipeline/bitcoin-tx-fetch-udf:v1.61.0
          resources: {}
        builtin: null
        groupBy: null
      imagePullSecrets:
        - name: registry-credentials
      limits:
        readBatchSize: 1
        bufferMaxLength: 10000
      scale:
        min: 4
        max: 15

    - name: tx-coinjoin
      udf:
        container:
          image: registry.gitlab.com/mycompany/pipeline/bitcoin-coinjoin-udf:v1.61.0
          resources: {}
        builtin: null
        groupBy: null
      imagePullSecrets:
        - name: registry-credentials
      limits:
        bufferMaxLength: 100000
      scale:
        min: 5
        max: 5
      partitions: 5

    - name: block-writer
      sink:
        udsink:
          container:
            image: registry.gitlab.com/mycompany/pipeline/bitcoin-block-sink:v1.61.0
            env:
              - name: CASSANDRA_HOSTS
                value: scylla-svc.scylla-helios-dev
              - name: RAW_KEYSPACE
                value: invalid_keyspace_missing_block_table
            resources: {}
      imagePullSecrets:
        - name: registry-credentials
      limits:
        readBatchSize: 50
        bufferMaxLength: 100000
      scale:
        min: 1
        max: 1

    - name: block-writer-witness
      sink:
        udsink:
          container:
            image: registry.gitlab.com/mycompany/pipeline/bitcoin-block-sink:v1.61.0
            env:
              - name: CASSANDRA_HOSTS
                value: scylla-svc.scylla-helios-dev
              - name: RAW_KEYSPACE
                value: btc_raw_dev_b
      imagePullSecrets:
        - name: registry-credentials
      limits:
        readBatchSize: 50
        bufferMaxLength: 100000
      scale:
        min: 1
        max: 1

    - name: tx-writer
      sink:
        udsink:
          container:
            image: registry.gitlab.com/mycompany/pipeline/bitcoin-tx-sink:v1.61.0
            env:
              - name: CASSANDRA_HOSTS
                value: scylla-svc.scylla-helios-dev
              - name: RAW_KEYSPACE
                value: invalid_keyspace_missing_block_table
            resources: {}
      imagePullSecrets:
        - name: registry-credentials
      scale:
        min: 5
        max: 5
      partitions: 5

  edges:
    - from: block-metadata
      to: block-tx-fetcher
      conditions: null

    - from: block-tx-fetcher
      to: tx-coinjoin
      conditions:
        tags:
          operator: or
          values:
            - transaction

    - from: block-tx-fetcher
      to: block-writer
      conditions:
        tags:
          operator: or
          values:
            - block

    - from: block-tx-fetcher
      to: block-writer-witness
      conditions:
        tags:
          operator: or
          values:
            - block

    - from: tx-coinjoin
      to: tx-writer
      conditions: null
```

**To Reproduce**
I don't have a working example that doesn't involve my freelance client's code yet, but I believe we can make one in the upcoming days with some trials and errors.
I tried and failed to narrow down the error to this much simpler pipeline where I send sequence of number through HTTP and  use a sink that has an envar to make it fail just like my previous pipeline sinks misconfigurations, but didn't succeed:
```yaml
kind: Pipeline
metadata:
  name: super-odd
  namespace: poseidon-dev
spec:
  vertices:
    - name: in
      source:
        http: {}
      scale:
        min: 1
        max: 1
    - name: shroedinger-cat
      partitions: 4
      sink:
        udsink:
          container:
            image: docker.io/faidideq/numaflow-shroedinger-cat-sink:v0.0.1
            env:
              - name: CAT_STATE
                value: dead  # <- I run it, let it fail a bit, and change this to "alive", which cat the output
      scale:
        min: 1
        max: 1
  edges:
    - from: in
      to: shroedinger-cat
```
This has led me to believe that it has to do with how full the buffers are, but I didn't have time to confirm it yet.

**Expected behavior**
**block-writer**'s output table has no missing blocks, just like **block-writer-witness**.

- Kubernetes: v1.26.1+k3s1
- Numaflow: v1.1.4

---

QuentinFAIDIDE:
might fix #1551 #1554 
We would need to ensure that there are no adverse consequence in the handling of the new write error that would happen in the surge scenario.


pkg/isbsvc/jetstream_service.go
```diff
@@ -114,7 +114,7 @@ func (jss *jetStreamSvc) CreateBuffersAndBuckets(ctx context.Context, buffers, b
 				Name:       streamName,
 				Subjects:   []string{streamName}, // Use the stream name as the only subject
 				Retention:  nats.RetentionPolicy(v.GetInt("stream.retention")),
-				Discard:    nats.DiscardOld,
+				Discard:    nats.DiscardNew,
```
vigith, line 117: can we make it overridable by the user?

QuentinFAIDIDE: Maybe we can set it when users specify "drop on full" ? It's pretty much the exact same behaviour.
Though according to @yhl25 DiscardNew can't be used, but I'm still trying to understand why in the other issue.

yhl25: +1 to make it configurable to the user, but let's set it to `DiscardOld` by default since we use the `Limits` Policy by default. Also, please add a comment saying `DiscardNew` can only be used with `WorkQueue` Policy.

vigith: since we cannot use `DiscardNew` with `Limits` policy. we should not let the pipeline even start and validation should fail.

QuentinFAIDIDE: Maybe we should also document the risk of data loss on surge when using DiscardOld on high throughput and be really transparent with why it happens. Right now, users like me who fiddle with the config might be in big trouble if some UDF/Sink create a silent data loss scenario on production.

QuentinFAIDIDE:
image

Something I currently experience with the "surge pipeline situation" with `DiscardNew` that may or may not be what @yhl25 is referring to:
- After letting the sink fail for a moment, I am letting it work and it does has a descent Ack rate
-  the buffer before the sink stays at 30k and the cpu usage for the `message-variator-udf` is high while logs are repeating: 
```
2024/04/03 18:51:32 | ERROR | {...,"msg":"Retrying failed messages","pipeline":"super-odd-8","vertex":"msg-variator","protocol":"uds-grpc-map-udf","errors":{"nats: maximum messages exceeded":31}...}
```
The `{"nats: maximum messages exceeded":31}` seems to be the new jetstream side "buffer full" error thrown because we now use DiscardNew. The number of these errors tends to slowly decrease and then go up again, indicating it's writing some, and then some other datum arrive from the source .
- buffer the source is writing to experiences the same down and up but with a bufferFull! error, which is the normal error we usually get on buffer being fulls.

Overall, the number of messages stays nearly stable (it decrease at super slow rate on source buffer) due to the huge pile of retries that tends to immediately refill any missing data. So either this is what gave the impression of undelivered messages staying in the pipe, or either I still didn't reproduce it.
I'm going to let it sit for some time and confirm I emptied the retries with no losses. I'll keep you updated.

QuentinFAIDIDE:
I added the doc changes to remove the retention policy parameter and default to WorkQueue as discussed.
Was not able to reproduce the "stucked messages" issue yet. 
My guess is that one of the following is true:
1) The new issue is due to the new "buffer over capacity" error from jetstream which is the only behavior change since we activated the WorkQueue/DiscardNew setting. The issue would then be a subset of the fixed lost datum issue, because the full buffer management is supposed to prevent that error from ever being returned in the first place.
2) The new issue lies in Jetstream (feels unlikely but I may be wrong).
3) The new issue is due to some other Jetstream behaviour change with WorkQueue/DiscardNew that I am not aware of. Is there likely to be any other than the new error ?

What do you guys think we should do ? I've been trying to reproduce the issue a few times with no luck, going to retry but let me know your input.

vigith:
as per https://github.com/nats-io/nats-server/issues/5148#issuecomment-2007087670 the issue seems to have been resolved in 2.10.12

vigith:
> So what's the plan, do we change the new "compatible" jetstream configmap to specify only the new version with the fix, or do we wait for someone to try to reproduce this error enough times to convince us that it's fixed ?

We can make this change a configurable option, with defaults set to what is currently being used since that is battle-tested. Eventually, this could be the default, but before we do that, we need to make sure it works as expected with a decent amount of run in the production.

vigith:
https://github.com/nats-io/nats-server/pull/5270 seems to fix the problem. 2.10.14 release of jetstream seems very promising for WorkQueue.

whynowy:
Closed by https://github.com/numaproj/numaflow/pull/1884.

PR #1628, opened 2024-04-02
feat: read only view for UI

veds-g:
- Added support to toggle Read Only View for Numaflow UI
- Update boolean `server.readonly` in `numaflow-cmd-params-config` to toggle view
- Added new endpoint - `/api/v1/readonlyinfo` to be used in the UI

image
image
image

### DIFF
--- a/cmd/commands/server.go
+++ b/cmd/commands/server.go
@@ -35,6 +35,7 @@ func NewServerCommand() *cobra.Command {
 		namespaced       bool
 		managedNamespace string
 		baseHref         string
+		readOnly         bool
 		disableAuth      bool
 		serverAddr       string
 	)
@@ -55,6 +56,7 @@ func NewServerCommand() *cobra.Command {
 				Namespaced:       namespaced,
 				ManagedNamespace: managedNamespace,
 				BaseHref:         baseHref,
+				ReadOnly:         readOnly,
 				DisableAuth:      disableAuth,
 				DexServerAddr:    common.NumaflowDexServerAddr,
 				ServerAddr:       serverAddr,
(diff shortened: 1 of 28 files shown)


whynowy:
How to enable/disable it?

whynowy:
> > How to enable/disable it?
> 
> today, default behaviour - current UI(edit), package(read). We need to modify a boolean variable in `App.tsx` component as mentioned in description above to enable/disable. Later, I am thinking of setting it up by accepting a variable from backend via API call, and set it up.

Maybe do it with an approach similar to the `localPath`?

veds-g:
> Maybe do it with an approach similar to the `localPath`?

I tried doing it that way but I realised this approach won't work for the package. For `localPath` we create a script with config and directly inject it into the UI server. This works for standalone Numaflow but to solve both cases I'm thinking of writing a separate API. wdyt?

mshakira (approved):
looks good

cmd/commands/server.go
```diff
@@ -69,6 +71,7 @@ func NewServerCommand() *cobra.Command {
 	command.Flags().BoolVar(&namespaced, "namespaced", sharedutil.LookupEnvBoolOr("NUMAFLOW_SERVER_NAMESPACED", false), "Whether to run in namespaced scope, defaults to false.")
 	command.Flags().StringVar(&managedNamespace, "managed-namespace", sharedutil.LookupEnvStringOr("NUMAFLOW_SERVER_MANAGED_NAMESPACE", sharedutil.LookupEnvStringOr("NAMESPACE", "numaflow-system")), "The namespace that the server watches when \"--namespaced\" is \"true\".")
 	command.Flags().StringVar(&baseHref, "base-href", sharedutil.LookupEnvStringOr("NUMAFLOW_SERVER_BASE_HREF", "/"), "Base href for Numaflow server, defaults to '/'.")
+	command.Flags().BoolVar(&readOnly, "readonly", sharedutil.LookupEnvBoolOr("NUMAFLOW_SERVER_READONLY", true), "Whether to enable read only view for the UX server, defaults to false.")
```
whynowy, line 74: This means default is true.

server/cmd/server/start.go
```diff
@@ -152,6 +156,7 @@ func UrlRewrite(r *gin.Engine) gin.HandlerFunc {
 func CreateAuthRouteMap(baseHref string) authz.RouteMap {
 	return authz.RouteMap{
 		"GET:" + baseHref + "api/v1/sysinfo":                                                         authz.NewRouteInfo(authz.ObjectPipeline, false),
+		"GET:" + baseHref + "api/v1/readonlyinfo":                                                    authz.NewRouteInfo(authz.ObjectEvents, false),
```
whynowy, line 159: Is it possible to put it into sysinfo and use that API?

cmd/commands/server.go
veds-g: typo, should be false

PR #1639, opened 2024-04-04
fix: Dedup not working for multi-partitioned edge

### LINKED ISSUE #1634: Track read metrics at partition level for user defined source
This will help us understand lagging and stuck readers and also the data distribution.

### LINKED ISSUE #1635: Dedup not working as expected when using multi-partitioned edges

yhl25:
fixes #1635 
fixes #1634

### DIFF
--- a/pkg/forwarder/interfaces.go
+++ b/pkg/forwarder/interfaces.go
@@ -29,9 +29,13 @@ type ToWhichStepDecider interface {
 	// where former means, forward to all the neighbouring steps and latter means do not forward anywhere.
 	//
 	// Parameters:
-	// - keys: Used by shuffle to decide which partition to write, if the toVertex is 'reduce' and has multiple partitions.
+	// - keys: Used by shuffle to decide which partition to write, if the toVertex is 'reduce' and has
+	// multiple partitions. It is deterministic messages with same set of keys will always go to the same partition.
+	//
 	// - tags: Used for conditional forwarding.
-	// - id: Used by shuffle to decide which partition to write, if the toVertex is a 'map' and has multiple partitions.
+	//
+	// - id: Used by shuffle to decide which partition to write, if the toVertex is a 'map' and has
+	// multiple partitions. It is deterministic messages with same id will always go to the same partition.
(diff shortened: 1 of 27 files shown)


pkg/sources/source.go
```diff
@@ -175,7 +175,7 @@ func (sp *SourceProcessor) Start(ctx context.Context) error {
 	var toVertexPartitionMap = make(map[string]int)
 	shuffleFuncMap := make(map[string]*shuffle.Shuffle)
 	for _, edge := range sp.VertexInstance.Vertex.Spec.ToEdges {
-		if edge.ToVertexType == dfv1.VertexTypeReduceUDF && edge.GetToVertexPartitionCount() > 1 {
+		if edge.GetToVertexPartitionCount() > 1 {
```
whynowy, line 178: Was this something we forgot to change when working on multi-partition support?

vigith: we are adding message-id based shuffling for every vertex. this is to have deterministic writing when we have multi-partitioned edges and dedupe will work correctly because dedupe is per stream.

vigith (approved):

pkg/forwarder/interfaces.go
```diff
@@ -25,17 +25,18 @@ type VertexBuffer struct {
 // ToWhichStepDecider decides which step to forward after applying the WhereTo function.
 type ToWhichStepDecider interface {
 	// WhereTo decides where to forward the result to based on the name of the step it returns.
+	// It is deterministic and will forward the same message-id to the partition using a hashing algorithm.
 	// It supports 2 addition keywords which need not be a step name. They are "ALL" and "DROP"
 	// where former means, forward to all the neighbouring steps and latter means do not forward anywhere.
-	WhereTo([]string, []string) ([]VertexBuffer, error)
+	WhereTo([]string, []string, string) ([]VertexBuffer, error)
```
whynowy, line 31: Can we rewrite the docs above, and explain the args accepted? I feel like the docs have drifted too much, and no one understands what they tell.

PR #1640, opened 2024-04-04
fix: readonly view

veds-g:
- Merged `readonly` flag with `sysInfo` API
- Removed the `/readonlyinfo` endpoint
- Read only defaults to `false`, edit shared config variable `server.readonly` to enable it


vigith (approved):

whynowy:
@veds-g - Thanks!

1. Comment this [line](https://github.com/numaproj/numaflow/blob/v1.2.0-rc2/config/base/shared-config/numaflow-cmd-params-config.yaml#L25) since by default it's false.
2. When it's `readonly`, all the Create/Update APIs should return `403 forbidden`.

vigith:
> 2. When it's `readonly`, all the Create/Update/Delete APIs should return `403 forbidden`.

very valid point :)

whynowy (approved):
LGTM

PR #1643, opened 2024-04-05
chore: update examples

yhl25:


vigith (approved):

KeranYang (dismissed):
why not use the `stable` tag? that way we don't need to update for each release.

yhl25:
> why not use the `stable` tag? that way we don't need to update for each release.

My understanding is that stable images from the Numaflow SDK are compatible with Numaflow main, which is why they are used in e2e tests. However, the examples should be compatible with the latest Numaflow release.

KeranYang:
> However, the examples should be compatible with the latest Numaflow release

@yhl25 , I see the examples are used in our quick-start.md file where we use `kubectl apply -n numaflow-system -f https://raw.githubusercontent.com/numaproj/numaflow/stable/config/install.yaml` to create numaflow. Does it mean that the examples should be compatible with the numaflow version in the install.yaml file?

In the [install.yaml](https://github.com/numaproj/numaflow/blob/main/config/install.yaml#L17365) file, the numaflow image points to the `latest` tag. At quay.io, it seems the latest tag is even newer than Nighty. Does it mean the latest tag points to the main branch?

Not sure why the raw install.yaml that we provide in quick-start uses `quay.io/numaproj/numaflow:v1.1.6` instead of `latest`, which is not in sync with our install.yaml. More importantly, it means our quick-start is not working right now. 

@vigith WDYT?

vigith:
This is a big concern; our [quick-start.md](https://github.com/numaproj/numaflow/blob/main/docs/quick-start.md) should always work. @yhl25 let's fix this ASAP. Our examples should we always be working with the latest stable release. Our tests can be based on the main.

Issue #1504, opened 2024-02-13
Surface SDK in-compatibility issue to users through logs


KeranYang:
# Summary

SDK in-compatibility is one of the most common issues that users encounter when they are working on their test pipelines. A very common answer that we provide on the numaflow-support channel is “upgrade your numaflow version to v1.x.x” or “use a newer sdk version v0.x.x ”. It would be great if we can surface such information through logs. “Numaflow platform is running on version a.b.c while the UDF is running on numaflow-go version x.y.z, they are not compatible, please update numaflow-go to version l.m.n”.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

kohlisid:
Just to clarify, In this issue we are trying to identify the Numaflow and SDK versions and then match whether they are compatible right?
And then provide the end user this information

KeranYang:
@kohlisid Yes, hopefully by having this, users can resolve it all by themselves without numa support involved.

ayildirim21:
meeting notes:

```
# Go SDK
# server info file contains:
version: x.x.x
minimum numaflow version: >= x2.x2.x2 # if this key-value doesnt exists then old SDK version, please update your SDK log

------------------------

# numaflow
version: z.z.z
# maintain this information for all languages in a matrix, yaml file (root dir?), parse yaml into a dict in Go code
minimum Go SDK version: y.y.y
if numaflow version 'latest' skip numaflow check as you already have latest changes

# check before grpc connection is established
if x.x.x < y.y.y  || if z.z.z < x2.x2.x2 # compatibility issue 
if x.x.x >= y.y.y && z.z.z >= x2.x2.x2 # compatible
```

PR #1644, opened 2024-04-05
chore: surface SDK compatibility issues through logs

ayildirim21:
When we upgrade to numaflow 1.2.0, sdk versions must be >= 0.7.0 in order to be compatible. This adds logic to surface this information to users, example output:

image

related to: [#1504](https://github.com/numaproj/numaflow/issues/1504)


version.go
```diff
@@ -21,6 +21,8 @@ import (
 	"runtime"
 )
 
+const ReleaseVersion = "1.2.0"
```
vigith, line 24: we might forget this and can cause problems.

pkg/shared/util/grpc_utils.go
```diff
@@ -28,10 +28,47 @@ import (
...
 )
 
+// isCompatible checks if the current numaflow version is compatible with the given language's SDK version
+func isCompatible(serverInfo *info.ServerInfo) error {
+	type sdkVersions map[info.Language]string
+	var versionMappingConfig = sdkVersions{
+		info.Go:     ">= 0.7.0-0",
+		info.Java:   ">= 0.7.0-0",
+		info.Python: ">= 0.7.0-0",
+	}
```
vigith, line 46: should we have to maintain a matrix?

1.0.0 supports SDK X1, Y1, Z1 versions.
1.1.0 supports SDK X2, Y2, Z2 versions.

kohlisid: A matrix would be more cleaner and extensible for sure. 
Screenshot 2024-04-05 at 11 10 41 AM

ayildirim21: is there a way that we can dynamically fetch the current version the client is running? Because with the way i have it right now in `version.go`, it would only make sense to embed the supported sdk version for the given numaflow version, i.e all the other rows in the matrix would never be used.

kohlisid: Is this the minimum supported version that we are considering or maximum?

vigith: minimum support version will be the lower bound. upper bound will be what is out there?

ayildirim21: yes

ayildirim21:
couple of notes: we dont need to check if the minimumNumaflowVersion field exists because we are using the latest numaflow-go dependency in the go.mod file, so it will always be present. Still not entirely sure how best to handle the check for local testing when numaflowVersion = "latest" currently we just skip the numaflow side check if that is the case, but what if an older version of the sdk is running, locally we will still have "latest"

ayildirim21:
tests are failing because redis-sink is using old numaflow-go version: PR to upgrade it: https://github.com/numaproj/numaflow-sinks/pull/37

pkg/shared/util/grpc_utils.go
```diff
@@ -20,18 +20,77 @@ import (
...
+	"github.com/numaproj/numaflow"
 	sdkerr "github.com/numaproj/numaflow/pkg/sdkclient/error"
 	resolver "github.com/numaproj/numaflow/pkg/sdkclient/grpc_resolver"
 )
 
+func checkConstraint(version *semver.Version, constraint string) error {
+	c, err := semver.NewConstraint(constraint)
+	if err != nil {
+		return fmt.Errorf("error parsing constraint: %w", err)
+	}
```
KeranYang, line 42:
```suggestion
	if c, err := semver.NewConstraint(constraint); err != nil || c == nil {
		return fmt.Errorf("error parsing constraint: %w", err)
	}
```

Same for below, this is a good syntax to simplify code.

pkg/shared/util/grpc_utils.go
```diff
@@ -20,18 +20,77 @@ import (
...
+
+	isValid, _ := c.Validate(version)
+	if !isValid {
+		return fmt.Errorf("version did not meet constraint requirement")
```
KeranYang, line 46: let's make the error message more useful by providing the content of `version` and `constraint`

pkg/shared/util/grpc_utils.go
```diff
@@ -20,18 +20,77 @@ import (
...
+}
+
+// isCompatible checks if the current numaflow version is compatible with the given language's SDK version
+func isCompatible(serverInfo *info.ServerInfo) error {
+	type sdkVersions map[info.Language]string
+	versionMappingConfig := sdkVersions{
+		info.Go:     "0.7.0-0",
+		info.Java:   "0.7.0-0",
+		info.Python: "0.7.0-0",
+	}
```
KeranYang, line 59: Can we instead of hard-coding, read it from the version-mapping file? We should always have a single source of truth for this mapping.

pkg/shared/util/grpc_utils.go
```diff
@@ -20,18 +20,77 @@ import (
...
+
+	numaflowVersionStr := numaflow.GetVersion().Version
```
KeranYang, line 61: just curious, how does this work? if I am releasing numaflow to v1.2.0, how do I ensure this call returns me `v1.2.0`?

pkg/shared/util/grpc_utils.go
```diff
@@ -20,18 +20,77 @@ import (
...
+	if !strings.Contains(numaflowVersionStr, "latest") {
+		numaflowVersion, err := semver.NewVersion(numaflowVersionStr)
+		if err != nil {
+			return fmt.Errorf("error parsing numaflow version: %w", err)
+		}
+
+		// Check if the numaflow version satisfies the minimum required numaflow version by SDK
+		numaflowConstraint := fmt.Sprintf(">= %s", serverInfo.MinimumNumaflowVersion)
+		if err := checkConstraint(numaflowVersion, numaflowConstraint); err != nil {
+			return fmt.Errorf("numaflow version %s must be upgraded to %s, in order to work with current SDK version", numaflowVersion, numaflowConstraint)
```
KeranYang, line 72: please also add current SDK version number to the error message.

KeranYang, line 72:
```suggestion
			return fmt.Errorf("numaflow version %s must be upgraded to %s or later, in order to work with current SDK version", numaflowVersion, numaflowConstraint)
```

pkg/shared/util/grpc_utils.go
```diff
@@ -20,18 +20,77 @@ import (
...
+	if err != nil {
+		return fmt.Errorf("error parsing SDK version: %w", err)
+	}
+
+	sdkConstraint, ok := versionMappingConfig[sdkLanguage]
+	if ok {
+		// Check if the SDK version satisfies the minimum required SDK version by the numaflow platform
+		sdkConstraint = fmt.Sprintf(">= %s", sdkConstraint)
+		if err := checkConstraint(sdkVersion, sdkConstraint); err != nil {
+			return fmt.Errorf("SDK version %s must be upgraded to %s, in order to work with current numaflow version", sdkVersion, sdkConstraint)
```
KeranYang, line 87:
```suggestion
			return fmt.Errorf("SDK version %s must be upgraded to %s or later, in order to work with current numaflow version", sdkVersion, sdkConstraint)
```

pkg/shared/util/grpc_utils_test.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+func TestIsCompatible(t *testing.T) {
+	tests := []struct {
+		name       string
+		serverInfo *info.ServerInfo
+		shouldErr  bool
+	}{
+		{
+			name: "Test with incompatible version",
```
KeranYang, line 17: can we test both cases? - sdk not compatible and numaflow

pkg/shared/util/grpc_utils_test.go
```diff
@@ -0,0 +1,48 @@
...
+			serverInfo: &info.ServerInfo{
+				Protocol: info.UDS,
+				Language: info.Python,
+				Version:  "0.6.0",
+				Metadata: nil,
+			},
+			shouldErr: true,
+		},
+		{
+			name: "Test with compatible version",
```
KeranYang, line 27: will this test fail if we update numaflow version or update version-mapping?

KeranYang (commented):
partial review...

ayildirim21:
PR to bump numaflow-go versions in side-input examples: https://github.com/numaproj-contrib/e2e-tests-go/pull/11

ayildirim21:
The failing unit tests are not because of the changes in this PR. When @yhl25 merges his changes in https://github.com/numaproj/numaflow/pull/1669/files they will be fixed (specifically the changes in the files `event_time_extractor_test.go`, `filter_test.go`, and `time_extraction_filter_test.go`)

pkg/shared/util/grpc_utils.go
ayildirim21: i need the c variable to be out of the scope of the if statement to use it with isValid i think

KeranYang: I think in this case u can access it in an else statement.

pkg/shared/util/grpc_utils.go
ayildirim21: we provide their contents after the function call

pkg/shared/util/grpc_utils.go
```diff
@@ -20,18 +20,77 @@ import (
...
+	isValid, _ := c.Validate(version)
+	if !isValid {
+		return fmt.Errorf("version did not meet constraint requirement")
+	}
+
+	return nil
+}
+
+// isCompatible checks if the current numaflow version is compatible with the given language's SDK version
+func isCompatible(serverInfo *info.ServerInfo) error {
```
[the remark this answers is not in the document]
ayildirim21: instead should I just change the name to checkCompatability instead, since i think returning the bool is redundant when im already using an error to indicate whether the versions are compatible? wdyt?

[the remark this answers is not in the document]
KeranYang: Makes sense

pkg/shared/util/grpc_utils_test.go
ayildirim21: yes so we would need to update this test each time, or we can just remove this test

KeranYang: We don't want either. To make the code more unit test friendly, we can consider change the checkCompatibility method signature to something like `checkCompatibility(serverInfo, versionMapping, numaVersion)`, such that we can mock the input in unit tests.

pkg/shared/util/grpc_utils_test.go
```diff
@@ -0,0 +1,107 @@
...
+	},
+}
+
+func TestCheckCompatibility(t *testing.T) {
+	tests := []struct {
+		name                 string
+		serverInfo           *info.ServerInfo
+		versionMappingConfig map[string]sdkConstraints
+		numaflowVersion      string
+		shouldErr            bool
```
KeranYang, line 29: please also check the content of the error message

PR #1647, opened 2024-04-06
fix: flaky TestDropOnFull 

yhl25:
We hash the message ID and distribute the messages to different partitions based on that hash value. Now messages can end up in any partition, not just the first one

### DIFF
--- a/test/e2e/functional_test.go
+++ b/test/e2e/functional_test.go
@@ -1,3 +1,5 @@
+//go:build test
+
 /*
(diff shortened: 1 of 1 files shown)


test/e2e/functional_test.go
```diff
@@ -1,5 +1,3 @@
-//go:build test
```
vigith, line 1: don't we need this?

yhl25: removed by mistake

vigith (approved):

PR #1656, opened 2024-04-09
chore: info level logging for dropped messages

yhl25:


vigith (commented):
any particular reason for moving away from Warn? Isn't this a real warning? Since our Info logging is a lot, I am thinking there could be users running numaflow at only warn level.

vigith (approved):

whynowy:
> any particular reason for moving away from Warn? Isn't this a real warning? Since our Info logging is a lot, I am thinking there could be users running numaflow at only warn level.

I assume this question has been addressed?

PR #1662, opened 2024-04-13
chore: Refactor Sinks to Avoid Implementing Forwarder Interface

yhl25:
Now that we have a separate forwarder for the sink, our sinks don't need to implement the forwarder interface.


whynowy (approved):
LGTM

PR #1664, opened 2024-04-13
feat: controller change for fallback sink

whynowy:
Explain what this PR does.


pkg/apis/numaflow/v1alpha1/const.go
```diff
@@ -67,12 +67,21 @@ const (
...
 	CtrUdSideInput       = "udsi"
 	CtrInitSideInputs    = "init-side-inputs"
 	CtrSideInputsWatcher = "side-inputs-synchronizer"
 
+	// user-defined container types
+	UDContainerFunction     = "udf"
+	UDContainerSink         = "udsink"
+	UDContainerFackbackSink = "fallback-udsink"
+	UDContainerTransformer  = "transformer"
+	UDContainerSource       = "udsource"
```
yhl25, line 82: fb-udsink? similar to line 70

whynowy: feel free to update

vigith (approved):

PR #1669, opened 2024-04-15
feat: Fallback Sink 

### LINKED ISSUE #1668: add fallback-sink to sinker

yhl25:
fixes #1668

### DIFF
--- a/pkg/sinks/forward/forward.go
+++ b/pkg/sinks/forward/forward.go
@@ -31,10 +31,10 @@ import (
 	dfv1 "github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
 	"github.com/numaproj/numaflow/pkg/isb"
 	"github.com/numaproj/numaflow/pkg/metrics"
-	"github.com/numaproj/numaflow/pkg/shared/idlehandler"
 	"github.com/numaproj/numaflow/pkg/shared/logging"
+	"github.com/numaproj/numaflow/pkg/sinks/sinker"
+	"github.com/numaproj/numaflow/pkg/sinks/udsink"
 	"github.com/numaproj/numaflow/pkg/watermark/fetch"
-	"github.com/numaproj/numaflow/pkg/watermark/publish"
 	"github.com/numaproj/numaflow/pkg/watermark/wmb"
 )
 
@@ -45,33 +45,23 @@ type DataForward struct {
 	// we need something more than a cancel().
 	cancelFn            context.CancelFunc
 	fromBufferPartition isb.BufferReader
-	toBuffer            isb.BufferWriter
+	sinkWriter          sinker.SinkWriter
 	wmFetcher           fetch.Fetcher
-	wmPublisher         publish.Publisher
 	opts                options
 	vertexName          string
 	pipelineName        string
 	vertexReplica       int32
-	// idleManager manages the idle watermark status.
-	idleManager wmb.IdleManager
 	// wmbChecker checks if the idle watermark is valid.
 	wmbChecker wmb.WMBChecker
 	Shutdown
 }
 
 // NewDataForward creates a sink data forwarder.
-func NewDataForward(
-	vertexInstance *dfv1.VertexInstance,
-	fromStep isb.BufferReader,
-	toStep isb.BufferWriter,
-	fetchWatermark fetch.Fetcher,
-	publishWatermark publish.Publisher,
-	idleManager wmb.IdleManager,
-	opts ...Option) (*DataForward, error) {
+func NewDataForward(vertexInstance *dfv1.VertexInstance, fromStep isb.BufferReader, sinkWriter sinker.SinkWriter, fetchWatermark fetch.Fetcher, opts ...Option) (*DataForward, error) {
 
-	options := DefaultOptions()
+	dOpts := DefaultOptions()
 	for _, o := range opts {
-		if err := o(options); err != nil {
+		if err := o(dOpts); err != nil {
 			return nil, err
 		}
 	}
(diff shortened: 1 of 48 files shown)


pkg/sinks/forward/forward.go
```diff
@@ -222,60 +219,34 @@ func (df *DataForward) forwardAChunk(ctx context.Context) {
...
-			messageToStep = append(messageToStep, message.Message)
-		}
+		writeMessages = append(writeMessages, m.Message)
 	}
 
-	// forward the message to the edge buffer (could be multiple edges)
-	writeOffsets, err := df.writeToBuffer(ctx, df.toBuffer, messageToStep)
+	// write the messages to the sink
+	writeOffsets, fallbackMessages, err := df.writeToSink(ctx, df.sinkWriter, writeMessages)
 	if err != nil {
```
vigith, line 230: comment that this error will only happen if there is `ctx.Done()`

pkg/sinks/udsink/udsink_grpc.go
```diff
@@ -28,6 +28,8 @@ import (
 	"github.com/numaproj/numaflow/pkg/shared/logging"
 )
 
+var WriteToFallbackErr = "write to fallback sink"
```
vigith, line 31: inline it?

vigith (approved):

pkg/sinks/forward/forward.go
```diff
@@ -222,60 +219,34 @@ func (df *DataForward) forwardAChunk(ctx context.Context) {
...
-		df.opts.logger.Errorw("failed to write to toBuffer", zap.Error(err))
+		df.opts.logger.Errorw("failed to write to sink", zap.Error(err))
 		df.fromBufferPartition.NoAck(ctx, readOffsets)
 		return
 	}
-	df.opts.logger.Debugw("writeToBuffers completed")
 
+	for len(fallbackMessages) > 0 {
+		if df.opts.fbSinkWriter == nil {
+			df.opts.logger.Errorw("Fallback sink is not specified, dropping messages", zap.Int("count", len(fallbackMessages)))
```
whynowy, line 238: Infinite loop?

pkg/sinks/sink.go
```diff
@@ -187,7 +219,22 @@ func (u *SinkProcessor) Start(ctx context.Context) error {
...
 
-		// NOTE: forwarder can make multiple sink writers when we introduce fallback sinks
+		// create the main sink writer
+		sinkWriter, err := u.createSinkWriter(ctx, &u.VertexInstance.Vertex.Spec.Sink.AbstractSink, sinkHandler)
+		if err != nil {
+			return fmt.Errorf("failed to find a sink, error: %w", err)
+		}
+
+		// create the fallback sink writer if fallback sink is present
+		if u.VertexInstance.Vertex.Spec.Sink.Fallback != nil {
```
whynowy, line 229: This requires a sink must be defined in the field `fallback`. We could argue this is a requirement, but usually this should not be enforced - if nothing is defined under `fallback`, we should consider there's no fallback specified.

pkg/sinks/forward/forward.go
vigith: yes, we have to make sure we write to fallback all the messages

pkg/sinks/forward/forward.go
```diff
@@ -222,60 +219,32 @@ func (df *DataForward) forwardAChunk(ctx context.Context) {
...
+		df.opts.logger.Errorw("failed to write to sink", zap.Error(err))
 		df.fromBufferPartition.NoAck(ctx, readOffsets)
 		return
 	}
-	df.opts.logger.Debugw("writeToBuffers completed")
 
+	// if we have any messages to write to the fallback sink, write them.
+	if len(fallbackMessages) > 0 {
+		df.opts.logger.Infow("Writing messages to fallback sink", zap.Int("count", len(fallbackMessages)))
+		_, fallbackMessages, err = df.writeToSink(ctx, df.opts.fbSinkWriter, fallbackMessages, true)
```
vigith, line 240:
```suggestion
		// write to sink is an infinite loop; it will return only if writes are successful or
		// ctx.Done happens due to shutdown.
		_, _, err = df.writeToSink(ctx, df.opts.fbSinkWriter, fallbackMessages, true)
```

pkg/sinks/sink.go
```diff
@@ -187,7 +219,23 @@ func (u *SinkProcessor) Start(ctx context.Context) error {
...
+		// create the main sink writer
+		sinkWriter, err := u.createSinkWriter(ctx, &u.VertexInstance.Vertex.Spec.Sink.AbstractSink, sinkHandler)
+		if err != nil {
+			return fmt.Errorf("failed to find a sink, error: %w", err)
+		}
+
+		// create the fallback sink writer if fallback sink is present
+		fbSink := u.VertexInstance.Vertex.Spec.Sink.Fallback
+		if fbSink != nil && fbSink.IsAnySinkSpecified() {
+			fbSinkWriter, err := u.createSinkWriter(ctx, u.VertexInstance.Vertex.Spec.Sink.Fallback, fbSinkHandler)
```
whynowy, line 231: Who handles the close of `gbSinkWriter`?

whynowy: Instead of using an option to pass in the fallback writer, is it better to explicitly add it as a property to the forwarder?

yhl25: Good find... The forwarder should close it, I missed that.

yhl25: fb sink writer is not a mandatory field for the forwarder to start, so kept it optional. Since it is optional for users as well

pkg/sinks/forward/forward.go
```diff
@@ -222,60 +228,34 @@ func (df *DataForward) forwardAChunk(ctx context.Context) {
...
+	// error will not be nil only when we get ctx.Done()
 	if err != nil {
-		df.opts.logger.Errorw("failed to write to toBuffer", zap.Error(err))
+		df.opts.logger.Errorw("failed to write to sink", zap.Error(err))
 		df.fromBufferPartition.NoAck(ctx, readOffsets)
 		return
 	}
-	df.opts.logger.Debugw("writeToBuffers completed")
 
+	// if we have any messages to write to the fallback sink, write them.
```
whynowy, line 246:
```suggestion
	// Only when fallback is configured, it is possible to return fallbackMessages. If there's any, write to the fallback sink.
```

whynowy (approved):

PR #1678, opened 2024-04-16
feat: add batch reading for Jetstream

### LINKED ISSUE #1675: Batching with Timeout not working as expected
**Describe the bug**
When given a timeout as T, and batch size at B we want numaflow to return either B messages or read up until T whichever occurs first

kohlisid:
Fixes #1675

### DIFF
--- a/pkg/isb/stores/jetstream/reader.go
+++ b/pkg/isb/stores/jetstream/reader.go
@@ -119,12 +119,27 @@ func (jr *jetStreamReader) Read(_ context.Context, count int64) ([]*isb.ReadMess
 	}(time.Now())
 	var err error
 	var result []*isb.ReadMessage
-	msgs, err := jr.sub.Fetch(int(count), nats.MaxWait(jr.opts.readTimeOut))
+	// We use FetchBatch to have batching in our code with two parameters, ie ReadBatchSize
+	// and ReadTimeout. This function should read from Jetstream until we have a read ReadBatchSize
+	// number of messages or the ReadTimeout duration has elapsed.
+	msgs, err := jr.sub.FetchBatch(int(count), nats.MaxWait(jr.opts.readTimeOut))
 	if err != nil && !errors.Is(err, nats.ErrTimeout) {
 		isbReadErrors.With(map[string]string{"buffer": jr.GetName()}).Inc()
 		return nil, fmt.Errorf("failed to fetch messages from jet stream subject %q, %w", jr.subject, err)
 	}
-	for _, msg := range msgs {
+
+	// Wait until the batch processing is done
+	<-msgs.Done()
+
+	// Check if there was any error while fetching the messages
+	// No error is returned from the FetchBatch method in case of pull request expiry (even if there are no messages).
+	if err = msgs.Error(); err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/isb/stores/jetstream/reader.go
```diff
@@ -119,12 +119,27 @@ func (jr *jetStreamReader) Read(_ context.Context, count int64) ([]*isb.ReadMess
...
-	for _, msg := range msgs {
+
+	// Wait until the batch processing is done
+	<-msgs.Done()
+
+	// Check if there was any error while fetching the messages
+	// No error is returned from the FetchBatch method in case of pull request expiry (even if there are no messages).
+	if err = msgs.Error(); err != nil {
+		return nil, err
+	}
```
vigith, line 138: we cannot do this; we will have to return partial read messages, or else it is a data loss scenario. Perhaps you can return err at the end.

pkg/isb/stores/jetstream/reader.go
```diff
@@ -119,12 +119,27 @@ func (jr *jetStreamReader) Read(_ context.Context, count int64) ([]*isb.ReadMess
...
+	// number of messages or the ReadTimeout duration has elapsed.
+	msgs, err := jr.sub.FetchBatch(int(count), nats.MaxWait(jr.opts.readTimeOut))
 	if err != nil && !errors.Is(err, nats.ErrTimeout) {
 		isbReadErrors.With(map[string]string{"buffer": jr.GetName()}).Inc()
 		return nil, fmt.Errorf("failed to fetch messages from jet stream subject %q, %w", jr.subject, err)
 	}
-	for _, msg := range msgs {
+
+	// Wait until the batch processing is done
+	<-msgs.Done()
```
whynowy, line 132: Be very cautious on this async fetch call, I don't trust the reliability based on our past experience.

vigith: @kohlisid, can you check whether it uses `Fetch` behind the scenes?

pkg/isb/stores/jetstream/reader.go
kohlisid: I can return the partial messages read along with the error.

Though one question, in case we return no messages, they won't be acked by us as well right? And ideally the whole batch should be retried in the next read?

pkg/isb/stores/jetstream/reader.go
kohlisid: It does not use Fetch call directly. Both of them have different ways of collating and reporting the data. 
@whynowy Any specific areas that we can verify to ensure the scenarios you are mentioning?

kohlisid: Additionally, if you feel there is high risk. We can use this method only when Batching is requested by the user, and the default could be `Fetch`

vigith:
what's the plan for this PR?

kohlisid:
@vigith I was thinking to get this done along with the new Jetstream API changes. With the changes in functionality from their side regarding the same.

vigith:
> @vigith I was thinking to get this done along with the new Jetstream API changes. With the changes in functionality from their side regarding the same.

close this PR then?

PR #1681, opened 2024-04-17
chore: Fallback Sink E2E test

yhl25:


test/e2e/testdata/simple-fallback.yaml
```diff
@@ -0,0 +1,32 @@
...
+      udf:
+        builtin:
+          name: cat
+    - name: output
+      scale:
+        min: 1
+      sink:
+        udsink:
+          container:
+            image: quay.io/numaio/numaflow-go/fb-sink-log:stable
```
KeranYang, line 22: nit: could you please add the source code to numaflow-go? seems you are manually building it by uncomment the simple sink. Might be better if we have a separate example so that we utilize all the benefits of CI improvements.

yhl25: https://github.com/numaproj/numaflow-go/pull/122

vigith (approved):

PR #1683, opened 2024-04-17
chore: add version compatibility check

ayildirim21:
Explain what this PR does.


pkg/shared/util/version_mapping.go
```diff
@@ -0,0 +1,41 @@
...
+		info.Go:     "0.7.0-0",
+		info.Python: "0.7.0a",
+		info.Java:   "0.7.1-0",
+	},
+	"1.2.0-rc3": {
+		info.Go:     "0.7.0-0",
+		info.Python: "0.7.0a",
+		info.Java:   "0.7.1-0",
+	},
+}
```
vigith, line 41: should we just maintain a YAML file for this and embed that into the build?

ayildirim21: if we maintain it in a yaml file then we still need to read it into a map so it seems redundant. Also we would need to make the file somehow available to the pods because we cant just do os.ReadFile. Wdyt Vigith?

pkg/shared/util/grpc_utils.go
```diff
@@ -20,18 +20,98 @@ import (
...
+	}
+
+	return nil
+}
+
+// checkCompatibility checks if the current numaflow version is compatible with the given language's SDK version
+func checkCompatibility(serverInfo *info.ServerInfo, versionMappingConfig map[string]sdkConstraints, numaflowVersion string) error {
+	// If we are in CI or testing locally, we do not need to check for compatibility issues
+	if !strings.Contains(numaflowVersion, "latest") && os.Getenv("GITHUB_ACTIONS") == "true" {
+		print("keran is testing, the version is " + numaflowVersion)
```
whynowy, line 59: Clean up?

pkg/shared/util/version_mapping.go
vigith: use go:embed ?

i am okay with code too

pkg/shared/util/grpc_utils.go
```diff
@@ -20,18 +20,98 @@ import (
...
+		// Check if server info contains the MinimumNumaflowVersion field
+		sdkVersion := serverInfo.Version
+		rv := reflect.ValueOf(serverInfo)
```
whynowy, line 62: What is the reason we can not use `serverInfo.minimumNumaflowVersion` to compare?

pkg/shared/util/version_mapping.go
```diff
@@ -0,0 +1,41 @@
...
+
+package util
+
+import (
+	"github.com/numaproj/numaflow-go/pkg/info"
+)
+
+type sdkConstraints map[info.Language]string
+
+var versionMappingConfig = map[string]sdkConstraints{
```
whynowy, line 25: why do we need to maintain history mappings?

KeranYang: The only benefit is that we can link this to the user-guide so that users can view history mappings.

pkg/shared/util/grpc_utils.go
KeranYang: If the user is using an old version of SDK, the ServerInfo struct might not have the field just yet.

whynowy: In that case, won't the `minimumNumaflowVersion` be empty string after parsing?

pkg/shared/util/version_mapping.go
ayildirim21: we are ok with just maintaining the mapping for the current version as well if that is perferred

pkg/shared/util/grpc_utils.go
KeranYang: @ayildirim21

ayildirim21: if you try accessing a field that doesnt exist in a struct wouldn't we get an error?

pkg/shared/util/version_mapping.go
whynowy: If that's the case, we do not need to maintain the history here - it might be too verbose to have everything including `-rc`. For documentation purpose, we should have well maintained doc for version compatibilities.

ayildirim21: ok we can add the historical mappings to release.md or create a separate yaml file in the root directory

pkg/shared/util/grpc_utils.go
whynowy: You should be using the struct which has the field right? How come you don't see the struct field?

pkg/shared/util/version_mapping.go
whynowy: > ok we can add the historical mappings to release.md or create a separate yaml file in the root directory

This could be a separated task.

pkg/shared/util/grpc_utils.go
KeranYang: @whynowy you are right. the struct is imported from go.mod. @ayildirim21 we should always have access to the struct. could you update?

ayildirim21: ok yeah that makes sense since numaflow is using latest version of numaflow-go it will always have that field, we can simply the check to just seeing if it is empty string in that case

pkg/shared/util/grpc_utils.go
```diff
@@ -78,6 +151,13 @@ func WaitForServerInfo(timeout time.Duration, filePath string) (*info.ServerInfo
 		return nil, fmt.Errorf("failed to read server info: %w", err)
 	}
 
+	// If we are testing locally or in CI, we skip checking for compatibility issues
+	if strings.Contains(numaflow.GetVersion().Version, "latest") || os.Getenv("GITHUB_ACTIONS") == "true" {
+		return serverInfo, nil
```
whynowy, line 156: In this case, checking if the sdk version meets the min requirement still applies.

pkg/shared/util/grpc_utils.go
```diff
@@ -20,18 +20,89 @@ import (
...
+	return nil
+}
+
+// checkNumaflowCompatibility checks if the current numaflow version is compatible with the given language's SDK version
+func checkNumaflowCompatibility(numaflowVersion string, minNumaflowVersion string) error {
+	// Check if server info contains MinimumNumaflowVersion
+	if minNumaflowVersion == "" {
+		return fmt.Errorf("server info does not contain minimum numaflow version. Upgrade SDK to latest version")
+	}
+
```
KeranYang, line 55: We can remove this piece of code. It should never be "", and even if it is, it will be caught by line 62.

pkg/shared/util/grpc_utils.go
```diff
@@ -78,6 +149,22 @@ func WaitForServerInfo(timeout time.Duration, filePath string) (*info.ServerInfo
 		return nil, fmt.Errorf("failed to read server info: %w", err)
 	}
 
+	sdkVersion := serverInfo.Version
+	numaflowVersion := numaflow.GetVersion().Version
+
+	// If we are testing locally or in CI, we can skip checking for numaflow compatibility issues
```
KeranYang, line 155:
```suggestion
	// If we are testing locally or in CI, we can skip checking for numaflow compatibility issues because both return us a version string that the version check library can't properly parse. (local: "*latest*" CI: commit SHA)
```

pkg/shared/util/grpc_utils.go
ayildirim21: if we get rid of this then in the error on line 62 message the minNumaflowVersion will be blank and might confuse users

KeranYang: It won't, right? user should immediately realize that the minNumaflowVersion is empty

KeranYang: You are right, we should check in the case when user use an old sdk version which don't support writing minNumaflowVersion. Please also surface to the user which exact version to upgrade to instead of latest. Because latest might not be compatible to current numaflow.

pkg/shared/util/version_mapping.go
```diff
@@ -0,0 +1,29 @@
...
+package util
+
+import (
+	"github.com/numaproj/numaflow-go/pkg/info"
+)
+
+type sdkConstraints map[info.Language]string
+
+var minimumSupportedSDKVersions = sdkConstraints{
+	info.Go:     "0.7.0-rc2",
```
whynowy, line 26: Quick question: is `0.7.0` > `0.7.0-rc2`?

ayildirim21: yes it is

pkg/shared/util/grpc_utils_test.go
```diff
@@ -0,0 +1,96 @@
...
+				assert.Error(t, err, "Expected error")
+				assert.Contains(t, err.Error(), tt.errMessage)
+			} else {
+				assert.NoError(t, err, "Expected no error")
+			}
+		})
+	}
+}
+
+func TestCheckSDKCompatibility(t *testing.T) {
```
KeranYang, line 59: In this unit test, can we also cover the non-python branch?

pkg/shared/util/version_mapping.go
KeranYang: we will use the code for now.

whynowy (approved):

PR #1700, opened 2024-04-23
docs: add references to numaproj-contrib

### LINKED ISSUE #1596: Describe numaproj-contrib org using github org profile
# Summary

[numaproj-contrib](https://github.com/numaproj-contrib) does not describe what it has or how to use the org. We need to document this.

# Use Cases

We should be able to hyperlink to the contrib repo from our numaflow repo. Currently, it is discoverable only by word of mouth.

KeranYang:
Closes #1596


docs/user-guide/sinks/user-defined-sinks.md
```diff
@@ -1,9 +1,19 @@
-# User Defined Sinks
+# User-defined Sinks
```
whynowy, line 1: Let's make it consistent, if `user-defined` is the term, then use it everywhere.

docs/user-guide/sinks/user-defined-sinks.md
```diff
@@ -25,3 +35,9 @@ Some environment variables are available in the user defined sink container:
 - `NUMAFLOW_REPLICA` - Replica index.
 - `NUMAFLOW_PIPELINE_NAME` - Name of the pipeline.
 - `NUMAFLOW_VERTEX_NAME` - Name of the vertex.
+
+## User-defined Sink contributed from the open source community
```
whynowy, line 39: `Sinks`

docs/user-guide/sinks/user-defined-sinks.md
KeranYang: Agree. I think `user-defined` is the right term. Will do.

whynowy (approved):

PR #1702, opened 2024-04-24
chore: UDS based multiproc server

kohlisid:
This transitions the multiproc client to use a UDS socket instead of TCP.
We will receive the number of clients to start in the ServerInfo metadata 

And we will use UDS sockets in the format
```
unix:///var/run/numaflow/multiproc#serv_num.sock
```

Client side changes for [this](https://github.com/numaproj/numaflow-python/issues/161)

### DIFF
--- a/pkg/sdkclient/grpc_resolver/client_resolver.go
+++ b/pkg/sdkclient/grpc_resolver/client_resolver.go
@@ -101,7 +101,7 @@ func buildConnAddrs(numServers int) []string {
 // on multiprocessing TCP or UDS connection
 func RegMultiProcResolver(svrInfo *info.ServerInfo) error {
 	// Extract the server ports from the server info file and convert it to a list
-	numServers, _ := strconv.Atoi(svrInfo.Metadata["NUM_SERVERS"])
+	numServers, _ := strconv.Atoi(svrInfo.Metadata["MULTIPROC"])
 	log.Println("Multiprocessing Servers :", numServers)
 	conn := buildConnAddrs(numServers)
 	res := newMultiProcResolverBuilder(conn)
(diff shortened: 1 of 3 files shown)


whynowy (approved):
LGTM - pending on python sdk review

pkg/sdkclient/grpc_resolver/client_resolver.go
```diff
@@ -80,14 +82,17 @@ func (*multiProcResolver) Close()                                  {}
...
 
 // buildConnAddrs Populate the connection list for the clients
-// Format (serverAddr, serverIdx) : (0.0.0.0:5551, 1), (0.0.0.0:5552, 2)
-func buildConnAddrs(servPorts []string) []string {
-	var conn = make([]string, len(servPorts))
-	for i := 0; i < len(servPorts); i++ {
-		// Use the server ports from the server info file and assign to each client
-		addr, _ := strconv.Atoi(servPorts[i])
+// Format (serverAddr, serverIdx) :
+// (unix:///var/run/numaflow/multiproc0.sock, 1),
```
vigith, line 86: Why is 0 mapping to idx 1 but 2 is mapped to idx 2? doc typo?

kohlisid: Typo. Fixed, thanks for pointing out

whynowy: What was the reason we originally use 0 -> 1, 1 -> 2, instead of 0 -> 0, 1 -> 1?

kohlisid: That was from legacy code. Have updated to match server name with indexes

vigith (approved):

Issue #1686, opened 2024-04-18
Add documentation to maintain historical version config mappings


ayildirim21:
# Summary

As per https://github.com/numaproj/numaflow/pull/1683#discussion_r1569625987. We should also in the same file mention that `version_mapping.go` should be updated after each new release.

# Use Cases

After a release, in case developer did not know which files to update.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

KeranYang:
> We should also in the same file mention that version_mapping.go should be updated after each new release.

I think it would be better that mention it in the releasing.md file.

PR #1705, opened 2024-04-24
chore: version compatibility docs

ayildirim21:
as per #1686


.gitignore
```diff
@@ -16,3 +16,4 @@ site/
 *.idea/
 *.pytest_cache/
 docs/APIs.html
+test/profile.cov.tmp
```
ayildirim21, line 19: Whenever I run `make test-coverage` this file gets created

docs/development/releasing.md
```diff
@@ -2,21 +2,32 @@
...
 
-1. Cherry-pick fixes to the release branch, skip this step if it's the first release in the branch.
-1. Run `make test` to make sure all test cases pass locally.
-1. Push to remote branch, and make sure all the CI jobs pass.
-1. Run `make prepare-release VERSION=v{x.y.z}` to update version in manifests, where `x.y.x` is the expected new version.
-1. Follow the output of last step, to confirm if all the changes are expected, and then run `make release VERSION=v{x.y.z}`.
-1. Follow the output, push a new tag to the release branch, GitHub actions will automatically build and publish the new release, this will take around 10 minutes.
-1. Test the new release, make sure everything is running as expected, and then recreate a `stable` tag against the latest release.
+1. Cherry-pick fixes to the release branch, skip this step if it's the first release in the branch
+2. Run `make test` to make sure all test cases pass locally
```
whynowy, line 21: Use `1.`, it will be interpreted correctly.

docs/development/releasing.md
```diff
@@ -2,21 +2,32 @@
 
 ## Release Branch
 
-Always create a release branch for the releases, for example branch `release-0.5` is for all the v0.5.x versions release. If it's a new release branch, simply create a branch from `main`.
+Always create a release branch for the releases. For example branch `release-0.5` is for all the `v0.5.x` version releases. If it's a new release branch, simply create a branch from `main`.
```
whynowy, line 5: `releasing.md` is supposed to be read by contributors/developers, we could mention something like "version mapping needs to be updated at xxx when needed". The real version mapping would be a doc for the users, instead of using a yaml file, it would be better to be in a doc file, maybe shown as a table.

docs/version-mapping.yaml
vigith, line 1: Since we have this YAML file, can we generate a human friendly table view for docs at build time? Also this YAML can power the manual mapping we maintain in code. There should only be a single source of truth. I am quite certain it will diverge otherwise.

yhl25: +1 to use the same YAML for version compatibility check in the code.

docs/version-mapping.yaml
```diff
@@ -0,0 +1,19 @@
...
+  - Python: 0.7.0a
+  - Java:   0.7.1-0
+1.2.0-rc3:
+  - Go:     0.7.0-0
+  - Python: 0.7.0a
+  - Java:   0.7.1-0
+1.2.0-rc4:
+  - Go:     0.7.0-rc2
+  - Python: 0.7.0a1
+  - Java:   0.7.2-0
```
yhl25, line 19: Let's rearrange the order so that the latest version is at the top so that it is more relatable to the user.

docs/version-mapping.yaml
whynowy: This is a good idea.

whynowy: Actually, not that good. Need to carefully think about the process of development and release.

1. How to make `main` branch work (`latest`).
2. How to update the file during the release (e.g. add the config for the to be released version).
3. How to bring the new config in the release back to `main` branch.

Details like above need to be addressed.

vigith: I think we need to invest in that, our docs should be versioned based on release.

whynowy: I agree, We need to think it through about how to maintain this file with automation.

PR #1706, opened 2024-04-25
chore: avoid no-ack when ctx is canceled

yhl25:
* fixes flaky reduce-e2e tests

### DIFF
--- a/pkg/reduce/pbq/pbq.go
+++ b/pkg/reduce/pbq/pbq.go
@@ -68,11 +68,13 @@ func (p *PBQ) Write(ctx context.Context, request *window.TimedWindowRequest, per
 	}
 
 	// write the request to the output channel
+	// since it is a blocking write, we should have a select with context,
 	select {
 	case p.output <- request:
-
 	case <-ctx.Done():
-		return ctx.Err()
+		// we can persist the message even if the context is done that way we will not rely on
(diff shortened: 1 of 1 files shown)


pkg/reduce/pbq/pbq.go
```diff
@@ -68,11 +68,13 @@ func (p *PBQ) Write(ctx context.Context, request *window.TimedWindowRequest, per
 	}
 
 	// write the request to the output channel
+	// since it is a blocking write, we should have a select with context,
 	select {
 	case p.output <- request:
-
 	case <-ctx.Done():
-		return ctx.Err()
+		// we can persist the message even if the context is done that way we will not rely on
```
vigith, line 75: How will we guarantee that the message will be written to disk?

yhl25: Writing to disk is a blocking call, and we only close the WAL once we're done processing all the messages in the batch.

yhl25: Just like No-Ack, WAL writes also don't depend on the ctx.

vigith: Please open an issue to track

vigith (approved):

whynowy (approved):

Issue #1564, opened 2024-03-14
improve forward-a-chunk loop to support low tail-latency


vigith:
# Summary

Today, we have to fix constraints
- [x] only an element can be processed at a time
- [x] we create a map-stream (gRPC) for every batch (meaning streaming inside unary)
- [x] our batch is a blocking batch and will not proceed if we have a slow message
- [x] Read, Execute, Forward, Ack is an ordered (sync) workflow for a batch, we need to make it out-of-order non-blocking
- [x] move away from blocking batch to bounded concurrency (X in a batch to always have X in the backlog)

We need to remove these constraints without affecting the correctness or completeness property.

# Use Cases

* Better throughput

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

vigith:
Resolved via Rust rewrite.

PR #1707, opened 2024-04-27
feat: Stream mapper implementation of `MapStreamBatchFn`. Fixes #1688

### LINKED ISSUE #1688: Batch Mapping & Sink
# Summary

I'mm looking for the ability for a mapper and sink to take a batch of items at once. Today these take single item at once, regardless of `readBatchSize` limiting the ability to take advantage of bulk APIs/processing,.

# Use Cases
I have three main use cases that interest me. All of these desire to operate at near real-time, and the messages aren't necessarily required to be ordered for histogram/statistic type metrics, Thus, horizontal scaling should be easy for my use cases

1. Bundling: I have a use case receiving 100k+ messages per sec, but each message is very small, < 1k. For throughput purposes it would make sense to take batches of those messages (say up to 1000), and group them into a single bundle. The overall MB of messages may not be changed, but it, from my rough experimentation, I can get more ISB throughput when operating on this smaller quantity of messages. Implementation of this bundling is of course user centric, but being able to receive mulitple messages at once would facilitate this.
    * Input would be N messages with 1-N responses (typically 1 in my personal use-case).
2. Bundled transactions: Some operations can be generally more efficient by sending multiple data items through at once vs a single call for each. Ex. Elasticsearch BatchAPI, or pykafka producer (ignoring numaflow's built-in kafka sink).
    * Typically more of a 'sink' use case, 
3. Batching for performance: In a language like python, if you're doing the same operation on a lot of datasets using numpy/scipy/etc, it's useful to group them together and call the math logic once. If we could pull a batch of items at once, we can organize them for faster processing, and then write back out the results as appropriate. 
    * This specific case would still be N messages in, N messages out, but could be combined with Bundling

If there's a way to accomplish any of this today, I'm happy to adopt that, but I don't think the reduce capability quite fits my expectations, particularly with the time based windows and limitations of horizontal scaling in order to meet the ordering expectations.

---

```[tasklist]
### Tasks
- [ ] https://github.com/numaproj/numaflow/issues/1786
- [ ] https://github.com/numaproj/numaflow/issues/1787
- [ ] https://github.com/numaproj/numaflow/issues/1788
- [ ] https://github.com/numaproj/numaflow/issues/1962
- [ ] https://github.com/numaproj/numaflow/issues/1789
- [ ] https://github.com/numaproj/numaflow/issues/1827
- [ ] https://github.com/numaproj/numaflow/issues/1810
- [ ] https://github.com/numaproj/numaflow/issues/1809
- [ ] https://github.com/numaproj/numaflow/issues/1812
- [ ] #1817
- [ ] https://github.com/numaproj/numaflow/issues/1820
- [ ] https://github.com/numaproj/numaflow/issues/1819
```

magelisk:
This implements backend for updated stream mapper (`MapStreamBatchFn`) implemented here: https://github.com/numaproj/numaflow-go/pull/129

It allows stream mapper to have readBatchSize > 1. 
* If `readBatchSize == 1`, the capability today is executed as-is
* if `readBatchSize >= 1`, the newer interface will be used.

Individual langauge servicers can be implemetned to have the default handler for `MapStreamBatchFn` simply iterate over the one-item implementation so that it _should_ be completely backwards compatible for everyone. But now those who wish to implement handling more than datum item at a time, the capability is now exposed

Fixes: #1688  and partial fix of #1564

### DIFF
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -2,7 +2,7 @@
   "swagger": "2.0",
   "info": {
     "title": "Numaflow",
-    "version": "latest"
+    "version": "vwarner"
   },
   "paths": {},
   "definitions": {
--- a/docs/APIs.md
+++ b/docs/APIs.md
@@ -1,23 +1,43 @@
 <p>
+
 Packages:
+
 </p>
+
 <ul>
+
 <li>
+
 <a href="#numaflow.numaproj.io%2fv1alpha1">numaflow.numaproj.io/v1alpha1</a>
+
 </li>
+
 </ul>
+
 <h2 id="numaflow.numaproj.io/v1alpha1">
+
 numaflow.numaproj.io/v1alpha1
+
 </h2>
+
 <p>
+
 </p>
+
 Resource Types:
+
 <ul>
+
 </ul>
+
 <h3 id="numaflow.numaproj.io/v1alpha1.AbstractPodTemplate">
+
 AbstractPodTemplate
+
 </h3>
+
 <p>
+
 (<em>Appears on:</em>
 <a href="#numaflow.numaproj.io/v1alpha1.AbstractVertex">AbstractVertex</a>,
 <a href="#numaflow.numaproj.io/v1alpha1.DaemonTemplate">DaemonTemplate</a>,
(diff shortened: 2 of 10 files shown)


magelisk:
This is not relevant based on other designs that are expected. Closing

PR #1708, opened 2024-04-29
chore: Refactor Source to Avoid Implementing Forwarder Interface 

### LINKED ISSUE #1663: Refactor Source Implementations to Avoid Implementing Forwarder Interface
Since we have a separate forwarder for the source, our sources don't need to implement the forwarder interface.

Similar to #1662

yhl25:
fixes #1663

### DIFF
--- a/pkg/sources/forward/data_forward.go
+++ b/pkg/sources/forward/data_forward.go
@@ -78,9 +78,9 @@ func NewDataForward(
 	toVertexWmStores map[string]store.WatermarkStore,
 	idleManager wmb.IdleManager,
 	opts ...Option) (*DataForward, error) {
-	defaultOptions := DefaultOptions()
+	dOpts := defaultOptions()
 	for _, o := range opts {
-		if err := o(defaultOptions); err != nil {
+		if err := o(dOpts); err != nil {
 			return nil, err
 		}
 	}
@@ -115,10 +115,10 @@ func NewDataForward(
 		Shutdown: Shutdown{
 			rwlock: new(sync.RWMutex),
 		},
-		opts: *defaultOptions,
+		opts: *dOpts,
 	}
 	// add logger from parent ctx to child context.
-	isdf.ctx = logging.WithLogger(ctx, defaultOptions.logger)
+	isdf.ctx = logging.WithLogger(ctx, dOpts.logger)
 	return &isdf, nil
 }
 
@@ -270,6 +270,7 @@ func (df *DataForward) forwardAChunk(ctx context.Context) {
 		messageToStep[toVertex] = make([][]isb.Message, len(df.toBuffers[toVertex]))
 	}
 
+	// FIXME: when the transformer is not defined, we should avoid doing this.
 	// user-defined transformer concurrent processing request channel
 	transformerCh := make(chan *readWriteMessagePair)
 	// transformerResults stores the results after user-defined transformer processing for all read messages. It indexes
--- a/pkg/sources/kafka/reader.go
+++ b/pkg/sources/kafka/reader.go
@@ -61,224 +61,6 @@ type kafkaSource struct {
 	saramaClient  sarama.Client              // sarama client
 }
 
-// kafkaOffset implements isb.Offset
-// we need topic information to ack the message
-type kafkaOffset struct {
-	offset       int64
-	partitionIdx int32
-	topic        string
-}
-
-func (k *kafkaOffset) String() string {
-	return fmt.Sprintf("%s:%d:%d", k.topic, k.offset, k.partitionIdx)
-}
-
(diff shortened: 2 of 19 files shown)


pkg/sources/forward/data_forward.go
```diff
@@ -270,6 +270,7 @@ func (df *DataForward) forwardAChunk(ctx context.Context) {
 		messageToStep[toVertex] = make([][]isb.Message, len(df.toBuffers[toVertex]))
 	}
 
+	// FIXME: when the transformer is not defined, we should avoid doing this.
```
KeranYang, line 273: yes. @ayildirim21 is working on this.

pkg/sources/source.go
```diff
@@ -200,69 +207,90 @@ func (sp *SourceProcessor) Start(ctx context.Context) error {
...
+
 	if sp.VertexInstance.Vertex.HasUDTransformer() {
 		// Wait for server info to be ready
 		serverInfo, err := sdkserverinfo.SDKServerInfo(sdkserverinfo.WithServerInfoFilePath(sdkclient.SourceTransformerServerInfoFile))
 		if err != nil {
 			return err
 		}
 
-		sdkClient, err = sourcetransformer.New(serverInfo, sdkclient.WithMaxMessageSize(maxMessageSize))
+		sdkClient, err := sourcetransformer.New(serverInfo, sdkclient.WithMaxMessageSize(maxMessageSize))
```
KeranYang, line 232:
```suggestion
		srcTranformerClient, err := sourcetransformer.New(serverInfo, sdkclient.WithMaxMessageSize(maxMessageSize))
```

pkg/sources/source.go
```diff
@@ -200,69 +207,90 @@ func (sp *SourceProcessor) Start(ctx context.Context) error {
...
 		if err != nil {
-			return fmt.Errorf("failed to create gRPC client, %w", err)
+			return fmt.Errorf("failed to create transformer gRPC client, %w", err)
 		}
 
-		transformerGRPCClient := transformer.NewGRPCBasedTransformer(sdkClient)
+		transformerClient = transformer.NewGRPCBasedTransformer(sdkClient)
```
KeranYang, line 237:
```suggestion
		srcTransformerGRPCClient = transformer.NewGRPCBasedTransformer(sdkClient)
```

nit: this is to be consistent with srcClient/udsGRPCClient above for udsource.

KeranYang (commented):
LGTM. I think we can merge once UT failures are fixed.

pkg/sources/kafka/reader.go
```diff
@@ -28,107 +28,109 @@ import (
...
+			return nil, err
+		}
+	}
 
-// AckIt acking is taken care by the consumer group
-func (k *kafkaOffset) AckIt() error {
-	// NOOP
-	return nil
-}
+	sarama.NewConfig()
```
whynowy, line 76: Is this line in-use? How was it introduced?

pkg/sources/sourcer/sourcer.go
```diff
@@ -40,12 +39,9 @@ type SourceReader interface {
...
 	Partitions(ctx context.Context) []int32
+	// Pending returns the pending messages number.
+	Pending(context.Context) (int64, error)
 }
 
-// Sourcer interface provides an isb.BufferReader abstraction over the underlying data source.
-// This is intended to be consumed by a connector like isb.forward
-type Sourcer interface {
-	SourceReader
-	forwarder.StarterStopper
```
whynowy, line 49: In this case, how to implement those logics for `preStop`, `preStart` or `postStart`? Like the `ready` flag used in HTTP source?

I understand for `preStart` or `postStart`, we could leverage the `New` function to sort of achieve (e.g. you start the kafka consumer Goroutine) that, however that is not an elegant implementation, and it relies on the `ctx` cancel to exit, and there's no way to do a clean shutdown.

whynowy: Synced up offline, we could go ahead.

vigith:
Though this PR removes some redundant code required for implementing a built-in source, this PR does not solve the problem of orchestrating a clean shutdown. An ideal source reader will always read ahead and be ready to hand over data, but when shutting down, it will ensure the read-ahead data is emptied.

Perhaps we need to rethink how to orchestrate a clean shutdown for a source (also for a sink), and this PR could remove the redundant code **without any regression**.

Using `ctx.Done` to do shutdown is like having a bull run through a shop. It will randomly pull down things without any proper order, and we will have to heavily rely on custom edge case handling.

vigith:
@whynowy how should we proceed? Should we merge this in and then make it cleaner in subsequent PRs?

whynowy (approved):

Issue #1612, opened 2024-03-28
Bypass transformer calls if source transformer is not used


vigith:
`source/data_forward.go` always assumes a transformer is associated, and in case it does not find one, we apply the following code. This has a performance hit because it creates unwanted go-routines and channels.

The [following code](https://github.com/numaproj/numaflow/blob/main/pkg/sources/forward/applier/sourcetransformer.go#L39-L46) is applied if no transformer is found, but there is heavy machinery behind the scenes. 
```go
var (
	// Terminal Applier do not make any change to the message
	Terminal = ApplySourceTransformFunc(func(ctx context.Context, msg *isb.ReadMessage) ([]*isb.WriteMessage, error) {
		return []*isb.WriteMessage{{
			Message: msg.Message,
		}}, nil
	})
)
```

PR #1709, opened 2024-04-30
chore: bypass transformer call if UDTransformer is not defined

ayildirim21:
related to: #1612


pkg/sources/forward/data_forward.go
```diff
@@ -278,35 +278,44 @@ func (df *DataForward) forwardAChunk(ctx context.Context) {
...
+				defer wg.Done()
+				df.concurrentApplyTransformer(ctx, transformerCh)
+			}()
+		}
+		concurrentTransformerProcessingStart := time.Now()
+
+		// send to transformer only the data messages
+		for idx, m := range readMessages {
+			// emit message size metric
+			metrics.ReadBytesCount.With(map[string]string{metrics.LabelVertex: df.vertexName, metrics.LabelPipeline: df.pipelineName, metrics.LabelVertexType: string(dfv1.VertexTypeSource), metrics.LabelVertexReplicaIndex: strconv.Itoa(int(df.vertexReplica)), metrics.LabelPartitionName: df.reader.GetName()}).Add(float64(len(m.Payload)))
```
KeranYang, line 297: We need this metric even when transformer is not specified

KeranYang: Also, the watermark assignment right below is needed.

KeranYang: Let me know if you are not sure about which piece is transformer-specific or global.

pkg/sources/forward/data_forward.go
```diff
@@ -192,6 +192,31 @@ type readWriteMessagePair struct {
 	transformerError error
 }
 
+func processReadMessages(hasUDTransformer bool, readMessages []*isb.ReadMessage, processorWM wmb.Watermark,
```
KeranYang, line 195: Please write a small paragraph as a comment about what this function does, especially what it **owns**. If it owns processing the messages from the ReadMessages list and generating WriteMessages list ready for writing to buffer, then strictly there should be no concept of UDTransformer outside of this function.

pkg/sources/forward/data_forward.go
```diff
@@ -192,6 +192,31 @@ type readWriteMessagePair struct {
...
+		// assign watermark to the message
+		m.Watermark = time.Time(processorWM)
+		transformerResults[idx].readMessage = m
+
+		if hasUDTransformer {
+			// send transformer processing work to the channel
+			transformerCh <- &transformerResults[idx]
+		} else {
+			// if no user-defined transformer exists, then the messages to write will be identical to the messages read from source
+			transformerResults[idx].writeMessages = []*isb.WriteMessage{{Message: m.Message}}
```
KeranYang, line 215: * With this change, is transformerResults still a concise name?
* It looks we are assigning writeMessages within processReadMessages. Does `processReadMessages` own populating writeMessages? If so, should it also populate when a transformer exists?

We can sync up on this tmr morning in the office.

KeranYang:
@ayildirim21, let's also think about how we can unit-test this change.

pkg/sources/forward/data_forward.go
```diff
@@ -270,50 +323,52 @@ func (df *DataForward) forwardAChunk(ctx context.Context) {
...
+
+		metrics.SourceTransformerConcurrentProcessingTime.With(map[string]string{
+			metrics.LabelVertex:             df.vertexName,
+			metrics.LabelPipeline:           df.pipelineName,
+			metrics.LabelVertexReplicaIndex: strconv.Itoa(int(df.vertexReplica)),
+			metrics.LabelPartitionName:      df.reader.GetName(),
+		}).Observe(float64(time.Since(concurrentTransformerProcessingStart).Microseconds()))
+		// transformer processing is done.
+	} else {
+		processReadMessages(false, readMessages, processorWM, readWriteMessagePairs, df, nil)
```
KeranYang, line 363: This function call has `hasUDTransformer` as `false` and `tranformerCh` as `nil`. I sense an opportunity for improvement. Can we include line L328->L364 in the processReadMessages method such that we can call `processReadMessages(df.transformer != nil, readMessages, processorWM, readWriteMessagePairs, df)` and inside of `processReadMessages`, conditionally create a transformerCh?

Two benefits: 1. more readable code as developers might not care about the transformer when looking through the code. 2. de-coupled input parameters of `processReadMessages`.

ayildirim21: I refactored a bit and if we create the channel inside the helper function we would essentially just move the entire logic into the function, so i decided to just keep it in the forwardAChunk method and make a helper for emitting the metrics.

pkg/sources/forward/data_forward.go
```diff
@@ -192,6 +195,17 @@ type readWriteMessagePair struct {
 	transformerError error
 }
 
+// emitMessageSizeMetric emits the message size metric
+func emitMessageSizeMetric(df *DataForward, m *isb.ReadMessage) {
```
KeranYang, line 199: I don't think we need to create a separate function for `ReadBytesCount`. For 2 reasons: 1. it's inconsistent with how we emit other metrics, one can argue if we create func for ReadBytesCount, why not all other metrics? 2. and mainly, functions should be [deep](https://medium.com/@alexturdean/shallow-functions-the-silent-killer-of-good-software-527a29ab56c9).

KeranYang (changes requested):

pkg/sources/source.go
```diff
@@ -271,7 +270,7 @@ func (sp *SourceProcessor) Start(ctx context.Context) error {
 	if sp.VertexInstance.Vertex.HasUDTransformer() {
 		sourceForwarder, err = sourceforward.NewDataForward(sp.VertexInstance, sourceReader, writersMap, sp.getTransformerGoWhereDecider(shuffleFuncMap), srcTransformerGRPCClient, fetchWatermark, sourceWmPublisher, toVertexWatermarkStores, idleManager, forwardOpts...)
 	} else {
-		sourceForwarder, err = sourceforward.NewDataForward(sp.VertexInstance, sourceReader, writersMap, sp.getSourceGoWhereDecider(shuffleFuncMap), applier.Terminal, fetchWatermark, sourceWmPublisher, toVertexWatermarkStores, idleManager, forwardOpts...)
+		sourceForwarder, err = sourceforward.NewDataForward(sp.VertexInstance, sourceReader, writersMap, sp.getSourceGoWhereDecider(shuffleFuncMap), nil, fetchWatermark, sourceWmPublisher, toVertexWatermarkStores, idleManager, forwardOpts...)
 	}
```
yhl25, line 274: make transformer an optional parameter? you can refer to fallback sink implementation.

pkg/sources/forward/data_forward.go
yhl25: +1

KeranYang:
@ayildirim21 awesome job on collecting the perf data!

KeranYang (approved):

vigith (approved):

vigith:
@yhl25, can you look into why idle-source for kafka is failing?

yhl25 (approved):

PR #1733, opened 2024-05-17
Tick generator blob - Closes #1732

### LINKED ISSUE #1732: User Defined data in generator
# Summary

Generator source is useful for very basic testing of ISB, but provides very little value to user pipelines. It would be useful to easily control data coming out of generator source to help feed user pipelines to do more realistic load testing without requiring users to setup full structure of their inputs.

# Use Cases

* Load/performance testing of user pipelines with predetermined data. 
* Regression testing of pipelines with known datasets

---

magelisk:
Updates generator source to allow setting `valueBlob` to explicitly provide the packet that will be sent. This allows for users to provide meaningful data and test their pipelines. Implements recommendation in #1732 . 

I was trying to profile a pipeline and found myself jumping through hoops to inject meaningful data, and I found it easy enough to implement this for my own testing and it notably simplified my setup and ability to tweak settings. 

In order to test locally, you will need to reinstall the pipeline and vertex CRD, then restart the controller. Then simply add a desired `valueBlob` to a pipeline (like `1-saimple-pipeline.yaml` ). 

Discussion question: I'm more than open a different name for the field than `valueBlob`. I know some people don't like the word `blob` but I did want to express (as best as possible) that it could be anything - not just a JSON message, but wasn't sure if it should include `b64` to indicate required formating


pkg/sources/generator/tickgen.go
```diff
@@ -132,6 +135,21 @@ func NewMemGen(ctx context.Context, vertexInstance *dfv1.VertexInstance, opts ..
 		jitter = vertexInstance.Vertex.Spec.Source.Generator.Jitter.Duration
 	}
 
+	var genFunction = recordGenerator
+	if vertexInstance.Vertex.Spec.Source.Generator.ValueBlob != nil {
+		logger.Info("ValueBlob was set, using provided value instead of randomly generated data")
+		emitBytes, err := base64.StdEncoding.DecodeString(*vertexInstance.Vertex.Spec.Source.Generator.ValueBlob)
+		if err != nil {
+			logger.Errorw("Error decoding provided ValueBlob as base64, cannot initialize generator", zap.Error(err))
```
whynowy, line 143: In this case, shall we return the error?

magelisk: Indeed, fixed to include return

vigith:
run `make codegen`?

docs/user-guide/sources/generator.md
```diff
@@ -31,3 +31,19 @@ spec:
     - from: p1
       to: out
 ```
+
+## User Defined Data
+The default data created by the generator is not likely to be useful in testing user pipelines with spefici business logic. To allow this to help with user testing, a user defined value can be provided which will be emitted for each of the generator.
```
vigith, line 36:
```suggestion
The default data created by the generator is not likely to be useful in testing user pipelines with specific business logic. To allow this to help with user testing, a user defined value can be provided which will be emitted for each of the generator.
```

vigith (approved):

pkg/sources/generator/tickgen.go
magelisk: done

PR #1744, opened 2024-05-30
feat: adding numaflow version to the UI

### LINKED ISSUE #1714: Display Numaflow Version Information on UI Home and Pipeline Pages
* The home page could display the Numaflow Server version it is currently running on.
* The pipeline page could display the Numaflow controller version being used by the pipeline.

veds-g:
Fixes #1714 

App (with `GitTreeState = dirty`) - 
image

Plugin (with `GitTreeState=clean`)- 
image


whynowy:
1. The use a grey color for the version info so that it won't be as outstanding as the `numaflow`?
Screenshot 2024-05-30 at 3 33 04 PM

2. I understand we are following Argo CD style, to put the version on the top left. For our case, our top right is also empty, will it be better to put it there, so that it will not impact the left side `numaflow`? This is my "unprofessional" thought, if putting on the right side looks weird/ugly or anything else, then just ignore it.
Screenshot 2024-05-30 at 3 33 32 PM

whynowy:
Use `Numaflow Server Version` instead of `Numaflow Version` to avoid confusion.

Screenshot 2024-05-30 at 3 39 16 PM

veds-g:
> 1. Use a grey color for the version info so that it won't be as outstanding as the `numaflow`?
> 
> Screenshot 2024-05-30 at 3 33 04 PM
> 2. I understand we are following Argo CD style, to put the version on the top left. For our case, our top right is also empty, will it be better to put it there, so that it will not impact the left side `numaflow`? This is my "unprofessional" thought, if putting on the right side looks weird/ugly or anything else, then just ignore it.
> 
> Screenshot 2024-05-30 at 3 33 32 PM

I have a concern around when user enables login, then an additional button will appear on the right corner, then we might want to change the arrangement, I want it to be somewhere where it can be constant. wdyt?
image

whynowy:
> > 1. Use a grey color for the version info so that it won't be as outstanding as the `numaflow`?
> > 
> > Screenshot 2024-05-30 at 3 33 04 PM
> > 2. I understand we are following Argo CD style, to put the version on the top left. For our case, our top right is also empty, will it be better to put it there, so that it will not impact the left side `numaflow`? This is my "unprofessional" thought, if putting on the right side looks weird/ugly or anything else, then just ignore it.
> > Screenshot 2024-05-30 at 3 33 32 PM
> 
> I have a concern around when user enables login, then an additional button will appear on the right corner, then we might want to change the arrangement, I want it to be somewhere where it can be constant. wdyt? image

You are right, I forgot about it.

whynowy (approved):

PR #1753, opened 2024-06-06
feat: publish to callback endpoint for tracking

yhl25:


pkg/udf/rpc/grpc_map.go
```diff
@@ -141,13 +145,18 @@ func (u *GRPCBasedMap) ApplyMap(ctx context.Context, readMessage *isb.ReadMessag
...
 	writeMessages := make([]*isb.WriteMessage, 0)
-	for _, result := range response.GetResults() {
+	for index, result := range response.GetResults() {
 		keys := result.Keys
 		taggedMessage := &isb.WriteMessage{
 			Message: isb.Message{
 				Header: isb.Header{
 					MessageInfo: parentMessageInfo,
 					Keys:        keys,
+					ID: isb.MessageID{
```
whynowy, line 155: `ID` was not set?

yhl25: it was set inside forward.go

pkg/shared/callback/callback.go
```diff
@@ -0,0 +1,275 @@
...
+	resp, err := client.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to send request: %w", err)
+	}
+
+	if resp.StatusCode > 299 {
+		return fmt.Errorf("received non-OK response status: %s", resp.Status)
+	}
+
+	_ = resp.Body.Close()
```
whynowy, line 243: Use `defer` before line 239.

pkg/shared/callback/callback.go
```diff
@@ -0,0 +1,275 @@
...
+		opt(dOpts)
+	}
+
+	clientCache, _ := lru.NewWithEvict[string, *http.Client](dOpts.cacheSize, func(key string, value *http.Client) {
+		// Close the client when it's removed from the cache
+		value.CloseIdleConnections()
+	})
+
+	if dOpts.callbackURL != "" {
+		client := &http.Client{
```
whynowy, line 41: This has been existing in `GetClient()`, do not need it here?

pkg/sources/source.go
```diff
@@ -74,7 +75,7 @@ func (sp *SourceProcessor) Start(ctx context.Context) error {
 		healthCheckers           []metrics.HealthChecker
 		idleManager              wmb.IdleManager
 		pipelineName             = sp.VertexInstance.Vertex.Spec.PipelineName
-		vertexName               = sp.VertexInstance.Vertex.Name
+		vertexName               = sp.VertexInstance.Vertex.Spec.Name
```
whynowy, line 78: Will existing pipelines get impacted with this fix?

pkg/isb/message.go
```diff
@@ -73,6 +74,23 @@ type Header struct {
...
 
+// MessageID is the message ID of the message which is used for exactly-once-semantics.
+type MessageID struct {
+	// VertexName is the name of the vertex
+	VertexName string
+	// Offset is the offset of the message
+	// NOTE: should be unique across the replicas of the vertex, that is the
+	// reason we don't have a separate replica field in the MessageID
+	Offset string
+	// Index is the index of the message
```
whynowy, line 85:
```suggestion
	// Index is the index of a flatmap message, otherwise use 0
```

pkg/isb/message.go
```diff
@@ -73,6 +74,23 @@ type Header struct {
 	Headers map[string]string
 }
 
+// MessageID is the message ID of the message which is used for exactly-once-semantics.
+type MessageID struct {
```
whynowy, line 78: Will this change impact existing pipelines?

yhl25: No

pkg/sources/source.go
yhl25: No

pkg/shared/callback/callback.go
yhl25: we don't need it

pkg/sinks/sink.go
```diff
@@ -87,7 +92,7 @@ func (u *SinkProcessor) Start(ctx context.Context) error {
 		// create reader for each partition. Each partition is a group in redis
 		for index, bufferPartition := range u.VertexInstance.Vertex.OwnedBuffers() {
 			fromGroup := bufferPartition + "-group"
-			consumer := fmt.Sprintf("%s-%v", u.VertexInstance.Vertex.Name, u.VertexInstance.Replica)
+			consumer := fmt.Sprintf("%s-%v", vertexName, u.VertexInstance.Replica)
```
whynowy, line 95: This is risky. `vertex.spec.Name` is not unique in one namespace, while `vertex.Name` is.

pkg/udf/map_udf.go
```diff
@@ -125,9 +130,13 @@ func (u *MapUDFProcessor) Start(ctx context.Context) error {
 		return fmt.Errorf("unrecognized isbsvc type %q", u.ISBSvcType)
 	}
 
-	enableMapUdfStream, err := u.VertexInstance.Vertex.MapUdfStreamEnabled()
-	if err != nil {
-		return fmt.Errorf("failed to parse UDF map streaming metadata, %w", err)
+	enableMapStreamStr := os.Getenv(dfv1.EnvMapStreaming)
```
whynowy, line 133: Use `pkg/shared/util.LookupEnvBoolOr()`

pkg/shared/callback/callback.go
```diff
@@ -0,0 +1,285 @@
...
+// GetClient returns a client for the given URL from the cache
+// If the client is not in the cache, a new one is created.
+func (u *Uploader) GetClient(url string) *http.Client {
+	// Check if the client is in the cache
+	if client, ok := u.clientsCache.Get(url); ok {
+		return client
+	}
+
+	// If the client is not in the cache, create a new one
+	client := &http.Client{
```
vigith, line 265: if the callback server (for this url) is restarted, will the http.Client auto-reconnect internally or will this client become invalid?

vigith (approved):

whynowy (approved):

examples/1-simple-pipeline.yaml
```diff
@@ -5,16 +5,20 @@ metadata:
...
   vertices:
     - name: in
+      scale:
+        min: 1
       source:
         # A self data generating source
-        generator:
-          rpu: 5
-          duration: 1s
+        http: {}
```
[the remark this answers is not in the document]
yhl25: This was added by mistake.

pkg/shared/callback/callback.go
yhl25: For the server restart case, If I am not wrong the client doesn't maintain a persistent connection to the server, so it simply creates a new connection when there is a post request, so ideally it should be fine. But still, the client caches the TCP connections. Let me run some tests, If I find any issues I will create a follow-up PR to fix it.

PR #1767, opened 2024-06-24
chore: Enable Rust Unit Tests And Code Coverage in CI

yhl25:


vigith (approved):
I am okay with renaming `Unit Tests` to `Unit Tests Go` because we should add `NodeJS` in the future.

whynowy (approved):

PR #1771, opened 2024-06-26
feat: use protobuf to store header and messages in ISB

### LINKED ISSUE #1768: Use protobuf for payload and header serde
# Summary

We use [encoding/binary](https://pkg.go.dev/encoding/binary) for serde. We need to standardize on a wire format that is scalable and also supports backward compatibility. `encoding/binary` encourages using `gob` or `protobuf`; I think `protobuf` has more benefits than `gob` because it is language agnostic and streaming is not implicit.

---

yhl25:
fixes #1768


.github/workflows/ci.yaml
```diff
@@ -61,6 +61,8 @@ jobs:
           sudo find /usr/local/include -type d | xargs sudo chmod a+rx
           rm -f $PROTOC_ZIP
           ls /usr/local/include/google/protobuf/
+          go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
```
whynowy, line 64: Move them to the proto gen script?

pkg/isb/serde.go
```diff
@@ -1,431 +1,126 @@
...
+// UnmarshalBinary decodes Message from the proto bytes.
+// Even though the header and body are pointers in the proto, we don't need
+// nil checks because while marshalling, we always set the default values.
+func (m *Message) UnmarshalBinary(data []byte) error {
+	pb := &isb.Message{}
+	if err := proto.Unmarshal(data, pb); err != nil {
+		return err
+	}
+
+	m.Header = Header{
```
vigith, line 43:
```suggestion
	// no nil checks are performed because Marshalling and Unmarshalling
	// are completely controlled by numaflow. It is best not to do nil checks
	// and get panics so we can fix the root cause.
	
	m.Header = Header{
```

pkg/isb/serde.go
```diff
@@ -1,431 +1,126 @@
...
-		return nil, fmt.Errorf("expected to write id size of %d but got %d", preamble.IDLen, n)
-	}
-	for i := 0; i < len(h.Keys); i++ {
-		if err = binary.Write(buf, binary.LittleEndian, int16(len(h.Keys[i]))); err != nil {
-			return nil, err
-		}
-		if err = binary.Write(buf, binary.LittleEndian, []byte(h.Keys[i])); err != nil {
-			return nil, err
-		}
+	h.MessageInfo = MessageInfo{
```
vigith, line 85:
```suggestion
	// no nil checks are performed because Marshalling and Unmarshalling
	// are completely controlled by numaflow. It is best not to do nil checks
	// and get panics so we can fix the root cause.
	
	h.MessageInfo = MessageInfo{
```

pkg/reduce/data_forward_test.go
```diff
@@ -647,7 +647,7 @@ func TestReduceDataForward_IdleWM(t *testing.T) {
 	assert.Equal(t, isb.WMB, msgs[0].Kind)
 	// the second message should be the data message from the closed window above
 	// in the test ApplyUDF above we've set the final message to have key="result"
-	for len(msgs[1].Keys) == 0 {
+	for len(msgs) < 2 || len(msgs[1].Keys) == 0 {
```
vigith, line 650: what made you add this `len(msgs) < 2`, check?

vigith (approved):
any performance regressions?

i see that the WAL size can come down, so that is good :)

vigith:
watermark propagation is still using `encoding/binary`; let's take that up too, next?

.github/workflows/ci.yaml
yhl25: Didn't work, got a permission denied error.

pkg/reduce/data_forward_test.go
yhl25: There was a bug in simple in memory buffer implementation, because of which this test was working fine without this check.  https://github.com/numaproj/numaflow/blob/491b6374b8ce0312eca08110ccc723acea863347/pkg/isb/stores/simplebuffer/buffer.go#L269

yhl25:
> any performance regressions?
> 
> i see that the WAL size can come down, so that is good :)

I didn't notice much difference with a low TPS pipeline. I have yet to perform some performance testing with a high TPS pipeline.

PR #1777, opened 2024-06-27
chore: propagate all the vertex template properties to vertex object

whynowy:
Currently, when creating vertex pods, we merge the vertex template definition in a pipeline object (`spec.templates.vertex.xxx`) with the vertex definitions (`spec.vertices[.].xxx`), to generate the final pod spec. And the vertex template definition will not be propagated to the `Vertex` object. There's nothing wrong, but it has some potential problems.

1. If someone updates the vertex template (e.g. add an annotation to all the vertex pods), the new pod spec will not take effect unless the vertex pod is recreated (because there's no `Vertex` object change);
2. `Vertex` object does not have finalized spec, and we could not use its spec to make runtime decision - e.g. to tell if there's an annotation on a vertex pod. @yhl25 

This PR updated the logic to propagate the vertex template definition to the vertex object, and stop merging the template definition when generating pod spec from a vertex object. It makes no change to the final vertex pod spec.


pkg/reconciler/vertex/controller.go
```diff
@@ -251,20 +251,6 @@ func (r *vertexReconciler) reconcile(ctx context.Context, vertex *dfv1.Vertex) (
 		if needToCreate {
 			labels := map[string]string{}
 			annotations := map[string]string{}
-			if pipeline.Spec.Templates != nil && pipeline.Spec.Templates.VertexTemplate != nil {
-				apt := pipeline.Spec.Templates.VertexTemplate.AbstractPodTemplate
-				apt.ApplyToPodSpec(podSpec)
```
vigith, line 256: who does the apply now?

vigith (approved):

pkg/reconciler/vertex/controller.go
whynowy: Do not need to apply anymore, the spec comes from vertex already has everything.

PR #1778, opened 2024-06-28
feat: implement map batch

kohlisid:
Partially implements #1787


pkg/udf/forward/forward.go
```diff
@@ -73,17 +73,7 @@ type InterStepDataForward struct {
...
-	fromStep isb.BufferReader,
-	toSteps map[string][]isb.BufferWriter,
-	fsd forwarder.ToWhichStepDecider,
-	applyUDF applier.MapApplier,
-	applyUDFStream applier.MapStreamApplier,
-	fetchWatermark fetch.Fetcher,
-	publishWatermark map[string]publish.Publisher,
-	idleManager wmb.IdleManager,
-	opts ...Option) (*InterStepDataForward, error) {
+func NewInterStepDataForward(vertexInstance *dfv1.VertexInstance, fromStep isb.BufferReader, toSteps map[string][]isb.BufferWriter, fsd forwarder.ToWhichStepDecider, applyUDF applier.MapApplier, applyUDFStream applier.MapStreamApplier, applyUDFBatchMap applier.BatchMapApplier, fetchWatermark fetch.Fetcher, publishWatermark map[string]publish.Publisher, idleManager wmb.IdleManager, opts ...Option) (*InterStepDataForward, error) {
```
vigith, line 76: can you merge all the appliers into a struct? the list is arg list is exploding :)

pkg/udf/forward/forward.go
```diff
@@ -249,50 +239,37 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
...
+	// assign watermark to data messages
+	for _, msg := range dataMessages {
+		msg.Watermark = time.Time(processorWM)
+	}
+
+	var udfResults []isb.ReadWriteMessagePair
 	var writeOffsets map[string][][]isb.Offset
-	if !isdf.opts.enableMapUdfStream {
+	// Check if map streaming mode is enabled
+	if isdf.opts.enableMapUdfStream {
```
vigith, line 250: this check will go away, right?

pkg/udf/map_udf.go
```diff
@@ -157,28 +160,49 @@ func (u *MapUDFProcessor) Start(ctx context.Context) error {
...
-		mapHandler = rpc.NewUDSgRPCBasedMap(vertexName, mapClient)
-
-		// Readiness check
-		if err := mapHandler.WaitUntilReady(ctx); err != nil {
-			return fmt.Errorf("failed on map UDF readiness check, %w", err)
-		}
-		defer func() {
-			err = mapHandler.CloseConn(ctx)
+		// if Batch Map mode is enabled create the client and handler for that accordingly
+		if enableBatchMapUdf {
```
vigith, line 171: this will also go away, right?

pkg/udf/rpc/tracker.go
vigith, line 1: why do we need a tracker for Batch if we are calling the rpc bidi for each batch?

pkg/udf/forward/forward.go
kohlisid: Not right now, in this PR we would not unify all the map functionalities. Would be better to do in a follow up PR as a separate task

pkg/udf/map_udf.go
kohlisid: Not right now, in this PR we would not unify all the map functionalities. Would be better to do in a follow up PR as a separate task

pkg/udf/rpc/tracker.go
kohlisid: We need the tracker to trace back the read message for a corresponding ID. Once we get responses back from the stream, we need some information like the `MessageInfo` and `Headers` to be written using the parent request.

kohlisid: We could potentially keep that message attached with each grpc request so that the sdk can add it to the response, this would then eliminate the need for the tracker here, but on the counterside have more data be streamed on grpc.

vigith: Let's do heavy lifting in Numaflow side so that the SDKs are lighter

pkg/udf/map_udf.go
vigith: Let's do it via follow up

pkg/isb/message.go
```diff
@@ -71,6 +71,7 @@ type Header struct {
 	Keys []string
 	// Headers is the headers of the message which can be used to store and propagate source headers like kafka headers,
 	// http headers and Numaflow platform headers like tracing headers etc.
+	//TODO(map-batch): can we rename this? Gets confusing for understanding headers under header
```
vigith, line 74:
```suggestion
	//TODO: can we rename this? Gets confusing for understanding headers under header
```

pkg/sdkclient/batchmapper/client.go
```diff
@@ -0,0 +1,178 @@
...
+
+	// read the response from the server stream and send it to responseCh channel
+	// any error is sent to errCh channel
+	go func() {
+		// close this channel to indicate that no more elements left to receive from grpc
+		// We do defer here on the whole go-routine as even during a error scenario, we
+		// want to close the channel and stop forwarding any more responses from the UDF
+		// as we would be replaying the current ones.
+		defer close(responseCh)
+		var resp *batchmappb.BatchMapResponse
```
vigith, line 118:
```suggestion

		var resp *batchmappb.BatchMapResponse
```

pkg/sdkclient/batchmapper/client.go
```diff
@@ -0,0 +1,178 @@
...
+// BatchMapFn is the handler for the gRPC client (Numa container)
+// It takes in a stream of input Requests, sends them to the gRPC server(UDF) and then streams the
+// responses received back on a channel asynchronously.
+// We spawn 2 goroutines here, one for sending the requests over the stream
+// and the other one for receiving the responses
+func (c *client) BatchMapFn(ctx context.Context, inputCh <-chan *batchmappb.BatchMapRequest) (<-chan *batchmappb.BatchMapResponse, <-chan error) {
+	// errCh is used to track and propagate any errors that might occur during the rpc lifecyle, these could include
+	// errors in sending, UDF errors etc
+	// These are propagated to the applier for further handling
+	errCh := make(chan error)
```
vigith, line 90: who is closing this?

pkg/sdkclient/batchmapper/client.go
```diff
@@ -0,0 +1,178 @@
...
+	}()
+
+	// Read from the read messages and send them individually to the bi-di stream for processing
+	// in case there is an error in sending, send it to the error channel for handling
+	go func() {
+		// TODO(map-batch): Should we check for ctx.Done here as well?
+		for inputMsg := range inputCh {
+			err = stream.Send(inputMsg)
+			if err != nil {
+				go func(sErr error) {
```
vigith, line 155: why do we need a `go func` for every message?

pkg/sdkclient/batchmapper/client.go
```diff
@@ -0,0 +1,178 @@
...
+				// Otherwise this would be a break
+				return
+			}
+		}
+		// CloseSend closes the send direction of the stream. This indicates to the
+		// UDF that we have sent all requests from the client, and it can safely
+		// stop listening on the stream
+		sendErr := stream.CloseSend()
+		if sendErr != nil && !errors.Is(sendErr, io.EOF) {
+			go func(sErr error) {
```
vigith, line 171: do we need a go func

pkg/udf/forward/forward.go
```diff
@@ -399,12 +371,77 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
...
+// processBatchMessages is used for processing the Batch Map mode UDF
+// batch map processing we send a list of N input requests together to the UDF and get the consolidated
+// response for all of them.
+func (isdf *InterStepDataForward) processBatchMessages(ctx context.Context, dataMessages []*isb.ReadMessage) []isb.ReadWriteMessagePair {
+	concurrentUDFProcessingStart := time.Now()
+	udfResults, err := isdf.mapAppliers.BatchMapUDF.ApplyBatchMap(ctx, dataMessages)
+	if err != nil {
+		// In case of an error received while processing the UDF call, we would do not handle partial failures,
+		// and hence would want to replay the whole batch currently.
+		// For achieving this, no-ack all the messages in the batch to force a re-read from the ISB
```
vigith, line 422: this cannot be held true all the time, ack can fail midway too. please add in the documentation that Batch Map can behave both way based on where the failure is.

pkg/sdkclient/batchmapper/client.go
kohlisid: For error channel we explicitly do not need to tell the reader that the channel has no more data. In that case this should be garbage collected when the routine ends right?

Or if its better to close I think we can add a close at
https://github.com/numaproj/numaflow/pull/1778/files/6b8271529878d5994cf1b20ffa15f09b9b5f4448#diff-bff711417b7f070d80a34939e929d12d12667a29719cc68e23671f3306a223daR117

pkg/udf/map_udf.go
```diff
@@ -156,13 +160,40 @@ func (u *MapUDFProcessor) Start(ctx context.Context) error {
...
+			return fmt.Errorf("failed on batch map UDF readiness check, %w", err)
+		}
+		defer func() {
+			err = batchMapHandler.CloseConn(ctx)
+			if err != nil {
+				log.Warnw("Failed to close gRPC client conn", zap.Error(err))
+			}
+		}()
 	} else {
+		// As a vertex can be either Map mode or Batch mode we reuse the same server-info file and socket file for both
```
whynowy, line 189: Is the comment correct?

pkg/udf/map_udf.go
```diff
@@ -130,9 +132,11 @@ func (u *MapUDFProcessor) Start(ctx context.Context) error {
 	}
 
 	maxMessageSize := sharedutil.LookupEnvIntOr(dfv1.EnvGRPCMaxMessageSize, sdkclient.DefaultGRPCMaxMessageSize)
-
 	enableMapUdfStream := sharedutil.LookupEnvBoolOr(dfv1.EnvMapStreaming, false)
+	enableBatchMapUdf := sharedutil.LookupEnvBoolOr(dfv1.EnvBatchMap, false)
```
whynowy, line 136: This options should be mutually exclusive, check it somewhere.

pkg/udf/rpc/tracker.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func NewTracker() *Tracker {
+	return &Tracker{
+		m: make(map[string]*isb.ReadMessage),
+	}
+}
+
+func (t *Tracker) AddRequest(msg *isb.ReadMessage) {
+	id := msg.ReadOffset.String()
```
whynowy, line 22: lock

pkg/udf/rpc/tracker.go
```diff
@@ -0,0 +1,75 @@
...
+
+func (t *Tracker) PrintAll() {
+	t.lock.RLock()
+	defer t.lock.RUnlock()
+	for k, v := range t.m {
+		log.Println("MYDEBUG: MAP VALS", k, " ", v)
+	}
+}
+
+func (t *Tracker) Get(key string) (*isb.ReadMessage, bool) {
```
whynowy, line 42: Do all the functions need to be public?

pkg/udf/rpc/tracker.go
```diff
@@ -0,0 +1,75 @@
+package rpc
+
+import (
+	"log"
+	"sync"
+
+	"github.com/numaproj/numaflow/pkg/isb"
+)
+
+type Tracker struct {
```
whynowy, line 10: private?

pkg/udf/map_udf.go
kohlisid: Done.

pkg/udf/rpc/tracker.go
kohlisid: I have kept the lock inside the `t.set` function. Do you see a need to lock here instead?

pkg/udf/rpc/tracker.go
kohlisid: Updated.

pkg/udf/rpc/tracker.go
kohlisid: Updated

pkg/udf/map_udf.go
kohlisid: It was an older comment when vertex was unified. Updated

pkg/udf/rpc/tracker.go
whynowy: Forgot to remove the comment.

pkg/isb/testutils/udf.go
```diff
@@ -78,3 +78,42 @@ func CopyUDFTestApplyStream(ctx context.Context, vertexName string, writeMessage
 	writeMessageCh <- isb.WriteMessage{Message: writeMessage}
 	return nil
 }
+
+func CopyUDFTestApplyBatchMap(ctx context.Context, vertexName string, readMessages []*isb.ReadMessage) ([]isb.ReadWriteMessagePair, error) {
+	udfResults := make([]isb.ReadWriteMessagePair, len(readMessages))
+
+	_ = ctx
+
```
yhl25, line 86: Can you remove this?

pkg/sdkclient/batchmapper/client.go
```diff
@@ -0,0 +1,175 @@
...
+// We spawn 2 goroutines here, one for sending the requests over the stream
+// and the other one for receiving the responses
+func (c *client) BatchMapFn(ctx context.Context, inputCh <-chan *batchmappb.BatchMapRequest) (<-chan *batchmappb.BatchMapResponse, <-chan error) {
+	// errCh is used to track and propagate any errors that might occur during the rpc lifecyle, these could include
+	// errors in sending, UDF errors etc
+	// These are propagated to the applier for further handling
+	errCh := make(chan error)
+
+	// response channel for streaming back the results received from the gRPC server
+	// TODO(map-batch): Should we keep try to keep this buffered?
```
yhl25, line 93: typo?

pkg/sdkclient/batchmapper/client.go
```diff
@@ -0,0 +1,175 @@
...
+
+	// Read from the read messages and send them individually to the bi-di stream for processing
+	// in case there is an error in sending, send it to the error channel for handling
+	go func() {
+		// TODO(map-batch): Should we check for ctx.Done here as well?
+		for inputMsg := range inputCh {
+			err = stream.Send(inputMsg)
+			if err != nil {
+				errCh <- sdkerr.ToUDFErr("c.grpcClt.BatchMapFn", err)
+				// TODO(map-batch): Check if we should still do a CloseSend on the stream even if we have
```
yhl25, line 157: let's check for `ctx.Done()`, to avoid sending requests when the `ctx` is cancelled.

pkg/sdkclient/batchmapper/client.go
```diff
@@ -0,0 +1,175 @@
...
+				// received an error. Ideally on an error we would be stopping any
+				// further processing and go for a replay so this should not be required.
+				// but would be good to verify.
+				// Otherwise this would be a break
+				return
```
yhl25, line 162: We don't need to again call `closeSend()`, if there are any errors `gRPC` stream will be closed. Please remove the TODO

pkg/sdkclient/batchmapper/client.go
```diff
@@ -0,0 +1,175 @@
...
+		index := 0
+		for {
+			select {
+			// handle a context done
+			case <-ctx.Done():
+				errCh <- ctx.Err()
+				return
+			default:
+				resp, recvErr = stream.Recv()
+				// check if this is EOF error, which indicates that no more responses left to process on the
```
yhl25, line 130: Do we need a `ctx.Done()` check here? Since we create the gRPC stream with `ctx`, it seems to me that gRPC should handle it by throwing an error in the Recv() method if the `ctx` is cancelled. If we don't have a `ctx` check while closing the stream, there is a possibility that we could invoke `closeSend` to signal EOF to the gRPC stream while the `ctx` is cancelled. In this case, gRPC might return the actual result, but since we don't `recv` when there is a `ctx.Done()` signal, we might end up not receiving the result. This whole context setup is quite confusing. My suggestion is to rely on gRPC to handle `ctx` since we pass it while creating the stream. I believe we should adopt this approach in the reduce and ud source as well.

pkg/sdkclient/batchmapper/client.go
yhl25: No, we can't close the error channel, because you are using select on the error channel and the response channel on the read side. So, if you close the error channel, it will get selected and there are chances that you never read from the response channel. So, in this case, we should set `errCh` to null. That way, the error channel will never get selected.

pkg/udf/forward/forward.go
```diff
@@ -45,6 +44,13 @@ import (
 	"github.com/numaproj/numaflow/pkg/watermark/wmb"
 )
 
+// MapAppliers consolidates the appliers for different map modes together
+type MapAppliers struct {
+	MapUDF       applier.MapApplier
+	MapStreamUDF applier.MapStreamApplier
+	BatchMapUDF  applier.BatchMapApplier
+}
+
```
yhl25, line 53: I am not fully in favor of this way of sending appliers to forward. This is because even if we need just `MapUDF`, we still have to set the other `UDFs`. So, I'm thinking, should we use an options approach for all appliers? This way, we just pass the option that we need.

test/sdks-e2e/testdata/flatmap-batch.yaml
```diff
@@ -0,0 +1,84 @@
...
+        http: {}
+    - name: go-split
+      metadata:
+        annotations:
+          numaflow.numaproj.io/batch-map: "true"
+      scale:
+        min: 1
+      udf:
+        container:
+          # Split input message into an array with comma, see https://github.com/numaproj/numaflow-go/tree/main/pkg/mapper/examples/flatmap
```
yhl25, line 18: let's have multiple partitions for testing.

pkg/udf/forward/forward.go
kohlisid: Right now, the other appliers would be nil as they are created under the flag. 
If we want to completely do away with the argument and use options I guess that is something we should keep consistent across all other forwards also then?

pkg/sdkclient/batchmapper/client.go
kohlisid: I have updated the implementation
1) On the sending side, check for context.Done, and return without stream.CloseSend
2) On the recv side, the grpc server should send a ctx error back to the client.

pkg/udf/map_udf.go
```diff
@@ -130,9 +132,16 @@ func (u *MapUDFProcessor) Start(ctx context.Context) error {
...
+
+	// We can have the vertex running only of the map modes
+	if enableMapUdfStream && enableBatchMapUdf {
```
whynowy, line 139: Improvement: Shall we use one `server-info` for all the mappers, and start the mode based what is written in the file?

pkg/udf/rpc/grpc_batch_map.go
```diff
@@ -0,0 +1,217 @@
...
+	log := logging.FromContext(ctx)
+	for {
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("failed on readiness check: %w", ctx.Err())
+		default:
+			if _, err := u.client.IsReady(ctx, &emptypb.Empty{}); err == nil {
+				return nil
+			} else {
+				log.Infof("waiting for map udf to be ready: %v", err)
```
whynowy, line 73: Nit - `Waiting`

pkg/udf/rpc/grpc_batch_map.go
```diff
@@ -0,0 +1,217 @@
...
+				WriteMessages: parsedResp,
+				Err:           nil,
+			}
+			udfResults = append(udfResults, responsePair)
+			trackerReq.removeRequest(msgId)
+		}
+	}
+	// check if there are elements left in the tracker. This cannot be an acceptable case as we want the
+	// UDF to send responses for all elements.
+	leftRequest := trackerReq.getItems()
```
whynowy, line 167: It looks like  a `isEmpty()` function is good enough, do not need to have a `getItems()`, right?

pkg/sdkclient/batchmapper/client.go
```diff
@@ -0,0 +1,180 @@
...
+
+	// BatchMapFn is a bidirectional streaming RPC
+	// We get a Map_BatchMapFnClient object over which we can send the requests,
+	// receive the responses asynchronously.
+	// TODO(map-batch): this creates a new gRPC stream for every batch,
+	// it might be useful to see the performance difference between this approach
+	// and a long-running RPC
+	stream, err := c.grpcClt.BatchMapFn(ctx)
+	if err != nil {
+		go func() {
```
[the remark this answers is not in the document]
kohlisid: We are trying to write to the channel, it will be a blocking call until we have a receiver end for the error channel.
We would be returning the error channel from this function itself. So as to not have a blocking situation in a case where the receiver has not started reading on the channel yet.

pkg/sdkclient/batchmapper/client.go
```diff
@@ -0,0 +1,180 @@
...
+			errCh <- sdkerr.ToUDFErr("c.grpcClt.BatchMapFn stream", err)
+		}()
+		return responseCh, errCh
```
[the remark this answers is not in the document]
kohlisid: If we see an error while establishing the rpc itself, then I felt starting the reading and writing go routines wouldn't be required. Hence, the early return in this case. 
Does something come to mind for the other way?

pkg/udf/map_udf.go
kohlisid: We can do that, Can take up in a followup PR?

pkg/udf/rpc/grpc_batch_map.go
kohlisid: Sure! I can update that.

pkg/udf/forward/applier/batchmap.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+// BatchMapApplier applies the batch map UDF on the read messages and gives back the responses for the whole
+// batch
+type BatchMapApplier interface {
+	ApplyBatchMap(ctx context.Context, messages []*isb.ReadMessage) ([]isb.ReadWriteMessagePair, error)
+}
+
+// BatchMapFunc utility function used to create a BatchMapApplier implementation
+type BatchMapFunc func(ctx context.Context, messages []*isb.ReadMessage) ([]isb.ReadWriteMessagePair, error)
```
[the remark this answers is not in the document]
kohlisid: Following the pattern from the other UDFs

pkg/udf/forward/forward.go
yhl25: other forwarders only have one UDF and other UDFs like fallback and transformer as passed as options.

kohlisid: Sure! If that seems more feasible we can go ahead with that approach. I can do that in a followup PR if that works?
Not to block this further

pkg/udf/forward/forward.go
```diff
@@ -399,12 +373,76 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
...
+			isdf.opts.logger.Errorw("batchMapUDF.Apply, Stop called during udf processing", zap.Error(err))
+			metrics.PlatformError.With(map[string]string{metrics.LabelVertex: isdf.vertexName, metrics.LabelPipeline: isdf.pipelineName, metrics.LabelVertexType: string(dfv1.VertexTypeMapUDF), metrics.LabelVertexReplicaIndex: strconv.Itoa(int(isdf.vertexReplica))}).Inc()
+		}
+		readOffsets := make([]isb.Offset, 0)
+		for _, msg := range dataMessages {
+			readOffsets = append(readOffsets, msg.ReadOffset)
+		}
+		isdf.fromBufferPartition.NoAck(ctx, readOffsets)
+		isdf.opts.logger.Panic("Got an error while processing Batch Map UDF", zap.Error(err))
+	}
```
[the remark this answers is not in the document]
kohlisid: I did think about that and left a TODO item there regarding the same. In some earlier discussions with @vigith we concluded that error scenarios might be rare and would be fine with restart for a clean setup again. 
Again, this is probably a behavior from our end that is open for discussion, I'm personally fine with setting up a retry mechanism if possible. 
I would let @whynowy @vigith weigh in if they have any opinions on this

vigith:
@kohlisid please create follow-up TODOs issues and track it under https://github.com/numaproj/numaflow/issues/1688?

kohlisid:
@vigith 
I have the following TODO's noted for now
1) Improvement: Shall we use one server-info for all the mappers, and start the mode based what is written in the file?
2) Use an options approach for all appliers

Could you go through the open comments once to see if there is something else that we need to take?

pkg/udf/rpc/tracker.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/numaproj/numaflow/pkg/isb"
+)
+
+type tracker struct {
+	lock sync.RWMutex
+	m    map[string]*isb.ReadMessage
+}
+
+func NewTracker() *tracker {
```
vigith, line 14: document this tracker

pkg/udf/rpc/grpc_batch_map.go
```diff
@@ -0,0 +1,217 @@
...
+				Err:           nil,
+			}
+			udfResults = append(udfResults, responsePair)
+			trackerReq.removeRequest(msgId)
+		}
+	}
+	// check if there are elements left in the tracker. This cannot be an acceptable case as we want the
+	// UDF to send responses for all elements.
+	trackerEmpty := trackerReq.isEmpty()
+	if !trackerEmpty {
```
vigith, line 168: merge these 2 lines

pkg/udf/rpc/grpc_batch_map.go
```diff
@@ -0,0 +1,217 @@
...
+		logger.Error("BatchMap response for all requests not received from UDF")
+		return nil, fmt.Errorf("batchMap response for all requests not received from UDF")
+	}
+	trackerReq.clear()
```
vigith, line 172: defer this?

pkg/udf/rpc/tracker.go
```diff
@@ -0,0 +1,63 @@
...
+
+func (t *tracker) delete(key string) {
+	t.lock.Lock()
+	defer t.lock.Unlock()
+	delete(t.m, key)
+}
+
+func (t *tracker) isEmpty() bool {
+	t.lock.Lock()
+	defer t.lock.Unlock()
```
vigith, line 54: use RLock

pkg/udf/forward/forward.go
kohlisid: https://github.com/numaproj/numaflow/issues/1809

pkg/udf/map_udf.go
kohlisid: https://github.com/numaproj/numaflow/issues/1810

pkg/udf/forward/forward.go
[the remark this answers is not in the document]
yhl25: I do not want to rely on no-ack that much, since we already have the batch why not just retry it like we do for sink and map?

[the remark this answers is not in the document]
vigith: @yhl25 #1812 to revisit this.

yhl25 (approved):
LGTM!

PR #1785, opened 2024-07-02
feat: crate for retry with backoff strategy

vigith:
crate to do retries with different backoff strategies as an iterator for any async function.

### DIFF
--- a/serving/.rustfmt.toml
+++ b/serving/.rustfmt.toml
@@ -1,2 +1,1 @@
 edition = "2021"
-indent_style = "Block"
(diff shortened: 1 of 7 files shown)


vigith:
This coverage report is wrong, it is checking doc comments and marking it as not covered even though I have doc tests.

```
   Doc-tests backoff

running 1 test
test backoff/src/lib.rs - (line 7) ... ok
```

yhl25 (commented):
Can we also update the redis store retry to use this?

vigith:
we can do that in a follow up PR

yhl25 (approved):
Nicely Done! LGTM!

PR #1793, opened 2024-07-04
test: reconfigure codecov settings

whynowy:
This solves the problem that codecov does not work for some scenarios.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -170,8 +170,8 @@ jobs:
        uses: codecov/codecov-action@v4
        with:
          files: ./serving/lcov.info
-        env:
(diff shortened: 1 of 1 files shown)


whynowy:
https://github.com/numaproj/numaflow/pull/1793

whynowy:
Follow up - The codecov report can be uploaded together so that it can generate correct report.

vigith (approved):

vigith:
> Follow up - The codecov report should be uploaded together so that it can generate correct report.

How can we do that? Run the tests together?

whynowy:
> > Follow up - The codecov report should be uploaded together so that it can generate correct report.
> 
> How can we do that? Run the tests together?

Run the tests in one job, and upload 2 files.

PR #1806, opened 2024-07-10
feat: Serving Source

### LINKED ISSUE #1813: controller to scaffold serving endpoint colocated with JetStream source

yhl25:
closes #1813


examples/15-serving-source-pipeline.yaml
```diff
@@ -0,0 +1,40 @@
+apiVersion: numaflow.numaproj.io/v1alpha1
+kind: Pipeline
+metadata:
+  name: simple-pipeline
+spec:
+  templates:
+    vertex:
+      metadata:
+        annotations:
+          numaflow.numaproj.io/callback: "true"
```
vigith, line 10: We will have to differentiate between a callback store and a serving output store.

pkg/apis/numaflow/v1alpha1/serving_source.go
```diff
@@ -0,0 +1,24 @@
...
+
+// Serving is the HTTP endpoint for Numaflow.
+type Serving struct {
+	// +optional
+	Auth *Authorization `json:"auth" protobuf:"bytes,1,opt,name=auth"`
+	// Whether to create a ClusterIP Service
+	// +optional
+	Service bool `json:"service" protobuf:"bytes,2,opt,name=service"`
+	// The header key from which the message id will be extracted
+	// Default
```
whynowy, line 11: Default?

pkg/apis/numaflow/v1alpha1/serving_source.go
```diff
@@ -0,0 +1,24 @@
...
+	// +optional
+	MsgIDHeaderKey string `json:"msgIDHeaderKey" protobuf:"bytes,3,opt,name=msgIDHeaderKey"`
```
whynowy, line 13: `*string`?

pkg/apis/numaflow/v1alpha1/serving_source.go
```diff
@@ -0,0 +1,24 @@
...
+	// Persistent store for the callbacks for serving and tracking
+	// +optional
+	Store *Store `json:"store" protobuf:"bytes,4,opt,name=store"`
+}
+
+// Store to track and store data and metadata for tracking and serving.
+type Store struct {
```
whynowy, line 20: `Store` is too generic.

pkg/apis/numaflow/v1alpha1/serving_source.go
```diff
@@ -0,0 +1,24 @@
+package v1alpha1
+
+// Serving is the HTTP endpoint for Numaflow.
+type Serving struct {
```
whynowy, line 4: `ServeringSource`

pkg/apis/numaflow/v1alpha1/vertex_types.go
```diff
@@ -236,6 +241,8 @@ func (v Vertex) GetPodSpec(req GetVertexPodSpecReq) (*corev1.PodSpec, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	containers[0].Ports = append(containers[0].Ports, corev1.ContainerPort{Name: VertexMetricsPortName, ContainerPort: VertexMetricsPort})
```
whynowy, line 245: there's already one in line 271.

whynowy (commented):
partial reviewed.

whynowy:
Which one is better?

1. `Serving` source orchestrated to be a `JetStream` source with a sidecar.
2. `Serving` source orchestrated to be a `JetStream` source  + a `Deployment`.

yhl25:
> Which one is better?
> 
> 1. `Serving` source orchestrated to be a `JetStream` source with a sidecar.
> 2. `Serving` source orchestrated to be a `JetStream` source  + a `Deployment`.

I prefer `Jetstream source + Deployment` because it's not tied to the source, but it will be tricky to manage the deployment, especially the scaling part and also we will have to make changes in the UI to monitor the deployment. For the initial phase, I think the sidecar approach should be good since we don't have to worry about the autoscaling part and also at this point serving is very much tied to the pipeline. But in the future, if we decide to support serving independently we can consider switching to a deployment.

whynowy:
> > Which one is better?
> > 
> > 1. `Serving` source orchestrated to be a `JetStream` source with a sidecar.
> > 2. `Serving` source orchestrated to be a `JetStream` source  + a `Deployment`.
> 
> I prefer `Jetstream source + Deployment` because it's not tied to the source, but it will be tricky to manage the deployment, especially the scaling part and also we will have to make changes in the UI to monitor the deployment. For the initial phase, I think the sidecar approach should be good since we don't have to worry about the autoscaling part and also at this point serving is very much tied to the pipeline. But in the future, if we decide to support serving independently we can consider switching to a deployment.

Scaling wouldn't be a problem, instead it would be better - independent scaling mechanism, but the UI is a problem. Anyways, go ahead with what you think is better for now.

yhl25:
> > > Which one is better?
> > > 
> > > 1. `Serving` source orchestrated to be a `JetStream` source with a sidecar.
> > > 2. `Serving` source orchestrated to be a `JetStream` source  + a `Deployment`.
> > 
> > 
> > I prefer `Jetstream source + Deployment` because it's not tied to the source, but it will be tricky to manage the deployment, especially the scaling part and also we will have to make changes in the UI to monitor the deployment. For the initial phase, I think the sidecar approach should be good since we don't have to worry about the autoscaling part and also at this point serving is very much tied to the pipeline. But in the future, if we decide to support serving independently we can consider switching to a deployment.
> 
> Scaling wouldn't be a problem, instead it would be better - independent scaling mechanism, but the UI is a problem. Anyways, go ahead with what you think is better for now.

That is true, but we should have a way to allow users to control the min and max and manage the deployment. Otherwise, the scaling could be done independently.

whynowy:
> > > > Which one is better?
> 
> > > > 
> 
> > > > 1. `Serving` source orchestrated to be a `JetStream` source with a sidecar.
> 
> > > > 2. `Serving` source orchestrated to be a `JetStream` source  + a `Deployment`.
> 
> > > 
> 
> > > 
> 
> > > I prefer `Jetstream source + Deployment` because it's not tied to the source, but it will be tricky to manage the deployment, especially the scaling part and also we will have to make changes in the UI to monitor the deployment. For the initial phase, I think the sidecar approach should be good since we don't have to worry about the autoscaling part and also at this point serving is very much tied to the pipeline. But in the future, if we decide to support serving independently we can consider switching to a deployment.
> 
> > 
> 
> > Scaling wouldn't be a problem, instead it would be better - independent scaling mechanism, but the UI is a problem. Anyways, go ahead with what you think is better for now.
> 
> 
> 
> That is true, but we should have a way to allow users to control the min and max and manage the deployment. Otherwise, the scaling could be done independently.

This is an http service, so HPA....

yhl25:
> > > > > Which one is better?
> > 
> > 
> > > > >
> > 
> > 
> > > > > 1. `Serving` source orchestrated to be a `JetStream` source with a sidecar.
> > 
> > 
> > > > > 2. `Serving` source orchestrated to be a `JetStream` source  + a `Deployment`.
> > 
> > 
> > > >
> > 
> > 
> > > >
> > 
> > 
> > > > I prefer `Jetstream source + Deployment` because it's not tied to the source, but it will be tricky to manage the deployment, especially the scaling part and also we will have to make changes in the UI to monitor the deployment. For the initial phase, I think the sidecar approach should be good since we don't have to worry about the autoscaling part and also at this point serving is very much tied to the pipeline. But in the future, if we decide to support serving independently we can consider switching to a deployment.
> > 
> > 
> > > 
> > 
> > 
> > > Scaling wouldn't be a problem, instead it would be better - independent scaling mechanism, but the UI is a problem. Anyways, go ahead with what you think is better for now.
> > 
> > 
> > That is true, but we should have a way to allow users to control the min and max and manage the deployment. Otherwise, the scaling could be done independently.
> 
> This is an http service, so HPA....

yes, we should allow users to control the min and max right? Since it is a separate component we can't put it inside the pipeline spec WDYT?

pkg/reconciler/pipeline/controller.go
```diff
@@ -294,7 +294,9 @@ func (r *pipelineReconciler) reconcileNonLifecycleChanges(ctx context.Context, p
 			bks = append(bks, k)
 		}
 		args := []string{fmt.Sprintf("--buffers=%s", strings.Join(bfs, ",")), fmt.Sprintf("--buckets=%s", strings.Join(bks, ","))}
+		// Why do we always create side input store?
```
whynowy, line 297: Remove it?

We always create to make the logic easier.

pkg/isbsvc/jetstream_service.go
```diff
@@ -262,6 +293,11 @@ func (jss *jetStreamSvc) ValidateBuffersAndBuckets(ctx context.Context, buffers,
 			return fmt.Errorf("failed to query side inputs store KV %q, %w", sideInputsKVName, err)
 		}
 	}
+	if servingSourceStream != "" {
+		if _, err := js.StreamInfo(servingSourceStream); err != nil {
```
whynowy, line 297: Just confirm - we are not doing validation in any place at this moment, right? Could be added to the vertex init container when it's a serving source.

pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -205,6 +205,10 @@ func (p Pipeline) GetSideInputsStoreName() string {
 	return fmt.Sprintf("%s-%s", p.Namespace, p.Name)
 }
 
+func (p Pipeline) GetServingSourceStreamName() string {
```
whynowy, line 208: We will not allow multiple Serving sources in one pipeline, correct?

vigith: there is nothing stopping us from doing so.

pkg/isbsvc/jetstream_service.go
```diff
@@ -103,6 +103,24 @@ func (jss *jetStreamSvc) CreateBuffersAndBuckets(ctx context.Context, buffers, b
...
 		}
 	}
+
+	if servingSourceStream != "" {
+		_, err := js.StreamInfo(servingSourceStream)
+		if err != nil {
+			if !errors.Is(err, nats.ErrStreamNotFound) {
+				return fmt.Errorf("failed to query information of stream %q during buffer creating, %w", servingSourceStream, err)
+			}
+			if _, err := js.AddStream(&nats.StreamConfig{
```
vigith, line 113: use explicit config.

pkg/isbsvc/jetstream_service.go
vigith: let's list down all the validations we have to do.

serving/src/app/callback/store.rs
```diff
@@ -31,4 +31,5 @@ pub(crate) trait LocalStore {
         id: &str,
     ) -> Result<Vec<Arc<CallbackRequest>>, crate::Error>;
     async fn retrieve_datum(&mut self, id: &str) -> Result<Vec<Vec<u8>>, crate::Error>;
+    async fn is_available(&mut self) -> bool;
```
vigith, line 34:
```suggestion
    async fn ready(&mut self) -> bool;
```

`is` and `mut` doesn't sound right :)

pkg/apis/numaflow/v1alpha1/pipeline_types.go
whynowy: Then the stream name should not be tied to pipeline name only.

vigith (approved):

pkg/apis/numaflow/v1alpha1/pipeline_types.go
yhl25: yes, I will tackle it in a separate PR. It needs more changes.

whynowy (approved):

PR #1816, opened 2024-07-11
chore: update map appliers to use options

### LINKED ISSUE #1809: Refactor appliers to use option when there are multiple UDFs
I am not fully in favor of this way of sending appliers to forward. This is because even if we need just `MapUDF`, we still have to set the other `UDFs`. So, I'm thinking, should we use an options approach for all appliers? This way, we just pass the option that we need.

_Originally posted by @yhl25 in https://github.com/numaproj/numaflow/pull/1778#discussion_r1665118156_

kohlisid:
fixes #1809


pkg/udf/forward/forward.go
```diff
@@ -78,14 +69,20 @@ type InterStepDataForward struct {
...
 
 	options := DefaultOptions()
 	for _, o := range opts {
 		if err := o(options); err != nil {
 			return nil, err
 		}
 	}
+
+	// verify if initialized with a valid map mode
+	if !isValidMapMode(options) {
```
KeranYang, line 82: Can we move this validation to the options.go itself? When we set MapMode, we directly set the value and then call this function before returning the options. If we found multiple modes being set, return error. This is more modular.

pkg/udf/forward/forward.go
```diff
@@ -113,7 +109,8 @@ func NewInterStepDataForward(vertexInstance *dfv1.VertexInstance, fromStep isb.B
 	// Add logger from parent ctx to child context.
 	isdf.ctx = logging.WithLogger(ctx, options.logger)
 
-	if isdf.opts.enableMapUdfStream && isdf.opts.readBatchSize != 1 {
+	// if map stream mode is enabled and batch size !=1
```
KeranYang, line 112: Please remove this comment, it's an exact repeat of the code below.

pkg/udf/forward/forward.go
```diff
@@ -753,3 +750,24 @@ func errorArrayToMap(errs []error) map[string]int64 {
...
+
+// check if the options provided are for a valid map mode
+// exactly one of the appliers should not be nil as only one mode can be active at a time, not more not less only 1
+// the way our map mode options work is that whenever one of the applier is set, it sets the others appliers to nil.
+// There is a case where none of them is set which cannot be allowed
+func isValidMapMode(opts *options) bool {
+	// if all the appliers are set, then it is an invalid scenario
+	if (opts.batchMapUdfApplier != nil) && (opts.unaryMapUdfApplier != nil) && (opts.streamMapUdfApplier != nil) {
+		return false
+	}
```
KeranYang, line 762: I think this case is covered by the if statement below, can we remove it?

pkg/udf/forward/forward.go
```diff
@@ -753,3 +750,24 @@ func errorArrayToMap(errs []error) map[string]int64 {
...
+
+	// One and one only of the appliers should be set, It also verifies that all of them cannot be empty
+	// equivalent to (A xor B xor C)
+	if (opts.batchMapUdfApplier != nil) != (opts.unaryMapUdfApplier != nil) != (opts.streamMapUdfApplier != nil) {
+		return true
+	} else {
+		//  invalid case
+		return false
+	}
+
```
KeranYang, line 772:
```suggestion
return (opts.batchMapUdfApplier != nil) != (opts.unaryMapUdfApplier != nil) != (opts.streamMapUdfApplier != nil)
```

pkg/udf/forward/options.go
```diff
@@ -88,26 +93,51 @@ func WithLogger(l *zap.SugaredLogger) Option {
...
+// WithUDFBatchMap enables the batch map for UDF if provided with a non-nil applier
+func WithUDFBatchMap(f applier.BatchMapApplier) Option {
 	return func(o *options) error {
-		o.cbPublisher = cp
+		if f != nil {
+			o.batchMapUdfApplier = f
+			o.unaryMapUdfApplier = nil
+			o.streamMapUdfApplier = nil
+		}
+		return nil
```
KeranYang, line 117: I suggest putting isValidMode check here. Same for below

pkg/udf/forward/options.go
```diff
@@ -36,23 +37,27 @@ type options struct {
...
 	// cbPublisher is the callback publisher for the vertex.
 	cbPublisher *callback.Uploader
-	// enableBatchMapUdf indicates whether the batch map mode is enabled for the map UDF
-	enableBatchMapUdf bool
+	// unaryMapUdfApplier is the UDF applier for a unary map mode
+	unaryMapUdfApplier applier.MapApplier
+	// streamMapUdfApplier is the UDF applier for a server streaming map mode
+	streamMapUdfApplier applier.MapStreamApplier
+	// batchMapUdfApplier is the UDF applier for a batch map mode
+	batchMapUdfApplier applier.BatchMapApplier
```
KeranYang, line 47: Now that we have appliers in options for both source transformer and map udf, should we apply the same for reduce?

KeranYang (changes requested):

pkg/udf/forward/forward.go
kohlisid: The corner case that occurs with moving this to the function like `WithUDFBatchMap` is that if we pass a value which is nil for the applier then this check will give as a invalid case. 
But some other option might be setting a valid applier later on, so there would be a mismatch.

```
Options [WithUDFBatchMap(nil), WithUDFUnaryMap(applier)]

So in such a case while processing the options, WithUDFBatchMap will error out if we add a check because all appliers will be empty. Even though WithUDFUnaryMap will be setting a correct applier in a later option.
```

Hence, wanted to keep the check only after all options have been set.

pkg/udf/forward/forward.go
kohlisid: The below check is a like a 3 input XOR gate 
`a XOR b XOR c `
When all of them are true -> the output will still be true, hence we need this extra check here

Example is here
https://go.dev/play/p/ha6pnfsq79u

pkg/udf/forward/options.go
kohlisid: We can do that for consistency!

pkg/udf/forward/forward.go
KeranYang: Can we disallow passing a value which is nil for the applier? If an applier is nil, we don't call With***

pkg/udf/forward/forward.go
KeranYang: Ah, it took me quite a while to figure this out and I was still wrong...

nit: in this case, maybe using a single counter to check only one is non-nil is more readable. Feel free to disagree.

pkg/udf/forward/options.go
KeranYang: Do we plan to do that as part of this PR or separately?

kohlisid: Separate PR is better for that

KeranYang (approved):

whynowy (approved):

PR #1823, opened 2024-07-12
docs: conditional forwarding from source is allowed

KeranYang:
Explain what this PR does.


pkg/apis/numaflow/v1alpha1/edge_types.go
```diff
@@ -21,7 +21,7 @@ import "fmt"
 type Edge struct {
 	From string `json:"from" protobuf:"bytes,1,opt,name=from"`
 	To   string `json:"to" protobuf:"bytes,2,opt,name=to"`
-	// Conditional forwarding, only allowed when "From" is a Sink or UDF.
+	// Conditional forwarding, only allowed when "From" is a Source or UDF.
```
whynowy, line 24: hmmmm. the `from` can only be a `Source` or `UDF`, right? do we have that validation?

whynowy: oh... this is a doc fix.

whynowy (approved):

PR #1826, opened 2024-07-15
feat: enable restful daemon client option for UX server

whynowy:
Added an option to use RESTful client to access daemon service.

```
  ### The protocol used to connect to the Pipeline daemon service from Numaflow UX server.
  #   Could be either 'grpc' or 'http', defaults to 'grpc'.
  # 
  # server.daemon.client.protocol: grpc
```

This is useful when there's a breaking change on daemon service protobuf, while the pipeline can not run the same version of code as the UX server (e.g. UX server is installed in cluster scope, while controllers are in namespace scope, UX server and controllers do not follow the same release cadence). In this case, using RESTful client will have better UX experience, which will not display errors.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -81,7 +81,7 @@ jobs:
       - name: Set up Go 1.x
         uses: actions/setup-go@v5
         with:
-          go-version: '1.21'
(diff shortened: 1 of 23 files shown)


vigith (approved):
nice tests generation :)

PR #1839, opened 2024-07-16
chore: Exposing pipeline processing lag metrics

samhith-kakarla:


pkg/daemon/server/daemon_server.go
```diff
@@ -180,6 +191,76 @@ func (ds *daemonServer) newGRPCServer(
...
+		case <-ticker.C:
+			// get general pipeline info
+			log := logging.FromContext(ctx)
+			pl := ds.pipeline
+
+			pl.Kind = v1alpha1.PipelineGroupVersionKind.Kind
+			pl.APIVersion = v1alpha1.SchemeGroupVersion.String()
+
+			// get pipeline source and sink vertex
+			var (
```
whynowy, line 209: This can be done before the for loop starts.

pkg/daemon/server/daemon_server.go
```diff
@@ -180,6 +191,76 @@ func (ds *daemonServer) newGRPCServer(
...
+				if vertex.IsASource() {
+					source[vertex.Name] = true
+				} else if vertex.IsASink() {
+					sink[vertex.Name] = true
+				}
+			}
+
+			resp, err := ds.metaDataQuery.GetPipelineWatermarks(ctx, &daemon.GetPipelineWatermarksRequest{Pipeline: ds.pipeline.Name})
+			if err != nil {
+				log.Errorw(" failed to calculate lag for pipeline", zap.Error(err))
```
whynowy, line 223: When there's an error, we should skip this reset.

pkg/daemon/server/daemon_server.go
```diff
@@ -47,6 +51,13 @@ import (
 	"github.com/numaproj/numaflow/pkg/watermark/fetch"
 )
 
+var (
+	lag = promauto.NewGauge(prometheus.GaugeOpts{
```
whynowy, line 55: `pipelineProcessingLag`

pkg/daemon/server/daemon_server.go
```diff
@@ -47,6 +51,14 @@ import (
 	"github.com/numaproj/numaflow/pkg/watermark/fetch"
 )
 
+var (
```
whynowy, line 54: It would be better to move all the metrics related items to a new file `metrics.go`.

pkg/daemon/server/metrics.go
```diff
@@ -0,0 +1,102 @@
...
+		}
+	}
+
+	for {
+		select {
+		case <-ticker.C:
+
+			resp, err := ds.metaDataQuery.GetPipelineWatermarks(ctx, &daemon.GetPipelineWatermarksRequest{Pipeline: ds.pipeline.Name})
+			if err != nil {
+				log.Errorw(" failed to calculate lag for pipeline", zap.Error(err))
```
whynowy, line 62:
```suggestion
				log.Errorw("Failed to calculate processing lag for pipeline", zap.Error(err))
```

pkg/daemon/server/metrics.go
```diff
@@ -0,0 +1,102 @@
...
+						}
+					}
+				}
+			}
+			// if the data hasn't arrived the sink vertex
+			// set the lag to be -1
+			if minWM == -1 {
+				pipelineProcessingLag.Set(-1)
+			} else {
+				pipelineProcessingLag.Set(float64(maxWM - minWM))
```
whynowy, line 96: Do we need to check if it's <0?

pkg/daemon/server/metrics.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"go.uber.org/zap"
+)
+
+var (
+	pipelineProcessingLag = promauto.NewGauge(prometheus.GaugeOpts{
+		Name:      "pipeline_lag_milliseconds",
+		Help:      "pipeline processing lag metrics in milliseconds.",
+		Subsystem: "daemon",
```
whynowy, line 34: Also missing `pipeline` label.

pkg/daemon/server/metrics.go
```diff
@@ -0,0 +1,174 @@
...
+
+var (
+	currentTimeToWatermark = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name:      "current_time_compared_to_watermark_milliseconds",
+		Help:      "Watermark compared with current time in milliseconds.",
+		Subsystem: "daemon",
+	}, []string{metrics.LabelPipeline})
+)
+
+func (ds *daemonServer) exposeCTMetrics(ctx context.Context) {
```
whynowy, line 48: We can do it in one routine, right?

pkg/daemon/server/metrics.go
```diff
@@ -0,0 +1,174 @@
...
+	pipelineProcessingLag = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name:      "pipeline_lag_milliseconds",
+		Help:      "pipeline processing lag metrics in milliseconds.",
+		Subsystem: "daemon",
+	}, []string{metrics.LabelPipeline})
+)
+
+var (
+	currentTimeToWatermark = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name:      "current_time_compared_to_watermark_milliseconds",
```
whynowy, line 42: We need a better metrics name...

whynowy: @yhl25 - this is the metric you want, right?

yhl25: yes

pkg/daemon/server/metrics.go
yhl25, line 48: please add a comment about what this function does.

yhl25: +1

pkg/daemon/server/metrics.go
```diff
@@ -0,0 +1,125 @@
...
+			} else {
+				if maxWM < minWM {
+					pipelineProcessingLag.WithLabelValues(ds.pipeline.Name).Set(-1)
+				} else {
+					pipelineProcessingLag.WithLabelValues(ds.pipeline.Name).Set(float64(maxWM - minWM))
+				}
+			}
+
+			if maxWM == math.MinInt64 {
+				watermarkCmpNow.WithLabelValues(ds.pipeline.Name).Set(0)
```
whynowy, line 115: Don't set it to 0, 0 means there's no difference. Let's do -1.

samhith-kakarla: ok

whynowy (approved):

PR #1845, opened 2024-07-17
chore: Configure TTL for Serving Source Store

yhl25:
fixes:
  * TTL support for serving store
  * update tonic to v0.12 in servesink


serving/src/app/callback/store/redisstore.rs
```diff
@@ -70,8 +71,16 @@ impl RedisConnection {
...
-            .await
-            .map(|_| ())
+            .await?;
+
+        // if the ttl is configured, set the key to expire after the specified time
+        if let Some(ttl) = config().redis.ttl_secs {
+            conn_manager
+                .send_packed_command(redis::cmd(EXPIRE).arg(key).arg(ttl))
+                .await?;
+        }
```
vigith, line 81: use pipelining? meaning, send both the commands together

serving/src/app/callback/store/redisstore.rs
```diff
@@ -264,4 +271,47 @@ mod tests {
...
+        )
+        .await
+        .expect("Failed to write to Redis");
+
+        // Immediately check existence
+        let exists: bool = conn_manager
+            .exists(&key)
+            .await
+            .expect("Failed to check existence immediately");
+        assert!(exists, "Key should exist immediately after saving");
```
vigith, line 305: use `TTL` command to see whether the value has been set. do not sleep. check for existence or TTL being set.

vigith (approved):

PR #1848, opened 2024-07-18
chore: add retry for batchmap UDF on error

### LINKED ISSUE #1812: Does panic on error affect the user experience
i am okay with a panic since UDF errors are rare, perhaps we should revisit this if this is affecting the user experience?

_Originally posted by @vigith in https://github.com/numaproj/numaflow/pull/1778#discussion_r1673212015_

kohlisid:
Fixes #1812 
fixes https://github.com/numaproj/numaflow/issues/1820


kohlisid:
@numaproj/numaflow-dev Opinions on having infinite retry vs max retry (and no ack after that)?

KeranYang:
> @numaproj/numaflow-dev Opinions on having infinite retry vs max retry (and no ack after that)?

@kohlisid , which one is consistent with unary/streaming map?

kohlisid:
@KeranYang, the unary-map will go for an infinite retry until shutdown
```
func (isdf *InterStepDataForward) applyUDF(ctx context.Context, readMessage *isb.ReadMessage) ([]*isb.WriteMessage, error) {
	for {
		writeMessages, err := isdf.opts.unaryMapUdfApplier.ApplyMap(ctx, readMessage)
		if err != nil {
			isdf.opts.logger.Errorw("mapUDF.Apply error", zap.Error(err))
			// TODO: implement retry with backoff etc.
			time.Sleep(isdf.opts.retryInterval)
			// keep retrying, I cannot think of a use case where a user could say, errors are fine :-)
			// as a platform we should not lose or corrupt data.
			// this does not mean we should prohibit this from a shutdown.
			if ok, _ := isdf.IsShuttingDown(); ok {
				isdf.opts.logger.Errorw("mapUDF.Apply, Stop called while stuck on an internal error", zap.Error(err))
				metrics.PlatformError.With(map[string]string{metrics.LabelVertex: isdf.vertexName, metrics.LabelPipeline: isdf.pipelineName, metrics.LabelVertexType: string(dfv1.VertexTypeMapUDF), metrics.LabelVertexReplicaIndex: strconv.Itoa(int(isdf.vertexReplica))}).Inc()
				return nil, err
			}
			continue
		}
		return writeMessages, nil
	}
}
```

KeranYang:
@kohlisid Cool, I prefer consistency so infinite retry until shutdown.

vigith:
> @kohlisid Cool, I prefer consistency so infinite retry until shutdown.

+1

pkg/udf/forward/forward.go
```diff
@@ -409,32 +409,58 @@ func (isdf *InterStepDataForward) processConcurrentMap(ctx context.Context, data
...
 		}
-		readOffsets := make([]isb.Offset, 0)
-		for _, msg := range dataMessages {
-			readOffsets = append(readOffsets, msg.ReadOffset)
+		// if no error, let's mark this as a success and break
+		success = true
+		break
+	}
+	if !success {
+		udfErr := fmt.Errorf("batchMapUDF.Apply got error during batch udf processing")
```
KeranYang, line 449: The error message is too generic, making it hard for debugging. Can we create and return the `udfResultsErr` object at each break statement and add more descriptions to the error message? That way we can also remove the success flag, making the code more readable.

pkg/udf/forward/forward.go
```diff
@@ -409,32 +409,58 @@ func (isdf *InterStepDataForward) processConcurrentMap(ctx context.Context, data
...
+					metrics.LabelPipeline: isdf.pipelineName, metrics.LabelVertexType: string(dfv1.VertexTypeMapUDF),
+					metrics.LabelVertexReplicaIndex: strconv.Itoa(int(isdf.vertexReplica))}).Inc()
+				break udfLoop
+			}
+			isdf.opts.logger.Errorw("batchMapUDF.Apply got error during batch udf processing", zap.Error(err))
+			select {
+			case <-ctx.Done():
+				// no point in retrying if the context is cancelled
+				break udfLoop
+			case <-time.After(1 * time.Second):
```
KeranYang, line 439: use `isdf.opts.retryInterval`?

pkg/udf/forward/forward.go
```diff
@@ -409,32 +409,58 @@ func (isdf *InterStepDataForward) processConcurrentMap(ctx context.Context, data
...
+			},
 		}
-		isdf.fromBufferPartition.NoAck(ctx, readOffsets)
-		isdf.opts.logger.Panic("Got an error while processing Batch Map UDF", zap.Error(err))
+		return udfResultsErr
 	}
 	isdf.opts.logger.Debugw("batch map applyUDF completed", zap.Int("concurrency", isdf.opts.udfConcurrency), zap.Duration("took", time.Since(concurrentUDFProcessingStart)))
 	metrics.ConcurrentUDFProcessingTime.With(map[string]string{metrics.LabelVertex: isdf.vertexName, metrics.LabelPipeline: isdf.pipelineName, metrics.LabelVertexType: string(dfv1.VertexTypeMapUDF), metrics.LabelVertexReplicaIndex: strconv.Itoa(int(isdf.vertexReplica))}).Observe(float64(time.Since(concurrentUDFProcessingStart).Microseconds()))
 	return udfResults
+
```
KeranYang, line 463:
```suggestion
```

KeranYang (approved):

whynowy (approved):

docs/user-guide/user-defined-functions/map/map.md
```diff
@@ -51,6 +51,29 @@ Check the links below to see the UDF examples in streaming mode for different la
 - [Golang](https://github.com/numaproj/numaflow-go/tree/main/pkg/mapstreamer/examples/flatmap_stream/)
 - [Java](https://github.com/numaproj/numaflow-java/tree/main/examples/src/main/java/io/numaproj/numaflow/examples/mapstream/flatmapstream/)
 
+### Batch Map Mode
+
+BatchMap is an interface that allows developers to process multiple data items in a UDF single call,
+rather than each item in separate calls.
+
+The BatchMap interface can be helpful in scenarios where performing operations on a group of data can be more efficient.
```
vigith, line 59: please add that the batch size might be smaller (as small as one element) and the size is best effort..

PR #1855, opened 2024-07-20
feat: add controller and pipeline info metrics

whynowy:
1. Added controller build info metrics
2. Added pipeline build info metrics
3. Moved `webhook` package to `./pkg/`


pkg/daemon/server/daemon_server.go
```diff
@@ -227,6 +236,77 @@ func (ds *daemonServer) newHTTPServer(ctx context.Context, port int, tlsConfig *
 		log.Info("Not enabling pprof debug endpoints")
 	}
 
-	go ds.exposeMetrics(ctx)
 	return &httpServer
 }
+
+// calculate processing lag and watermark_delay to current time using watermark values.
+func (ds *daemonServer) exposeMetrics(ctx context.Context) {
```
vigith, line 243: are we not emitting watermarks for the rest of the vertices?

vigith: I see that we only have for sources and sinks, did i get that wrong?

whynowy: We emits the diff of the source and sink, which is the lag we show on the ui, also emits source diff from now.

whynowy: Maybe we could emit watermark for each vertex directly...

vigith (approved):

Issue #1774, opened 2024-06-26
Status of Numaflow resource should track the status of underlying child resource


xdevxy:
# Summary

To truly reflect the status of a Numaflow resource, e.g. ISBService, the underlying status of children `StatefuleSet`.

# Use Cases

User, e.g. numaplane can rely solely on the status of Numaflow resource to truly reflect the overall status including the children.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

juliev0:
It looks like ISBService does in fact have a `phase`. Is it not being reset to `Failed` in the case that the underlying StatefulSet has failed?

juliev0:
This is probably a more generic issue that we need to check if both ISBService and Pipeline will be reset to Failed phase if they become unhealthy and not just at the point that they're being deployed initially (i.e. for Pipeline is there a Vertex in a Failed state? (which requires that Vertex phase take into account the phase of the underlying Pods))

cc @whynowy

whynowy:
I would recommend to check how `StatefulSet` and `Deployment` do. 

`StatefulSet` -> `Pod`
`Deployment` -> `ReplicaSet` -> `Pod`

For us:

`InterStepBufferService` -> `StatefulSet`
`Pipeline` -> `Vertex` -> `Pod`

PR #1860, opened 2024-07-22
feat: Sync the health status of  ISBService, pipeline and vertex

chandankumar4:
Explain what this PR does.

Add the feature to sync the numaflow child resource status to their ownerreference.
for example the ownerreference is currently is listed below
- ISBService -> StatefulSets -> Pod
- Pipeline -> Deployment -> Replicaset -> Pod (daemon/Sideinput)
- Vertex -> Pods

Design Docs: https://docs.google.com/document/d/1VCHA-tHjUQsC_9G-gNLRnR4_uQ0q7p07ccGdbbfnLPA

Fixes: #1774


pkg/reconciler/isbsvc/watcher.go
```diff
@@ -0,0 +1,106 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
+	dfv1 "github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
+)
+
+// StatefulSetStatusHandler updates the InterStepBufferService status based on the StatefulSet status
+func StatefulSetStatusHandler(mgr manager.Manager, logger *zap.SugaredLogger) handler.EventHandler {
```
whynowy, line 19: Instead of doing the operations in the `EventHandler`, we should handle them in `Reconcile()`. 

The purpose of interface `EventHandler` is to `enqueue`.

```
// EventHandler enqueues reconcile.Requests in response to events (e.g. Pod Create).  EventHandlers map an Event
// for one object to trigger Reconciles for either the same object or different objects - e.g. if there is an
// Event for object with type Foo (using source.KindSource) then reconcile one or more object(s) with type Bar.
```

pkg/reconciler/cmd/start.go
```diff
@@ -182,18 +181,7 @@ func Start(namespaced bool, managedNamespace string) {
...
-			predicate.Funcs{
-				UpdateFunc: func(e event.UpdateEvent) bool {
-					if e.ObjectOld == nil || e.ObjectNew == nil {
-						return true
-					}
-					old, _ := e.ObjectOld.(*dfv1.Vertex)
-					new, _ := e.ObjectNew.(*dfv1.Vertex)
-					return !reflect.DeepEqual(new.Spec.WithOutReplicas(), old.Spec.WithOutReplicas())
-				}},
-		)); err != nil {
```
chandankumar4, line 196: Have to watch for replica update event as well for calculating the status of linked pod

pkg/reconciler/isbsvc/watcher.go
chandankumar4: Made the changes accordingly

pkg/reconciler/pipeline/watcher.go
```diff
@@ -0,0 +1,135 @@
...
+	"context"
+	"fmt"
+
+	appv1 "k8s.io/api/apps/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	dfv1 "github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
+)
+
+func setResourceStatus(c client.Client, pipeline *dfv1.Pipeline) error {
```
whynowy, line 13: Can we move it to `pipelineReconciler` as a function `checkChildrenResourceStatus()`?

pkg/reconciler/pipeline/watcher.go
```diff
@@ -0,0 +1,135 @@
...
+	if err := c.Get(context.TODO(), client.ObjectKey{
+		Namespace: pipeline.GetNamespace(),
+		Name:      pipeline.GetDaemonDeploymentName(),
+	}, &daemonDeployment); err != nil {
+		return err
+	}
+	if msg, serviceType, reason, status := getDeploymentStatus(&daemonDeployment); status {
+		pipeline.Status.MarkServiceHealthy(serviceType, reason, msg)
+	} else {
+		pipeline.Status.MarkServiceNotHealthy(serviceType, reason, msg)
```
whynowy, line 25: If it's not healthy, return and requeue.

pkg/reconciler/pipeline/watcher.go
```diff
@@ -0,0 +1,135 @@
...
+			Namespace: pipeline.GetNamespace(),
+			Name:      pipeline.GetSideInputsManagerDeploymentName(sideInput.Name),
+		}, &sideInputDeployment); err != nil {
+			return err
+		}
+
+		if msg, serviceType, reason, status := getDeploymentStatus(&sideInputDeployment); status {
+			pipeline.Status.MarkServiceHealthy(serviceType, reason, msg)
+		} else {
+			pipeline.Status.MarkServiceNotHealthy(serviceType, reason, msg)
```
whynowy, line 41: If it's not healthy, return a requeue.

pkg/reconciler/pipeline/watcher.go
```diff
@@ -0,0 +1,135 @@
...
+		return err
+	}
+	if msg, serviceType, reason, status := getDeploymentStatus(&daemonDeployment); status {
+		pipeline.Status.MarkServiceHealthy(serviceType, reason, msg)
+	} else {
+		pipeline.Status.MarkServiceNotHealthy(serviceType, reason, msg)
+	}
+
+	// get the side input deployments and update the status of them to the pipeline
+	for _, sideInput := range pipeline.Spec.SideInputs {
```
whynowy, line 29: If `len()==0`, mark successful directly.

pkg/reconciler/pipeline/watcher.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	return nil
+}
+
+// getVertexStatus will calculate the status of the vertices and return the status and reason
+func getVertexStatus(c client.Client, pipeline *dfv1.Pipeline) (string, string, error) {
+	totalVertices := len(pipeline.Spec.Vertices)
+	healthyVertices := 0
+	for _, pipelineVertex := range pipeline.Spec.Vertices {
```
whynowy, line 65: Instead of querying one by one, use list call with label selector.

pkg/reconciler/pipeline/controller.go
```diff
@@ -90,6 +90,9 @@ func (r *pipelineReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 			return result, err
 		}
 	}
+	if err := setResourceStatus(r.client, plCopy); err != nil {
```
whynowy, line 93: Shall this go to `reconcileNonLifecycleChanges()`, after `pl.Status.SetPhase(pl.Spec.Lifecycle.GetDesiredPhase(), "")`?

pkg/reconciler/isbsvc/controller.go
```diff
@@ -75,6 +75,9 @@ func (r *interStepBufferServiceReconciler) Reconcile(ctx context.Context, req ct
 			return reconcile.Result{}, err
 		}
 	}
+	if err := setStatefulSetStatus(r.client, isbSvcCopy); err != nil {
```
whynowy, line 78: Shall this go to `installer.go`, under `Install()`, after `isbSvc.Status.Config = *bufferConfig`?

whynowy: Not all the `InterStepBufferService` implementations have `StatefulSet`, shall we add a function to the interface `Installer` to check the children resource status, and call it in the `Install()` function?

pkg/apis/numaflow/v1alpha1/isbsvc_types.go
```diff
@@ -51,6 +35,9 @@ const (
 	// ISBSvcConditionDeployed has the status True when the InterStepBufferService
 	// has its RestfulSet/Deployment as well as services created.
 	ISBSvcConditionDeployed ConditionType = "Deployed"
+
+	// ISBSvcConditionStatefulSetResourceHealthy has the status True when the child resources is healthy.
+	ISBSvcConditionStatefulSetResourceHealthy ConditionType = "StatefulSetHealthy"
```
whynowy, line 40: Maybe `ChildrenResourcesHealthy`? Considering not all of the implementations have children resources.

pkg/reconciler/vertex/controller.go
```diff
@@ -77,6 +77,9 @@ func (r *vertexReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		log.Errorw("Reconcile error", zap.Error(err))
 	}
 
+	if err = setPodStatus(r.client, vertexCopy); err != nil {
```
whynowy, line 80: Similarly, move it to `reconcile()`, after `vertex.Status.MarkPhaseRunning()`.

pkg/reconciler/vertex/watcher.go
```diff
@@ -0,0 +1,48 @@
...
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	dfv1 "github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
+)
+
+func setPodStatus(c client.Client, vertex *dfv1.Vertex) error {
```
whynowy, line 13: Move it to `vertexReconciler`.

pkg/reconciler/vertex/watcher.go
```diff
@@ -0,0 +1,48 @@
...
+		client.InNamespace(vertex.GetNamespace()),
+		client.MatchingLabels{dfv1.KeyAppName: vertex.GetName()},
+	); err != nil {
+		return err
+	}
+
+	if msg, reason, status := getVertexStatus(vertex, &podList); status {
+		vertex.Status.MarkPodHealthy(reason, msg)
+	} else {
+		vertex.Status.MarkPodNotHealthy(reason, msg)
```
whynowy, line 26: Not healthy, return and requeue.

pkg/apis/numaflow/v1alpha1/isbsvc_types.go
whynowy: All the new added conditions need to be added to the initialization.

pkg/reconciler/pipeline/watcher.go
chandankumar4: Just thinking, do we need to requeue in case of not healthy, as this fn will be going to triggered everytime if there is any change in status of daemon deployment due to `predicate.ResourceVersionChangedPredicate{}` ?

pkg/reconciler/pipeline/controller.go
chandankumar4: yes, I can be, I'll move it

pkg/reconciler/pipeline/watcher.go
whynowy: Makes sense.

pkg/reconciler/pipeline/controller.go
```diff
@@ -935,3 +938,54 @@ func (r *pipelineReconciler) safeToDelete(ctx context.Context, pl *dfv1.Pipeline
...
+	}
+
+	// get the side input deployments and update the status of them to the pipeline
+	if len(pipeline.Spec.SideInputs) == 0 {
+		pipeline.Status.MarkServiceHealthy(dfv1.PipelineConditionSideInputServiceHealthy,
+			"SideInputNotAvailable", "No any side input attached to the pipeline")
+	} else {
+		for _, sideInput := range pipeline.Spec.SideInputs {
+			var sideInputDeployment appv1.Deployment
+			if err := c.Get(ctx, client.ObjectKey{
```
whynowy, line 964: Minor: could use list call.

pkg/reconciler/pipeline/controller.go
```diff
@@ -935,3 +938,54 @@ func (r *pipelineReconciler) safeToDelete(ctx context.Context, pl *dfv1.Pipeline
...
+				Namespace: pipeline.GetNamespace(),
+				Name:      pipeline.GetSideInputsManagerDeploymentName(sideInput.Name),
+			}, &sideInputDeployment); err != nil {
+				return err
+			}
+			if msg, reason, status := getDeploymentStatus(&sideInputDeployment); status {
+				pipeline.Status.MarkServiceHealthy(dfv1.PipelineConditionSideInputServiceHealthy, reason, msg)
+			} else {
+				pipeline.Status.MarkServiceNotHealthy(dfv1.PipelineConditionSideInputServiceHealthy, reason, msg)
```
whynowy, line 973: If it's not healthy, shall we break the loop?

pkg/reconciler/pipeline/watcher.go
```diff
@@ -0,0 +1,76 @@
...
+func getVertexStatus(ctx context.Context, c client.Client, pipeline *dfv1.Pipeline) (string, string, error) {
+	totalVertices := len(pipeline.Spec.Vertices)
+	healthyVertices := 0
+	var vertices dfv1.VertexList
+	if err := c.List(ctx, &vertices, client.InNamespace(pipeline.GetNamespace()),
+		client.MatchingLabels{dfv1.KeyPipelineName: pipeline.GetName()}); err != nil {
+		return "", "", err
+	}
+	for _, vertex := range vertices.Items {
+		for _, condition := range vertex.Status.Conditions {
```
whynowy, line 23: Add a `IsHealthy()` function for `Vertex.Status`?

pkg/reconciler/vertex/controller.go
```diff
@@ -434,3 +437,22 @@ func (r *vertexReconciler) markPhaseLogEvent(vertex *dfv1.Vertex, log *zap.Sugar
 	vertex.Status.MarkPhaseFailed(reason, message)
 	r.recorder.Event(vertex, corev1.EventTypeWarning, reason, message)
 }
+
+func checkChildrenResourceStatus(ctx context.Context, c client.Client, vertex *dfv1.Vertex) error {
+	// fetch the pods for calculating the status of child resources
+	var podList corev1.PodList
+	if err := c.List(ctx, &podList,
+		client.InNamespace(vertex.GetNamespace()),
+		client.MatchingLabels{dfv1.KeyAppName: vertex.GetName()},
```
whynowy, line 446: LabelSelector should be `dfv1.KeyPipelineName=pipelineName,dfv1.KeyVertexName=vertex.Spec.Name`

pkg/reconciler/vertex/watcher.go
```diff
@@ -0,0 +1,25 @@
...
+	"fmt"
+
+	dfv1 "github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// getVertexStatus calculate the status by iterating over pods objects
+func getVertexStatus(vertex *dfv1.Vertex, pods *corev1.PodList) (string, string, bool) {
+	if len(pods.Items) == 0 {
+		return "No Pods found", "CooldownPeriod", true
```
whynowy, line 13: Not `CooldownPeriod`, maybe just put `NoPodsFound` for now.

pkg/reconciler/vertex/watcher.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"fmt"
+
+	dfv1 "github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// getVertexStatus calculate the status by iterating over pods objects
+func getVertexStatus(vertex *dfv1.Vertex, pods *corev1.PodList) (string, string, bool) {
+	if len(pods.Items) == 0 {
```
whynowy, line 12: Please add a comment `Need to revisit later.`

pkg/reconciler/vertex/watcher.go
```diff
@@ -0,0 +1,25 @@
...
+		return "No Pods found", "CooldownPeriod", true
+	} else if *vertex.Spec.Replicas != int32(len(pods.Items)) {
```
whynowy, line 14: Maybe we do not need to check this. It could be in transient status (e.g. rolling update).

pkg/reconciler/vertex/watcher.go
```diff
@@ -0,0 +1,25 @@
...
+		return "Number of pods are not equal to replicas", "Processing", false
+	} else {
+		for _, pod := range pods.Items {
+			if pod.Status.Phase != corev1.PodRunning {
```
whynowy, line 18: since auto scaling happens from time to time, maybe we should ignore this, but only check if there's any pod in crash status, or keeps crashing?

pkg/reconciler/isbsvc/installer/watcher.go
```diff
@@ -0,0 +1,36 @@
...
+	appv1 "k8s.io/api/apps/v1"
+)
+
+// getStatefulSetStatus returns a message describing statefulset status, and a bool value indicating if the status is considered done.
+// Borrowed at kubernetes/kubectl/rollout_status.go https://github.com/kubernetes/kubernetes/blob/cea1d4e20b4a7886d8ff65f34c6d4f95efcb4742/staging/src/k8s.io/kubectl/pkg/polymorphichelpers/rollout_status.go#L130
+func getStatefulSetStatus(sts *appv1.StatefulSet) (string, bool) {
+	if sts.Status.ObservedGeneration == 0 || sts.Generation > sts.Status.ObservedGeneration {
+		return "Waiting for statefulset spec update to be observed...\n", false
+	}
+	if sts.Spec.Replicas != nil && sts.Status.ReadyReplicas < *sts.Spec.Replicas {
```
juliev0, line 15: I just want to check what this `ReadyReplicas` means. Does it mean "readiness" in the sense of the "readiness probe", where it means something throughout the life of the Pod, or does it just mean something for when the Pod is starting up? 
Mainly, I want to verify that we're returning false, which is used to set "Progressing", when we are in fact progressing through the initial deployment. 

I know the logic is borrowed from the code you're pointing to above (nice find). Is it being used to indicate something like "progressing" there?

pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -206,6 +211,10 @@ func (p Pipeline) GetSideInputsStoreName() string {
 	return fmt.Sprintf("%s-%s", p.Namespace, p.Name)
 }
 
+func (p Pipeline) GetVertexName(defname string) string {
```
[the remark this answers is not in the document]
chandankumar4: not using now, was using earlier, I'll remove it

pkg/reconciler/isbsvc/installer/watcher.go
chandankumar4: ReadyReplicas is the number of pods created for this StatefulSet with a Ready Condition. it for throughout the life of the pod.
"progressing" denotes here either the initial deployment or new pod is coming up, in the reference code they are not mentioned about "Progressing" they just returning status as false/true

pkg/apis/numaflow/v1alpha1/isbsvc_types.go
```diff
@@ -160,3 +147,13 @@ func (isbsvc *InterStepBufferServiceStatus) IsHealthy() bool {
...
 	return isbsvc.IsReady()
 }
+
+// MarkChildrenResourceNotHealthy marks the children resources as not healthy
+func (isbsvc *InterStepBufferServiceStatus) MarkChildrenResourceNotHealthy(reason, message string) {
+	isbsvc.MarkFalse(ISBSvcConditionChildrenResourcesHealthy, reason, message)
+}
+
+// MarkChildrenResourceHealthy marks the children resources as healthy
+func (isbsvc *InterStepBufferServiceStatus) MarkChildrenResourceHealthy(reason, message string) {
```
KeranYang, line 157: when marking healthy, do we need to provide the reason and message? I don't see them in markDeployed and markConfigured.

pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -47,6 +47,11 @@ const (
 	// PipelineConditionDeployed has the status True when the Pipeline
 	// has its Vertices and Jobs created.
 	PipelineConditionDeployed ConditionType = "Deployed"
+
+	// PipelineConditionDaemonServiceHealthy has the status True when the daemon service is healthy.
+	PipelineConditionDaemonServiceHealthy    ConditionType = "DaemonServiceHealthy"
+	PipelineConditionSideInputServiceHealthy ConditionType = "SideInputManagerHealthy"
+	PipelineConditionVerticesServiceHealthy  ConditionType = "VerticesHealthy"
```
KeranYang, line 54: To match the enum name

```suggestion
	PipelineConditionSideInputServiceHealthy ConditionType = "SideInputServiceHealthy"
	PipelineConditionVerticesServiceHealthy  ConditionType = "VerticesServiceHealthy"
```

pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -725,6 +731,16 @@ func (pls *PipelineStatus) IsHealthy() bool {
 	}
 }
 
+// MarkServiceNotHealthy marks the Daemon, SideInput and Vertex Service as not healthy with the given reason and message.
```
KeranYang, line 734: We don't necessarily need to add concrete service names in the comment, in the future we might remove some of them or add more, in that case, we don't want to come back here and update the comment.

```suggestion
// MarkServiceNotHealthy marks a service as not healthy with the given reason and message.
```

pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -725,6 +731,16 @@ func (pls *PipelineStatus) IsHealthy() bool {
...
+func (pls *PipelineStatus) MarkServiceNotHealthy(conditionType ConditionType, reason, message string) {
+	pls.MarkFalse(conditionType, reason, message)
+}
+
+// MarkServiceHealthy marks the Daemon, SideInput and Vertex Service as healthy with the given reason and message.
```
KeranYang, line 739:
```suggestion
// MarkServiceHealthy marks a service as healthy with the given reason and message.
```

pkg/reconciler/isbsvc/installer/installer_test.go
```diff
@@ -193,6 +194,7 @@ func TestInstall(t *testing.T) {
 		testObj := testJetStreamIsbSvc.DeepCopy()
 		err := Install(ctx, testObj, cl, kubeClient, fakeConfig, zaptest.NewLogger(t).Sugar(), record.NewFakeRecorder(64))
 		assert.NoError(t, err)
+		testObj.Status.MarkChildrenResourceHealthy("RolloutFinished", "partitioned roll out complete: 3 new pods have been updated...")
```
KeranYang, line 197: can we add one more unit test to cover the failure case where children resources are not healthy?

pkg/reconciler/pipeline/controller_test.go
```diff
@@ -862,3 +911,37 @@ func Test_copyVertexTemplate(t *testing.T) {
...
+		assert.NoError(t, err)
+		err = checkChildrenResourceStatus(ctx, cl, testObj)
+		assert.NoError(t, err)
+		for _, c := range testObj.Status.Conditions {
+			if c.Type == string(dfv1.PipelineConditionDaemonServiceHealthy) {
+				assert.Equal(t, string(corev1.ConditionTrue), string(c.Status))
+			} else if c.Type == string(dfv1.PipelineConditionSideInputServiceHealthy) {
+				assert.Equal(t, string(corev1.ConditionTrue), string(c.Status))
+			} else if c.Type == string(dfv1.PipelineConditionVerticesServiceHealthy) {
+				assert.Equal(t, string(corev1.ConditionFalse), string(c.Status))
```
KeranYang, line 943: why PipelineConditionVerticesServiceHealthy corresponds to ConditionFalse instead of True?

pkg/reconciler/pipeline/watcher.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+// getVertexStatus will calculate the status of the vertices and return the status and reason
+func getVertexStatus(vertices *dfv1.VertexList) (string, string, error) {
+	for _, vertex := range vertices.Items {
+		if !vertex.Status.IsHealthy() {
+			return "False", "Progressing", nil
+		}
+	}
+	return "True", "Successful", nil
```
KeranYang, line 18: Returning a string "True"/"False" is a bit confusing to me. Could you please follow the same method signature as `getDeploymentStatus` below, it's more clear.

pkg/reconciler/vertex/controller.go
```diff
@@ -320,6 +320,9 @@ func (r *vertexReconciler) reconcile(ctx context.Context, vertex *dfv1.Vertex) (
 	vertex.Status.Selector = selector.String()
 
 	vertex.Status.MarkPhaseRunning()
+	if err = checkChildrenResourceStatus(ctx, r.client, vertex); err != nil {
```
KeranYang, line 323: unit test?

pkg/apis/numaflow/v1alpha1/isbsvc_types.go
chandankumar4: For `kind: Deployment` in k8s is providing `reason` and `message` in healthy case as well as example below and here it's checking the Deployment, Statefulset or Pod status only, So have followed the same.

```
  conditions:
  - lastTransitionTime: "2024-07-25T15:39:30Z"
    lastUpdateTime: "2024-07-25T15:39:30Z"
    message: Deployment has minimum availability.
    reason: MinimumReplicasAvailable
    status: "True"
    type: Available
```

pkg/reconciler/isbsvc/installer/watcher.go
juliev0: Thanks @chandankumar4. I'm not sure if I totally understand. So, will `ReadyReplicas` decrease in the case where the StatefulSet was last deployed a long time ago and then something cases one of the Pods to get into a bad state?

pkg/reconciler/isbsvc/installer/installer_test.go
chandankumar4: Added

pkg/reconciler/vertex/controller.go
chandankumar4: Added

pkg/reconciler/pipeline/controller_test.go
chandankumar4: After reconciler ran, It's setting the vertex pod health as "False", Let me check the reason behind it.

whynowy:
@chandankumar4 - can you check why many test cases failed? was it because we have stricter health check for ISBSvc which causes timeout?

chandankumar4:
> @chandankumar4 - can you check why many test cases failed? was it because we have stricter health check for ISBSvc which causes timeout?

Checking it, was resolving the review comment so pushed the changes.

pkg/reconciler/isbsvc/installer/watcher.go
chandankumar4: Yes, For above case `ReadyReplicas` will get decreased.
For example: I have deleted one of the sts pod `isbsvc-default-js-0` and `ReadyReplicas` count decreased to 2 from 3
```
  status:
    availableReplicas: 2
    collisionCount: 0
    currentReplicas: 3
    currentRevision: isbsvc-default-js-597b7f74d7
    observedGeneration: 1
    readyReplicas: 2
    replicas: 3
    updateRevision: isbsvc-default-js-597b7f74d7
    updatedReplicas: 3
```
```
default           isbsvc-default-js-0                        1/3     Running     0             36s
default           isbsvc-default-js-1                        3/3     Running     0             7m7s
default           isbsvc-default-js-2                        3/3     Running     0             7m7s
```

juliev0: Thanks for checking and explaining. So, for Numaplane we need to know “has the ISB Service been fully reconciled?” For that, I would need to be able to distinguish between “just got deployed and not up yet” vs “already got deployed and was initially fine but then the pod started becoming unavailable”. It’s really just the first one I need. That’s what I was thinking “Progressing” might mean. But what do you think?

pkg/reconciler/pipeline/controller_test.go
chandankumar4: Here vertex reconciler is not running, so the status of vertex pod is not getting updated and getting set as false by default

test/fixtures/e2e_suite.go
```diff
@@ -42,7 +42,7 @@ const (
 	Label          = "numaflow-e2e"
 	LabelValue     = "true"
 	ISBSvcName     = "numaflow-e2e"
-	defaultTimeout = 60 * time.Second
+	defaultTimeout = 90 * time.Second
```
chandankumar4, line 45: Didn't found any issue for the failure e2e test cases, just seen that it was getting timeout while checking ISB Service creation and deletion.

pkg/reconciler/isbsvc/installer/watcher.go
juliev0: Discussed offline with @chandankumar4 just now. I am fine if we address my request above in a subsequent PR so as to not hold up this one.

test/fixtures/e2e_suite.go
whynowy: I think this is expected when having more sub resources check.

pkg/apis/numaflow/v1alpha1/vertex_types.go
```diff
@@ -37,6 +37,9 @@ const (
 	VertexPhaseRunning   VertexPhase = "Running"
 	VertexPhaseSucceeded VertexPhase = "Succeeded"
 	VertexPhaseFailed    VertexPhase = "Failed"
+
+	// VertexConditionPodHealthy has the status True when all the vertex pods are healthy.
+	VertexConditionPodHealthy ConditionType = "PodHealthy"
```
whynowy, line 42: `PodsHealthy`

pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -47,6 +47,11 @@ const (
 	// PipelineConditionDeployed has the status True when the Pipeline
 	// has its Vertices and Jobs created.
 	PipelineConditionDeployed ConditionType = "Deployed"
+
+	// PipelineConditionDaemonServiceHealthy has the status True when the daemon service is healthy.
+	PipelineConditionDaemonServiceHealthy    ConditionType = "DaemonServiceHealthy"
+	PipelineConditionSideInputServiceHealthy ConditionType = "SideInputServiceHealthy"
+	PipelineConditionVerticesServiceHealthy  ConditionType = "VerticesServiceHealthy"
```
whynowy, line 54: `VerticesHealthy`

pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -47,6 +47,11 @@ const (
 	// PipelineConditionDeployed has the status True when the Pipeline
 	// has its Vertices and Jobs created.
 	PipelineConditionDeployed ConditionType = "Deployed"
+
+	// PipelineConditionDaemonServiceHealthy has the status True when the daemon service is healthy.
+	PipelineConditionDaemonServiceHealthy    ConditionType = "DaemonServiceHealthy"
+	PipelineConditionSideInputServiceHealthy ConditionType = "SideInputServiceHealthy"
```
whynowy, line 53: `SideInputsManagersHealthy`

pkg/reconciler/pipeline/controller.go
```diff
@@ -942,3 +945,52 @@ func (r *pipelineReconciler) safeToDelete(ctx context.Context, pl *dfv1.Pipeline
 	}()
 	return daemonClient.IsDrained(ctx, pl.Name)
 }
+
+func checkChildrenResourceStatus(ctx context.Context, c client.Client, pipeline *dfv1.Pipeline) error {
+	// get the daemon deployment and update the status of it to the pipeline
+	var daemonDeployment appv1.Deployment
+	if err := c.Get(ctx, client.ObjectKey{Namespace: pipeline.GetNamespace(), Name: pipeline.GetDaemonDeploymentName()},
+		&daemonDeployment); err != nil && !apierrors.IsNotFound(err) {
```
whynowy, line 953: Shall we add `apierrors.IsNotFound(err)`? that means daemon service does not exist.

whynowy: When returning error, we should also mark unhealthy with reason.

pkg/reconciler/pipeline/controller.go
```diff
@@ -942,3 +945,52 @@ func (r *pipelineReconciler) safeToDelete(ctx context.Context, pl *dfv1.Pipeline
...
+
+	// get the side input deployments and update the status of them to the pipeline
+	if len(pipeline.Spec.SideInputs) == 0 {
+		pipeline.Status.MarkServiceHealthy(dfv1.PipelineConditionSideInputServiceHealthy,
+			"NoSideInputs", "No Side Inputs attached to the pipeline")
+	} else {
+		var sideInputs appv1.DeploymentList
+		selector, _ := labels.Parse(dfv1.KeyPipelineName + "=" + pipeline.Name + "," + dfv1.KeyComponent + "=" + dfv1.ComponentSideInputManager)
+		if err := c.List(ctx, &sideInputs, &client.ListOptions{Namespace: pipeline.Namespace, LabelSelector: selector}); err != nil {
+			return err
```
whynowy, line 970: Same here.

pkg/reconciler/pipeline/controller.go
```diff
@@ -942,3 +945,52 @@ func (r *pipelineReconciler) safeToDelete(ctx context.Context, pl *dfv1.Pipeline
...
+				break
+			}
+		}
+	}
+
+	// calculate the status of the vertices and update the status of them to the pipeline
+	var vertices dfv1.VertexList
+	selector, _ := labels.Parse(dfv1.KeyPipelineName + "=" + pipeline.GetName() + "," + dfv1.KeyComponent + "=" + dfv1.ComponentVertex)
+	if err := c.List(ctx, &vertices, &client.ListOptions{Namespace: pipeline.Namespace, LabelSelector: selector}); err != nil {
+		return err
```
whynowy, line 986: Same here.

whynowy (commented):
Overall, the logic looks good to me, I also checked out the code, tested on my local, the major features are working, some of the details can be refined later.

whynowy:
@numaproj/numaflow-dev - please review, let's get this one in ASAP.

pkg/reconciler/pipeline/controller.go
chandankumar4: Just after creating daemon deployment, it's not able to find it instant and was giving error here, so have added check `!apierrors.IsNotFound(err)`

whynowy: hmmm.. Logically, skipping it is not right, maybe put a word like "Deployment not found, might be still under creation" when no found?

pkg/reconciler/pipeline/controller.go
whynowy: missing this.

pkg/reconciler/pipeline/controller.go
chandankumar4: Do we need to reconcile also for `apierrors.IsNotFound(err)` error as returning `err`, I have checked other places it's not reconciling for `apierrors.IsNotFound(err)`?

api/json-schema/schema.json
```diff
@@ -670,6 +670,9 @@
           "description": "Standard list metadata. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds"
         }
       },
+      "required": [
+        "items"
+      ],
```
chandankumar4, line 675: Not sure, why this change is required from today?

pkg/reconciler/pipeline/controller.go
chandankumar4: Done

xdevxy (approved):

whynowy (approved):

KeranYang (approved):

PR #1874, opened 2024-07-29
chore: publishing extension image

### LINKED ISSUE #1857: publish serve image

yhl25:
Using Alpine has performance implications
References:
* https://medium.com/inside-sumup/stop-using-alpine-docker-images-fbf122c63010
* https://news.ycombinator.com/item?id=38616023
* https://news.ycombinator.com/item?id=38798233

fixes #1857


yhl25:
@BulkBeing please review.

serving/Makefile
```diff
@@ -0,0 +1,48 @@
...
+test:
+	cargo test
+
+.PHONY: all-tests
+all-tests:
+	cargo test --all-features
+
+set-qemu:
+	$(DOCKER) pull tonistiigi/binfmt:latest
+	$(DOCKER) run --rm --privileged tonistiigi/binfmt:latest --install amd64,arm64
```
vigith, line 32: what is this for?

vigith (approved):

whynowy (changes requested):
Is [this](https://github.com/numaproj/numaflow/pull/1882) needed?

Please finalize the PR.

vigith:
> Is [this](https://github.com/numaproj/numaflow/pull/1882) needed?

That is a test PR to a different branch, not to the main. It is only for testing.

whynowy:
I have couple of concerns:

1. Image build time. 

Following information is from my laptop.

First time build: `7m46.093s` - this might now be a concern since it's 1st time
Second time build with any slight change in rust code: `4m29.952s`

As a reference, current image build time on `main`: `2m7.347s`

2. Image size. 
  `alpine` based image on local: `130MB`.
  New image: `251MB`

Questions:

1. Can we build the rust code on the local instead of in docker, and then copy to the image? I believe this will speed up a lot.
2. Can we use `scratch` for official release, `alpine` for local build? If the challenge is about CERT generation, I believe there's way to do it in rust.
3. Is multi platform image support there?

vigith:
rust build/compile is notoriously slow

whynowy:
> rust build/compile is notoriously slow

Understoodable. But local build should be way faster than in the docker.

vigith:
> Understoodable. But local build should be way faster than in the docker.

make it configurable, perhaps? building locally doesn't sound right 😄

whynowy:
> > Understoodable. But local build should be way faster than in the docker.
> 
> make it configurable, perhaps? building locally doesn't sound right 😄

Why? It should be able to build any format for any platform.

vigith:
> Why? It should be able to build any format for any platform from anywhere.

my concern is, what if we forget to locally build and copy?

whynowy:
> > Why? It should be able to build any format for any platform from anywhere.
> 
> my concern is, what if we forget to locally build and copy?

Everything should be automated, locally build does not mean we build manually. This is the way we build golang and node code, which makes it way faster.

vigith:
what is the status of this?

Makefile
```diff
@@ -161,15 +161,19 @@ ui-build:
 ui-test: ui-build
 	./hack/test-ui.sh
 
+.PHONY: ext-build
+ext-build:
+	cd serving && $(MAKE) build
+
 .PHONY: image
-image: clean ui-build dist/$(BINARY_NAME)-linux-amd64
+image: clean ui-build dist/$(BINARY_NAME)-linux-amd64 ext-build
```
whynowy, line 169: Since we still build binaries in docker,  so we don't need `ext-build`, right?

Makefile
```diff
@@ -161,15 +161,19 @@ ui-build:
...
 	DOCKER_BUILDKIT=1 $(DOCKER) build --build-arg "ARCH=amd64" --build-arg "BASE_IMAGE=$(DEV_BASE_IMAGE)" $(DOCKER_BUILD_ARGS) -t $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION) --target $(BINARY_NAME) -f $(DOCKERFILE) .
 	@if [[ "$(DOCKER_PUSH)" = "true" ]]; then $(DOCKER) push $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION); fi
 ifdef IMAGE_IMPORT_CMD
 	$(IMAGE_IMPORT_CMD) $(IMAGE_NAMESPACE)/$(BINARY_NAME):$(VERSION)
 endif
 
-image-multi: ui-build set-qemu dist/$(BINARY_NAME)-linux-arm64.gz dist/$(BINARY_NAME)-linux-amd64.gz
+image-multi: ui-build set-qemu dist/$(BINARY_NAME)-linux-arm64.gz dist/$(BINARY_NAME)-linux-amd64.gz ext-build
```
whynowy, line 176: Same here, `ex-build` is not needed.

Dockerfile
```diff
@@ -1,27 +1,64 @@
...
+COPY ./serving/src ./src
+COPY ./serving/servesink/src ./servesink/src
+COPY ./serving/extras/upstreams/src ./extras/upstreams/src
+COPY ./serving/backoff/src ./backoff/src
+
+# Build the real binaries
+RUN touch src/main.rs servesink/main.rs extras/upstreams/main.rs && \
+    cargo build --release
+
+COPY --from=base /bin/numaflow /bin/numaflow
```
whynowy, line 48: Is this needed?

Dockerfile
```diff
@@ -1,27 +1,64 @@
-ARG BASE_IMAGE=scratch
+ARG BASE_IMAGE=gcr.io/distroless/cc-debian12
```
whynowy, line 1: What is the effort to bring it back to `scratch`?

whynowy:
Something helpful - https://github.com/bjornmolin/rust-minimal-docker/tree/master

serving/Dockerfile
```diff
@@ -1,44 +1,42 @@
-FROM rust:1.79-bookworm as build
+# Use multi-stage builds to keep the final image small
```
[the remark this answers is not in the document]
yhl25: I thought it would be useful to keep it if we just want to build the serving image.

Makefile
yhl25: we copy the Cargo.lock file which will be generated after we build

PR #1875, opened 2024-07-29
chore: use appropriate image for serving container

yhl25:


yhl25:
Decided to go with the same Numaflow image but a different binary.

Issue #1878, opened 2024-07-29
Read bytes metric is missing for sinks


whynowy:
**Describe the bug**

<img width="797" alt="Screenshot 2024-07-29 at 1 34 44 PM" src="https://github.com/user-attachments/assets/ca7c2f8b-bab3-4697-ab98-90c5cf697c03">

**To Reproduce**

1. ....
2. ....
3. ....

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**

- Kubernetes: [e.g. v1.18.6]
- Numaflow: [e.g. v0.5.1]

**Additional context**

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

>For quick help and support, join our [slack](https://numaproj.slack.com/join/shared_invite/zt-19svuv47m-YKHhsQ~~KK9mBv1E7pNzfg#/shared-invite/email) channel.

whynowy:
@samhith-kakarla

PR #1879, opened 2024-07-30
chore: add readBytesTotal metric

samhith-kakarla:
Closes: https://github.com/numaproj/numaflow/issues/1878


pkg/sinks/forward/forward.go
```diff
@@ -213,6 +213,14 @@ func (df *DataForward) forwardAChunk(ctx context.Context) {
 	// store the offsets of the messages we read from ISB
 	var readOffsets = make([]isb.Offset, len(readMessages))
 	for idx, m := range readMessages {
+		metrics.ReadBytesCount.With(map[string]string{
```
whynowy, line 216: It might be better to follow the pattern in `reduce/data_forward.go`, to use a variable to add them up, and then emit the metric outside the for loop.

whynowy: Could you please also update the other places (source, udf) to use the same patten?

pkg/udf/forward/forward.go
```diff
@@ -242,9 +243,12 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
 	for _, msg := range dataMessages {
 		msg.Watermark = time.Time(processorWM)
 		// emit message size metric
-		metrics.ReadBytesCount.With(map[string]string{metrics.LabelVertex: isdf.vertexName, metrics.LabelPipeline: isdf.pipelineName, metrics.LabelVertexType: string(dfv1.VertexTypeMapUDF), metrics.LabelVertexReplicaIndex: strconv.Itoa(int(isdf.vertexReplica)), metrics.LabelPartitionName: isdf.fromBufferPartition.GetName()}).Add(float64(len(msg.Payload)))
+
+		totalBytes += len(msg.Payload)
```
kohlisid, line 247: @whynowy Here we are increasing the metric only for data messages. 
We should use this metric for all read messages right? Data  + ctrl

I realized in the UDF we have been updating it only for data messags, and other types have for all

whynowy: Good point!

whynowy: This metric is only for data messages, we should make it consistent. Same for message count.

kohlisid: In that case for all other UDFs we should update this, we are publishing for both data + ctrl.

pkg/reduce/data_forward.go
```diff
@@ -366,8 +366,23 @@ func (df *DataForward) forwardAChunk(ctx context.Context) {
 			m.Message.Keys = []string{dfv1.DefaultKeyForNonKeyedData}
 		}
 		m.Watermark = time.Time(processorWM)
+	}
+
+	var dataMessages = make([]*isb.ReadMessage, 0, len(readMessages))
+
+	// store the offsets of the messages we read from ISB
+	var readOffsets = make([]isb.Offset, len(readMessages))
+	for idx, m := range readMessages {
```
vigith, line 375: do we really have to do this?

samhith-kakarla: you are right readOffsets is not used anywhere in the method

pkg/reduce/data_forward.go
```diff
@@ -366,7 +366,13 @@ func (df *DataForward) forwardAChunk(ctx context.Context) {
 			m.Message.Keys = []string{dfv1.DefaultKeyForNonKeyedData}
 		}
 		m.Watermark = time.Time(processorWM)
-		totalBytes += len(m.Payload)
+	}
+
+	// store the offsets of the messages we read from ISB
+	for _, m := range readMessages {
```
whynowy, line 372: can we do it in the for loop above?

pkg/sinks/forward/forward.go
```diff
@@ -230,10 +232,19 @@ func (df *DataForward) forwardAChunk(ctx context.Context) {
 
 	writeMessages := make([]isb.Message, 0, len(dataMessages))
 	for _, m := range dataMessages {
+		totalBytes += len(m.Payload)
```
whynowy, line 235: Move it to the for loop in line 216+?

pkg/metrics/metrics.go
```diff
@@ -58,6 +58,12 @@ var (
 		Help:      "Total number of bytes read",
 	}, []string{LabelVertex, LabelPipeline, LabelVertexType, LabelVertexReplicaIndex, LabelPartitionName})
 
+	ReadDataBytesCount = promauto.NewCounterVec(prometheus.CounterOpts{
+		Subsystem: "forwarder",
+		Name:      "data_read_bytes_total",
+		Help:      "Total number of bytes read",
```
whynowy, line 64:
```suggestion
		Help:      "Total number of Data message bytes read",
```

pkg/sinks/forward/forward.go
```diff
@@ -234,6 +239,22 @@ func (df *DataForward) forwardAChunk(ctx context.Context) {
 		writeMessages = append(writeMessages, m.Message)
 	}
 
+	metrics.ReadBytesCount.With(map[string]string{
```
whynowy, line 242: Move them up to line 228.

pkg/sources/forward/data_forward.go
```diff
@@ -331,24 +330,33 @@ func (df *DataForward) forwardAChunk(ctx context.Context) {
...
 			m.Watermark = time.Time(processorWM)
 			readWriteMessagePairs[idx].ReadMessage = m
 			// if no user-defined transformer exists, then the messages to write will be identical to the message read from source
 			// thus, the unmodified read message will be stored as the corresponding writeMessage in readWriteMessagePairs
 			readWriteMessagePairs[idx].WriteMessages = []*isb.WriteMessage{{Message: m.Message}}
 		}
+
 	}
 
+	metrics.ReadBytesCount.With(map[string]string{
```
whynowy, line 345: Nit: Move them up to line 276.

pkg/udf/forward/forward.go
```diff
@@ -242,9 +245,12 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
 	for _, msg := range dataMessages {
 		msg.Watermark = time.Time(processorWM)
 		// emit message size metric
-		metrics.ReadBytesCount.With(map[string]string{metrics.LabelVertex: isdf.vertexName, metrics.LabelPipeline: isdf.pipelineName, metrics.LabelVertexType: string(dfv1.VertexTypeMapUDF), metrics.LabelVertexReplicaIndex: strconv.Itoa(int(isdf.vertexReplica)), metrics.LabelPartitionName: isdf.fromBufferPartition.GetName()}).Add(float64(len(msg.Payload)))
+
+		dataBytes += len(msg.Payload)
```
whynowy, line 249: Do it at line 231?

pkg/udf/forward/forward.go
```diff
@@ -242,9 +245,12 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) {
...
 	}
 
+	metrics.ReadBytesCount.With(map[string]string{metrics.LabelVertex: isdf.vertexName, metrics.LabelPipeline: isdf.pipelineName, metrics.LabelVertexType: string(dfv1.VertexTypeMapUDF), metrics.LabelVertexReplicaIndex: strconv.Itoa(int(isdf.vertexReplica)), metrics.LabelPartitionName: isdf.fromBufferPartition.GetName()}).Add(float64(totalBytes))
```
whynowy, line 252: And then move these 2 up to line 236?

whynowy (approved):

PR #1884, opened 2024-07-31
fix: configure discard policy for WorkQueue/Interest 

kohlisid:


vigith:
should we add a check so that nobody will turn on `DiscardOld` with `WorkQueue` *or* `DiscardNew` with `Limits` ?

kohlisid:
In that case do you want to tie both the properties together instead of configurable?
1) If Limits/Interest is given -> use discardOld
2) If workqueue is given -> use discardNew

pkg/isbsvc/jetstream_service.go
```diff
@@ -120,11 +120,17 @@ func (jss *jetStreamSvc) CreateBuffersAndBuckets(ctx context.Context, buffers, b
 			if !errors.Is(err, nats.ErrStreamNotFound) {
 				return fmt.Errorf("failed to query information of stream %q during buffer creating, %w", streamName, err)
 			}
+			retention := nats.RetentionPolicy(v.GetInt("stream.retention"))
+			discard := nats.DiscardPolicy(v.GetInt("stream.discardPolicy"))
+			// we cannot use DiscardNew with Limits policy.
+			if retention == nats.LimitsPolicy && discard == nats.DiscardNew {
```
Krithika3, line 126: Shoudl we also check for the other? WorkQueue with DiscardOld?

Krithika3: nevermind, just saw that @vigith has already made that comment.

vigith:
> In that case do you want to tie both the properties together instead of configurable?
> 
> 1. If Limits/Interest is given -> use discardOld
> 2. If workqueue is given -> use discardNew

I would think so, @whynowy ?

config/install.yaml
```diff
@@ -18329,6 +18329,8 @@ data:
             storage: 0
             replicas: 3
             duplicates: 60s
+            # 0: DiscardOld, 1: DiscardNew
+            discardPolicy: 0
```
Krithika3, line 18333: just curious, why do we have multiple configs where we have this field set?

kohlisid: We use the different installation strategies for the numaflow controller, so replicating it for all.

pkg/isbsvc/jetstream_service.go
```diff
@@ -120,11 +120,23 @@ func (jss *jetStreamSvc) CreateBuffersAndBuckets(ctx context.Context, buffers, b
...
+			retention := nats.RetentionPolicy(v.GetInt("stream.retention"))
+			// default discard policy is set to DiscardOld
+			discard := nats.DiscardOld
+
+			// Based on the retention policy we use the following discard policy
+			// 1) Limits/Interest Policy -> DiscardOld
+			// 2) WorkQueuePolicy -> DiscardNew
+			if retention == nats.WorkQueuePolicy {
+				discard = nats.DiscardNew
+			}
```
vigith, line 133: what about Interest? Why should it be `DiscardOld`? please document

kohlisid: I went ahead as it was default, but I realised it didn't seem to be correct.
We should use it with DiscardNew instead
I have added few lines to document, please take a look @vigith

vigith (approved):

whynowy (approved):

PR #1892, opened 2024-08-01
feat: Add ObservedGeneration field in vertex status and use it for calculating status

chandankumar4:
Explain what this PR does.

- Add ObservedGeneration field in vertex status and use it for calculating the status of vertex
- Update the `reason` for deployment status calculation

Verified the changes in local kind cluster


pkg/reconciler/isbsvc/installer/watcher.go
```diff
@@ -8,29 +8,29 @@ import (
...
-		return fmt.Sprintf("Waiting for %d pods to be ready...\n", *sts.Spec.Replicas-sts.Status.ReadyReplicas), false
+		return fmt.Sprintf("Waiting for %d pods to be ready...\n", *sts.Spec.Replicas-sts.Status.ReadyReplicas), "Unavailable", false
 	}
 	if sts.Spec.UpdateStrategy.Type == appv1.RollingUpdateStatefulSetStrategyType && sts.Spec.UpdateStrategy.RollingUpdate != nil {
 		if sts.Spec.Replicas != nil && sts.Spec.UpdateStrategy.RollingUpdate.Partition != nil {
 			if sts.Status.UpdatedReplicas < (*sts.Spec.Replicas - *sts.Spec.UpdateStrategy.RollingUpdate.Partition) {
 				return fmt.Sprintf(
 					"Waiting for partitioned roll out to finish: %d out of %d new pods have been updated...\n",
-					sts.Status.UpdatedReplicas, *sts.Spec.Replicas-*sts.Spec.UpdateStrategy.RollingUpdate.Partition), false
+					sts.Status.UpdatedReplicas, *sts.Spec.Replicas-*sts.Spec.UpdateStrategy.RollingUpdate.Partition), "Unavailable", false
```
juliev0, line 27: It sounds like you changed this from "Progressing" to "Unavailable" based on your testing? Can you clarify the testing that caused you to change that? Did you find that you hit this when not in the initial startup phase? (If so, I guess the message "Waiting for partitioned roll out to finish: %d out of %d new pods have been updated" should probably be changed.)

pkg/reconciler/isbsvc/installer/watcher.go
```diff
@@ -8,29 +8,29 @@ import (
...
 			}
 		}
 		return fmt.Sprintf("partitioned roll out complete: %d new pods have been updated...\n",
-			sts.Status.UpdatedReplicas), true
-	}
-	if sts.Status.UpdateRevision != sts.Status.CurrentRevision {
-		return fmt.Sprintf("waiting for statefulset rolling update to complete %d pods at revision %s...\n",
-			sts.Status.UpdatedReplicas, sts.Status.UpdateRevision), false
+			sts.Status.UpdatedReplicas), "RolloutFinished", true
```
juliev0, line 31: nit: maybe we should change "RolloutFinished" to say something more generic like "Healthy" since it's now distinguished both from "Progressing" and "Unavailable"

pkg/reconciler/isbsvc/installer/watcher.go
chandankumar4: Ohh, It was my bad, I didn't mean to change it here, Let me update it

whynowy:
Is it ready for review? It looks good to me.

pkg/reconciler/pipeline/watcher.go
```diff
@@ -11,9 +11,12 @@ import (
...
 	for _, vertex := range vertices.Items {
-		if !vertex.Status.IsHealthy() {
+		if vertex.Status.ObservedGeneration == 0 || vertex.Generation > vertex.Status.ObservedGeneration {
 			return false, "Progressing"
 		}
+		if !vertex.Status.IsHealthy() {
+			return false, "Unavailable"
+		}
 	}
 	return true, "Successful"
```
juliev0, line 21: sorry - if we're using "Healthy" in the other places, should we use that here too for consistency instead of "Successful"?

pkg/reconciler/vertex/controller.go
```diff
@@ -320,6 +320,7 @@ func (r *vertexReconciler) reconcile(ctx context.Context, vertex *dfv1.Vertex) (
 	vertex.Status.Selector = selector.String()
 
 	vertex.Status.MarkPhaseRunning()
+	vertex.Status.SetObservedGeneration(vertex.Generation)
```
whynowy, line 323: Set at the beginning?

whynowy: I have a follow up PR, will address it there.

whynowy (approved):

pkg/reconciler/vertex/controller.go
chandankumar4: okay, Thanks

PR #1904, opened 2024-08-06
feat: source to sink with an optional transformer without ISB

vigith:
Source to Sink via an optional transformer without ISB.

### Manual Testing

```
kind: MonoVertex
metadata:
  name: ceres-mono-vertex
spec:
  replicas: 5
  source:
    udsource:
      container:
        image: docker.xxx.com/eventbus-numaflow:0.5.1
        args:
          - --spring.config.location=file:/conf/eventbus-source-config.yaml
          - --numaflow.handler=eventbus-source
        env:
          - name: config_file_path
            value: /conf/eventbus-source-config.yaml
        volumeMounts:
          - name: eventbus-source-config-volume
            mountPath: /conf

  sink:
    udsink:
      container:
        image: docker.xxx.com/blackhole-sink:stable

  volumes:
    - name: eventbus-source-config-volume
      configMap:
        name: eventbus-source-config
        items:
          - key: configuration
            path: eventbus-source-config.yaml
```


Dockerfile
```diff
@@ -44,21 +50,29 @@ COPY ./serving/servesink/src ./servesink/src
...
-    cargo build --release
+RUN touch src/main.rs servesink/src/main.rs numaflow-models/src/main.rs source-sink/src/main.rs && \
+    cargo build --workspace --all --release
 
 ####################################################################################################
 # numaflow
 ####################################################################################################
 ARG BASE_IMAGE
-FROM ${BASE_IMAGE} as numaflow
+FROM debian:bookworm as numaflow
```
whynowy, line 65: This is going out of control.

1. Image build time in the CI increased from `<2 mins` to `> 7 mins`
2. Image size and security vulunabilities.

serving/source-sink/build.rs
```diff
@@ -0,0 +1,13 @@
+fn main() {
+    tonic_build::configure()
+        .build_server(true)
+        .compile(
+            &[
+                "proto/source.proto",
```
whynowy, line 6: What is the reason the client needs to keep the proto files?

vigith: we do not need to, we can copy at build time. this was the reason for #1910

Dockerfile
vigith: we can handle this in follow up PR, already created an action item.

vigith:
i am reviewing this PR, please do not merge.

serving/source-sink/build.rs
vigith: we will handle this in follow up PR.

whynowy (approved):

PR #1905, opened 2024-08-06
chore: changing health status thresholds

samhith-kakarla:
Explain what this PR does.


samhith-kakarla:
Pipline is reaching warning/critical state when we stop scaling: 
pipeline_with_config_1
Screenshot 2024-08-06 at 2 37 11 PM

samhith-kakarla:
unit tests were erroring because of change in thresholds so i changed the cutoff values in the tests to be dependent on DefaultBufferUsageLimit and match the new thresholds. Also added TestHealthThresholds to show that thresholds are actually changing as expected when BufferUsageLimit changes

pkg/daemon/server/service/health_status.go
```diff
@@ -59,11 +59,11 @@ const (
 )
 
 // HealthThresholds are the thresholds used to compute the health of a vertex
-const (
+var (
 	// criticalBufferThreshold is the threshold above which the health of a vertex is critical
-	criticalBufferThreshold = 95
+	criticalBufferThreshold = v1alpha1.DefaultBufferUsageLimit * float64(90) // 95
```
whynowy, line 64: The threshold for different vertex should be different - based on the vertex `limits` configuration, is this right? @kohlisid

kohlisid: Yes now that we have it relative, it can vary vertex to vertex. 
The code will need changes to compare each timeline using the limits for the corresponding vertex limits.

KeranYang:
Hey @samhith-kakarla , Yashash and I just [made changes](https://github.com/numaproj/numaflow/commit/7b85e89f307c2286c5b57daf3fb17d70a08107c0) to our e2e tests and committed to the main. If you continue working on your branch, you might see some un-expected e2e testing failures. To fix them, please rebase. Thanks!

PR #1910, opened 2024-08-08
chore: update the proto files

vigith:
copied proto files from numaflow-go


pkg/apis/proto/mapstream/v1/mapstream.proto
```diff
@@ -17,11 +17,11 @@ limitations under the License.
 syntax = "proto3";
 
 option go_package = "github.com/numaproj/numaflow-go/pkg/apis/proto/mapstream/v1";
-option java_package = "io.numaproj.numaflow.mapstream.v1";
```
whynowy, line 20: Keep all of these `java_packages`, eventually we will use them as the source of truth.

whynowy (approved):

PR #1933, opened 2024-08-13
fix: retry failed messages for MonoVertex sink

yhl25:


vigith (commented):
let's add a retry-interval too?

yhl25:
> let's add a retry-interval too?

I have added it to the config, but I guess we will have to add this to the mono vertex spec that way user can configure it. I will tackle that in a separate PR since it requires controller changes.

vigith (approved):

rust/monovertex/src/forwarder.rs
```diff
@@ -106,21 +107,61 @@ impl Forwarder {
...
+
+                                let successful_offsets: Vec<Offset> = retry_messages.iter()
+                                    .filter(|msg| !failed_ids.contains(&msg.id))
+                                    .map(|msg| msg.offset.clone())
+                                    .collect();
+
+
+                                // ack the successful offsets
+                                let n = successful_offsets.len();
+                                self.source_client.ack_fn(successful_offsets).await?;
```
vigith, line 133: we cannot do this, in Kafka if we ack `n`th message, everything upto `n` is considered as processing. If the pod crashes, Kafka will not redeliver those messages that we are still retrying.

Issue #1906, opened 2024-08-07
containerTemplate gets propagated to initContainer in spec


kohlisid:
**Describe the bug**
While applying an containerTemplate to a vertex, it gets partially propagated to the initContainer spec as well causing a mismatch in spec and issues in reconciler. 

**To Reproduce**

Used this spec
```
vertices:
- name: in
  scale:
    min: 1
  # A self data generating source
  source:
    generator:
      rpu: 50
      duration: 1s
      # 1Mb
      msgSize: 1000000
  initContainerTemplate:
    resources:
      limits:
        cpu: "4"
        memory: 1Gi
      requests:
        cpu: "4"
        memory: 1Gi
  containerTemplate:
    resources:
      limits:
        cpu: "4"
        memory: 8Gi
      requests:
        cpu: "4"
        memory: 8Gi

```

**Expected behavior**
Container template should only be applied to the numa container only.

**Screenshots**
```
skohli@macos-JQWR9T560R manifests % kubectl get vtx
NAME                  PHASE    REASON            MESSAGE                                                                                                                                                            DESIRED   CURRENT
simple-pipeline-in    Failed   CreatePodFailed   Pod "simple-pipeline-in-0-4nmul" is invalid: spec.initContainers[0].resources.requests: Invalid value: "8Gi": must be less than or equal to memory limit of 1Gi    1         0
simple-pipeline-out   Failed   CreatePodFailed   Pod "simple-pipeline-out-0-yg7w1" is invalid: spec.initContainers[0].resources.requests: Invalid value: "8Gi": must be less than or equal to memory limit of 1Gi   1         0
```

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

>For quick help and support, join our [slack](https://numaproj.slack.com/join/shared_invite/zt-19svuv47m-YKHhsQ~~KK9mBv1E7pNzfg#/shared-invite/email) channel.

kohlisid:
For anyone looking at this, as a workaround please allocate same resources to init container and container template. That should not block the pod startup.

PR #1948, opened 2024-08-15
fix: default resources mutated when applying templates

whynowy:
Fixes: #1906


vigith (approved):

kohlisid (approved):
@whynowy Could there be other fields which might get affected due to similar behavior?

PR #1954, opened 2024-08-16
feat: add health for monovertex

kohlisid:
fixes https://github.com/numaproj/numaflow/issues/1952
For data health, this adds critical status, warning status to be added in a follow up


pkg/apis/numaflow/v1alpha1/mono_vertex_types.go
```diff
@@ -529,14 +529,21 @@ func (mvs *MonoVertexStatus) MarkPhaseRunning() {
 }
 
 // IsHealthy indicates whether the MonoVertex is in healthy status
-func (mvs *MonoVertexStatus) IsHealthy() bool {
+// It returns an error if any issues exists
+// No errors indicate that the MonoVertex is healthy
+// TODO: Add support for paused whenever added in MonoVtx?
+func (mvs *MonoVertexStatus) IsHealthy() error {
```
whynowy, line 535: Instead changing this function, use the `reason` and `message` for unhealthy message.

kohlisid: To extract the `reason` and `message` from the conditions I had updated the message. Do you prefer extracting that directly from the new function `mvs.IsReadyWithError()` instead for that? And not passing it through this one

kohlisid: Okay, Ignore the above understood what you meant. Looked into the spec.

I had looked into healthy status message, thus reason and message were not present hence I didn't realise that we would set those fields are also present when there is an error. 

Will update the code. 

```
status:
  conditions:
  - lastTransitionTime: "2024-08-13T17:09:16Z"
    message: Successful
    reason: Successful
    status: "True"
    type: DaemonHealthy
  - lastTransitionTime: "2024-08-13T17:09:10Z"
    message: Successful
    reason: Successful
    status: "True"
    type: Deployed
  - lastTransitionTime: "2024-08-13T17:09:10Z"
    message: All vertex pods are healthy
    reason: Running
    status: "True"
    type: PodsHealthy
  lastScaledAt: "2024-08-13T17:09:10Z"
  lastUpdated: null
  observedGeneration: 1
  phase: Running
  replicas: 1
  selector: app.kubernetes.io/component=mono-vertex,numaflow.numaproj.io/mono-vertex-name=simple-mono-vertex
```

pkg/apis/proto/mvtxdaemon/mvtxdaemon.proto
```diff
@@ -37,11 +37,30 @@ message GetMonoVertexMetricsResponse {
...
 
+// MonoVertexStatus is used to provide information about the mono vertex status.
+message MonoVertexStatus {
+  string status = 1;
+  string message = 2;
+  string code = 3;
+}
+
+message GetMonoVertexStatusRequest {
+  string monovertex = 1;
```
whynowy, line 48: Using `google.protobuf.Empty` for the request is good enough.

kohlisid: Sure! For now I had copied over the pipeline template and was just adding the implementation. 
We wouldn't need it the pipeline API as well I checked. 
I'll update for both

kohlisid:
```
  // GetPipelineWatermarks return the watermark of the given pipeline
  rpc GetPipelineWatermarks (GetPipelineWatermarksRequest) returns (GetPipelineWatermarksResponse) {
    option (google.api.http).get = "/api/v1/pipelines/{pipeline}/watermarks";
  };

  rpc GetPipelineStatus (GetPipelineStatusRequest) returns (GetPipelineStatusResponse) {
    option (google.api.http).get = "/api/v1/pipelines/{pipeline}/status";
  };
```
Actually, we won't need for these 2 pipeline level APIs to daemon as well

whynowy: This is right.

kohlisid: @whynowy 
Sorry just to clarify, you mean that you are good with the current implementation or are you good to change them to empty?
We don't use the request so I believe we can update to empty

whynowy: Let's use empty for MonoVertex. For Pipeline, we need to do several iterations if we want to correct it, leave it for now.

kohlisid: Sure, Will create an issue to track that and refactor for later!

kohlisid:
```
https://localhost:8443/api/v1/namespaces/default/mono-vertices/simple-mono-vertex/health
{
    "data": {
        "resourceHealthStatus": "healthy",
        "dataHealthStatus": "healthy",
        "resourceHealthMessage": "mono vertex \"simple-mono-vertex\" is healthy",
        "dataHealthMessage": "MonoVertex data flow is healthy",
        "resourceHealthCode": "M1",
        "dataHealthCode": "D1"
    },
    "errMsg": null
}
```

kohlisid:
```
https://localhost:4327/api/v1/status
{
  "status": {
    "status": "healthy",
    "message": "MonoVertex data flow is healthy",
    "code": "D1"
  }
}
```

pkg/mvtxdaemon/server/service/health_status.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+// getMonoVertexDataCriticality is used to provide the data criticality of the MonoVertex
+// They can be of the following types:
+// 1. Healthy: The MonoVertex is working as expected
+// 2. Warning: The MonoVertex is working but there could be a lag in the data movement
+// 3. Critical: The MonoVertex is not working as expected
+// We need to check the following things to determine the data criticality of the MonoVertex:
+// At any given instant of time what is the desired number of replicas required by the MonoVertex
+// to clear out the backlog in the target state time.
```
vigith, line 125: What about the case when pods are continuously being restarted? or what if the TPS is low and pending and is growing very slowly?

vigith (approved):

pkg/mvtxdaemon/server/service/health_status.go
vigith: It looks like cases like pod restarting will be caught by resource health.

whynowy (approved):

Issue #1960, opened 2024-08-19
MonoVertex gets autoscaled to zero replicas when the pending is zero and processing rate is not zero


yhl25:
**Describe the bug**
MonoVertex gets autoscaled to zero replicas when the pending is zero and the processing rate is not zero

**To Reproduce**

* Create a monovertex with the udsource that returns pending as zero and the read rate is non zero

**Expected behavior**
Since the processing rate is not zero, it should not autoscale to zero replicas.

kohlisid:
Can you share the spec that you used?

I tried with our example spec and latest main and was seeing the field getting populated. 

Maybe was there any other steps that I'm missing here. Or should I wait for some more time etc?


```
spec:
  replicas: 1
  sink:
    udsink:
      container:
        image: quay.io/numaio/numaflow-java/simple-sink:stable
  source:
    transformer:
      container:
        image: quay.io/numaio/numaflow-rs/source-transformer-now:stable
    udsource:
      container:
        image: quay.io/numaio/numaflow-java/source-simple-source:stable
```

```
{
  "metrics": {
    "monoVertex": "simple-mono-vertex",
    "processingRates": {
      "15m": 16117.8571428571,
      "1m": 16091.6666666667,
      "5m": 16117.8571428571,
      "default": 16133.3333333333
    },
    "pendings": {
      "15m": "0",
      "1m": "0",
      "5m": "0",
      "default": "0"
    }
  }
}
```

whynowy:
Is it the latest code?

kohlisid:
Triaged the issue, the following are the observations

1)To reproduce this, we need to apply the spec in a way which causes the pod to restart
2) In a restart scenario, the daemon server is also restarted and hence the rater metrics do not have the processing rates updated initially.
3) With the source always sending pending as 0, the scaler will set the desired replicas to 0 ([rate = 0 and pending =0 ](https://github.com/numaproj/numaflow/blob/42671138250d67f6eacddf33d4b5d5e069e5674f/pkg/reconciler/monovertex/scaling/scaling.go#L273))
4) This causes the scale down

The rater should be sending a negative value to indicate that the ProcessingRate is not available, 0 can be misleading.

On seeing a negative value we will not scale

PR #1966, opened 2024-08-19
chore: send negative value rateNotAvailable

kohlisid:
fixes https://github.com/numaproj/numaflow/issues/1960

When the rater cannot calculate the processing rate due to any missing data, we should send a negative value to indicate that the rate is not available instead of 0
This helps with a corner case where the scaler can consider this as the TPS to be 0 and cause unnecessary scale downs.

### DIFF
--- a/pkg/daemon/server/service/rater/helper.go
+++ b/pkg/daemon/server/service/rater/helper.go
@@ -17,6 +17,7 @@ limitations under the License.
 package rater
 
 import (
+	"math"
 	"time"
 
 	sharedqueue "github.com/numaproj/numaflow/pkg/shared/queue"
@@ -24,6 +25,9 @@ import (
 
 const (
 	indexNotFound = -1
+	// rateNotAvailable is returned when the processing rate cannot be derived from the currently
+	// available pod data, a negative min is returned to indicate this.
+	rateNotAvailable = float64(math.MinInt)
 )
 
 // UpdateCount updates the count of processed messages for a pod at a given time
@@ -48,14 +52,14 @@ func UpdateCount(q *sharedqueue.OverflowQueue[*TimestampedCounts], time int64, p
 func CalculateRate(q *sharedqueue.OverflowQueue[*TimestampedCounts], lookbackSeconds int64, partitionName string) float64 {
 	counts := q.Items()
 	if len(counts) <= 1 {
-		return 0
+		return rateNotAvailable
 	}
 	startIndex := findStartIndex(lookbackSeconds, counts)
 	// we consider the last but one element as the end index because the last element might be incomplete
 	// we can be sure that the last but one element in the queue is complete.
 	endIndex := len(counts) - 2
 	if startIndex == indexNotFound {
-		return 0
+		return rateNotAvailable
 	}
 
 	// time diff in seconds.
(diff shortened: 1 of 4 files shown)


vigith (commented):
how is this handled in `Pipeline`?

whynowy (approved):

kohlisid:
@vigith 
Yes, same behavior in Pipeline

The flow is the same for that as well as documented here
https://github.com/numaproj/numaflow/issues/1960#issuecomment-2297415677

pkg/daemon/server/service/rater/helper.go
```diff
@@ -17,13 +17,17 @@ limitations under the License.
...
 	"time"
 
 	sharedqueue "github.com/numaproj/numaflow/pkg/shared/queue"
 )
 
 const (
 	indexNotFound = -1
+	// rateNotAvailable is returned when the processing rate cannot be derived from the currently
+	// available pod data, a negative min is returned to indicate this.
+	rateNotAvailable = float64(math.MinInt)
```
vigith, line 30: please make sure that the -ve val returned [by ud-source gRPC](https://github.com/numaproj/numaflow/blob/main/pkg/apis/proto/source/v1/source.proto#L139-L141) is set as `rateNotAvailable`

vigith (approved):

kohlisid:
I see another case in Pipeline though which we should potentially address, There can be scenario that we have a lookup timeline but there is no entry for a specific partition in that. But in the calculation we would send 0 as we do a simple lookup

Just trying to think whether sending 0 as rate here can have any side effects or not.

https://github.com/numaproj/numaflow/blob/42671138250d67f6eacddf33d4b5d5e069e5674f/pkg/daemon/server/service/rater/helper_test.go#L330

pkg/daemon/server/service/rater/helper_test.go
```diff
@@ -122,14 +122,14 @@ func TestCalculateRate(t *testing.T) {
 	t.Run("givenCollectedTimeLessThanTwo_whenCalculateRate_thenReturnZero", func(t *testing.T) {
```
KeranYang, line 122:
```suggestion
	t.Run("givenCollectedTimeLessThanTwo_whenCalculateRate_thenReturnRateNotAvailable", func(t *testing.T) {
```

pkg/mvtxdaemon/server/service/rater/helper_test.go
```diff
@@ -109,14 +109,14 @@ func TestCalculateRate(t *testing.T) {
 	t.Run("givenCollectedTimeLessThanTwo_whenCalculateRate_thenReturnZero", func(t *testing.T) {
```
KeranYang, line 109:
```suggestion
	t.Run("givenCollectedTimeLessThanTwo_whenCalculateRate_thenReturnRateNotAvailable", func(t *testing.T) {
```

KeranYang (commented):
We have the following calls `result[n] = wrapperspb.Double(rate)`, could you test to verify it works well with MinInt?

kohlisid:
@KeranYang I have added it to the Unit tests, they seem to be passing correctly. 
I do wrap the MinInt with a float64 as well for consistency

PR #1967, opened 2024-08-19
fix: minor perf improvements of mvtx fallback sink

vigith:
Explain what this PR does.


rust/monovertex/src/forwarder.rs
```diff
@@ -205,67 +210,62 @@ impl Forwarder {
...
+                    // if all messages are successfully written, break the loop
+                    if messages_to_send.is_empty() {
+                        break;
                     }
+
+                    warn!(
+                        "Retry attempt {} due to retryable error. Errors: {:?}",
+                        attempts, error_map
+                    );
+                    error_map.clear();
```
vigith, line 258: if we clear here, how will we return it in the `Err` ?

yhl25: updated to handle that case.

yhl25 (approved):

PR #1975, opened 2024-08-20
fix: remove coloring in logs

vigith:
Explain what this PR does.


vigith:
is there a way to detect whether there is ptty and turn on coloring conditionally?

whynowy (approved):

PR #1985, opened 2024-08-21
chore: metrics and https support for serving

yhl25:


vigith:
we need to move to tower 0.5.0, but it has a few breaking changes. perhaps as a standalone PR.

rust/serving/src/config.rs
```diff
@@ -33,12 +35,23 @@ pub fn config() -> &'static Settings {
     })
 }
 
+static GLOBAL_TLS_CONFIG: OnceLock<(Certificate, KeyPair)> = OnceLock::new();
```
vigith, line 38: move it to utils?

vigith (commented):
LGTM

vigith (approved):

PR #1997, opened 2024-08-23
fix: e2e testing isbsvc deletion timeout issue

KeranYang:
Based on logs, there can be a case when a) isbsvc CRD is deleted, AND b) stateful set is deleted, but one of the isbsvc-managed pods still stay in RUNNING status, causing the e2e test cleanup time out. The timeout issue has been transient but frequent. Since the fix is for e2e testing hence not critical, I choose to force delete pods as a workaround.

Testing: ran it 10+ times, no more isbsvc deletion timeout issue.


test/fixtures/expect.go
```diff
@@ -95,6 +101,26 @@ func (t *Expect) ISBSvcDeleted(timeout time.Duration) *Expect {
...
+			_, err = t.kubeClient.AppsV1().StatefulSets(Namespace).Get(ctx, fmt.Sprintf("isbsvc-%s-js", ISBSvcName), metav1.GetOptions{})
+			stsIsDeleted := apierr.IsNotFound(err)
+			for _, pod := range podList.Items {
+				podHasFinalizer := len(pod.Finalizers) > 0
+				podIsRunning := pod.Status.Phase == "Running"
+				if stsIsDeleted && !podHasFinalizer && podIsRunning {
+					// the stateful set has been deleted, the pod has no finalizer and is running
+					// we can safely force to delete the pod
+					t.t.Logf("The statefulset has been deleted, the pod %s has no finalizer and is running, force to delete it", pod.Name)
+					deleteOpt := metav1.DeleteOptions{GracePeriodSeconds: new(int64)}
```
whynowy, line 115: This approach is too hacky.

test/fixtures/e2e_suite.go
```diff
@@ -139,13 +142,18 @@ func (s *E2ESuite) TearDownSuite() {
 		When().
 		Wait(5 * time.Second).
 		DeleteISBSvc().
-		Wait(3 * time.Second).
+		Wait(3 * time.Second)
+	// force deleting the ISB svc pods because we have seen pods stuck in terminating state after CRD deletion,
+	// which causes e2e tests to timeout, this is a workaround to avoid the issue.
+	deleteISBPodsCMD := fmt.Sprintf("kubectl delete pods -n %s -l %s=%s,%s=%s --ignore-not-found=true", Namespace, dfv1.KeyComponent, dfv1.ComponentISBSvc, dfv1.KeyISBSvcName, ISBSvcName)
```
whynowy, line 148: `--grace-period=0 --force`

whynowy (approved):

PR #1999, opened 2024-08-23
chore: fix servesink

yhl25:


rust/servesink/src/lib.rs
```diff
@@ -30,43 +34,189 @@ impl Logger {
...
+    use super::*;
+    use numaflow::sink::{SinkRequest, Sinker};
+    use std::collections::HashMap;
+    use tokio::io::{AsyncReadExt, AsyncWriteExt};
+    use tokio::net::TcpListener;
+    use tokio::sync::mpsc;
+
+    #[tokio::test]
+    async fn test_serve_sink_without_url_header() {
+        let logger = ServeSink::new();
```
vigith, line 102: typo?

vigith (approved):
Should we move this code inside serving in a follow up PR?

rust/servesink/src/lib.rs
yhl25: fixed.

PR #2000, opened 2024-08-23
fix: log format with config load error

xdevxy:
A quick fix for log format error when loading config in the controller.


pkg/reconciler/cmd/start.go
```diff
@@ -51,7 +51,7 @@ import (
 func Start(namespaced bool, managedNamespace string) {
 	logger := logging.NewLogger().Named("controller-manager")
 	config, err := reconciler.LoadConfig(func(err error) {
-		logger.Errorf("Failed to reload global configuration file", zap.Error(err))
+		logger.Errorf("Failed to reload global configuration file: %v", zap.Error(err))
```
whynowy, line 54: Maybe just update it to `Errorw`.

Issue #2007, opened 2024-08-27
Is it possible to support DRA(Dynamic Resource Allocation)?


sesame0224:
# Summary
In recent years, accelerators have become essential for data processing infrastructure. Is there a plan to support accelerators in the numaflow project?

Recently, [DRA](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/) has been proposed as a mechanism to easily use accelerators in Kubernetes pods. DRA allocates accelerator resources to pods using a manifest file.

In the following example, a fictional resource driver is defined as a "DeviceClass," and the spec and some metadata for the fictional resources (e.g., GPU, FPGA) managed by this driver are defined as a "ResourceClaimTemplate."

Then, in the pod manifest, a request for the allocation of fictional accelerator resources is made via a "ResourceClaim."
```
apiVersion: resource.k8s.io/v1alpha3
kind: DeviceClass
name: resource.example.com
spec:
  selectors:
  - cel:
      expression: device.driver == "resource-driver.example.com"
---
apiVersion: resource.k8s.io/v1alpha2
kind: ResourceClaimTemplate
metadata:
  name: large-black-cat-claim-template
spec:
  spec:
    devices:
      requests:
      - name: req-0
        deviceClassName: resource.example.com
        selectors:
        - cel:
           expression: |-
              device.attributes["resource-driver.example.com"].color == "black" &&
              device.attributes["resource-driver.example.com"].size == "large"              
–--
apiVersion: v1
kind: Pod
metadata:
  name: pod-with-cats
spec:
  containers:
  - name: container0
    image: ubuntu:20.04
    command: ["sleep", "9999"]
    resources:
      claims:
      - name: cat-0
  - name: container1
    image: ubuntu:20.04
    command: ["sleep", "9999"]
    resources:
      claims:
      - name: cat-1
  resourceClaims:
  - name: cat-0
    resourceClaimTemplateName: large-black-cat-claim-template
  - name: cat-1
    resourceClaimTemplateName: large-black-cat-claim-template
```
In this case, resources are allocated to the pod if the device's attributes match, such as having a black color and a large size.

A more practical example can be found [here](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/4381-dra-structured-parameters/README.md#partial-gpu-allocation).

With the use of DRA, is it possible that UDFs in numaflow will support the use of accelerator resources?

# Use Cases

[...]

PR #2009, opened 2024-08-28
feat: enable resourceClaims for vertex and monovtx

whynowy:
Closes: #2007

Can be added in the spec of `Pipeline` or `MonoVertex`

```yaml
kind: Pipeline
spec:
  vertices:
  - name: my-vtx
      resourceClaims:
      - name: cat-0
        source: xxxxx
    containerTemplate: # for "numa" container
      resources:
        claims:
        - name: cat-0
    udf:
      container:
        resources:
          claims:
          - name: cat-xxx
```


vigith (approved):
add this in the doc?

PR #2010, opened 2024-08-29
feat: allow configurable retryStrategy

### LINKED ISSUE #1951: Expose Retry Count and Retry Timeout For Sink
# Summary

Currently, we have infinite retry with a fixed interval of 1 ms, which can lead to problems when the sink is in a bad state. We should provide an option for the user to configure max retry timeout interval. 

We should also expose Retry Count so the user can either drop the message based on the number of retries or write to the fallback sink.

Note: Retry is an optimistic retry. The max retry can be >= the user-configured retry count because the counter will be reset each time the pod is restarted.

# Use Cases

When would you use this?
* Users want to drop the message after 'N' attempts.
* Users don't want to write retry logic.

---

```[tasklist]
### Tasks
- [ ] Add retryStrategy to Sink CRD
- [ ] Fixed retry for Pipeline
- [ ] Fixed retry for MonoVertex
- [ ] Exponential retry for Pipeline
- [x] ~Exponential retry for MonoVertex~
- [ ] https://github.com/numaproj/numaflow/issues/2016
- [ ] Expose retry count from sinks
```

kohlisid:
fixes #1951 

For testing refer - https://github.com/numaproj/numaflow/issues/2016


pkg/apis/numaflow/v1alpha1/retry_strategy.go
```diff
@@ -0,0 +1,58 @@
...
+limitations under the License.
+*/
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type OnFailRetryStrategy string
```
whynowy, line 20: OnFailureRetryStrategy

pkg/apis/numaflow/v1alpha1/retry_strategy.go
```diff
@@ -0,0 +1,58 @@
+/*
```
whynowy, line 1: This header is different from others.

pkg/sinks/forward/forward.go
```diff
@@ -357,76 +357,120 @@ func (df *DataForward) ackFromBuffer(ctx context.Context, offsets []isb.Offset)
...
-				if errors.Is(err, &udsink.WriteToFallbackErr) && df.opts.fbSinkWriter != nil && !isFbSinkWriter {
-					fallbackMessages = append(fallbackMessages, msg)
-					continue
-				}
+		err = wait.ExponentialBackoffWithContext(ctx, backoffCond, func(_ context.Context) (done bool, err error) {
+			// Note: this is an unwanted memory allocation during a happy path. We want only minimal allocation
+			// since using failedMessages is an unlikely path.
+			var failedMessages []isb.Message
+			needRetry := false
+			df.opts.logger.Info("MYDEBUG: Trying sink write at: ", time.Now().String())
```
vigith, line 385: remove these?

pkg/sinks/forward/forward.go
```diff
@@ -357,76 +357,120 @@ func (df *DataForward) ackFromBuffer(ctx context.Context, offsets []isb.Offset)
...
+		// Check if we still have messages left to be processed
+		if len(messagesToTry) > 0 {
+			df.opts.logger.Infof("%d tries exhausted in sink, messagesLeft %d, Next strategy %s",
+				backoffCond.Steps, len(messagesToTry), failStrategy)
+			// Check what is the failure strategy to be followed after retry exhaustion
+			switch failStrategy {
+			case dfv1.OnFailRetry:
+				// If on failure, we keep on retrying then lets continue the loop and try all again
+				continue
+			case dfv1.OnFailFallback:
```
vigith, line 455: add metrics for Fallback?

rust/monovertex/src/config.rs
```diff
@@ -19,8 +19,9 @@ const DEFAULT_LAG_CHECK_INTERVAL_IN_SECS: u16 = 5;
 const DEFAULT_LAG_REFRESH_INTERVAL_IN_SECS: u16 = 3;
 const DEFAULT_BATCH_SIZE: u64 = 500;
 const DEFAULT_TIMEOUT_IN_MS: u32 = 1000;
-const DEFAULT_MAX_SINK_RETRY_ATTEMPTS: u16 = 10;
-const DEFAULT_SINK_RETRY_INTERVAL_IN_MS: u32 = 1;
+pub const DEFAULT_MAX_SINK_RETRY_ATTEMPTS: u16 = u16::MAX - 1;
+pub const DEFAULT_SINK_RETRY_INTERVAL_IN_MS: u32 = 1;
```
vigith, line 23: can we avoid this to be `pub`?

rust/monovertex/src/config.rs
```diff
@@ -117,9 +120,33 @@ impl Settings {
             settings.is_fallback_enabled = mono_vertex_obj
                 .spec
                 .sink
+                .clone()
                 .ok_or(Error::ConfigError("Sink not found".to_string()))?
                 .fallback
                 .is_some();
+
+            if let Some(retry_strategy) = mono_vertex_obj.spec.sink.clone().unwrap().retry_strategy
```
vigith, line 128:
```suggestion
            if let Some(retry_strategy) = mono_vertex_obj.spec.sink.clone().expect("sink should not be empty").retry_strategy
```

vigith (commented):
need to review rust code, just skimmed over. will review after tests are added.

pkg/sinks/forward/forward.go
kohlisid: Added seperate metrics for FB on vertex and monovertex

pkg/metrics/metrics.go
```diff
@@ -86,7 +86,7 @@ var (
 		Help:      "Total number of bytes written",
 	}, []string{LabelVertex, LabelPipeline, LabelVertexType, LabelVertexReplicaIndex, LabelPartitionName})
 
-	// WriteMessagesError is used to indicate the number of errors messages written
+	// WriteMessagesError is used to indicate the number of errors while writing to a sink
```
whynowy, line 89: Why is `writing to a sink`?

pkg/sinks/sink.go
```diff
@@ -246,6 +246,14 @@ func (u *SinkProcessor) Start(ctx context.Context) error {
 			forwardOpts = append(forwardOpts, sinkforward.WithCallbackUploader(cbPublisher))
 		}
 
+		// Derive the retryStrategy from the spec and add to the forwarder if it is valid
+		retryStrategy := u.VertexInstance.Vertex.Spec.Sink.GetRetryStrategy()
+
+		if ok := u.VertexInstance.Vertex.Spec.Sink.IsValidSinkRetryStrategy(retryStrategy); ok != nil {
```
whynowy, line 252: These functions do not make sense - you already have the sink object, you get the retry strategy from the object, and then call a function of the sink object to tell if the strategy is valid...

kohlisid: My bad, oversight.
Have updated to check for validation while fetching retryStrategy

pkg/apis/numaflow/v1alpha1/vertex_types.go
```diff
@@ -643,7 +643,7 @@ func (av AbstractVertex) HasUDTransformer() bool {
 }
 
 func (av AbstractVertex) HasFallbackUDSink() bool {
-	return av.IsASink() && av.Sink.Fallback != nil && av.Sink.Fallback.UDSink != nil
+	return av.IsASink() && av.Sink.hasValidFallbackSink()
```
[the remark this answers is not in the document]
kohlisid: `hasValidFallbackSink()` function just checks those original two conditions together, the same logic. 

```
// hasValidFallbackSink checks if the Sink vertex has a valid fallback sink configured
func (s *Sink) hasValidFallbackSink() bool {
	return s.Fallback != nil && s.Fallback.UDSink != nil
}
```

pkg/sinks/forward/forward.go
```diff
@@ -357,82 +357,190 @@ func (df *DataForward) ackFromBuffer(ctx context.Context, offsets []isb.Offset)
...
+// updateSinkWriteMetrics updates metrics related to data writes to a sink.
+// Metrics are updated based on whether the operation involves the primary or fallback sink.
+func (df *DataForward) updateSinkWriteMetrics(writeCount int, writeBytes float64, sinkWriterName string, isFallback bool) {
+	// Use default metrics for primary sink operations
+	writeCountMetric := metrics.WriteMessagesCount
+	writeBytesMetric := metrics.WriteBytesCount
+
+	// Assign fallback-specific metrics for both write count and write bytes
+	// if this is for Fallback Sink
+	if isFallback {
```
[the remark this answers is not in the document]
kohlisid: So you want WriteMessages to be the total for fallback + primary?
I was keeping this for primary sink and fallback metric for the other sink separately

pkg/apis/numaflow/v1alpha1/sink.go
```diff
@@ -134,3 +139,20 @@ func (s Sink) getFallbackUDSinkContainer(mainContainerReq getContainerReq) corev
...
 }
+
+// hasValidFallbackSink checks if the Sink vertex has a valid fallback sink configured
+func (s *Sink) hasValidFallbackSink() bool {
+	return s.Fallback != nil && s.Fallback.UDSink != nil
+}
+
+// HasValidSinkRetryStrategy checks if the provided RetryStrategy is valid based on the sink's configuration.
+// This validation ensures that the retry strategy is compatible with the sink's current setup
+func (s Sink) HasValidSinkRetryStrategy() error {
```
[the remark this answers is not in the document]
kohlisid: Can move to validation

pkg/sinks/sink.go
```diff
@@ -246,6 +246,9 @@ func (u *SinkProcessor) Start(ctx context.Context) error {
 			forwardOpts = append(forwardOpts, sinkforward.WithCallbackUploader(cbPublisher))
 		}
 
+		// Derive the retryStrategy from the spec and add to the forwarder
+		forwardOpts = append(forwardOpts, sinkforward.WithRetryStrategy(u.VertexInstance.Vertex.Spec.Sink.RetryStrategy))
```
whynowy, line 250: Minor: we pass `u.VertexInstance` to `NewDataForward()`, which means everything in the spec is available there.

pkg/sinks/forward/forward.go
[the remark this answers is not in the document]
whynowy: yes. or we have to sum these 2 metrics to know the total writes for sinks.

pkg/sinks/forward/forward.go
```diff
@@ -443,3 +551,20 @@ func errorArrayToMap(errs []error) map[string]int64 {
...
+		return wait.Backoff{
+			Duration: dfv1.DefaultRetryInterval,
+			// Always account for the first try
+			Steps: dfv1.DefaultRetrySteps + 1,
+		}, dfv1.OnFailureRetry
+	}
+	// Initial interval duration and number of retries are taken from DataForward settings.
+	backoff := df.opts.retryStrategy.GetBackoff()
+	// Adding a step for the first try
+	backoff.Steps = backoff.Steps + 1
```
whynowy, line 568: If we need to do this trick, why don't we correct the definition in the spec?

kohlisid: I had thought about it, just one corner case where the user gives 0 somehow in the spec. Then there will be no sink writes happening.
This just ensures that we try `atleast` once, regardless of what the user gives us. 
I couldn't think a valid case where the user would not want that.

rust/monovertex/src/forwarder.rs
```diff
@@ -222,85 +224,140 @@ impl Forwarder {
...
+                    }
+                }
+            }
+            break;
         }
 
+        let fallback_msgs_count = fallback_msgs.len() as u64;
         // If there are fallback messages, write them to the fallback sink
-        if !fallback_msgs.is_empty() {
+        if fallback_msgs_count > 0 {
```
vigith, line 286: why did you change this?

kohlisid: `let fallback_msgs_count = fallback_msgs.len() as u64;`
I had extracted the count because I had to use it for metrics, have updated there. So can revert this now

rust/monovertex/src/forwarder.rs
```diff
@@ -222,67 +223,43 @@ impl Forwarder {
...
-                    warn!(
-                        "Retry attempt {} due to retryable error. Errors: {:?}",
-                        attempts, error_map
-                    );
-                    sleep(tokio::time::Duration::from_millis(
-                        config().sink_retry_interval_in_ms as u64,
-                    ))
+        // only breaks out of this loop based on the retry strategy unless all the messages have been written to sink
+        // successfully.
+        loop {
```
vigith, line 228: remove this loop as there are mutually exclusion operations happening inside the same loop.

rust/monovertex/src/config.rs
```diff
@@ -17,8 +17,9 @@ const DEFAULT_LAG_CHECK_INTERVAL_IN_SECS: u16 = 5;
 const DEFAULT_LAG_REFRESH_INTERVAL_IN_SECS: u16 = 3;
 const DEFAULT_BATCH_SIZE: u64 = 500;
 const DEFAULT_TIMEOUT_IN_MS: u32 = 1000;
-const DEFAULT_MAX_SINK_RETRY_ATTEMPTS: u16 = 10;
+const DEFAULT_MAX_SINK_RETRY_ATTEMPTS: u16 = u16::MAX;
 const DEFAULT_SINK_RETRY_INTERVAL_IN_MS: u32 = 1;
+const DEFAULT_SINK_RETRY_ON_FAIL_STRATEGY: &str = "retry";
```
vigith, line 22: convert this to a ENUM

vigith (approved):
feel free to create follow up PRs for these.

whynowy (approved):

PR #2011, opened 2024-08-29
chore: use `kube_codegen.sh` instead in codegen.

xdevxy:
`make codegen` has the following warning,
```
WARNING: generate-internal-groups.sh is deprecated.
WARNING: Please use k8s.io/code-generator/kube_codegen.sh instead.

```
which can cause similar permission denied issue as https://github.com/numaproj/numaplane/pull/214

### DIFF
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -17,8 +17,6 @@ cd "${FAKE_REPOPATH}"
 
 CODEGEN_PKG=${CODEGEN_PKG:-$(cd "${FAKE_REPOPATH}"; ls -d -1 ./vendor/k8s.io/code-generator 2>/dev/null || echo ../code-generator)}
 
-chmod +x ${CODEGEN_PKG}/*.sh
-
 subheader "running codegen"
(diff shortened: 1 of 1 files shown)


kohlisid:
Hey @xdevxy 
Just curios did any other things change along with the script if you have taken a look into that?
Otherwise I would do a spike on that just to understand the transition a bit better

hack/update-codegen.sh
```diff
@@ -20,12 +20,12 @@ CODEGEN_PKG=${CODEGEN_PKG:-$(cd "${FAKE_REPOPATH}"; ls -d -1 ./vendor/k8s.io/cod
 chmod +x ${CODEGEN_PKG}/*.sh
 
 subheader "running codegen"
-bash -x ${CODEGEN_PKG}/generate-groups.sh "deepcopy" \
+bash -x ${CODEGEN_PKG}/kube_codegen.sh "deepcopy" \
```
whynowy, line 23: Thank you! We don't need line 20 any more.

whynowy (approved):

vigith (approved):

PR #2029, opened 2024-09-05
feat: rolling update for MonoVertex

whynowy:
Rolling update for MonoVertex, which defaults to following strategy:

```yaml
  updateStrategy:
    rollingUpdate:
      maxUnavailable: 25% # Could be an integer or percentage
```

1. Rolling replace `maxUnavailable` pods, starting from replica 0, until all the pods are done.
2. Will not start the next round until the last batch has been done with all pods in `Ready` status;
3. When rolling update starts, autoscaling will be paused until the update is done.


pkg/mvtxdaemon/server/service/health_status.go
```diff
@@ -132,20 +132,18 @@ func (hc *HealthChecker) setCurrentHealth(status *dataHealthResponse) {
...
-	// Get the current state of the MonoVertex replicas
-	currentReplicas := hc.monoVertex.GetReplicas()
 	maxReplicas := int(hc.monoVertex.Spec.Scale.GetMaxReplicas())
 	// default status is healthy
 	status := v1alpha1.MonoVertexStatusHealthy
-	// If the current replicas are equal to the max replicas, and the desired replicas are more than the max replicas,
+	// If the desired replicas are more than the max replicas,
 	// the data criticality is Critical.
-	if currentReplicas == maxReplicas && desiredReplicas > maxReplicas {
+	if desiredReplicas > maxReplicas {
```
kohlisid, line 146: If we do not use the current replicas here, this could often be a false alarm right? As in sudden spikes the desired can go beyond maxReplicas many times

whynowy: @kohlisid - can you create an issue for tracking?

vigith:
@whynowy why not add few with the newer hash and then remove old? if we add new and start rotating, we will have to scale down at the end (because we have contiguous pod indexing), but it will avoid tail latency spikes.

whynowy:
> @whynowy why not add few with the newer hash and then remove old? if we add new and start rotating, we will have to scale down at the end (because we have contiguous pod indexing), but it will avoid tail latency spikes.

It will be very complex to do that since we need to use different strategies for MonoVertex, Non-reduce Vertex and Reduce Vertex. Actually there's no way to implement that based on our current model after some experiments - we currently use `MonoVertex -> Pod` and `Vertex -> Pod`, which is very hard to manage multiple pods with same replica index - because of the nature of k8s informer cache, that it can not reflect any create/update operation immediately, so we can not easily know if there's already a new pod created for each reconciliation. If we want to do like that, we have to introduce `ReplicaSet` in between, this is what `Deployment` and `Argo Rollout` do. The way we do right now, is like `StatefulSet`, which directly manages `Pods`, which is okay for this strategy.

vigith:
Could you please add a doc?

whynowy:
> Could you please add a doc?
Doc will be added when Vertex one is done.

vigith (approved):

PR #2048, opened 2024-09-09
chore: pin kube version in rust

whynowy:
Explain what this PR does.

### DIFF
--- a/rust/numaflow-models/Cargo.toml
+++ b/rust/numaflow-models/Cargo.toml
@@ -8,7 +8,7 @@ edition = "2021"
 
 [dependencies]
 k8s-openapi = { version = "0.22.0", features = ["v1_29"] }
-kube = "0.93.1"
+kube = "0.94.0"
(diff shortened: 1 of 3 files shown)


rust/numaflow-models/Cargo.toml
```diff
@@ -8,7 +8,7 @@ edition = "2021"
 
 [dependencies]
 k8s-openapi = { version = "0.22.0", features = ["v1_29"] }
-kube = "0.93.1"
+kube = "0.94.0"
```
vigith, line 11: any particular reason? we have marked `k8s-openapi` to include only v1_29 feature. we might have to pin `k8s-openapi` too.

whynowy: To avoid unexpected new release caused build problem

KeranYang:
Thanks!

KeranYang (approved):

Issue #2054, opened 2024-09-11
Using patch instead of update


whynowy:
# Summary

As a principle, we should never update the spec that comes from the user other than the `status`. Today, we are doing a `Update` call the add/remove a finalizer, which will also update the spec even though there's no spec change (but it might update the default values, e.g. an integer filed has no value in etcd, but an update will set it to 0).

The solution would be using patch to update the finalizer.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #2059, opened 2024-09-13
chore: patch instead of update and bugfix

whynowy:
Closes: #2054

1. Patch instead of update for finalizers change to avoid some of the fields getting updated unexpectedly (e.g. `cpu: 2000m` is updated to `cpu: 2`)
2. Use explicit spec change to patch for pipeline pause/resume, scale replicas, etc.
3. Fixed a bug for pipeline vertex, that every time there's a spec change, it updates the replica to `min`.
4. Fixed a corner case bug, that somehow replicas goes to 0 (might have manually scaled to 0, or some weird testing case), and `updateHash != currentHash`, it then it gets stuck.

### DIFF
--- a/pkg/reconciler/isbsvc/controller.go
+++ b/pkg/reconciler/isbsvc/controller.go
@@ -18,17 +18,20 @@ package isbsvc
 
 import (
 	"context"
+	"strings"
 
 	"go.uber.org/zap"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/yaml"
 
 	dfv1 "github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
 	"github.com/numaproj/numaflow/pkg/reconciler"
@@ -58,7 +61,7 @@ func (r *interStepBufferServiceReconciler) Reconcile(ctx context.Context, req ct
 	isbSvc := &dfv1.InterStepBufferService{}
 	if err := r.client.Get(ctx, req.NamespacedName, isbSvc); err != nil {
 		if apierrors.IsNotFound(err) {
-			return reconcile.Result{}, nil
+			return ctrl.Result{}, nil
 		}
 		r.logger.Errorw("Unable to get ISB Service", zap.Any("request", req), zap.Error(err))
 		return ctrl.Result{}, err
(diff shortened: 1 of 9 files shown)


pkg/reconciler/vertex/controller.go
```diff
@@ -259,6 +259,8 @@ func (r *vertexReconciler) orchestratePods(ctx context.Context, vertex *dfv1.Ver
 		vertex.Status.CurrentHash = vertex.Status.UpdateHash
 	} else { // Update scenario
 		if updatedReplicas >= desiredReplicas {
+			vertex.Status.UpdatedReplicas = uint32(desiredReplicas)
```
whynowy, line 262: Fixing 4.)

pkg/reconciler/pipeline/controller.go
```diff
@@ -292,7 +295,9 @@ func (r *pipelineReconciler) reconcileFixedResources(ctx context.Context, pl *df
 			r.recorder.Eventf(pl, corev1.EventTypeNormal, "CreateVertexSuccess", "Created vertex %s successfully", vertexName)
 		} else {
 			if oldObj.GetAnnotations()[dfv1.KeyHash] != newObj.GetAnnotations()[dfv1.KeyHash] { // need to update
+				originReplicas := oldObj.Spec.Replicas
```
whynowy, line 298: Fixing 3.)

pkg/reconciler/pipeline/controller.go
```diff
@@ -292,7 +295,9 @@ func (r *pipelineReconciler) reconcileFixedResources(ctx context.Context, pl *df
 			r.recorder.Eventf(pl, corev1.EventTypeNormal, "CreateVertexSuccess", "Created vertex %s successfully", vertexName)
 		} else {
 			if oldObj.GetAnnotations()[dfv1.KeyHash] != newObj.GetAnnotations()[dfv1.KeyHash] { // need to update
+				originalReplicas := oldObj.Spec.Replicas
```
whynowy, line 298: Fixing 3.)

pkg/reconciler/isbsvc/controller.go
```diff
@@ -122,13 +126,14 @@ func (r *interStepBufferServiceReconciler) reconcile(ctx context.Context, isbSvc
 	return installer.Install(ctx, isbSvc, r.client, r.kubeClient, r.config, log, r.recorder)
 }
 
-func (r *interStepBufferServiceReconciler) needsUpdate(old, new *dfv1.InterStepBufferService) bool {
-	if old == nil {
-		return true
+func needsToPatchFinalizers(old, new *dfv1.InterStepBufferService) bool {
+	if old == nil { // This is a weird scenario, nothing we can do. Theoretically it will never happen.
```
vigith, line 130: since it is a weird scenario should be log so it might help whoever is debugging?

pkg/reconciler/pipeline/controller.go
```diff
@@ -51,6 +51,8 @@ import (
 
 const (
 	finalizerName = dfv1.ControllerPipeline
+
+	pauseTimestampPath = `/metadata/annotations/numaflow.numaproj.io~1pause-timestamp`
```
vigith, line 55: `~1` what does that mean?

vigith (approved):

pkg/reconciler/isbsvc/controller.go
whynowy: Cleaned it up since there's no need to add this function.

pkg/reconciler/pipeline/controller.go
whynowy: To escape `/`. https://datatracker.ietf.org/doc/html/rfc6901#section-3

Issue #2066, opened 2024-09-16
Readiness probe config customization


whynowy:
# Summary

1. Readiness probe for Vertex and MonoVertex pods are too aggressive, need to update the default values.
2. A way to customize the probe for both readiness and liveness (only for `InitialDelaySeconds`, `TimeoutSeconds`, `PeriodSeconds`, `FailureThreshold` and `SuccessThreshold`).

# Use Cases

When would you use this?

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

kohlisid:
What is the scope for which we want to allow this for?
Also, pipeline level defaults and vertex/container level overrides as well correct?

whynowy:
> What is the scope for which we want to allow this for? Also, pipeline level defaults and vertex/container level overrides as well correct? @whynowy

correct.

PR #2068, opened 2024-09-17
feat: allow customization on readyz and livez config

whynowy:
Closes: #2066


pkg/apis/numaflow/v1alpha1/vertex_types.go
```diff
@@ -253,6 +253,18 @@ func (v Vertex) GetPodSpec(req GetVertexPodSpecReq) (*corev1.PodSpec, error) {
 		return nil, err
 	}
 
+	var readyzInitDeploy, readyzPeriodSeconds, readyzTimeoutSeconds, readyzFailureThreshold int32 = 5, 10, 30, 6 // default values
```
kohlisid, line 256: Should we store these default numeric values as constants, i see them getting reused in multiple places

whynowy: done.

vigith (approved):

PR #2069, opened 2024-09-17
fix: support version compatibility check for pre-release versions

### LINKED ISSUE #2064: SDK version compatibility check fails for pre-releases
```
panic: SDK version v0.8.2-0.20240913163521-4910018031a7 does not satisfy the minimum required by numaflow version latest+0d2230f: SDK version 0.8.2-0.20240913163521-4910018031a7 must be upgraded to at least 0.8.0, in order to work with current numaflow version: version 0.8.2-0.20240913163521-4910018031a7 did not meet constraint requirement >= 0.8.0
```

```
SDK version 0.8.1-rc1 must be upgraded to at least v0.8.0, in order to work with current numaflow version: version 0.8.1-rc1 did not meet constraint requirement >= v0.8.0

```
https://github.com/Masterminds/semver/issues/177

KeranYang:
TODO - once this PR gets approved, I will update the `minNumaflowVersion` in each SDK to add more guidelines on what version string to choose.

closes #2064


kohlisid:
Hey @KeranYang 
Thanks for making these changes! :D
Have you tested if we would we need similar changes in the Monovertex versioning as well? If yes, We can create an issue for that and do in a follow up PR.

KeranYang:
@kohlisid thank you. I will make similar changes to mono vertex using this PR. Please stay tuned for new commits.

rust/monovertex/src/server_info.rs
```diff
@@ -79,6 +79,39 @@ pub async fn check_for_server_compatibility(
 
 /// Checks if the given version meets the specified constraint.
 fn check_constraint(version: &Version, constraint: &str) -> error::Result<()> {
+    let binding = version.to_string();
+    // extract the major.minor.patch version
+    let mmp_version = binding.split('-').next().unwrap();
```
vigith, line 84: what is there is no `-`?

vigith (commented):
please do no use `unwrap`, it is like `panic`. use `map_error` instead and convert to `Result` and short-circuit the return using `?`.

rust/monovertex/src/server_info.rs
KeranYang: it returns the original string itself.

KeranYang:
@kohlisid monovertex change added, please review.

rust/monovertex/src/server_info.rs
```diff
@@ -105,16 +143,19 @@ fn check_numaflow_compatibility(
...
+    let numaflow_version_semver = Version::parse(numaflow_version_stripped)
         .map_err(|e| Error::ServerInfoError(format!("Error parsing Numaflow version: {}", e)))?;
 
     // Create a version constraint based on the minimum numaflow version
     let numaflow_constraint = format!(">={}", min_numaflow_version);
     check_constraint(&numaflow_version_semver, &numaflow_constraint).map_err(|e| {
         Error::ServerInfoError(format!(
             "numaflow version {} must be upgraded to at least {}, in order to work with current SDK version {}",
-            numaflow_version_semver, min_numaflow_version, e
+            numaflow_version_semver, human_readable(min_numaflow_version.to_string()), e
```
vigith, line 158: why are you calling `to_string()`? just implement Debug for `Version` via [`NewType`](https://doc.rust-lang.org/rust-by-example/generics/new_types.html).

rust/monovertex/src/server_info.rs
```diff
@@ -176,6 +217,27 @@ fn check_sdk_compatibility(
     Ok(())
 }
 
+// human_readable returns the human-readable minimum supported version.
+// it's used for logging purposes.
+// it translates the version we used in the constraints to the real minimum supported version.
+// e.g., if the given version is "0.8.0rc100", human-readable version is "0.8.0".
+// if the given version is "0.8.0-z", "0.8.0".
+// if "0.8.0-rc1", "0.8.0-rc1".
+fn human_readable(ver: String) -> String {
```
vigith, line 226: `impl Debug` via the above asked NewType?

pkg/sdkclient/serverinfo/types.go
```diff
@@ -27,11 +29,72 @@ const (
...
+
+There are two cases to consider when updating the map:
+
+1. The minimum supported version is a pre-release version.
+In this case, directly put the exact pre-release version in the map.
+E.g., if the minimum supported version is "0.8.0-rc1", then put "0.8.0-rc1" for java, go, rust.
+"0.8.0b1", "0.8.0b1" for python.
+2. The minimum supported version is a stable version.
+In this case, put (almost) the largest available pre-release version of the stable version in the map.
+E.g., if the minimum supported version is "0.8.0", then put "0.8.0-z" for java, go, rust, "0.8.0rc100" for python.
```
whynowy, line 64: Curious, why using `-z`?

KeranYang: Yes, it's mainly because of the golang semver [library](https://github.com/Masterminds/semver?tab=readme-ov-file#checking-version-constraints) "considering pre-releases to be invalid if the ranges does not include one". So we have to put a pre-release version in the constraint. To do so, we need to put the largest pre-release, hence `-z` and `rc100` for python.

I will improve the docs accordingly.

rust/monovertex/src/server_info.rs
KeranYang: @vigith , thanks for introducing to me the Debug trait. After learning a bit of it, I would like to keep my change as it is. My use case is to convert the version string used by constraint to a human-readable string. I don't want users to see "please update numaflow to 1.3.1-z", which is confusing, I want them to see "1-3-1".

This is a very simple string conversion. Using Debug will require me to create a new type which wraps a string and create object of that struct during compatibility check, which is extra code. Please let me know if my understanding is off.

vigith: isn't "human-readable string" all about `Debug` Trait is?

KeranYang: As discussed offline, I created a separate issue to track. https://github.com/numaproj/numaflow/issues/2076

whynowy (approved):

vigith (approved):

PR #2071, opened 2024-09-18
feat: Use gRPC bidirectional streaming for source transformer

BulkBeing:
Related PRs:
https://github.com/numaproj/numaflow-go/pull/147
https://github.com/numaproj/numaflow-rs/pull/91


rust/monovertex/src/config.rs
```diff
@@ -53,17 +54,14 @@ impl OnFailureStrategy {
...
     }
+}
 
-    /// Converts the `OnFailureStrategy` enum variant to a String.
-    /// This facilitates situations where the enum needs to be displayed or logged as a string.
-    ///
-    /// # Returns
-    /// A string representing the `OnFailureStrategy` enum variant.
-    fn to_string(&self) -> String {
+impl Display for OnFailureStrategy {
```
vigith, line 59: thank you :)

rust/monovertex/src/transformer.rs
```diff
@@ -1,61 +1,160 @@
...
-    client: SourceTransformClient<Channel>,
+    read_tx: mpsc::Sender<SourceTransformRequest>,
+    resp_stream: Streaming<SourceTransformResponse>,
 }
 
 impl SourceTransformer {
-    pub(crate) async fn new(client: SourceTransformClient<Channel>) -> Result<Self> {
-        Ok(Self { client })
+    pub(crate) async fn new(mut client: SourceTransformClient<Channel>) -> Result<Self> {
+        let (read_tx, read_rx) = mpsc::channel(config().batch_size as usize);
```
vigith, line 31: NIT: referencing `config` directly can make testing difficult if you want to change the value  for testing.

rust/monovertex/src/transformer.rs
```diff
@@ -1,61 +1,160 @@
...
+        tracing::info!("Sending stream request");
+        let transform_fn_with_timeout = tokio::time::timeout(
+            Duration::from_secs(3),
+            client.source_transform_fn(Request::new(read_stream)),
+        );
+        let Ok(resp_stream) = transform_fn_with_timeout.await else {
+            return Err(Error::TransformerError(
+                "connection to transformer gRPC server timed out".to_string(),
+            ));
+        };
```
vigith, line 52: only 3 sec timeout? will this whole thing become flaky in a busy system?

rust/monovertex/src/transformer.rs
```diff
@@ -1,61 +1,160 @@
...
+                    offset: msg_info.offset.clone(),
+                    event_time: utc_from_timestamp(result.event_time),
+                    headers: msg_info.headers.clone(),
+                };
+                messages.push(message);
+            }
         }
 
-        Ok(Some(messages))
+        let sender_task = tokio::time::timeout(Duration::from_millis(2), sender);
```
vigith, line 139: please document your thought on why you did this approach.

pkg/udf/rpc/tracker.go
```diff
@@ -6,68 +6,68 @@ import (
 	"github.com/numaproj/numaflow/pkg/isb"
 )
 
-// tracker is used to store a key value pair for string and *isb.ReadMessage
+// Tracker is used to store a key value pair for string and *isb.ReadMessage
```
whynowy, line 9: Move this file to `pkg/isb/` and rename to `message_tracker.go`, also rename the struck name?

rust/monovertex/src/transformer.rs
```diff
@@ -1,61 +1,160 @@
-use crate::error::Result;
-use crate::message::Message;
+use std::collections::HashMap;
+
+use crate::config::config;
+use crate::error::{Error, Result};
+use crate::message::{Message, Offset};
```
whynowy, line 5: I suppose this `Message` will be moved to a separate package that will be shared by both MonoVertex and Pipeline.

rust/monovertex/src/transformer.rs
```diff
@@ -1,61 +1,160 @@
...
-            messages.push(message);
+            let Some((msg_id, msg_info)) = tracker.remove_entry(&resp.id) else {
+                return Err(Error::TransformerError(format!(
+                    "Received message with unknown ID {}",
+                    resp.id
+                )));
+            };
+            for (i, result) in resp.results.into_iter().enumerate() {
+                if result.tags.iter().any(|x| x == DROP) {
+                    continue;
```
whynowy, line 125: TODO: metrics

rust/monovertex/src/transformer.rs
yhl25: Why do we need timeout?

BulkBeing: Since we create the stream at the beginning and we don't create multiple streams, this should succeed immediately. I added the timeout because the tests were stuck when there was some issue with the server (server started,but there was a mismatch  in the protobuf files on the client and the server). I will remove the timeouts.

pkg/sdkclient/sourcetransformer/client.go
```diff
@@ -53,18 +57,76 @@ func New(serverInfo *serverinfo.ServerInfo, inputOptions ...sdkclient.Option) (C
...
+				time.Sleep(100 * time.Millisecond)
+				continue waitUntilReady
+			}
+			break waitUntilReady
+		}
+	}
+
+	c.stream, err = c.grpcClt.SourceTransformFn(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("creating a gRPC stream for source transform: %w", err)
```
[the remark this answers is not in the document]
BulkBeing: I've seen somewhere (Go blog / talks), to not include "failed" in an error message, since `error` type itself indicates that it failed. If the error gets wrapped again on the caller's side, it could result in a final message of below pattern:
```
Failed to create user: Failed to create new account: Failed to INSERT new account record: database not connected
```

Compared to:
```
Failed to create user: creating new account:  INSERT new account record: database not connected
```

pkg/udf/rpc/tracker.go
```diff
@@ -6,68 +6,68 @@ import (
...
 }
 
-// NewTracker initializes a new instance of a tracker
-func NewTracker() *tracker {
-	return &tracker{
+// NewTracker initializes a new instance of a Tracker
+func NewTracker() *Tracker {
+	return &Tracker{
 		m:    make(map[string]*isb.ReadMessage),
 		lock: sync.RWMutex{},
```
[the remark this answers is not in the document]
BulkBeing: Mutex operations (unless there is too much contention) are much faster than network calls (nanoseconds vs micro/milli seconds). Unary calls has some overhead. As per the docs:

> Use streaming RPCs when handling a long-lived logical flow of data from the client-to-server, server-to-client, or in both directions. Streams can avoid continuous RPC initiation, which includes connection load balancing at the client-side, starting a new HTTP/2 request at the transport layer, and invoking a user-defined method handler on the server side.

pkg/sources/forward/applier/sourcetransformer.go
```diff
@@ -25,13 +25,13 @@ import (
 // SourceTransformApplier applies the source transform on the read message and gives back a new message. Any UserError will be retried here, while
 // InternalErr can be returned and could be retried by the callee.
 type SourceTransformApplier interface {
-	ApplyTransform(ctx context.Context, message *isb.ReadMessage) ([]*isb.WriteMessage, error)
+	ApplyTransform(ctx context.Context, messages []*isb.ReadMessage) ([]isb.ReadWriteMessagePair, error)
```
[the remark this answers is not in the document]
BulkBeing: Currently, we run the below func in goroutines and call `ApplyTransform` within it: 
```go
concurrentApplyTransformer(ctx context.Context, readMessagePair <-chan *isb.ReadWriteMessagePair)
```
It retrieves message from the `readMessagePair` channel, apply transformer, and store the `WriteMessages` in the same element.

With bidrectional streaming, the concurrency is moved to the server side. It might be difficult to have a function signature that takes a message and returns the output since we are not sure of the result order from the stream. Instead, we send all messages, wait for all output and aggregate the result in the new implementation.

pkg/sources/transformer/grpc_transformer.go
```diff
@@ -75,103 +73,102 @@ func (u *GRPCBasedTransformer) CloseConn(ctx context.Context) error {
...
+			responsePair := isb.ReadWriteMessagePair{
+				ReadMessage:   parentMessage,
+				WriteMessages: taggedMessages,
+				Err:           nil,
+			}
+			transformResults = append(transformResults, responsePair)
+			tracker.RemoveRequest(msgId)
+			messageCount--
+			if messageCount == 0 {
+				break loop
```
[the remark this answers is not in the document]
BulkBeing: `IsEmpty` has an additional overhead of acquiring lock.

yhl25 (approved):
LGTM!

PR #2075, opened 2024-09-19
chore: suppress some misleading logs from raters

whynowy:
Explain what this PR does.

### DIFF
--- a/pkg/daemon/server/service/rater/rater.go
+++ b/pkg/daemon/server/service/rater/rater.go
@@ -33,6 +33,10 @@ import (
 	sharedqueue "github.com/numaproj/numaflow/pkg/shared/queue"
 )
 
+const (
+	readTotalMetricName = "forwarder_data_read_total"
+)
+
 type Ratable interface {
 	Start(ctx context.Context) error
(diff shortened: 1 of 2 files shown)


pkg/daemon/server/service/rater/rater.go
```diff
@@ -217,13 +221,11 @@ func sleep(ctx context.Context, duration time.Duration) {
 // getPodReadCounts returns the total number of messages read by the pod
 // since a pod can read from multiple partitions, we will return a map of partition to read count.
 func (r *Rater) getPodReadCounts(vertexName, podName string) *PodReadCount {
-	readTotalMetricName := "forwarder_data_read_total"
```
KeranYang, line 220: LGTM. nit: I intentionally put this var inside this particular func instead of const, to make it unaccessible from other funcs. A small abstraction.

KeranYang (approved):

pkg/daemon/server/service/rater/rater.go
whynowy: moved it back.

PR #2077, opened 2024-09-20
feat: add pause for monovertex

### LINKED ISSUE #2005: support pause feature for monovertex
this is required for semantic completion wrt pipeline spec

kohlisid:
fixes #2005


whynowy (commented):
In general, I don't think we need to add all of those resume/pause functions, as long a we add the `desiredPhase` check in `CalculateReplicas()`, all the rest should be automatically taken care of.

kohlisid:
In the current solution, while pausing a MonoVertex we are not patching the Spec to update the replicas.
Only the Status replicas are updated. 
Patching will be added in the future if the logic requires. 
cc @whynowy

pkg/apis/numaflow/v1alpha1/mono_vertex_types.go
```diff
@@ -80,6 +81,10 @@ func (mv MonoVertex) getReplicas() int {
 
 func (mv MonoVertex) CalculateReplicas() int {
 	desiredReplicas := mv.getReplicas()
+	// If we are pausing the MonoVertex then we should have the desired replicas as 0
```
whynowy, line 84: Move up.

pkg/reconciler/monovertex/controller.go
```diff
@@ -132,13 +130,21 @@ func (mr *monoVertexReconciler) reconcile(ctx context.Context, monoVtx *dfv1.Mon
 
 	monoVtx.Status.MarkDeployed()
 
-	// Mark it running before checking the status of the pods
-	monoVtx.Status.MarkPhaseRunning()
+	// Update the phase based on the DesiredPhase from the lifecycle, this should encompass
+	// the Paused and running states.
+	currentPhase := monoVtx.Status.Phase
```
whynowy, line 135: Nit: Naming it `originalPhase` ?

pkg/apis/numaflow/v1alpha1/mono_vertex_types.go
```diff
@@ -621,3 +634,18 @@ type MonoVertexList struct {
...
+type MonoVertexLifecycle struct {
+	// DesiredPhase used to bring the pipeline from current phase to desired phase
+	// +kubebuilder:default=Running
+	// +optional
+	DesiredPhase MonoVertexPhase `json:"desiredPhase,omitempty" protobuf:"bytes,1,opt,name=desiredPhase"`
+}
+
+// GetDesiredPhase is used to fetch the desired lifecyle phase for a MonoVertex
+func (lc MonoVertexLifecycle) GetDesiredPhase() MonoVertexPhase {
+	if string(lc.DesiredPhase) != "" {
```
whynowy, line 647: Using `switch/case/default` to avoid no surprise.

pkg/apis/numaflow/v1alpha1/mono_vertex_types.go
```diff
@@ -609,6 +621,8 @@ func (mvs *MonoVertexStatus) IsHealthy() bool {
 	// We check if all the required conditions are true for it to be healthy
 	case MonoVertexPhaseRunning:
 		return mvs.IsReady()
+	case MonoVertexPhasePaused:
```
whynowy, line 624: For paused phase, we also need to check `IsReady()` right? because daemon is still there.

kohlisid: I kept it consistent with our pipeline health check. 

https://github.com/numaproj/numaflow/blob/b8c61debb784db0eaae31352138fd9b447f9d151/pkg/apis/numaflow/v1alpha1/pipeline_types.go#L782

Do we want to add that now for both?

kohlisid: Ideally, we should have check for both to ensure no changes in the conditions

whynowy (approved):

PR #2084, opened 2024-09-24
chore: check for pause timeout after errors

kohlisid:


pkg/reconciler/pipeline/controller.go
```diff
@@ -845,37 +845,49 @@ func (r *pipelineReconciler) pausePipeline(ctx context.Context, pl *dfv1.Pipelin
...
+
+	// Check for the daemon to obtain the buffer draining information, in case we see an error trying to
+	// retrieve this we do not exit prematurely to allow honoring the pause timeout for a consistent error
+	// - In case the timeout has not occurred we would trigger a requeue
+	// - If the timeout has occurred even after getting the drained error, we will try to pause the pipeline
+	daemonClient, daemonError := daemonclient.NewGRPCDaemonServiceClient(pl.GetDaemonServiceURL())
+	r.logger.Info(daemonClient.ListPipelineBuffers(ctx, pl.Name))
 	if err != nil {
-		return true, err
+		daemonError = err
```
juliev0, line 858: so, if I'm reading this correctly, this is the `err` from the call above `r.scaleDownSourceVertices()`, right? Is that a daemon error?

juliev0: but actually, it looks like we return above if `err != nil` so I believe this code is unreachable ?

pkg/reconciler/pipeline/controller.go
```diff
@@ -845,37 +845,49 @@ func (r *pipelineReconciler) pausePipeline(ctx context.Context, pl *dfv1.Pipelin
 		return updated, err
```
juliev0, line 845: I guess if for some reason we return from here repeatedly and indefinitely, we will not honor pause timeout. Is that possible?

kohlisid: Here if we are not able to scale down then there is some issue at the K8s end. We should not force a pause in such a scenario

juliev0: Okay, I suppose I'm okay with that if that's what the numaflow team prefers. In my mind, I imagined keeping it simple and just saying a pause timeout is honored in all cases, especially since this function and/or the call to `scaleDownSourceVertices()` could morph over time and potentially include other calls that could fail and I wouldn't want to accidentally get into the same situation then. But maybe I'm being too paranoid.

I see what you're saying about not wanting to risk data loss because of a kubernetes issue

juliev0: I think the PR looks good to me in that case. Thanks for doing this!

kohlisid: I would say that if there are issues arising from K8s itself which are involved while doing `scaleDownSourceVertices` we have bigger problems at hand for the pipeline and they should get reflected in the status. 

_morph over time_ - This we will take care that the design is maintained :)

I will let @whynowy add any other opinions on this in case he has something else to add.

juliev0 (approved):

juliev0:
@kohlisid - should we move this into "Ready for Review" and maybe @whynowy can review?

pkg/reconciler/pipeline/controller.go
```diff
@@ -845,37 +845,45 @@ func (r *pipelineReconciler) pausePipeline(ctx context.Context, pl *dfv1.Pipelin
...
 	// if drain is completed, or we have exceeded the pause deadline, mark pl as paused and scale down
 	if time.Now().After(pauseTimestamp.Add(time.Duration(pl.Spec.Lifecycle.GetPauseGracePeriodSeconds())*time.Second)) || drainCompleted {
-		_, err := r.scaleDownAllVertices(ctx, pl)
+		_, err = r.scaleDownAllVertices(ctx, pl)
 		if err != nil {
 			return true, err
 		}
-		// if the drain completed succesfully, then set the DrainedOnPause field to true
+		if daemonError != nil {
+			r.logger.Errorf("error in fetching Drained status, Pausing due to timeout: %v", zap.Error(err))
```
whynowy, line 877: Nil: `Error in fetching Drained status, pausing due to ...`

whynowy (approved):

pkg/reconciler/pipeline/controller.go
```diff
@@ -845,37 +845,45 @@ func (r *pipelineReconciler) pausePipeline(ctx context.Context, pl *dfv1.Pipelin
...
 	// if drain is completed, or we have exceeded the pause deadline, mark pl as paused and scale down
 	if time.Now().After(pauseTimestamp.Add(time.Duration(pl.Spec.Lifecycle.GetPauseGracePeriodSeconds())*time.Second)) || drainCompleted {
-		_, err := r.scaleDownAllVertices(ctx, pl)
+		_, err = r.scaleDownAllVertices(ctx, pl)
 		if err != nil {
 			return true, err
 		}
-		// if the drain completed succesfully, then set the DrainedOnPause field to true
+		if daemonError != nil {
+			r.logger.Errorf("Error in fetching Drained status, Pausing due to timeout: %v", zap.Error(err))
```
whynowy, line 877: This is wrong. `Errorf` vs `Errorw`.

`r.logger.Errorf("Error in fetching Drained status, pausing due to timeout: %v", err`

whynowy (approved):

PR #2087, opened 2024-09-25
feat: container-type level version compatibility check

KeranYang:
introduces one more dimension when checking version compatibly - container type.

### DIFF
--- a/pkg/sdkclient/serverinfo/types.go
+++ b/pkg/sdkclient/serverinfo/types.go
@@ -29,6 +29,8 @@ const (
 
 type ContainerType string
 
+// the string content matches the corresponding server info file name.
+// DO NOT change it unless the server info file name is changed.
 const (
 	sourcer           ContainerType = "sourcer"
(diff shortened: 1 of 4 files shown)


pkg/sdkclient/serverinfo/types.go
```diff
@@ -70,14 +86,54 @@ A constraint ">=0.8.0-z" will match any pre-release version of 0.8.0, including
 More details about version comparison can be found in the PEP 440 and semver documentation.
 */
 var minimumSupportedSDKVersions = sdkConstraints{
```
vigith, line 88: we need to figure out how we can read this from a YAML file. maintaining the code is tricky.

vigith (approved):

pkg/sdkclient/serverinfo/types.go
KeranYang: maybe we can use the go embed. we can revisit at a later time.

PR #2112, opened 2024-10-02
fix: create buffers and buckets before updating Vertices

### LINKED ISSUE #2083: No "Create Job" executed - Possible Race Condition
**Describe the bug**
Normally, Numaplane's e2e test passes, but this captures an instance in which it did not. 

The sequence of events which Numaplane did:
1. Create ISBService
2. Create Pipeline
3. After Pipeline is running, update it in a trivial way that doesn't require pausing
4. After Pipeline is updated, pause it so it can be updated
5. Update Pipeline's topology from `in->out`  to  `in->cat->out` and keep its `desiredPhase=Paused`
6. Once topology change is reconciled, update `desiredPhase=Running`

Normally, this works fine. Normally step 5 causes a Creation Job and a Deletion Job:
```
{"level":"info","ts":"2024-09-23T20:08:28.842186617Z","logger":"numaflow.controller-manager","caller":"pipeline/controller.go:337","msg":"Created a job successfully for ISB creating","namespace":"numaplane-system","pipeline":"test-pipeline-rollout","buffers":["numaplane-system-test-pipeline-rollout-cat-0"],"buckets":["numaplane-system-test-pipeline-rollout-cat-out","numaplane-system-test-pipeline-rollout-in-cat"],"servingStreams":[]}
{"level":"info","ts":"2024-09-23T20:08:28.850466284Z","logger":"numaflow.controller-manager","caller":"pipeline/controller.go:356","msg":"Created ISB Svc deleting job successfully","namespace":"numaplane-system","pipeline":"test-pipeline-rollout","buffers":[],"buckets":["numaplane-system-test-pipeline-rollout-in-out"]}
```
(above is extracted from a good run)

However, in this run the Creation Job was not executed, which caused the Daemon Pods unable to get past the init container `isbsvc-validate`check for buffers and buckets. 

** What I suspect **
In the log I see that the `in` Vertex was successfully updated, the `cat` vertex was created, but when the `out` vertex was supposed to be created, a Resource Version conflict occurred [here](https://github.com/numaproj/numaflow/blob/v1.3.1/pkg/reconciler/pipeline/controller.go#L277). This may have happened prior to the Creation Job being created. This should cause Numaflow Controller to return and then re-reconcile idempotently.

However, note that the Creation Job is dependent on [newBuffers and newBuckets](https://github.com/numaproj/numaflow/blob/v1.3.1/pkg/reconciler/pipeline/controller.go#L300), which is dependent on [these](https://github.com/numaproj/numaflow/blob/v1.3.1/pkg/reconciler/pipeline/controller.go#L233-L241) Vertex values. In this bug, the `in` and `cat` Vertices were updated successfully on the previous reconciliation. So, the current state of the Vertex no longer reflects the new buffers and new buckets which need to be added.

I will add logs. 

- Numaflow: quay.io/numaio/numaflow-rc:v0.0.12

---

juliev0:
Fixes #2083 

I simply swapped 1. the Job creation to occur prior to 2. the creation/deletion/Update of the Vertex CRs. 

Previously the creation of the Jobs was dependent on the Vertex specs ([newBuffers and newBuckets](https://github.com/numaproj/numaflow/blob/v1.3.2/pkg/reconciler/pipeline/controller.go#L325) are [calculated](https://github.com/numaproj/numaflow/blob/v1.3.2/pkg/reconciler/pipeline/controller.go#L254-L282) based on Vertex specs).

So if there's a transient error that occurs after or in the middle of creating/updating the Vertex CRs, the next reconciliation will incorrectly calculate the Jobs it needs to perform, as some or all of the Vertex specs have changed.

### DIFF
--- a/pkg/reconciler/pipeline/controller.go
+++ b/pkg/reconciler/pipeline/controller.go
@@ -280,46 +280,6 @@ func (r *pipelineReconciler) reconcileFixedResources(ctx context.Context, pl *df
 			newBuckets[b] = b
 		}
 	}
-	newObjs := buildVertices(pl)
-	for vertexName, newObj := range newObjs {
-		if oldObj, existing := existingObjs[vertexName]; !existing {
-			if err := r.client.Create(ctx, &newObj); err != nil {
-				if apierrors.IsAlreadyExists(err) { // probably somebody else already created it
-					continue
-				} else {
-					r.recorder.Eventf(pl, corev1.EventTypeWarning, "CreateVertexFailed", "Failed to create vertex: %w", err.Error())
-					return fmt.Errorf("failed to create vertex, err: %w", err)
-				}
-			}
-			log.Infow("Created vertex successfully", zap.String("vertex", vertexName))
-			r.recorder.Eventf(pl, corev1.EventTypeNormal, "CreateVertexSuccess", "Created vertex %s successfully", vertexName)
-		} else {
-			if oldObj.GetAnnotations()[dfv1.KeyHash] != newObj.GetAnnotations()[dfv1.KeyHash] { // need to update
-				originalReplicas := oldObj.Spec.Replicas
-				oldObj.Spec = newObj.Spec
-				oldObj.Spec.Replicas = originalReplicas
-				oldObj.Annotations[dfv1.KeyHash] = newObj.GetAnnotations()[dfv1.KeyHash]
-				if err := r.client.Update(ctx, &oldObj); err != nil {
-					r.recorder.Eventf(pl, corev1.EventTypeWarning, "UpdateVertexFailed", "Failed to update vertex: %w", err.Error())
-					return fmt.Errorf("failed to update vertex, err: %w", err)
-				}
-				log.Infow("Updated vertex successfully", zap.String("vertex", vertexName))
-				r.recorder.Eventf(pl, corev1.EventTypeNormal, "UpdateVertexSuccess", "Updated vertex %s successfully", vertexName)
-			}
-			delete(existingObjs, vertexName)
-		}
-	}
-	for _, v := range existingObjs {
-		if err := r.client.Delete(ctx, &v); err != nil {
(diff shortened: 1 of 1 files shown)


whynowy:
This looks good to me. @kohlisid and @KeranYang , can you double check if there's any issue with the order switch?

juliev0:
I actually haven't tested it to tell you the truth yet. I wanted to see if it would pass the CI. If we can build an image for it, I can start to incorporate it into Numaplane, which is aggressively testing this. We can keep it in Draft stage for a little while until we feel confident.

juliev0:
What this does mean is that the Buffers and Buckets that the vertices are currently using will disappear and those Vertices will not be able to work with the isbsvc. Then the vertices will get updated and they will get reconciled by Vertex Controller to replace the old Pods with the new ones, and I think those should come up. 

Of course, if the Pipeline is paused first, then the Vertices will be down when the buffers and buckets get wiped and replaced, so nothing should be broken then.

KeranYang:
Thanks Julie for making the change. I don't think this change will break anything so should be safe to merge. My only concern is that we didn't validate the fix. +1 on using numaplane e2e to aggressively test it. Meanwhile, we should develop our own numaflow test case to catch such issue - a test with pipeline topology change would work.

@juliev0 , if we merge this in, will numaplane e2e immediately get it? If so, we can monitor.

kohlisid:
Thanks @juliev0 
Looks good on a high level! +1 on testing it out before rolling it.
Meanwhile will think of defining some cases for verification on our end as well that can be incorporated to help with such changes in the future.

juliev0:
I saw this same thing happen again when running our e2e test so I guess it's not that rare. Meanwhile, I don't believe I've seen any adverse consequences of not doing it. Therefore, I'll make this "ready for review". Feel free to merge either now or we can give it a little more time if you prefer.

whynowy (approved):

KeranYang:
Thanks, @juliev0 . Please continue monitoring and let us know if this fixes.

PR #2122, opened 2024-10-04
chore: Unit tests for metrics.rs

### LINKED ISSUE #2038: metrics.rs has no tests

BulkBeing:
Resolves #2038

### DIFF
--- a/rust/numaflow-core/src/shared/server_info.rs
+++ b/rust/numaflow-core/src/shared/server_info.rs
@@ -1,6 +1,6 @@
 use std::collections::HashMap;
 use std::fs;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::time::Duration;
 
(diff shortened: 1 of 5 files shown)


rust/numaflow-core/src/shared/server_info.rs
```diff
@@ -265,7 +265,7 @@ fn trim_after_dash(input: &str) -> &str {
 
 /// Extracts the container type from the server info file.
 /// The file name is in the format of <container_type>-server-info.
-fn get_container_type(server_info_file: &PathBuf) -> Option<&str> {
```
KeranYang, line 268: just curious, how does this type conversion work and why do we need to make this line of change?

rust/numaflow-core/src/shared/server_info.rs
```diff
@@ -190,12 +190,12 @@ fn human_readable(ver: &str) -> String {
         return String::new();
     }
     // semver
-    if ver.ends_with("-z") {
```
KeranYang, line 193: thank you

rust/numaflow-core/src/shared/server_info.rs
BulkBeing: The conversion works because of Deref coersion. `PathBuf` implements `Deref` to `Path` https://doc.rust-lang.org/std/path/struct.PathBuf.html#impl-Deref-for-PathBuf (similar to `String` and `str`).
The reason for doing it was that it was showing in `cargo clippy -p numaflow-core`

KeranYang (approved):

vigith (approved):

PR #2126, opened 2024-10-05
feat: generator based on ticker

vigith:
fixes: #2125


rust/numaflow-core/src/source/generator.rs
```diff
@@ -0,0 +1,354 @@
...
+}
+
+impl source::SourceAcker for GeneratorAck {
+    async fn ack(&mut self, _: Vec<Offset>) -> crate::error::Result<()> {
+        Ok(())
+    }
+}
+
+pub(crate) struct GeneratorLagReader {}
+
```
yhl25, line 281: LagReader should be cloneable.

yhl25 (approved):

PR #2131, opened 2024-10-08
fix: pipeline pausing race conditions of draining and terminating source

kohlisid:
Explain what this PR does.


juliev0:
I'm curious - are you able to explain now what could've caused `pl.Status.Phase` to be set to "Paused" [here](https://github.com/numaproj/numaflow/blob/v1.3.2/pkg/reconciler/pipeline/controller.go#L194) but not reflected in etcd Pipeline Status?

kohlisid:
From the logs I captured, I was able to see the status getting reflected to Paused in memory for sure. In the next reconciliation process (which begins almost immediately) that got triggered we started from running again and due to a race we stay in pausing as we are not able to drain out the buffers.

pkg/reconciler/pipeline/controller.go
```diff
@@ -884,7 +890,21 @@ func (r *pipelineReconciler) pausePipeline(ctx context.Context, pl *dfv1.Pipelin
...
+}
+
+// sourceVerticesRunning checks whether any source vertex has running replicas
+func (r *pipelineReconciler) sourceVerticesRunning(ctx context.Context, pl *dfv1.Pipeline) (bool, error) {
+	existingVertices, err := r.findExistingVertices(ctx, pl)
+	if err != nil {
+		return false, err
+	}
+	for _, vertex := range existingVertices {
+		if sourceVertexFilter(vertex) && vertex.Status.Replicas != 0 {
```
whynowy, line 903: this is not accurate, we should check the pods.

juliev0:
> From the logs I captured, I was able to see the status getting reflected to Paused in memory for sure. In the next reconciliation process (which begins almost immediately) that got triggered we started from running again and due to a race we stay in pausing as we are not able to drain out the buffers.

That's interesting but I still don't understand what would cause the write to etcd not to occur (unless it is writing and the watch is skipping changes).

pkg/reconciler/pipeline/controller.go
```diff
@@ -830,39 +830,45 @@ func (r *pipelineReconciler) resumePipeline(ctx context.Context, pl *dfv1.Pipeli
...
+		err            error
+	)
 	// check that annotations / pause timestamp annotation exist
 	if pl.GetAnnotations() == nil || pl.GetAnnotations()[dfv1.KeyPauseTimestamp] == "" {
 		patchJson := `{"metadata":{"annotations":{"` + dfv1.KeyPauseTimestamp + `":"` + time.Now().Format(time.RFC3339) + `"}}}`
 		if err := r.client.Patch(ctx, pl, client.RawPatch(types.MergePatchType, []byte(patchJson))); err != nil && !apierrors.IsNotFound(err) {
 			return true, err
 		}
+		pl.Status.MarkPhasePausing()
+		updated, err := r.scaleDownSourceVertices(ctx, pl)
```
whynowy, line 845: If scaling down source fails, but annotating succeeds, then it will not be able to enter this secion.

juliev0:
LGTM as far as I can tell

pkg/reconciler/pipeline/controller.go
```diff
@@ -830,39 +831,48 @@ func (r *pipelineReconciler) resumePipeline(ctx context.Context, pl *dfv1.Pipeli
...
-		// If there's an error, or scaling down happens, requeue the request
-		// This is to give some time to process the new messages, otherwise check IsDrained directly may get incorrect information
-		return updated, err
-	}
-
-	var daemonError error
-	var drainCompleted = false
-
+	// Check if all the source vertices have scaled down to zero
+	sourcesScaled, pauseError := r.sourceVerticesNotRunning(ctx, pl)
```
whynowy, line 857: Variable names are confusing: `sourcePodsTerminated, err`.

pkg/reconciler/pipeline/controller.go
```diff
@@ -884,7 +894,20 @@ func (r *pipelineReconciler) pausePipeline(ctx context.Context, pl *dfv1.Pipelin
 		pl.Status.MarkPhasePaused()
 		return false, nil
 	}
-	return true, daemonError
+	return true, pauseError
+}
+
+// sourceVerticesNotRunning checks whether any source vertex has running replicas
+func (r *pipelineReconciler) sourceVerticesNotRunning(ctx context.Context, pl *dfv1.Pipeline) (bool, error) {
```
whynowy, line 901: Function name, `noSourceVertexPodsRunning`?

whynowy (approved):
The logic change looks good to me, it would be better to have more appropriate variable and function names.

whynowy (approved):

PR #2135, opened 2024-10-09
feat: expose ports for user defined containers

whynowy:
Expose `containerPort` for user defined containers so that it can be used for prometheus metrics scraping configuration.


vigith (approved):

whynowy:
This needs to go to the patch release of v1.3

PR #2139, opened 2024-10-10
feat: Unify Batch Map and Unary Map Operations Using a Shared gRPC Protocol

yhl25:


vigith (approved):

kohlisid:
From earlier batchmap implementation
```
	// check if there are elements left in the Tracker. This cannot be an acceptable case as we want the
	// UDF to send responses for all elements.
	if !trackerReq.IsEmpty() {
		logger.Error("BatchMap response for all requests not received from UDF")
		return nil, fmt.Errorf("batchMap response for all requests not received from UDF")
	}
```

I see that we are not checking for this now, messages for which no response was sent.

So now are we going to drop them in such a case? Because they will get acked.

vigith:
> I see that we are not checking for this now, messages for which no response was sent.
So now are we going to drop them in such a case? Because they will get acked.

If their return from their code, doesn't it implicitly mean that they are done with the processing?

Issue #2142, opened 2024-10-11
Kafa sink not retaining the key value.


vegesana:
**Describe the bug**
I added key to my message (kafka usecase) in UDF vertex as part of mapping. I pinned two sink vertices to the UDF - one to sink log and one to kafka sink.
 - In the sink logs i see that the key exists in messages. But the message sent by kafka sink to the kafka topic does not have the key.
Basically i lost the key after it is sent by kafka sink. It looks like kafka sink is not setting the headers.

**To Reproduce**

1) Send message to kafka sink via UDF with key value set for the message.
2) Check the message on hte kafka topic on to which the kafka sink is sending.
3) Key is not present.

**Expected behavior**
Need key value to be ratained by the kafka sink. Else with multiple partitions in kafka topic will cause out of order situations

**Screenshots**

**Environment (please complete the following information):**

- Numaflow: [v1.3.3]
- Client Version: [v1.31.0]
- Kustomize Version: [v5.4.2]
- Server Version: [v1.31.0]

**Additional context**

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

>For quick help and support, join our [slack](https://numaproj.slack.com/join/shared_invite/zt-19svuv47m-YKHhsQ~~KK9mBv1E7pNzfg#/shared-invite/email) channel.

vigith:
how do you want the keys to be added to the header? keys is an array, while header is key/value.

vegesana:
Thats true.. when i started making changes to fix this, even i realized it. Like addkeys() and addtags() for messages, we might need to add addHeaders() api as well. For key, we might need to add new field in the header. That is my opinion which i am planning to tryOn Oct 10, 2024, at 7:55 PM, Vigith Maurice ***@***.***> wrote:﻿
how do you want the keys to be added to the header? keys is an array, while header is key/value.

vigith:
closed via #2143 2143

PR #2143, opened 2024-10-11
feat: add keys into kafka header while producing

vigith:
partially solves: #2142 

we will be adding `__key_len` in the header to indicate how many keys are going to be in the header. `__key_%d` will have the value at an index.

### DIFF
--- a/pkg/sinks/kafka/kafka.go
+++ b/pkg/sinks/kafka/kafka.go
@@ -150,10 +150,32 @@ func (tk *ToKafka) Write(_ context.Context, messages []isb.Message) ([]isb.Offse
 			}
 		}
 	}()
+
 	for index, msg := range messages {
+		// insert keys in the header.
+		// since keys is an array, to decompose it, we need len and key at each index.
+		var headers []sarama.RecordHeader
+		// insert __key_len
+		keyLen := sarama.RecordHeader{
+			Key:   []byte("__key_len"),
+			Value: []byte(fmt.Sprintf("%d", len(msg.Keys))),
+		}
+		headers = append(headers, keyLen)
+
+		// write keys into header if length > 0
+		if len(msg.Keys) > 0 {
+			for idx, key := range msg.Keys {
+				headers = append(headers, sarama.RecordHeader{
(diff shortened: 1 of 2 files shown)


pkg/sinks/kafka/kafka.go
```diff
@@ -150,10 +150,32 @@ func (tk *ToKafka) Write(_ context.Context, messages []isb.Message) ([]isb.Offse
...
+			Key:   []byte("__key_len"),
+			Value: []byte(fmt.Sprintf("%d", len(msg.Keys))),
+		}
+		headers = append(headers, keyLen)
+
+		// write keys into header if length > 0
+		if len(msg.Keys) > 0 {
+			for idx, key := range msg.Keys {
+				headers = append(headers, sarama.RecordHeader{
+					Key:   []byte(fmt.Sprintf("__key_%d", idx)),
```
whynowy, line 169: Document this in the Kafka sink? https://numaflow.numaproj.io/user-guide/sinks/kafka/

whynowy (approved):

pkg/sinks/kafka/kafka.go
vegesana: Hi,

filling Headers in the producer message is good. This is needed for retaining the headers. But my requirement is the **"key"** part of producer message. My understanding is that this key will decide on which partition of kafka the message will be put in. 

type ProducerMessage struct {
	Topic string // The Kafka topic for this message.
	// The partitioning key for this message. Pre-existing Encoders include
	// StringEncoder and ByteEncoder.
	Key Encoder                 **<< THis is the one i am looking for**
	// The actual message to store in Kafka. Pre-existing Encoders include
	// StringEncoder and ByteEncoder.
	Value Encoder

	// The headers are key-value pairs that are transparently passed
	// by Kafka between producers and consumers.
	Headers []RecordHeader

vegesana (changes requested):
Need a way to send the key value that is added in UDF to be retained till the end where kafka sink produces into the topic.

pkg/sinks/kafka/kafka.go
vigith: i will add via a separate PR.

Issue #2151, opened 2024-10-14
structured concurrency is lost in actor pattern


vigith:

vigith:
the moment we add handler to struct, `Clone` will no longer be applicable

BulkBeing:
I think we should rely on message passing to shutdown tasks. Another possible option is to return join handles in a wrapper struct and abort tasks within the `Drop` implementation of it.

PR #2155, opened 2024-10-15
chore: Document how actor tasks are terminated

BulkBeing:
resolves https://github.com/numaproj/numaflow/issues/2151


vigith:
no longer valid.

PR #2156, opened 2024-10-15
chore: multi controller awareness on validating webhook

whynowy:
Webhook will block following operations:

1. Creating a Pipeline which references an ISBSvc having different instance annotation
2. Updating instance annotation of a Pipeline
3. Updating instance annotation of an ISBSvc

### DIFF
--- a/pkg/webhook/validator/isbsvc.go
+++ b/pkg/webhook/validator/isbsvc.go
@@ -47,21 +47,21 @@ func (v *isbsvcValidator) ValidateUpdate(_ context.Context) *admissionv1.Admissi
 	if err := isbsvccontroller.ValidateInterStepBufferService(v.newISBService); err != nil {
 		return DeniedResponse(err.Error())
 	}
-	switch {
-	case v.oldISBService.Spec.JetStream != nil:
-		// check the type of ISB Service is not changed
-		if v.newISBService.Spec.Redis != nil {
-			return DeniedResponse("can not change ISB Service type from Jetstream to Redis")
-		}
+	// chck if the instance annotation is changed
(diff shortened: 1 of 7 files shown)


pkg/webhook/validator/isbsvc.go
```diff
@@ -72,6 +72,7 @@ func (v *isbsvcValidator) ValidateUpdate(_ context.Context) *admissionv1.Admissi
 		if oldRedisNative, newRedisNative := v.oldISBService.Spec.Redis.Native, v.newISBService.Spec.Redis.Native; oldRedisNative != nil && newRedisNative != nil && !equality.Semantic.DeepEqual(oldRedisNative.Persistence, newRedisNative.Persistence) {
 			return DeniedResponse("can not change persistence of Redis ISB Service")
 		}
+	default:
```
KeranYang, line 75: is it ok if an isbsvc changes from a valid one to unknown?

pkg/webhook/validator/pipeline.go
```diff
@@ -82,6 +82,10 @@ func (v *pipelineValidator) checkISBSVCExists(ctx context.Context, isbSvcName st
 	if err != nil {
 		return err
 	}
+	// Check if they have the same instance annotation
```
KeranYang, line 85: can we rename the method to `checkISBSVCIsValid`? and also unit test it.

vigith (approved):

pkg/webhook/validator/pipeline.go
whynowy: done.

pkg/webhook/validator/isbsvc.go
KeranYang: oh, sorry I missed. It should be caught by L55 already. Thanks!

KeranYang (approved):

PR #2159, opened 2024-10-16
chore: rename numaflow-grpc to numaflow-pb

whynowy:
Which makes more sense now.


vigith (approved):

vigith:
@yhl25 you might get conflicts

PR #2162, opened 2024-10-16
chore: Make TickGen compatible with current builtin

### LINKED ISSUE #2158: make tickgen compatible with current builtin

BulkBeing:
Resolves  #2158


rust/numaflow-core/src/config.rs
```diff
@@ -314,11 +327,44 @@ impl Settings {
...
+
+                    if let Some(msg_size) = generator_source.msg_size {
+                        if msg_size < 0 {
+                            warn!("'msgSize' can not be negative, using default value (8 bytes)");
+                        } else {
+                            config.msg_size_bytes = msg_size as u32;
+                        }
+                    }
+
+                    config.value = generator_source.value;
```
yhl25, line 352: if content is set, should we set this to None?

rust/numaflow-core/src/source/generator.rs
```diff
@@ -43,44 +47,140 @@ mod stream_generator {
...
+                padding: Vec<u8>,
+            }
+
+            let mut padding = vec![];
+            if msg_size_bytes > 8 {
+                let size = msg_size_bytes - 8;
+                let mut bytes = vec![0; size as usize];
+                rand::thread_rng().fill(&mut bytes[..]);
+                padding = bytes;
+            }
```
yhl25, line 101:
```suggestion
            let padding: Vec<u8> = (msg_size_bytes > 8)
                .then(|| {
                    let size = (msg_size_bytes - 8);
                    let mut bytes = vec![0; size as usize];
                    rand::thread_rng().fill(&mut bytes[..]);
                    bytes
                })
                .unwrap_or_default();
```

rust/numaflow-core/src/source/generator.rs
```diff
@@ -43,44 +47,140 @@ mod stream_generator {
...
+                            None => vec![],
+                        };
+                        data.push(Self::create_message(
+                            this.content.clone(),
+                            keys,
+                            *this.value,
+                            *this.msg_size_bytes,
+                            jitter,
+                        ));
+                    }
```
yhl25, line 181: can we extract this out? so that it can be used inside poll pending also?

rust/numaflow-core/src/source/generator.rs
```diff
@@ -43,44 +47,140 @@ mod stream_generator {
         /// the amount of credits used for the current time-period.
         /// remaining = (rpu - used) for that time-period
         used: usize,
+
+        value: Option<i64>,
+        msg_size_bytes: u32,
+        keys: (Vec<String>, usize),
```
yhl25, line 53: should we reset the usize after every tick?

yhl25 (approved):
LGTM!

vigith (changes requested):
we need to pass self

vigith:
try something like this? the key point is, `this` and `self` cannot be having overlapping lifetimes.

```rust
use std::{
    pin::Pin,
    task::{Context, Poll},
    time::Duration,
};

use futures::{Stream, StreamExt};
use pin_project::pin_project;

#[pin_project]
struct Generator {
    size: usize,
    #[pin]
    interval: tokio::time::Interval,
}

impl Generator {
    fn new(size: usize, duration: Duration) -> Self {
        Self {
            size,
            interval: tokio::time::interval(duration),
        }
    }

    fn increment_size(&mut self) -> usize {
        self.size += 1;
        self.size
    }
}

impl Stream for Generator {
    type Item = usize;

    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
        let mut this = self.as_mut().project();
        match this.interval.poll_tick(cx) {
            Poll::Ready(_) => {
                let size = self.increment_size();
                Poll::Ready(Some(size))
            }
            Poll::Pending => Poll::Pending,
        }
    }
}

#[tokio::main]
async fn main() {
    let mut gen = Generator::new(0, Duration::from_millis(100));
    for _ in 0..10 {
        let l = &gen.next().await.unwrap();
        println!("{}", l);
    }
}
```

vigith (approved):

PR #2185, opened 2024-10-24
fix: refine vertex/mvtx pod clean up logic

whynowy:
In some edged cases, e.g. the scaling operations on vertex/mvtx are too frequent (multiple times in 1 second, not caused by human, but some other issue like [this](https://github.com/numaproj/numaplane/issues/353)), the controller can not catch up with each change, and then it can end up with uncleaned up pods running (e.g. the mvtx only has 1 replica, but there are extra pods 3 and 4 running).

This PR fixes the problem.

### DIFF
--- a/pkg/reconciler/monovertex/controller.go
+++ b/pkg/reconciler/monovertex/controller.go
@@ -19,6 +19,7 @@ package monovertex
 import (
 	"context"
 	"fmt"
+	"math"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


vigith (approved):

pkg/reconciler/monovertex/controller.go
```diff
@@ -201,11 +202,11 @@ func (mr *monoVertexReconciler) orchestratePods(ctx context.Context, monoVtx *df
 		monoVtx.Status.UpdatedReadyReplicas = 0
 	}
 
-	// Manually or automatically scaled down
+	// Manually or automatically scaled down, in this case, we need to clean up extra pods if there's any
+	if err := mr.cleanUpPodsFromTo(ctx, monoVtx, desiredReplicas, math.MaxInt); err != nil {
+		return fmt.Errorf("failed to clean up mono vertex pods [%v, ∞): %w", desiredReplicas, err)
+	}
 	if currentReplicas := int(monoVtx.Status.Replicas); currentReplicas > desiredReplicas {
```
KeranYang, line 209: do we still need this if statement?

whynowy: Yes.

KeranYang (approved):

PR #2195, opened 2024-10-30
feat: metrics visualiser for mono vertex

veds-g:
config driven metrics framework to display existing metrics in the UI

- Currently supports histogram metrics for MonoVertex


server/apis/v1/handler.go
```diff
@@ -1180,6 +1247,90 @@ func (h *handler) GetMonoVertexHealth(c *gin.Context) {
...
+	if err != nil {
+		h.respondWithError(c, fmt.Sprintf("Failed to execute the prometheus query, %s", err.Error()))
+		return
+	}
+	c.JSON(http.StatusOK, NewNumaflowAPIResponse(nil, result))
+}
+
+// GetMetricsDiscovery is used to provide a metrics list for each
+// dimension along with necessary params and filters for a given object
+func (h *handler) GetMetricsDiscovery(c *gin.Context) {
```
whynowy, line 1288: `DiscoverMetrics`

cmd/commands/server.go
```diff
@@ -76,6 +78,7 @@ func NewServerCommand() *cobra.Command {
 	command.Flags().StringVar(&managedNamespace, "managed-namespace", sharedutil.LookupEnvStringOr("NUMAFLOW_SERVER_MANAGED_NAMESPACE", sharedutil.LookupEnvStringOr("NAMESPACE", "numaflow-system")), "The namespace that the server watches when \"--namespaced\" is \"true\".")
 	command.Flags().StringVar(&baseHref, "base-href", sharedutil.LookupEnvStringOr("NUMAFLOW_SERVER_BASE_HREF", "/"), "Base href for Numaflow server, defaults to '/'.")
 	command.Flags().BoolVar(&readOnly, "readonly", sharedutil.LookupEnvBoolOr("NUMAFLOW_SERVER_READONLY", false), "Whether to enable read only view for the UX server, defaults to false.")
+	command.Flags().BoolVar(&disableMetricsCharts, "disable-metrics-charts", sharedutil.LookupEnvBoolOr("NUMAFLOW_SERVER_DISABLE_METRICS_CHARTS", true), "Whether to disable metrics charts for the UX server, defaults to true.")
```
whynowy, line 81: Can we reuse the the configmap for this purpose? e.g. if the `url` is not configured, it means it's not enabled.

veds-g: we are passing this info to the UI via `/sysInfo` endpoint for which we create the object in `server/start.go`. We are defining the `PromQLService` inside `handler.go`. Should we move it outside at the top level?

veds-g: or similar to Dex Server, if we can move the `url` in shared config map, and use it to create the PromQLService object. wdyt?

whynowy: Use the 1st option?

whynowy:
When there are many pods.


Screenshot 2024-11-05 at 5 02 45 PM

whynowy:
Mouse hover on the dotted line, the cursor does not always show it's clickable.

Screenshot 2024-11-05 at 5 06 43 PM

whynowy:
Move container info on top of pod info? The pod info remains unchanged when click different container icon but the container info, users need to scroll down to see the changes.

In general, the pod and container info take too much space, reducing it would be better.

Screenshot 2024-11-05 at 5 13 10 PM

whynowy:
What is the reason the MonoVertex can not be deleted/paused? we don't have the UI features?

Screenshot 2024-11-05 at 5 16 59 PM

vigith:
also reduce the spacing between each entities (elements pod and container info)?

veds-g:
> What is the reason the MonoVertex can not be deleted/paused? we don't have the UI features?
> 
> Screenshot 2024-11-05 at 5 16 59 PM

we did not implement it as mono vertex did not support pause / resume initially.

vigith:
> we did not implement it as mono vertex did not support pause / resume initially. also for 
delete / edit the monovertex, we need to add backend support as well.

i think we can do that in follow up PR

whynowy (dismissed):

whynowy (approved):

PR #2207, opened 2024-11-06
feat: unify metrics ( cleanup and add missing metrics )

### LINKED ISSUE #2218: unify metrics to avoid duplicate metrics

adarsh0728:
Resolves #2218 
The PR aims to have unified metrics. 
Currently we are exposing metrics for different sources/sinks(kafka, http etc) in addition to forwarder metrics. 
Cleanup of such metrics as well as addition of any missing metrics


vigith:
How have you tested it?

pkg/sources/kafka/reader.go
```diff
@@ -159,11 +157,6 @@ loop:
 	for i := int64(0); i < count; i++ {
 		select {
 		case m := <-ks.handler.messages:
-			kafkaSourceReadCount.With(map[string]string{
-				metrics.LabelVertex:        ks.vertexName,
-				metrics.LabelPipeline:      ks.pipelineName,
-				metrics.LabelPartitionName: strconv.Itoa(int(m.Partition)),
```
whynowy, line 165: Not having this means there won't be read metrics for each Kafka partition?

adarsh0728: Yes, I thought about it. Since this metric(`kafka_source_read_total`) is implemented at message level we have the partition number but cannot be implemented for `forwarder_data_read_total`. Should we keep this?

whynowy: Kafka partition is not the same partition number at message level.

adarsh0728: Screenshot 2024-11-15 at 2 37 33 PM
Screenshot 2024-11-15 at 2 37 42 PM

`LabelPartitionName`: `m.Partition` could be 0, 1.. (kafka partition) in case of `kafka_source_read_total`
But in case of `forwarder_data_read_total`, we cannot distinguish read metrics based on kafka partition. Is my understanding correct?

whynowy: Can we use `Offset.PartitionIdx()` as partition?

adarsh0728: Screenshot 2024-11-18 at 3 03 55 PM

as discussed, brought read counter metrics for source inside for loop:
```
		metrics.ReadDataMessagesCount.With(map[string]string{
			metrics.LabelVertex:             df.vertexName,
			metrics.LabelPipeline:           df.pipelineName,
			metrics.LabelVertexType:         string(dfv1.VertexTypeSource),
			metrics.LabelVertexReplicaIndex: strconv.Itoa(int(df.vertexReplica)),
			metrics.LabelPartitionName:      strconv.Itoa(int(m.ID.Index)), // this is equal to readOffset.PartitionIdx() which is equal to m.Partition
		}).Inc()
```

pkg/sources/forward/data_forward.go
```diff
@@ -258,49 +259,57 @@ func (df *DataForward) forwardAChunk(ctx context.Context) error {
...
+	df.srcIdleHandler.Reset()
 
 	// store the offsets of the messages we read from source
-
-	// store the offsets of the messages we read from ISB
 	var readOffsets = make([]isb.Offset, len(readMessages))
 	for idx, m := range readMessages {
-		totalBytes += len(m.Payload)
+		totalBytes := len(m.Payload)
+
```
whynowy, line 277: ItoA replica out of the loop, and do the partitionIdx here to improve performance?

whynowy:
@KeranYang @kohlisid @yhl25 - please review.

pkg/sinks/forward/forward.go
```diff
@@ -524,10 +545,14 @@ func (df *DataForward) updateSinkWriteMetrics(writeCount int, writeBytes float64
...
 	metrics.WriteBytesCount.With(labels).Add(writeBytes)
 
+	// Add write processing time metric
+	metrics.WriteProcessingTime.With(labels).Observe(float64(time.Since(writeStart).Microseconds()))
+
 	// if this is for Fallback Sink, increment specific metrics as well
 	if isFallback {
 		metrics.FbSinkWriteMessagesCount.With(labels).Add(float64(writeCount))
 		metrics.FbSinkWriteBytesCount.With(labels).Add(writeBytes)
+		metrics.FbSinkWriteProcessingTime.With(labels).Observe(float64(time.Since(writeStart).Microseconds()))
```
KeranYang, line 555: if it's fallback sink, do we publish for both WriteProcessingTime and FbSinkWriteProcessingTime?

pkg/sinks/kafka/metrics.go
```diff
@@ -23,20 +23,6 @@ import (
...
-}, []string{metrics.LabelVertex, metrics.LabelPipeline})
-
-// kafkaSinkWriteCount is used to indicate the number of messages written to kafka
-var kafkaSinkWriteCount = promauto.NewCounterVec(prometheus.CounterOpts{
-	Subsystem: "kafka_sink",
-	Name:      "write_total",
-	Help:      "Total number of errors on NewToKafka",
-}, []string{metrics.LabelVertex, metrics.LabelPipeline})
-
 var kafkaSinkWriteTimeouts = promauto.NewCounterVec(prometheus.CounterOpts{
```
KeranYang, line 26: can we also move this to forwarder?

pkg/sources/forward/data_forward.go
```diff
@@ -258,49 +259,57 @@ func (df *DataForward) forwardAChunk(ctx context.Context) error {
...
+		metrics.ReadDataMessagesCount.With(map[string]string{
+			metrics.LabelVertex:             df.vertexName,
+			metrics.LabelPipeline:           df.pipelineName,
+			metrics.LabelVertexType:         string(dfv1.VertexTypeSource),
+			metrics.LabelVertexReplicaIndex: strconv.Itoa(int(df.vertexReplica)),
+			metrics.LabelPartitionName:      strconv.Itoa(int(m.ReadOffset.PartitionIdx())),
```
KeranYang, line 283: This is the first time we use partition index as partition name? and I think it's only for source. Can we introduce a new label say metrics.LabelSourcePartitionIndex, for better naming?

pkg/sources/kafka/metrics.go
```diff
@@ -23,27 +23,6 @@ import (
...
-
-// kafkaSourceOffsetAckErrors is used to indicate the number of errors while reading from kafka source with offsets
-var kafkaSourceOffsetAckErrors = promauto.NewCounterVec(prometheus.CounterOpts{
-	Subsystem: "kafka_source",
-	Name:      "ack_error_total",
-	Help:      "Total number of Kafka ID Errors",
-}, []string{metrics.LabelVertex, metrics.LabelPipeline})
-
 // kafkaPending is used to indicate the number of messages pending in the kafka source
 var kafkaPending = promauto.NewGaugeVec(prometheus.GaugeOpts{
```
KeranYang, line 27: can we also remove this one?

pkg/sources/udsource/metrics.go
```diff
@@ -23,13 +23,6 @@ import (
...
 
-// udsourceReadCount is used to indicate the number of messages read
-var udsourceReadCount = promauto.NewCounterVec(prometheus.CounterOpts{
-	Subsystem: "udsource",
-	Name:      "read_total",
-	Help:      "Total number of messages Read",
-}, []string{metrics.LabelVertex, metrics.LabelPipeline, metrics.LabelVertexReplicaIndex, metrics.LabelPartitionName})
-
 // udsourcePending is used to indicate the number of messages pending in the user defined source
 var udsourcePending = promauto.NewGaugeVec(prometheus.GaugeOpts{
```
KeranYang, line 27: same as above, can pending also be generalized?

pkg/sinks/forward/forward.go
adarsh0728: Yes, for fallback sink, normal sink metrics are being published in addition to fb sink.
For eg: if we are writing to fb sink, then `WriteMessagesCount` is incremented (successfully written to fb sink after try with normal sink), and we increment `FbSinkWriteMessagesCount` as well to know the message count written specifically to fb sink.

pkg/sources/forward/data_forward.go
adarsh0728: We were using partition idx as partition name in kafka source. Since this is only for source, thought of keeping it as partition name only. Because I think for sink, having `metrics.LabelSourcePartitionIndex` would then be redundant as `LabelPartitionName`(below) would already capture the partition idx of buffer. 
`metrics.LabelPartitionName:      df.fromBufferPartition.GetName(),`

```
	for i := 0; i < av.GetPartitionCount(); i++ {
		r = append(r, GenerateBufferName(namespace, pipeline, av.Name, i))
	}
```

pkg/sources/kafka/metrics.go
adarsh0728: we have `vertex_pending_messages` but I'm doubtful whether it can replace `kafkaPending` and `udsourcePending`. as per my understanding, 

```
if pending, err := lagReader.Pending(ctx); err != nil {
```
here we are calling below to build up pending info

```
func (ks *kafkaSource) Pending(_ context.Context) (int64, error) {
	if ks.adminClient == nil || ks.saramaClient == nil {
```

@whynowy can you please help here

pkg/sinks/forward/forward.go
KeranYang: I see, so for a batch size of 100, if I successfully wrote 80 to `primary-sink` and 20 to `fallback` sink, the metric will show `WriteMessageCount` as 100 and and `FbSinkWriteMessagesCount` 20. The label is the vertex name so it makes sense. Thanks.

pkg/sources/forward/data_forward.go
adarsh0728: Got your point. I'm inclined to go with option b. (metrics.LabelPartitionName: "source-partition-" + strconv.Itoa(int(m.ReadOffset.PartitionIdx()))

KeranYang (changes requested):
You probably noticed that one of the e2e tests is failing. It's because the vertex processing rate calculation stopped working. This is because in numaflow, for each vertex, we use `forwarder_data_read_total` metric to calculate processing rate. See [here](https://github.com/numaproj/numaflow/blob/main/pkg/daemon/server/service/rater/rater.go#L220). Please make sure that we don't change this name for ALL type of vertices. @adarsh0728

adarsh0728:
> You probably noticed that one of the e2e tests is failing. It's because the vertex processing rate calculation stopped working. This is because in numaflow, for each vertex, we use `forwarder_data_read_total` metric to calculate processing rate. See [here](https://github.com/numaproj/numaflow/blob/main/pkg/daemon/server/service/rater/rater.go#L220). Please make sure that we don't change this name for ALL type of vertices. @adarsh0728


Thanks, checking

pkg/sources/kafka/metrics.go
whynowy: Let's keep it for now? In case ppl want to see the real time kafka pending.

pkg/sources/udsource/metrics.go
adarsh0728: https://github.com/numaproj/numaflow/pull/2207#discussion_r1851336667

adarsh0728:
> @adarsh0728 please take a look at https://github.com/numaproj/numaflow/blob/main/pkg/daemon/server/service/rater/rater_test.go#L63 to understand how the metric is used to calculate data processing rate and see if your PR somehow breaks the contract.

@KeranYang @whynowy: I think I got the issue. 
1. `TestUserDefinedSourceSuite` test case is failing which involves testing `testSimpleSource`. To calculate processing rates, `client.GetVertexMetrics(context.Background(), pipelineName, vertexName)` is called [here](https://github.com/numaproj/numaflow/blob/c5afc906995c6cf2b4f0fa8f4b9c7d8c54577f5f/test/udsource-e2e/udsource_test.go#L126)

2. [Here](https://github.com/numaproj/numaflow/blob/c5afc906995c6cf2b4f0fa8f4b9c7d8c54577f5f/pkg/daemon/server/service/pipeline_metrics_query.go#L132), in `GetVertexMetrics` function we are assuming source to have single partition (vertex name itself) and subsequently calling [getRates](https://github.com/numaproj/numaflow/blob/c5afc906995c6cf2b4f0fa8f4b9c7d8c54577f5f/pkg/daemon/server/service/pipeline_metrics_query.go#L143) for this partition. 
3. With PR changes, we have changed source's partition name to 'source-partition-[partitionIdx]' which is causing the issue.

What should be the best possible way out here? Should I introduce a new label for partition idx as suggested by Keran before.
Screenshot 2024-11-21 at 6 02 39 PM
Screenshot 2024-11-21 at 6 04 16 PM

KeranYang:
@adarsh0728 thank you for diving deep into the issue. The problem is introduced because we move Kafka per-partition metrics to source forwarder. I think we can simplify it by keeping Kafka per-partition metric in Kafka source and continue using vertex name as source partition name. cc: @whynowy

pkg/sources/forward/data_forward.go
```diff
@@ -334,14 +312,8 @@ func (df *DataForward) forwardAChunk(ctx context.Context) error {
 			zap.Duration("took", time.Since(transformerProcessingStart)),
 		)
 
-		metrics.SourceTransformerProcessingTime.With(map[string]string{
-			metrics.LabelVertex:             df.vertexName,
-			metrics.LabelPipeline:           df.pipelineName,
-			metrics.LabelVertexReplicaIndex: strconv.Itoa(int(df.vertexReplica)),
-			metrics.LabelPartitionName:      df.reader.GetName(),
-		}).Observe(float64(time.Since(transformerProcessingStart).Microseconds()))
+		metrics.SourceTransformerProcessingTime.With(map[string]string{metrics.LabelVertex: df.vertexName, metrics.LabelPipeline: df.pipelineName, metrics.LabelVertexReplicaIndex: replicaIndex, metrics.LabelPartitionName: df.reader.GetName()}).Observe(float64(time.Since(transformerProcessingStart).Microseconds()))
```
KeranYang, line 315: can we add vertex type so that we can re-use `metricLabelsWithPartition`?

pkg/sources/forward/data_forward.go
```diff
@@ -258,49 +257,28 @@ func (df *DataForward) forwardAChunk(ctx context.Context) error {
...
-	}).Add(float64(len(readMessages)))
-
 	// store the offsets of the messages we read from source
-
-	// store the offsets of the messages we read from ISB
 	var readOffsets = make([]isb.Offset, len(readMessages))
 	for idx, m := range readMessages {
-		totalBytes += len(m.Payload)
-
+		mssgLabels := map[string]string{
```
KeranYang, line 267: the suggestion to use `metricLabelsWithPartition` so that it won't break rate calculation.

adarsh0728:
> > @adarsh0728 thank you for diving deep into the issue. The problem is introduced because we move Kafka per-partition metrics to source forwarder. I think we can simplify it by keeping Kafka per-partition metric in Kafka source and continue using vertex name as source partition name. cc: @whynowy
> 
> 
> 
> sounds good to me.

thanks @whynowy and @KeranYang . Will do the necessary changes and update.

yhl25 (approved):
LGTM!

KeranYang (approved):
LGTM. Please double check `metrics.md` in sync with your changes.

PR #2020, opened 2024-09-01
chore: convert onFailureStrategy to enum in rust

kohlisid:

PR #2208, opened 2024-11-06
feat: add sdk infomation metrics

whynowy:
For MonoVertex:
```
# HELP sdk_info A metric with a constant value '1', labeled by SDK information such as version, language, and type.
# TYPE sdk_info gauge
sdk_info{component="mono-vertex",component_name="simple-mono-vertex",language="rust",version="0.1.1",type="sourcer"} 1
sdk_info{component="mono-vertex",component_name="simple-mono-vertex",language="rust",version="0.1.1",type="sinker"} 1
sdk_info{component="mono-vertex",component_name="simple-mono-vertex",language="rust",version="0.1.1",type="sourcetransformer"} 1
```

For Vertex:
```
# HELP sdk_info A metric with a constant value '1', labeled by SDK information such as version, language, and type
# TYPE sdk_info gauge
sdk_info{component="vertex",component_name="simple-pipeline-cat",language="go",type="mapper",version="v0.9.0"} 1
```

Similarly, there's a metric emitted for side inputs manager.


rust/numaflow-core/src/shared/server_info.rs
```diff
@@ -18,6 +18,17 @@ use crate::shared::server_info::version::SdkConstraints;
 // Equivalent to U+005C__END__.
 const END: &str = "U+005C__END__";
 
+pub const CONTAINER_TYPE_SOURCER: &str = "sourcer";
```
whynowy, line 21: What is the best way to do this in rust?

kohlisid: Enum?

whynowy: Enum with string? Not seems to work.

kohlisid: Do you want something like this?
https://github.com/numaproj/numaflow/pull/2020

KeranYang:
do we need extra pipeline and vertex name dimensions? one pipeline can have multiple vertices using different SDK versions

KeranYang:
And just curious, what is the main use case of introducing this metric?

rust/numaflow-core/src/shared/server_info.rs
whynowy: Constant seems to be simpler for this specific use case.

whynowy:
> do we need extra pipeline and vertex name dimensions? one pipeline can have multiple vertices using different SDK versions

No need for that I think.

rust/numaflow-core/src/metrics.rs
```diff
@@ -366,6 +400,21 @@ pub(crate) fn forward_pipeline_metrics() -> &'static PipelineMetrics {
     PIPELINE_METRICS.get_or_init(PipelineMetrics::new)
 }
 
+
+// forward_metrics_labels is a helper function used to fetch the
+// SDK_INFO_LABELS object
```
vigith, line 404: wrong comment?

rust/numaflow-core/src/pipeline.rs
```diff
@@ -267,11 +270,10 @@ async fn create_transformer(
         if let config::components::transformer::TransformerType::UserDefined(ud_transformer) =
             &transformer_config.transformer_type
         {
-            check_for_server_compatibility(
-                ud_transformer.socket_path.clone().into(),
-                cln_token.clone(),
-            )
-            .await?;
+            _ = sdk_server_info(ud_transformer.socket_path.clone().into(), cln_token.clone())
+                .await?;
```
vigith, line 274:
```suggestion
            _ = sdk_server_info(ud_transformer.socket_path.clone().into(), cln_token.clone())
                .await?;
```

do we need this change?

rust/numaflow-core/src/shared/server_info.rs
vigith: idiomatic rust is to use enum and `impl std::fmt::Display` for getting the string version string.

rust/numaflow-core/src/shared/utils.rs
```diff
@@ -186,16 +186,15 @@ pub(crate) async fn create_sink_handle(
             None,
         )),
         SinkType::UserDefined(ud_config) => {
-            check_for_server_compatibility(
-                ud_config.server_info_path.clone().into(),
-                cln_token.clone(),
-            )
-            .await?;
+            _ = sdk_server_info(ud_config.server_info_path.clone().into(), cln_token.clone())
+                .await?;
```
vigith, line 190: any reason for this change?

rust/numaflow-core/src/pipeline.rs
whynowy: Do we need to handle the the return in rust? Also, I assume something needs to be added to for the returned server info later on - this is for pipeline.

rust/numaflow-core/src/shared/utils.rs
whynowy: Same reason as the other one.

rust/numaflow-core/src/pipeline.rs
vigith: `?` does the return on error. the success is a zero-size-type `()` which has no meaning.

whynowy: The success is changed to ServerInfo.

vigith: then it is fine, i missed that

vigith (approved):

PR #2221, opened 2024-11-15
fix: add decimal precision to CPU and memory metrics in pod info

### LINKED ISSUE #2225: Add decimal precision to CPU and memory metrics in pod info
Bug:
Pod Info box in the UI has CPU/Memory values which are rounding to integers.

Reproduce:
Select any vertex of a pipeline or a monovertex, the Pod Info box shows cpu/mem as integer values

Expected behavior
We should show values upto 2 decimal places to have parity with other values (container info and hexagonal heat map)

Changes Required:
Backend changes are only required as the API ( `pods-info` ) handles the logic completely.

Screenshot 2024-11-18 at 7 36 17 PM

adarsh0728:
Fixes #2225 

Pod Info Box:
- Format CPU millicores with 2 decimal places
- Format memory Mi with 2 decimal places

Changes: Only Backend changes are done in this PR as logic is completely handled in the API ( `pods-info` ) 

Screenshot 2024-11-15 at 10 09 51 PM


adarsh0728:
Currently, Container Info's CPU/Mem usage and HexagonHeatMap(container level usage) comes from `pods` API. Pod Info's CPU/Mem usage comes from `pods-info` API. In future, we would have all the CPU/Mem data from `pods-info` API. Changes in UI code would be required, as `pods-info` API already send container level data.

server/apis/v1/handler.go
```diff
@@ -1684,3 +1684,24 @@ func (h *handler) getContainerStatus(state corev1.ContainerState) string {
 		return "Unknown"
 	}
 }
+
+func cpuToMillicores(quantityStr string) string {
```
whynowy, line 1688: Why don't take `Quantity` directly so that there's no need to parse?

server/apis/v1/handler.go
```diff
@@ -1684,3 +1684,24 @@ func (h *handler) getContainerStatus(state corev1.ContainerState) string {
...
+	var value float64
+	var format string
+
+	// Parse the quantity string
+	if _, err := fmt.Sscanf(quantityStr, "%f%s", &value, &format); err != nil {
+		fmt.Println("Error parsing cpu quantity string:", err)
```
whynowy, line 1694: If there's a need for error handling, return it and handle it in upper layer.

whynowy:
Do we need to keep 2 decimals? I think no decimal is okay.

adarsh0728:
> Do we need to keep 2 decimals? I think no decimal is okay.

We already have decimals in container info box and hexagonal heat map. This is done to bring parity. Also, sometimes a pod can have only one container - in pod info we are showing lets say 3m but in container box it's 2.7m (these should be equal)

server/apis/v1/handler.go
adarsh0728: [Quantity](https://pkg.go.dev/k8s.io/apimachinery@v0.31.2/pkg/api/resource#Quantity) is like:
`` <quantity> ::= <signedNumber><suffix>. To get the suffix and calculate accordingly we're parsing the string.
For eg: 2847159n this is the cpuQuantity, `cpuQuantity.MilliValue()` will give us 3, but we want 2.84m as our final result. For that we divide 2847159/1e6

PR #2226, opened 2024-11-18
chore: rename and respect pipeline.deletionGracePeriodSeconds

whynowy:
1. It should be called `deletionGracePeriodSeconds` as a common practice in `spec.lifecycle` and fix the doc.
2. Respect `pipeline.deletionGracePeriodSeconds` as well.


vigith (approved):

pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -419,39 +419,50 @@ func (p Pipeline) GetPipelineLimits() PipelineLimits {
...
+	if p.Spec.Lifecycle.DeletionGracePeriodSeconds != nil {
+		return *p.Spec.Lifecycle.DeletionGracePeriodSeconds
+	}
+	if p.Spec.Lifecycle.DeprecatedDeleteGracePeriodSeconds != nil {
+		return *p.Spec.Lifecycle.DeprecatedDeleteGracePeriodSeconds
+	}
+	if p.DeletionGracePeriodSeconds != nil {
+		return *p.DeletionGracePeriodSeconds
 	}
 	return 30
```
kohlisid, line 452: Could we make this as a constant

pkg/apis/numaflow/v1alpha1/pipeline_types.go
```diff
@@ -419,39 +419,50 @@ func (p Pipeline) GetPipelineLimits() PipelineLimits {
...
 
 // return PauseGracePeriodSeconds if set
-func (lc Lifecycle) GetPauseGracePeriodSeconds() int32 {
-	if lc.PauseGracePeriodSeconds != nil {
-		return *lc.PauseGracePeriodSeconds
+func (p Pipeline) GetPauseGracePeriodSeconds() int64 {
+	if p.Spec.Lifecycle.PauseGracePeriodSeconds != nil {
+		return *p.Spec.Lifecycle.PauseGracePeriodSeconds
 	}
 	return 30
```
kohlisid, line 467: Could we make this as a constant

kohlisid (approved):
LGTM!
Small nits

kohlisid:
@whynowy +1 for not adding the complexity layer of checking here. 
I wanted to confirm our expectation

pkg/apis/numaflow/v1alpha1/pipeline_types.go
whynowy: Not a big fan of making it as a constant if it's only used in one place - considering there's some other things need to be updated (but could not use the constant, `// +kubebuilder:default=30`) when we want to make a change. In that case, prefer to keep them as close as possible. No strong opinion though.

PR #2232, opened 2024-11-20
fix: honor lookbackSeconds for mvtx

kohlisid:
Currently for monovertex if the lookbackSeconds is specified in the spec, the metric lag reader does not honor that value and keeps using default 
```
metadata:
  name: simple-mono-vertex
spec:
  scale:
    lookbackSeconds: 100
```

```
2024-11-20T23:39:37.561609Z  INFO numaflow_core::monovertex::forwarder: Forwarded 26200 messages at time 2024-11-20 23:39:37.561606851 UTC
2024-11-20T23:39:38.568245Z  INFO numaflow_core::monovertex::forwarder: Forwarded 25600 messages at time 2024-11-20 23:39:38.568242351 UTC
2024-11-20T23:39:39.535464Z  INFO numaflow_core::metrics: MYDEBUG LOOKBACK 1m 60
2024-11-20T23:39:39.535513Z  INFO numaflow_core::metrics: MYDEBUG LOOKBACK default 120
```

With this fix it should update the value based on the config 

```
"uds", language: "rust", minimum_numaflow_version: "1.3.1-z", version: "0.1.1", metadata: Some({}) }
2024-11-20T23:12:56.429612Z  INFO numaflow_core::shared::server_info: Server info file: ServerInfo { protocol: "uds", language: "rust", minimum_numaflow_version: "1.3.1-z", version: "0.1.1", metadata: Some({}) }
2024-11-20T23:12:56.469090Z  INFO numaflow_core::metrics: MYDEBUG LOOKBACK 1m 60
2024-11-20T23:12:56.469104Z  INFO numaflow_core::metrics: MYDEBUG LOOKBACK default 100
```

### DIFF
--- a/pkg/apis/numaflow/v1alpha1/mono_vertex_types.go
+++ b/pkg/apis/numaflow/v1alpha1/mono_vertex_types.go
@@ -318,7 +318,7 @@ func (mv MonoVertex) simpleCopy() MonoVertex {
 
 func (mv MonoVertex) GetPodSpec(req GetMonoVertexPodSpecReq) (*corev1.PodSpec, error) {
 	copiedSpec := mv.simpleCopy()
-	copiedSpec.Spec.Scale = Scale{}
+	copiedSpec.Spec.Scale = Scale{LookbackSeconds: mv.Spec.Scale.LookbackSeconds}
 	monoVtxBytes, err := json.Marshal(copiedSpec)
 	if err != nil {
 		return nil, errors.New("failed to marshal mono vertex spec")
--- a/rust/numaflow-core/src/config.rs
+++ b/rust/numaflow-core/src/config.rs
@@ -45,7 +45,6 @@ pub(crate) struct Settings {
 /// LookbackStruct is a struct to hold the lookback window map for the monovertex/pipeline.
 #[derive(Default)]
 pub(crate) struct LookbackStruct {
-    // It is okay to use std mutex because we register each metric only one time.
     window_map: parking_lot::Mutex<[(&'static str, i64); 4]>,
 }
 
--- a/rust/numaflow-core/src/config.rs
+++ b/rust/numaflow-core/src/config.rs
@@ -1,9 +1,14 @@
-use std::env;
-use std::sync::OnceLock;
-
+use futures::future::Lazy;
+use log::info;
 use monovertex::MonovertexConfig;
+use parking_lot::lock_api::MutexGuard;
+use parking_lot::RawMutex;
+use prometheus_client::registry::Registry;
+use std::env;
+use std::sync::{Mutex, OnceLock};
 
 use crate::config::pipeline::PipelineConfig;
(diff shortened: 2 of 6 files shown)


rust/numaflow-core/src/config.rs
```diff
@@ -42,20 +42,63 @@ pub(crate) struct Settings {
     pub(crate) custom_resource_type: CustomResourceType,
 }
 
+/// LookbackStruct is a struct to hold the lookback window map for the monovertex/pipeline.
+#[derive(Default)]
+pub(crate) struct LookbackStruct {
+    window_map: parking_lot::Mutex<[(&'static str, i64); 4]>,
```
vigith, line 48: why do you need parking_lot?

pkg/apis/numaflow/v1alpha1/mono_vertex_types.go
```diff
@@ -318,7 +318,7 @@ func (mv MonoVertex) simpleCopy() MonoVertex {
 
 func (mv MonoVertex) GetPodSpec(req GetMonoVertexPodSpecReq) (*corev1.PodSpec, error) {
 	copiedSpec := mv.simpleCopy()
-	copiedSpec.Spec.Scale = Scale{}
+	copiedSpec.Spec.Scale = Scale{LookbackSeconds: mv.Spec.Scale.LookbackSeconds}
```
whynowy, line 321: Didn't we have it already?

whynowy: ha, ignore it.

rust/numaflow-core/src/config.rs
```diff
@@ -42,20 +42,63 @@ pub(crate) struct Settings {
...
+}
+
+impl LookbackStruct {
+    fn new() -> Self {
```
whynowy, line 52: `fun new(defaultLookback: i64) -> Self`?

rust/numaflow-core/src/config.rs
kohlisid: We might be moving to a dynamic approach where this could be updated at runtime. 
Hence, added that for now

rust/numaflow-core/src/config.rs
kohlisid: I was exploring that as well, just one issue that I observed was in the testing flow. Where because the config is not available its difficult to different for the call here what to pass
https://github.com/numaproj/numaflow/pull/2232/files#diff-4057e535d3ca33f1b08e56689bfca36f9ce5b7845bc1a6915aa3d343f0aefa5aR748

rust/numaflow-core/src/config.rs
vigith: `LookbackStruct` is built at the startup, it doesn't make sense to make it mutable via mutex. There shouldn't be a reason for the mutable reference to be alive along with a shared reference.

PR #2233, opened 2024-11-21
chore: avoid creating kv buckets when watermark is disabled

yhl25:


vigith (approved):

whynowy:
If watermark is enabled later, will the buckets be created?

whynowy:
And what is the reason we want to skip creating?

whynowy (changes requested):
Please address the concerns.

vigith:
this could cause regressions.

PR #2230, opened 2024-11-19
feat: use sidecar for ud containers [Breaking k8s < v1.29]

whynowy:
Use [Sidecar](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/) to run ud containers, to make sure they will be started before and terminated after numa container.

PR #2236, opened 2024-11-22
chore: make sidecar container change backward compatible

whynowy:
https://github.com/numaproj/numaflow/pull/2230

Tested in k8s < 1.29, it worked.


pkg/apis/numaflow/v1alpha1/deprecated.go
```diff
@@ -0,0 +1,33 @@
...
+
+package v1alpha1
+
+import (
+	"os"
+	"strconv"
+)
+
+// TODO: (k8s 1.29) Remove this once we deprecate the support for k8s < 1.29
+func isSidecarSupported() bool {
```
KeranYang, line 25: I see the EnvK8sServerVersion is set when we start reconciller. Will it ever be possible that we call isSidecarSupported before reconciller cmd start?

whynowy: Good question. I think it's only used in the controller, should be okay.

KeranYang (approved):

vigith (approved):

PR #2237, opened 2024-11-22
feat: Built-in Pulsar source

BulkBeing:
Status: Tested by running a Pipeline and a Monovertex with Pulsar as built-in source


rust/numaflow-core/src/message.rs
```diff
@@ -119,11 +120,32 @@ impl TryFrom<async_nats::Message> for Message {
...
+            id: MessageID {
+                vertex_name: get_vertex_name().to_string(),
+                offset: offset.to_string(),
+                index: 0,
+            },
+            headers: message.headers,
+        })
+    }
+}
+
```
vigith, line 143: move this to `source/pulsar.rs`

pkg/apis/numaflow/v1alpha1/pulsar_source.go
```diff
@@ -0,0 +1,25 @@
...
+
+package v1alpha1
+
+type PulsarSource struct {
+	ServerAddr       string `json:"serverAddr" protobuf:"bytes,1,name=server_addr"`
+	Topic            string `json:"topic" protobuf:"bytes,2,name=topic"`
+	ConsumerName     string `json:"consumerName" protobuf:"bytes,3,name=consumerName"`
+	SubscriptionName string `json:"subscriptionName" protobuf:"bytes,4,name=subscriptionName"`
+	MaxUnack         uint32 `json:"maxUnack,omitempty" protobuf:"bytes,5,opt,name=maxUnack"`
+}
```
vigith, line 25: please add TLS and Auth too, if you prefer you can do that in a follow up PR.

rust/numaflow-core/src/config/components.rs
```diff
@@ -29,51 +29,87 @@ pub(crate) mod source {
...
-                })
-                .ok_or_else(|| Error::Config("Source type not found".to_string()))?
+        fn try_from(mut source: Box<Source>) -> Result<Self> {
+            if let Some(generator) = source.generator.take() {
+                return Ok(generator.into());
+            }
+
+            if source.udsource.is_some() {
+                return Ok(SourceType::UserDefined(UserDefinedConfig::default()));
+            }
```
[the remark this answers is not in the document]
BulkBeing: `Option.take()` returns the inner value as an option. We can't use it in `if` statement.

yhl25 (approved):

PR #2251, opened 2024-12-03
feat: Asynchronous Streaming for Source and Sink in Numaflow Core

yhl25:


rust/numaflow-core/src/message.rs
```diff
@@ -103,7 +79,7 @@ impl TryFrom<async_nats::Message> for Message {
         let event_time = Utc::now();
         let offset = None;
         let id = MessageID {
-            vertex_name: get_vertex_name().to_string(),
+            vertex_name: config::get_vertex_name().to_string(),
```
whynowy, line 82: Having `get_vertex_name()` in this `try_from` function is weird. Do we have to implement `try_from`, or a until function to convert Message (with vertexName in the signature) also works?

rust/numaflow-core/src/message.rs
```diff
@@ -319,7 +303,7 @@ impl TryFrom<read_response::Result> for Message {
             offset: Some(source_offset.clone()),
             event_time: utc_from_timestamp(result.event_time),
             id: MessageID {
-                vertex_name: get_vertex_name().to_string(),
+                vertex_name: config::get_vertex_name().to_string(),
```
whynowy, line 306: Same question.

rust/numaflow-core/src/config/monovertex.rs
```diff
@@ -94,6 +94,7 @@ impl MonovertexConfig {
             .as_ref()
             .and_then(|source| source.transformer.as_ref())
             .map(|_| TransformerConfig {
+                concurrency: batch_size as usize, // FIXME: introduce a new config called udf concurrency in the spec
```
whynowy, line 97: Do we need to?

rust/numaflow-core/src/message.rs
vigith: what do you mean by an `until` function?

rust/numaflow-core/src/config/monovertex.rs
vigith: this is to make sure we can achieve X workers in parallel immaterial of the messages from previous vertex. we do read-ahead to achieve X concurrency/parallelism.

vigith (approved):
please remove self.clone() in source.rs

PR #2261, opened 2024-12-05
fix: Fix Sink Config to respect Fallback

yhl25:


pkg/apis/numaflow/v1alpha1/mono_vertex_types.go
```diff
@@ -472,10 +472,9 @@ func (mvspec MonoVertexSpec) buildContainers(req getContainerReq) []corev1.Conta
 	if mvspec.Sink.UDSink != nil { // Only support UDSink for now.
 		containers = append(containers, mvspec.Sink.getUDSinkContainer(req))
 	}
-	if mvspec.Sink.Fallback != nil {
+	if mvspec.Sink.Fallback != nil && mvspec.Sink.Fallback.UDSink != nil {
 		containers = append(containers, mvspec.Sink.getFallbackUDSinkContainer(req))
 	}
```
vigith, line 477: don't we support builtins now?

yhl25: we do support, this check is to avoid creating sidecar container when the fallback sink is not udsink

rust/numaflow-core/src/sink/user_defined.rs
```diff
@@ -1,14 +1,15 @@
...
 use tokio::sync::mpsc;
 use tokio_stream::wrappers::ReceiverStream;
 use tonic::transport::Channel;
 use tonic::{Request, Streaming};
 
+use crate::message::{Message, ResponseFromSink};
+use crate::sink::Sink;
+use crate::Error;
+use crate::Result;
+
```
vigith, line 12: revert this change?

pkg/apis/numaflow/v1alpha1/mono_vertex_types.go
vigith: gotcha, this was a bug

PR #2265, opened 2024-12-06
fix: Sink Config to respect Fallback

yhl25:


vigith (approved):

rust/numaflow-core/src/config/components.rs
```diff
@@ -189,14 +189,29 @@ pub(crate) mod sink {
...
     }
 
-    impl TryFrom<Box<Sink>> for SinkType {
-        type Error = Error;
-
+    impl SinkType {
         // FIXME(cr): why is sink.fallback Box<AbstrackSink> vs. sink Box<Sink>. This is coming from
         //   numaflow-models. Problem is, golang has embedded structures and rust does not. We might
         //   have to AbstractSink for sink-configs while Sink for real sink types.
         //   NOTE: I do not see this problem with Source?
```
yhl25, line 196: @whynowy Does this FIXME makes sense to you?

PR #2266, opened 2024-12-06
chore: Define common dependency versions in root Cargo.toml

BulkBeing:
- Removes unused dependencies
- Specify common workspace dependency versions in root manifest file.

To do:
- Move to using virtual manifest
  -  Advantage: No need to pass `--workspace` for all cargo command. We can still run commands for individual packages either by running it from their directory or by passing `-p` option. https://matklad.github.io/2021/08/22/large-rust-workspaces.html
  - `src/main.rs` will have to be moved to be a workspace member (with package name `numaflow`). Similar to `rust-analyzer` https://github.com/rust-lang/rust-analyzer/tree/master/crates/rust-analyzer
-  Move from `numaflow-extns/pulsar` to `numaflow-extns-pulsar` as `cargo` has no meaning for sub-directories(The package name is just `pulsar`)
- Add more lints https://github.com/tokio-rs/axum/blob/7c871e9591bd997a04104fb0cfcd52c40ce15ad7/Cargo.toml#L13-L49

---
Update: All changes are done. Extensions will be placed under `extns` directory. Eg: `extns/numaflow-pulsar` with package name being `numaflow-puslar`


vigith:
this change looks good to me.

BulkBeing:
> how will it look like if extensions are not flattened?

Currently, we rename the crate in numaflow-core's Cargo.toml as numaflow_pulsar

vigith:
> Currently, we rename the crate in numaflow-core's Cargo.toml as numaflow_pulsar

my worry is that there many too many extension folders, as per the blog post you shared it fine, but still 😀

@BulkBeing what do you recommend? I am okay either way

yhl25 (approved):
LGTM!

vigith (approved):

PR #2270, opened 2024-12-09
feat: Conditional Forwarding In Asynchronous Pipeline 

yhl25:


rust/numaflow-core/src/pipeline/forwarder/source_forwarder.rs
```diff
@@ -11,22 +11,22 @@ use crate::transformer::Transformer;
 pub(crate) struct SourceForwarder {
     source: Source,
     transformer: Option<Transformer>,
-    writer: ISBWriter,
+    writer: JetstreamWriter,
```
whynowy, line 14: Why do we use JetstreamWriter directly?

vigith: we do not have a trait defined yet.

vigith (approved):

PR #2273, opened 2024-12-11
gRPC connector that allows self-signed certs

BulkBeing:
This approach works on Mac with Golang gRPC server listening on https (self-signed certs). But it failing in k8s with below error:
```
2024-12-11T15:33:23.693266Z  INFO numaflow_core::metrics: Connecting to Monovertex daemon uri=https://10.96.76.169:4327/
thread 'tokio-runtime-worker' panicked at /numaflow/numaflow-core/src/metrics.rs:896:14:
daemon should be up: Status { code: Unknown, message: "client error (UserUnsupportedVersion)", source: Some(hyper_util::client::legacy::Error(UserUnsupportedVersion)) }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2024-12-11T15:33:23.693934Z  WARN hyper_util::client::legacy::client: Connection is HTTP/1, but request requires HTTP/2
```

<s>This may be because of ALPN https://github.com/seanmonstar/reqwest/discussions/2350#discussioncomment-10020959
However, the unit tests shows ALPN protocols are set when we use `enable_http2()` method.</s>

It seems hyper's client pool is http1 and tonic is trying to send http2 request https://docs.rs/hyper-util/0.1.10/src/hyper_util/client/legacy/client.rs.html#289

UPDATE:
This was because of https://github.com/golang/go/issues/46602 and https://github.com/golang/go/issues/14619


pkg/mvtxdaemon/server/daemon_server.go
```diff
@@ -65,20 +65,18 @@ func (ds *daemonServer) Run(ctx context.Context) error {
...
+		return fmt.Errorf("failed to listen: %v", err)
 	}
 
 	cer, err := sharedtls.GenerateX509KeyPair()
 	if err != nil {
 		return fmt.Errorf("failed to generate cert: %w", err)
 	}
 
-	tlsConfig := &tls.Config{Certificates: []tls.Certificate{*cer}, MinVersion: tls.VersionTLS12}
+	tlsConfig := &tls.Config{Certificates: []tls.Certificate{*cer}, MinVersion: tls.VersionTLS12, NextProtos: []string{"h2", "http/1.1"}}
```
whynowy, line 79: do not make this change.

vigith: i am moving this PR to draft state.

BulkBeing: > do not make this change.

Removed.

BulkBeing:
When we specify`tls.Config.NextProtos = []string{"h2", "http/1.1"}`, kubelet's health check tries to do health check probes with HTTP/2 request. The `cmux` we use thinks it is a gRPC request and results in `415 Unsupported content type`. Since it breaks the k8s health checks, we won't be making this change.

Currently, this PR would only be adding a utility hyper client for future use with gRPC servers that use self-signed certs.

vigith (approved):

vigith:
@kohlisid please make the changes @whynowy suggested.

PR #2283, opened 2024-12-13
fix: set max decode size of proto message, add mvtx metrics

yhl25:


vigith (commented):
let's also think whether (1000ms x 5) retry is enough

yhl25:
> let's also think whether (1000ms x 5) retry is enough

I think 5 is less, I have increased the retry count to 60 (1000ms * 60).

rust/numaflow-core/src/shared/grpc.rs
```diff
@@ -90,7 +90,7 @@ pub(crate) fn prost_timestamp_from_utc(t: DateTime<Utc>) -> Option<Timestamp> {
 
 pub(crate) async fn create_rpc_channel(socket_path: PathBuf) -> error::Result<Channel> {
     const RECONNECT_INTERVAL: u64 = 1000;
-    const MAX_RECONNECT_ATTEMPTS: usize = 5;
+    const MAX_RECONNECT_ATTEMPTS: usize = 60;
```
vigith, line 93: does this print for every retry? else users might think why it is hanging. please print a `error!` for every retry?

yhl25: I added a `warn! `between the retries, if the retries are exhausted we print `error!`.

vigith (approved):

PR #2288, opened 2024-12-15
chore: always publish watermark to sink OT

whynowy:
Explain what this PR does.


vigith (approved):
i am not sure why we weren't sending it earlier.

PR #2291, opened 2024-12-17
chore: make serving config compatible with numaflow-core configs

BulkBeing:
Closes https://github.com/numaproj/numaflow/issues/2280

- Removes files in serving/config/ and moved the values to corresponding struct's `Default` trait implementation
- Removes usage of global static configuration. Instead, parse the config in the starting of the application and pass it around.

Tested by running the pipeline https://github.com/numaproj/numaproj-demo/blob/main/serving-demo/README.md


rust/serving/src/app.rs
```diff
@@ -149,19 +151,16 @@ async fn graceful_shutdown(handle: Handle) {
...
-    let js_config = &config().jetstream;
-
+async fn create_js_context(js_config: &JetStreamConfig) -> crate::Result<Context> {
     // Connect to Jetstream with user and password if they are set
-    let js_client = match (
-        env::var(ENV_NUMAFLOW_SERVING_JETSTREAM_USER),
-        env::var(ENV_NUMAFLOW_SERVING_JETSTREAM_PASSWORD),
-    ) {
-        (Ok(user), Ok(password)) => {
+    let js_client = match js_config.auth.as_ref() {
```
vigith, line 156: this will go away eventually right? we will directly write to ISB via the pipeline framework.

BulkBeing: Yes, for now I'm trying to get it to a working state with minimal changes.

rust/numaflow-core/src/config/components.rs
```diff
@@ -96,6 +97,51 @@ pub(crate) mod source {
...
+                    setting.api_auth_token = Some(secret);
+                } else {
+                    tracing::warn!("Authentication token for Serving API is specified, but the secret is empty");
+                };
+            }
+
+            if let Some(ttl) = cfg.store.ttl {
+                if ttl.is_negative() {
+                    return Err(Error::Config(format!(
+                        "TTL value for Redis store can not be negative. Provided value = {ttl:?}"
```
vigith, line 126:
```suggestion
                        "TTL value for the store can not be negative. Provided value = {ttl:?}"
```

removed the word `Redis` because `DDB` night come too.

rust/serving/src/app/jetstream_proxy.rs
```diff
@@ -33,34 +32,31 @@ use crate::config;
...
 // }
 
-const ID_HEADER_KEY: &str = "X-Numaflow-Id";
 const CALLBACK_URL_KEY: &str = "X-Numaflow-Callback-Url";
-
 const NUMAFLOW_RESP_ARRAY_LEN: &str = "Numaflow-Array-Len";
 const NUMAFLOW_RESP_ARRAY_IDX_LEN: &str = "Numaflow-Array-Index-Len";
 
-#[derive(Clone)]
 struct ProxyState<T> {
```
vigith, line 39: don't the state need to be `Clone` anymore? make sure it is a cheap clone if there is a need.

rust/serving/src/app/jetstream_proxy.rs
```diff
@@ -71,27 +67,28 @@ pub(crate) async fn jetstream_proxy<T: Clone + Send + Sync + Store + 'static>(
...
-    let id = extract_id_from_headers(&headers);
+    let id = extract_id_from_headers(&proxy_state.tid_header, &headers);
 
     // Register the ID in the callback proxy state
-    let notify = proxy_state.callback.register(id.clone());
+    let notify = proxy_state.callback.clone().register(id.clone());
 
     if let Err(e) = publish_to_jetstream(
-        proxy_state.stream,
+        proxy_state.stream.clone(),
```
vigith, line 80: is this clone too much for every request? we know that `String` is a `const` since it is stream name.

rust/serving/src/app/jetstream_proxy.rs
```diff
@@ -71,27 +67,28 @@ pub(crate) async fn jetstream_proxy<T: Clone + Send + Sync + Store + 'static>(
...
-        id.clone(),
+        proxy_state.context.clone(),
+        proxy_state.tid_header.as_str(),
+        id.as_str(),
     )
     .await
     {
         // Deregister the ID in the callback proxy state if writing to Jetstream fails
-        let _ = proxy_state.callback.deregister(&id).await;
+        let _ = proxy_state.callback.clone().deregister(&id).await;
```
vigith, line 91: same for callback.clone

rust/serving/src/config.rs
```diff
@@ -1,40 +1,24 @@
...
-                panic!("Failed to load configuration: {:?}", e);
-            }
-        }
-    })
-}
+const ENV_NUMAFLOW_SERVING_HOST_IP: &str = "NUMAFLOW_SERVING_HOST_IP";
+const ENV_NUMAFLOW_SERVING_APP_PORT: &str = "NUMAFLOW_SERVING_APP_LISTEN_PORT";
+const ENV_NUMAFLOW_SERVING_JETSTREAM_USER: &str = "NUMAFLOW_ISBSVC_JETSTREAM_USER";
+const ENV_NUMAFLOW_SERVING_JETSTREAM_PASSWORD: &str = "NUMAFLOW_ISBSVC_JETSTREAM_PASSWORD";
+const ENV_NUMAFLOW_SERVING_AUTH_TOKEN: &str = "NUMAFLOW_SERVING_AUTH_TOKEN";
```
vigith, line 21: create a mod called serving and move serving items there? config is getting bloated

rust/serving/src/config.rs
vigith, line 1: coverage is getting low

rust/serving/src/app/jetstream_proxy.rs
BulkBeing: The router now takes `Arc<ProxyState>` as the state.

rust/serving/src/app/jetstream_proxy.rs
BulkBeing: Currently, a lot of clones are happening for Router states. I will work on a separate PR to avoid those. Need to figure out whether to use `Arc` of the struct or `Arc` of individual fields, some handlers only need part of the application state. Since config lives for the duration of the application, we could make all strings to `&'static str` with `Box::leak` to avoid all cloning or use `Arc<str>` instead of `String` for cheaper cloning.

rust/serving/src/config.rs
BulkBeing: This is the `serving::config` module. As of now, it not that big. I think single module is fine for now.

rust/serving/src/app/jetstream_proxy.rs
vigith: please do as a follow up

vigith (approved):

rust/serving/src/app/jetstream_proxy.rs
BulkBeing: https://github.com/numaproj/numaflow/issues/2296

yhl25 (approved):

PR #2297, opened 2024-12-19
chore: Expose env to disable read ahead in source

yhl25:
Using the below env, we can disable the read ahead in source for both Monovertex and async pipeline
```
  containerTemplate:
    env:
      - name: READ_AHEAD
        value: "false"
```

### DIFF
--- a/rust/numaflow-core/src/source.rs
+++ b/rust/numaflow-core/src/source.rs
@@ -258,7 +258,8 @@ impl Source {
                 }
 
                 if !read_ahead_enabled {
-                    // Acquire the semaphore permit before reading the next batch
+                    // Acquire the semaphore permit before reading the next batch to make
+                    // sure we are not reading ahead and all the inflight messages are acked.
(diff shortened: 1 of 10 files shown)


vigith (commented):
isn't it better to set to false by default?

rust/numaflow-core/src/source.rs
```diff
@@ -266,6 +280,12 @@ impl Source {
                     }
                 };
                 let n = messages.len();
+
+                // If no messages are read, continue to read the next batch.
+                if n == 0 {
+                    continue;
+                }
```
vigith, line 287: will this end up as "missing metric" if the source does not have any data at all?

yhl25: yeah, it will end up in missing metric. I think its better publish metrics even when no messages are read. Reverting the change.

vigith (approved):

PR #2304, opened 2024-12-25
chore: use domain-qualified finalizer name

whynowy:
Using non domain-qualified finalizer name might cause a warning.

https://github.com/kubernetes/kubernetes/blob/f836773540cc83488250a12637e7d924078ef284/staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresource/validator.go#L89

### DIFF
--- a/pkg/reconciler/isbsvc/controller.go
+++ b/pkg/reconciler/isbsvc/controller.go
@@ -41,7 +41,9 @@ import (
 )
 
 const (
-	finalizerName = dfv1.ControllerISBSvc
+	finalizerName = "numaflow.numaproj.io/" + dfv1.ControllerISBSvc
+	// TODO: clean up the deprecated finalizer in v1.7
+	deprecatedFinalizerName = dfv1.ControllerISBSvc
 )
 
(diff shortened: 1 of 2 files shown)


pkg/reconciler/isbsvc/controller.go
```diff
@@ -41,7 +41,9 @@ import (
 )
 
 const (
-	finalizerName = dfv1.ControllerISBSvc
+	finalizerName = "numaflow.numaproj.io/" + dfv1.ControllerISBSvc
+	// TODO: clean up the deprecated finalizer in v1.7
```
vigith, line 45:
```suggestion
	// DEPRECATE(1.7): clean up the deprecated finalizer in v1.7
```
Is this a simpler search pattern?

vigith (approved):

pkg/reconciler/isbsvc/controller.go
whynowy: Might get confused with `Deprecated` annotation. We could search `Deprecated` in the future.

PR #2323, opened 2025-01-12
fix: lastScaledAt not updated during autoscaling down

whynowy:
Explain what this PR does.


pkg/reconciler/vertex/controller.go
```diff
@@ -300,7 +301,6 @@ func (r *vertexReconciler) orchestratePods(ctx context.Context, vertex *dfv1.Ver
 		}
 	}
 
-	currentReplicas := int(vertex.Status.Replicas)
```
whynowy, line 303: This is the place to set `lastScaledAt `. Should not use current `vertex.Status.Replicas` as it has been updated above.

vigith (approved):

PR #2327, opened 2025-01-13
chore: skip publishing the last empty struct when generator stops 

### LINKED ISSUE #2285: UDSink receives one unexpected message during shutdown
**Describe the bug**

I have a testing pipeline `generator -> cat -> java udsink`. I killed the pipeline using `kubectl delete -f` and at sink, I observed this message being received.  `id: cat-49-0-0, value: [], header: {}, eventtime: 2024-12-13T18:41:39.424879807Z, watermark: 0001-01-01T00Z`.

We should never expose to UDFs messages that are not produced by previous vertices.

**To Reproduce**

Numaflow Version: 1.4.0+
Java SDK Version: 0.9.0

Use the following pipeline spec. run `kubectl delete -f`.

```
kind: Pipeline
metadata:
  name: test-pipeline
spec:
  vertices:
    - name: in
      scale:
        min: 1
      source:
        generator:
          rpu: 1
          duration: 1s
    - name: cat
      scale:
        min: 1
      udf:
        builtin:
          name: cat
    - name: sink
      scale:
        min: 1
      limits:
        readBatchSize: 5
      sink:
        udsink:
          container:
            image: quay.io/numaio/numaflow-go/sink-log:stable
            imagePullPolicy: Always
  edges:
    - from: in
      to: cat
    - from: cat
      to: sink
```

See the last message being printed out:

```
User Defined Sink: {"Data":{"value":1734119381121003775},"Createdts":1734119381121003775}
User Defined Sink: {"Data":{"value":1734119382121008150},"Createdts":1734119382121008150}
User Defined Sink: {"Data":{"value":1734119383121005400},"Createdts":1734119383121005400}
User Defined Sink: {"Data":{"value":1734119384121005775},"Createdts":1734119384121005775}
User Defined Sink: {"Data":{"value":1734119385121005192},"Createdts":1734119385121005192}
User Defined Sink: 
2024/12/13 19:49:45 gracefully stopping grpc server
2024/12/13 19:49:45 error receiving from sink stream: context canceled
2024/12/13 19:49:45 error sending sink response: rpc error: code = Canceled desc = context canceled
2024/12/13 19:49:45 Stopping the SinkFn with err, context canceled
```

KeranYang:
closes #2285 

There is race condition when we terminate the generator vertex. The message channel can be closed followed by a read from the channel. [Reading from a closed channel](https://gobyexample.com/closing-channels#:~:text=Reading%20from%20a%20closed%20channel,channel%20is%20closed%20and%20empty.) will return the zero value of the struct, resulting in an empty message being sent to the next vertex. This change prevent it from being sent so that we have a consistent data structure for downstream vertices.


pkg/sources/generator/tickgen.go
```diff
@@ -198,6 +199,13 @@ loop:
 		// we implement Read With Wait semantics
 		select {
 		case r := <-mg.srcChan:
+			// when the channel is closed and empty, go will still read from the channel once and return the zero value.
+			// exclude the zero value from being passed to the next vertex,
+			// ensuring all messages produced by generator follow the same data schema.
+			if util.IsZeroStruct(r) {
```
whynowy, line 205: Why is it `zerostruct`?

KeranYang: Thanks! I updated the PR description and addressed the comment. Don't need the isZero check, just use the second parameter should work.

pkg/sources/generator/tickgen.go
```diff
@@ -197,7 +197,14 @@ loop:
...
 		// we implement Read With Wait semantics
 		select {
-		case r := <-mg.srcChan:
+		case r, ok := <-mg.srcChan:
+			// when the channel is closed and empty, go will still read from the channel once and return the zero value.
+			// exclude the zero value from being passed to the next vertex,
+			// ensuring all messages produced by generator follow the same data schema.
+			if !ok {
+				mg.logger.Info("Zero value read from the generator channel, skipping...")
+				continue
```
whynowy, line 206: In this case, shall we return instead of continue?

whynowy (approved):

PR #2332, opened 2025-01-15
feat: cpu memory metrics

adarsh0728:
- Adds Memory and CPU pattern in metrics config at pod level
- Adds Memory and CPU pattern in metrics config at container level [filter values added for container dropdown for a particular pod]

Screenshot 2025-01-20 at 2 35 37 PM
Screenshot 2025-01-20 at 2 35 46 PM
Screenshot 2025-01-20 at 2 36 05 PM


ui/src/components/pages/Pipeline/partials/Graph/partials/NodeInfo/partials/Pods/partials/PodDetails/index.tsx
```diff
@@ -115,6 +116,7 @@ export function PodDetail({
                 pipelineId={pipelineId}
                 type={type}
                 vertexId={vertexId}
+                podDetails={podDetails}
```
veds-g, line 119: since we are using pod name out of podDetails, can we just pass it instead?

config/base/numaflow-server/numaflow-server-metrics-proxy-config.yaml
```diff
@@ -178,4 +178,165 @@ data:
             - name: pod
               filters:
                 - name: pod
-                  required: false
\ No newline at end of file
+                  required: false
+    - name: mono_vertex_pod_cpu_memory_utilization
```
veds-g, line 182: can we merge monovertex and pipeline patterns for pod and containers, since it is repetitive?

veds-g: lets make it `Objects` which acts as a llist.

config/base/numaflow-server/numaflow-server-metrics-proxy-config.yaml
```diff
@@ -184,4 +189,117 @@ data:
...
+          required: false
+      metrics: 
+        # set your cpu metric name here
+        - metric_name: namespace_pod_cpu_utilization
+          # set display name as per metric name
+          display_name: CPU Utilization per Pod
+          metric_description: This metric represents the percentage utilization of cpu usage over cpu resource limits for a pod.
+          required_filters:
+            - namespace
+            - pod
```
veds-g, line 215: do we need this if we are defining filters below?

veds-g (approved):

server/apis/v1/handler.go
```diff
@@ -1297,6 +1297,16 @@ func (h *handler) GetMetricData(c *gin.Context) {
 	c.JSON(http.StatusOK, NewNumaflowAPIResponse(nil, result))
 }
 
+// Helper function to check if an object is in the list
+func isObjectInList(objects []string, object string) bool {
```
whynowy, line 1301: Use `slices.Contains()`

server/apis/v1/handler.go
```diff
@@ -1326,6 +1327,14 @@ func (h *handler) DiscoverMetrics(c *gin.Context) {
 				// Computing dimension data for each metric
 				var dimensionData []Dimensions
 				for _, dimension := range metric.Dimensions {
+					// Check if the object is "mono-vertex", skip the "vertex"(pipeline) dimension
+					if object == "mono-vertex" && dimension.Name == "vertex" {
```
whynowy, line 1331: Why is there some logic like this?

adarsh0728: I introduced `Objects` field which can be a list now instead of `Object`, because there was too much duplicity of patterns in config(handling pipeline and monoVertex separately).

Now, when, frontend wants to know the metrics for monoVertex, we don't want to send the ones with dimension vertex(pipeline). Hence the check

whynowy (approved):
Discussed offline, need to rethink and refactor the approach of doing configuration as well as the frontend implementation. Approving this PR for now to unblock.

PR #2335, opened 2025-01-17
chore: expose metrics for min and max replicas

whynowy:
Expose these metrics per Intuit ask.

### DIFF
--- a/pkg/reconciler/metrics.go
+++ b/pkg/reconciler/metrics.go
@@ -79,6 +79,20 @@ var (
 		Help:      "A metric indicates the current replicas of a Vertex",
 	}, []string{metrics.LabelNamespace, metrics.LabelPipeline, metrics.LabelVertex})
 
+	// VertexMinReplicas indicates the min replicas of a Vertex.
+	VertexMinReplicas = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: "controller",
+		Name:      "vertex_min_replicas",
+		Help:      "A metric indicates the min replicas of a Vertex",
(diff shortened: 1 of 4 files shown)


vigith (approved):

pkg/reconciler/metrics.go
```diff
@@ -92,11 +106,26 @@ var (
...
+		Help:      "A metric indicates the max replicas of a MonoVertex",
+	}, []string{metrics.LabelNamespace, metrics.LabelMonoVertexName})
 )
 
 func init() {
 	ctrlmetrics.Registry.MustRegister(BuildInfo, ISBSvcHealth, PipelineHealth,
 		MonoVertexHealth, JetStreamISBSvcReplicas, RedisISBSvcReplicas,
-		VertexDesiredReplicas, VertexCurrentReplicas, MonoVertexDesiredReplicas,
-		MonoVertexCurrentReplicas)
+		VertexDesiredReplicas, VertexCurrentReplicas, VertexMinReplicas,
```
apratinav-intuit, line 128: can we also add Pipeline Rollout object?

vigith: what do you mean by "add Pipeline Rollout object"? which metric are you looking for?

apratinav-intuit: I was referring to pipeline object following fields: spec.vertices[].scale.min and spec.vertices[].scale.max

https://numaflow.numaproj.io/user-guide/reference/autoscaling/#numaflow-autoscaling

whynowy: > can we also add Pipeline Rollout object?

No need to do that. PipelineRollout will be orchestrated to Pipeline.

PR #2349, opened 2025-01-22
chore: Upgrade Rust to 1.84

BulkBeing:
Also upgrades some of the dependencies, notably Axum to 0.8.2.
Removes `rust/Dockerfile`. We do not use this file.

Verified by running Monovertex `test/monovertex-e2e/testdata/mono-vertex-with-transformer.yaml`


vigith:
we also need to upgrade tower to 0.5.2, currently it is [0.4.x](https://github.com/numaproj/numaflow/blob/main/rust/serving/Cargo.toml#L26)

vigith (approved):

yhl25 (approved):

PR #2301, opened 2024-12-23
feat: Implement `Sourcer` traits for serving source

BulkBeing:
closes: https://github.com/numaproj/numaflow/issues/2281

PR #2357, opened 2025-01-23
feat: Controller changes to support Serving as a builtin source

BulkBeing:
fixes https://github.com/numaproj/numaflow/issues/2319

Tested the validation with below pipeline:
```yaml
kind: Pipeline
metadata:
  name: simple-pipeline
spec:
  vertices:
    - name: serving-in
      containerTemplate:
        imagePullPolicy: Never
      scale:
        min: 1
      source:
        serving:
          service: true
          msgIDHeaderKey: "X-Numaflow-Id"
          store:
            url: "redis://redis:6379"

    - name: compute-sum
      udf:
        container:
          # compute the sum
          image: quay.io/numaio/numaflow-go/reduce-sum:stable
        groupBy:
          window:
            fixed:
              length: 60s
          keyed: true
          storage:
            emptyDir: {}
      partitions: 2

    - name: serve-sink
      containerTemplate:
        imagePullPolicy: Never
      scale:
        min: 1
      sink:
        udsink:
          container:
            image: servesink:0.1
            imagePullPolicy: Never
            env:
              - name: NUMAFLOW_CALLBACK_URL_KEY
                value: "X-Numaflow-Callback-Url"
              - name: NUMAFLOW_MSG_ID_HEADER_KEY
                value: "X-Numaflow-Id"

  edges:
    - from: serving-in
      to: compute-sum
    - from: compute-sum
      to: serve-sink
```

```
➜  kubectl apply -f simple-pipeline-test.yaml
Error from server (BadRequest): error when creating "simple-pipeline-test.yaml": admission webhook "webhook.numaflow.numaproj.io" denied the request: pipeline has a Serving source "serving-in" and a reduce vertex "compute-sum". Reduce is not supported with Serving source
```


yhl25:
Add validation checks to not support reduce when serving source is configured?

Makefile
```diff
@@ -105,7 +105,7 @@ test-coverage:
 
 .PHONY: test-coverage-with-isb
 test-coverage-with-isb:
-	go test -v -timeout 7m -covermode=atomic -coverprofile=test/profile.cov -tags=isb_redis $(shell go list ./... | grep -v /vendor/ | grep -v /numaflow/test/ | grep -v /pkg/client/ | grep -v /pkg/proto/ | grep -v /hack/)
+	go test -timeout 7m -covermode=atomic -coverprofile=test/profile.cov -tags=isb_redis $(shell go list ./... | grep -v /vendor/ | grep -v /numaflow/test/ | grep -v /pkg/client/ | grep -v /pkg/proto/ | grep -v /hack/)
```
whynowy, line 108: What is the reason of disabling `-v`?

pkg/apis/numaflow/v1alpha1/vertex_types.go
```diff
@@ -228,7 +226,18 @@ func (v Vertex) GetPodSpec(req GetVertexPodSpecReq) (*corev1.PodSpec, error) {
 	envVars := []corev1.EnvVar{
 		{Name: EnvVertexObject, Value: encodedVertexSpec},
 	}
-	envVars = append(envVars, v.commonEnvs()...)
+
+	commonEnvVars := v.commonEnvs()
+	for _, vtx := range req.PipelineSpec.Vertices {
+		if vtx.IsASource() && vtx.Source.Serving != nil {
+			commonEnvVars = append(commonEnvVars,
+				corev1.EnvVar{Name: EnvCallbackEnabled, Value: "true"},
```
whynowy, line 234: It seems like we changed the logic from getting the information from annotations.

Are we expecting it's always `true`?

pkg/apis/numaflow/v1alpha1/vertex_types.go
```diff
@@ -343,13 +352,61 @@ func (v Vertex) GetPodSpec(req GetVertexPodSpecReq) (*corev1.PodSpec, error) {
 	}
 
 	if v.IsASource() && v.Spec.Source.Serving != nil {
-		servingContainer, err := v.getServingContainer(req)
+		servingSource := v.Spec.Source.Serving
```
whynowy, line 355: What is the reason of removing the abstracted function `getServingContainer()`?

pkg/apis/numaflow/v1alpha1/vertex_types.go
```diff
@@ -228,7 +226,18 @@ func (v Vertex) GetPodSpec(req GetVertexPodSpecReq) (*corev1.PodSpec, error) {
...
 		{Name: EnvVertexObject, Value: encodedVertexSpec},
 	}
-	envVars = append(envVars, v.commonEnvs()...)
+
+	commonEnvVars := v.commonEnvs()
+	for _, vtx := range req.PipelineSpec.Vertices {
+		if vtx.IsASource() && vtx.Source.Serving != nil {
+			commonEnvVars = append(commonEnvVars,
+				corev1.EnvVar{Name: EnvCallbackEnabled, Value: "true"},
+				corev1.EnvVar{Name: EnvCallbackURL, Value: KeyMetaCallbackURL},
```
whynowy, line 235: Is this correct? The value is a constant.

Makefile
BulkBeing: I found it difficult to identify the failed test in previous commit when Go unit test was failing, since it outputs too many logs: https://github.com/numaproj/numaflow/actions/runs/12943800158/job/36103700507?pr=2357
Looking it again, searching for `--- FAIL` works. Will revert this change.

pkg/apis/numaflow/v1alpha1/vertex_types.go
BulkBeing: This PR  removes the need for having an annotation to enable callbacks. Instead, if a serving source is used, the environment variable to enable callbacks will be set automatically. The `EnvCallbackEnabled` is checked to construct the callback http client, the callback only happens if the message contains the `X-Numaflow-Callback-Url` header.

pkg/apis/numaflow/v1alpha1/vertex_types.go
BulkBeing: Till now, if Serving source is specified, we used to have a second container for starting Serving http server. Since we made Serving as a [builtin](https://github.com/numaproj/numaflow/pull/2301) [source](https://github.com/numaproj/numaflow/pull/2311), this second container is not needed.
Instead, we set the env variables to enable callback if Serving is used as a source.

pkg/apis/numaflow/v1alpha1/vertex_types.go
BulkBeing: We were using this env variable to let the user specify an alternate service URL to which the callback can be sent, if callback sent to the Serving source pod fails. The value I used is incorrect. Also, we do not use it now (after the recent changes in Serving). I will remove the `EnvCallbackURL` env variable.

pkg/apis/numaflow/v1alpha1/vertex_types.go
```diff
@@ -343,13 +349,61 @@ func (v Vertex) GetPodSpec(req GetVertexPodSpecReq) (*corev1.PodSpec, error) {
...
+				},
+			},
+			corev1.EnvVar{
+				Name:  EnvServingStoreTTL,
+				Value: strconv.Itoa(int(servingSource.Store.GetTTL().Seconds())),
+			},
+		)
+
+		// if auth is configured, set the auth token in the environment
+		if servingSource.Auth != nil && servingSource.Auth.Token != nil {
```
whynowy, line 393: This should have been mounted as a volume to the container, there's no need to add env vars to it.

pkg/apis/numaflow/v1alpha1/vertex_types.go
```diff
@@ -343,13 +349,61 @@ func (v Vertex) GetPodSpec(req GetVertexPodSpecReq) (*corev1.PodSpec, error) {
 	}
 
 	if v.IsASource() && v.Spec.Source.Serving != nil {
-		servingContainer, err := v.getServingContainer(req)
+		servingSource := v.Spec.Source.Serving
+		servingSourceBytes, err := json.Marshal(servingSource)
```
whynowy, line 353: There's already an env var `EnvVertexObject`, no need to do this.

pkg/reconciler/validator/pipeline_validate.go
```diff
@@ -88,6 +88,21 @@ func ValidatePipeline(pl *dfv1.Pipeline) error {
...
+	var servingSource *dfv1.AbstractVertex
+	for _, srcVtx := range sources {
+		if srcVtx.Source.Serving != nil {
+			servingSource = &srcVtx
+			break
+		}
+	}
+	if servingSource != nil {
+		for _, v := range pl.Spec.Vertices {
+			if v.UDF != nil && v.UDF.GroupBy != nil {
```
kohlisid, line 100: We have a utility function IsReduceUDF()

whynowy (approved):
@kohlisid - please review.

pkg/apis/numaflow/v1alpha1/vertex_types.go
```diff
@@ -343,13 +349,33 @@ func (v Vertex) GetPodSpec(req GetVertexPodSpecReq) (*corev1.PodSpec, error) {
...
 		}
-		containers = append(containers, servingContainer)
-		// set the serving source stream name in the environment because the numa container will be reading from it
-		containers[0].Env = append(containers[0].Env, corev1.EnvVar{Name: EnvServingJetstreamStream, Value: req.ServingSourceStreamName})
+		encodedPipelineSpec := base64.StdEncoding.EncodeToString(pipelineSpecBytes)
+
+		containers[0].Env = append(
+			containers[0].Env,
+			// set the serving source stream name in the environment because the numa container will be reading from it
+			corev1.EnvVar{Name: EnvServingJetstreamStream, Value: req.ServingSourceStreamName},
```
yhl25, line 367: we don't need js stream anymore right?

rust/serving/src/config.rs
```diff
@@ -138,37 +146,62 @@ impl TryFrom<HashMap<String, String>> for Settings {
...
+        }
+        #[derive(Deserialize)]
+        struct Spec {
+            source: Source,
+        }
+
+        #[derive(Deserialize)]
+        struct VertexObject {
+            spec: Spec,
+        }
```
yhl25, line 172: Is it possible to use the generated models?

yhl25 (approved):

PR #2359, opened 2025-01-23
fix(test): use different approach to start jetstream service

whynowy:
From [2.10.24](https://github.com/bitnami/containers/pull/76503/files), there's no way to run `bitnami/nats` for JetStream in Github Action Services any more, have to use a different approach to do it.

The problem is there's no way to run `/nats-server -js` with Github Services - can not specify args, can specify endpoint but can not specify endpoint with args.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -76,7 +76,7 @@ jobs:
         ports:
           - 4222:4222
         options: >-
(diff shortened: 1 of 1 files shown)


whynowy:
Need to check why it fails starting from 2.10.24.

vigith (approved):

KeranYang (approved):

PR #2371, opened 2025-01-28
fix: get status of init containers in pods-info API call

adarsh0728:
In the call to `pods-info` API for UI, get status of sidecar containers (udsink, udsource etc) as well.

### DIFF
--- a/server/apis/v1/handler.go
+++ b/server/apis/v1/handler.go
@@ -1647,9 +1647,9 @@ func (h *handler) getContainerDetails(pod corev1.Pod) map[string]ContainerDetail
 	containerDetailsMap := make(map[string]ContainerDetails, totalContainers)
 
 	// Helper function to process container statuses
-	processContainerStatus := func(status corev1.ContainerStatus) {
-		// Skip the "init" container
-		if status.Name == "init" {
+	processContainerStatus := func(status corev1.ContainerStatus, isInitContainer bool) {
+		// Skip init containers without Always restart policy
+		if isInitContainer && h.isNotSidecarContainer(status.Name, pod) {
 			return
 		}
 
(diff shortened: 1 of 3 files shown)


yhl25 (approved):
Thanks for fixing this! LGTM!

server/apis/v1/handler.go
```diff
@@ -1643,8 +1643,16 @@ func (h *handler) getPodDetails(pod corev1.Pod) (PodDetails, error) {
...
 func (h *handler) getContainerDetails(pod corev1.Pod) map[string]ContainerDetails {
-	var containerDetailsMap = make(map[string]ContainerDetails)
-	for _, status := range pod.Status.ContainerStatuses {
+	totalContainers := len(pod.Spec.Containers) + len(pod.Spec.InitContainers)
+	containerDetailsMap := make(map[string]ContainerDetails, totalContainers)
+
+	// Helper function to process container statuses
+	processContainerStatus := func(status corev1.ContainerStatus) {
+		// Skip the "init" container
+		if status.Name == "init" {
```
whynowy, line 1652: does it intend to filter out the init containers (non-sidecar containers)?

whynowy (changes requested):

server/apis/v1/handler.go
whynowy: If this is the case, should not check the name, instead, check `restartPolicy`.

adarsh0728: on the UI side, we have been filtering out the "init" container (we do not show init container details). Hence, filtered out here as well.

adarsh0728: > restartPolicy

sure will make the changes

whynowy: > on the UI side, we have been filtering out the "init" container (we do not show init container details). Hence, filtered out here as well.

Then the logic on the UI side also needs to be updated.

Basically, the logic should be:

Anyone in the `initContainers`, `restartPolicy` is not `Always` can be excluded.

whynowy (approved):
Thank you!

Issue #2274, opened 2024-12-11
Numaflow should call Source::Pending only for replica-id 0 (mvtx)


vigith:
# Summary

We need to see whether every pod should make a pending call to the upstream source. We only read from pod-0, this will remove lot of pressure on the external source. I am worried whether we might in log run bring down the external source if the pending call is expensive. Also, for any pods other than replica-id 0, this information is meaningless.

# Use Cases

* Every deployment

### Tasks
- [x] only replica-id 0 pod in MonoVertex and Pipeline makes pending call
- [x] move pending logs to debug! for pod-0 and omit for others

vigith:
Will open a new one for Pipeline

PR #2373, opened 2025-01-28
feat: adaptive lookback for monovertex

kohlisid:
This build on top of our current rater to derive relevant information and calculate the required lookback window

This would encapsulate the lookback for two scenarios
1. Slow processing vertex
2. Slow data source - data arrives after long intervals

`lookbackSeconds - How many seconds to lookback for vertex average processing rate (tps) and pending messages calculation, defaults to 120. Rate and pending messages metrics are critical for autoscaling, you might need to tune this parameter a bit to see better results. For example, your data source only have 1 minute data input in every 5 minutes, and you don't want the vertices to be scaled down to 0. In this case, you need to increase lookbackSeconds to overlap 5 minutes, so that the calculated average rate and pending messages won't be 0 during the silent period, in order to prevent from scaling down to 0.`

https://numaflow.numaproj.io/user-guide/reference/autoscaling/#numaflow-autoscaling

**Follow up work**

- Move the pending calculation and the lag reader to the daemon server 
- https://github.com/numaproj/numaflow/issues/2274
- Test and confirm functionality with async data movement.

**Operational Flow:**
Data Entry: Pods report their processed message counts periodically, which are saved into a TimestampedCounts structure and pushed onto a queue.

**Lookback Adjustment Process:**

1. The system periodically triggers a routine to review recent data entries.
2. The routine calculates the maximum duration for which any pod's message count remains unchanged, using a function called CalculateMaxLookback.
3. Based on this calculation, another function, updateDynamicLookbackSecs, decides if the current lookback period needs adjustment to better align with observed pod activity.

When the value for a pod metric changes, new data is read
1. This occurs when the source has new data
2. The processing has completed, allowing more data to be read.


pkg/mvtxdaemon/server/service/rater/rater.go
```diff
@@ -63,8 +67,9 @@ type Rater struct {
 	podTracker *PodTracker
 	// timestampedPodCounts is a queue of timestamped counts for the MonoVertex
 	timestampedPodCounts *sharedqueue.OverflowQueue[*TimestampedCounts]
-	// userSpecifiedLookBackSeconds is the user-specified lookback seconds for that MonoVertex
-	userSpecifiedLookBackSeconds int64
+	// userSpecifiedLookBackSeconds the current lookback seconds for the monovertex
+	// this can be updated dynamically, defaults to user-specified value in the spec
+	userSpecifiedLookBackSeconds *atomic.Float64
```
whynowy, line 72: Use a different name since its meaning changed?

kohlisid: `currentLookBackWindow`/ `currentLookBackSecs` ?

whynowy: Simply `lookBackSeconds`?

pkg/mvtxdaemon/server/service/rater/rater.go
```diff
@@ -269,3 +283,67 @@ func sleep(ctx context.Context, duration time.Duration) {
...
+	}
+}
+
+// updateDynamicLookbackSecs continuously adjusts ths lookback duration based on the current
+// processing time of the MonoVertex system.
+func (r *Rater) updateDynamicLookbackSecs() {
+	counts := r.timestampedPodCounts.Items()
+	currentLookback := r.lookBackSeconds.Load()
+	vertexName := r.monoVertex.Name
+	if len(counts) <= 1 {
```
whynowy, line 308: nit: move it up.

pkg/mvtxdaemon/server/service/rater/helper.go
```diff
@@ -125,3 +125,77 @@ func calculatePodDelta(tc1, tc2 *TimestampedCounts) float64 {
...
+				if lastSeenData.count != count {
+					// Calculate the duration for which the count was unchanged.
+					duration := curTime - lastSeenData.seenTime
+					// Update maxDuration for the pod if this duration is the longest seen so far.
+					// TODO: Can check if average or EWMA works better than max
+					if currentMax, ok := maxDuration[key]; !ok || duration > currentMax {
+						maxDuration[key] = duration
+					}
+					// Update the last seen count and timestamp for the pod.
+					lastSeen[key] = struct {
```
whynowy, line 164: This is same as in the `else`, merge them?

pkg/mvtxdaemon/server/service/rater/helper.go
```diff
@@ -125,3 +125,77 @@ func calculatePodDelta(tc1, tc2 *TimestampedCounts) float64 {
...
+				}{count, curTime}
+			}
+		}
+	}
+
+	// Fetch the last timestamp used in the analysis to check unmodified runs.
+	endVals := counts[endIndex].PodCountSnapshot()
+	lastTime := counts[endIndex].PodTimestamp()
+
+	// Check for pods that did not change at all during the iteration,
```
whynowy, line 183: What will happen if some pods never get traffic?

kohlisid: That would make the lookback seconds increase until the limit is hit or the pods starts getting data and the value is adjust then.

whynowy: Would it be better if we ignore those ones having `duration > MaxLookback`?

pkg/mvtxdaemon/server/service/rater/helper_test.go
```diff
@@ -286,3 +287,149 @@ func TestCalculateRate(t *testing.T) {
...
+				newTimestampedCounts(100, map[string]float64{"pod1": 300, "pod2": 400}),
+				newTimestampedCounts(130, map[string]float64{"pod1": 500, "pod2": 400}),
+				newTimestampedCounts(160, map[string]float64{"pod1": 500, "pod2": 600}),
+			},
+			startIndex:  0,
+			endIndex:    2,
+			expectedMax: 60, // for pod2 between 1,3
+		},
+		{
+			name: "Gaps in timestamps",
```
KeranYang, line 373: we don't need this one, do we? from a calculation perspective, 1000 is no difference with 10

pkg/mvtxdaemon/server/service/rater/helper_test.go
```diff
@@ -286,3 +287,149 @@ func TestCalculateRate(t *testing.T) {
...
+				newTimestampedCounts(100, map[string]float64{"pod1": 100}),
+				newTimestampedCounts(1000, map[string]float64{"pod1": 100}), // Large gap with no change
+				newTimestampedCounts(1100, map[string]float64{"pod1": 200}),
+			},
+			startIndex:  0,
+			endIndex:    2,
+			expectedMax: 1000,
+		},
+		{
+			name: "Large number of pods",
```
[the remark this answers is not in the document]
KeranYang: same for `Large number of pods - No change`

pkg/mvtxdaemon/server/service/rater/rater.go
```diff
@@ -19,21 +19,25 @@ package rater
...
 
 	"github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
+	"github.com/numaproj/numaflow/pkg/metrics"
 	"github.com/numaproj/numaflow/pkg/shared/logging"
 	sharedqueue "github.com/numaproj/numaflow/pkg/shared/queue"
 )
 
 const CountWindow = time.Second * 10
 const monoVtxReadMetricName = "monovtx_read_total"
+const MaxLookback = time.Minute * 10
```
KeranYang, line 40: magic number 10 minutes, can we comment on why we chose this duration?

pkg/mvtxdaemon/server/service/rater/rater.go
```diff
@@ -95,20 +100,22 @@ func NewRater(ctx context.Context, mv *v1alpha1.MonoVertex, opts ...Option) *Rat
 			},
 			Timeout: time.Second * 1,
 		},
-		log:     logging.FromContext(ctx).Named("Rater"),
-		options: defaultOptions(),
+		log:             logging.FromContext(ctx).Named("Rater"),
+		options:         defaultOptions(),
+		lookBackSeconds: atomic.NewFloat64(float64(mv.Spec.Scale.GetLookbackSeconds())),
```
KeranYang, line 105: do we need nil check for GetLookbackSeconds?

pkg/mvtxdaemon/server/service/rater/rater.go
```diff
@@ -203,7 +210,10 @@ func (r *Rater) GetRates() map[string]*wrapperspb.DoubleValue {
 }
 
 func (r *Rater) buildLookbackSecondsMap() map[string]int64 {
-	lookbackSecondsMap := map[string]int64{"default": r.userSpecifiedLookBackSeconds}
+	// as the lookback value can be changing dynamically,
```
KeranYang, line 213:
```suggestion
	// as the default lookback value can be changing dynamically,
```

pkg/mvtxdaemon/server/service/rater/rater.go
```diff
@@ -269,3 +283,65 @@ func sleep(ctx context.Context, duration time.Duration) {
...
+		// no action required here
+		return
+	}
+	maxProcessingTime := CalculateMaxLookback(counts, startIndex, endIndex)
+	// round up to the nearest minute, also ensure that while going up and down we have the consistent value for
+	// a given processingTimeSeconds, then convert back to seconds
+	roundedMaxLookback := 60.0 * (math.Ceil(float64(maxProcessingTime) / 60.0))
+	// Based on the value received we can have two cases
+	// 1. Step up case (value is > than current):
+	// 	  Do not allow the value to be increased more than the MaxLookback allowed (10mins)
```
KeranYang, line 332: so user shouldn't declare a value that's higher than 10mins. let's make sure we document it.

pkg/mvtxdaemon/server/service/rater/helper.go
kohlisid: the lookback will be gated at MaxLookback it the previous caller. 
You mean that we add that check here as well?
Or do not include pods having `duration > MaxLookback` in the calculation itself?

pkg/mvtxdaemon/server/service/rater/helper.go
```diff
@@ -125,3 +125,77 @@ func calculatePodDelta(tc1, tc2 *TimestampedCounts) float64 {
...
+// 2. Slow data source - data arrives after long intervals
+func CalculateMaxLookback(counts []*TimestampedCounts, startIndex, endIndex int) int64 {
+	// Map to keep track of the last seen count and timestamp of each pod.
+	lastSeen := make(map[string]struct {
+		count    float64
+		seenTime int64
+	})
+
+	// Map to store the maximum duration for which the value of any pod was unchanged.
+	maxDuration := make(map[string]int64)
```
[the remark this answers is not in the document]
kohlisid: `maxUnchangedDuration` instead of idling?

pkg/mvtxdaemon/server/service/rater/rater.go
kohlisid: We have one inside the function 
```
func (s Scale) GetLookbackSeconds() int {
	if s.LookbackSeconds != nil {
		return int(*s.LookbackSeconds)
	}
	return DefaultLookbackSeconds
}
```

pkg/mvtxdaemon/server/service/rater/helper.go
whynowy: > the lookback will be gated at MaxLookback it the previous caller. You mean that we add that check here as well? Or do not include pods having `duration > MaxLookback` in the calculation itself?

What I meant was `do not include pods having duration > MaxLookback in the calculation`, does that make sense?

pkg/mvtxdaemon/server/service/rater/helper.go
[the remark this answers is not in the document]
kohlisid: I have kept a single function for easier understanding of logic right now. Can change if you feel required.

pkg/mvtxdaemon/server/service/rater/rater.go
kohlisid: Have added a condition where the spec value will get gated to 10mins itself.

pkg/mvtxdaemon/server/service/rater/helper.go
kohlisid: Discussed offline!

docs/user-guide/reference/autoscaling.md
```diff
@@ -49,6 +49,8 @@ spec:
   don't want the vertices to be scaled down to `0`. In this case, you need to increase `lookbackSeconds` to overlap
   5 minutes, so that the calculated average rate and pending messages won't be `0` during the silent period, in order to prevent from
   scaling down to 0.
+  The max value allowed is  `10mins ~ 600s`.
```
whynowy, line 52: Use a different format? Might confuse users that `Nmins` or `Ns` can be used.

whynowy (approved):
@KeranYang - please review.

pkg/mvtxdaemon/server/service/rater/helper_test.go
```diff
@@ -286,3 +287,141 @@ func TestCalculateRate(t *testing.T) {
...
+				newTimestampedCounts(0, map[string]float64{"pod1": 60}),
+				newTimestampedCounts(60, map[string]float64{"pod1": 60}),
+				newTimestampedCounts(120, map[string]float64{"pod1": 70}),
+				newTimestampedCounts(180, map[string]float64{"pod1": 0}),
+				newTimestampedCounts(240, map[string]float64{"pod1": 25}),
+				newTimestampedCounts(300, map[string]float64{"pod1": 25}),
+			},
+			startIndex:  0,
+			endIndex:    5,
+			expectedMax: 120, // here idx 0,3 should be used, after going to zero it resets
```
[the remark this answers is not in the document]
kohlisid: Change is from 60 -> 70 for the timestamps (0 - 120)
For timestamp(180-300) value changes from 0->25->25
So differences will be 240-180=60 and then 300-240 = 60

pkg/mvtxdaemon/server/service/rater/helper_test.go
```diff
@@ -286,3 +287,141 @@ func TestCalculateRate(t *testing.T) {
...
+		},
+		{
```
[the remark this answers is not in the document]
kohlisid: this one just checks if one pod and no changes, then it should be 0.
The one below is a super set, added them for more understanding and checks

pkg/mvtxdaemon/server/service/rater/helper_test.go
```diff
@@ -286,3 +287,141 @@ func TestCalculateRate(t *testing.T) {
...
+				newTimestampedCounts(100, map[string]float64{"pod1": 500, "pod2": 400}),
+				newTimestampedCounts(130, map[string]float64{"pod1": 600, "pod2": 400}),
+				newTimestampedCounts(160, map[string]float64{"pod1": 600, "pod2": 600}),
+			},
+			startIndex:  0,
+			endIndex:    2,
+			expectedMax: 60,
+		},
+		{
+			name: "Pod goes to zero",
```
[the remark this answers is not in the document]
kohlisid: Have updated the logic.
Max should be 240-120 (where value is 25) = 120

KeranYang (approved):

Issue #2350, opened 2025-01-22
Monovertex changes to support serving as a builtin source


yhl25:
Some changes to the user facing APIs also will be made in this PR:
- `POST /v1/process/sync` - Receive payload and return result synchronously. If client gets disconnects (network error or timeouts), the results can still be retrieved using `/v1/process/serve` endpoint
- `POST /v1/process/async` - Receive payload and process asynchronously.
- `GET /v1/process/serve` - Receive processing status and/or results. The request may be submitted using /sync or /async endpoint. 
- `GET /v1/process/message` - Retrive message graph for a request id

When submitting the request (either `/sync` or `/async`), the request id must not already exist in the store (for the duration of the store TTL). This is needed to avoid accidental overwrite of the results. A query parameter can be provided to allow overwriting.

The callback info will be saved in memory. It will be saved to Redis when the pod is shutting down or when all callbacks for a request id is received.

PR #2382, opened 2025-02-06
feat: Serving as a builtin source with Monovertex

BulkBeing:
Fixes https://github.com/numaproj/numaflow/issues/2350

APIs:
----
- `POST /v1/process/sync`
- `POST /v1/process/async`
- `GET /v1/process/fetch?id=req_id`
- `GET /v1/process/status?id=req_id`

The responsibilty of generating a unique request id is moved to the Store implementation if the user doesn't provide one with `X-Numaflow-Id` header. If one is provided and if it already exist in the Store, we return 409 conflict status.

Since request id is generated by Store (or validated if user provides it), we use different tid for tracing (logs).

A request (`/sync` and `/async`) is registered in the store as the first step. With Redis store implementation, this adds an entry:
```
request:${request_id}:status  = 'pending'
``` 
When the processing is completed and all callbacks are received, the removal of request id from the tracker implementation causes the Redis status key to change value to 'completed'.

The callbacks will be stored in Redis in a list with the key `request:${request}:callbacks` and the result will be saved using the key `request:${request}:results`.

Example: The entries in Redis follows below structure:
```
request:01951c1a-418f-74f3-a2b2-ee480d877f80:status  = "completed"
request:01951c1a-418f-74f3-a2b2-ee480d877f80:results = {"hello":"world"}

request:01951c1a-418f-74f3-a2b2-ee480d877f80:callbacks = [
    {"id":"01951c1a-418f-74f3-a2b2-ee480d877f80","vertex":"serve-sink","cb_time":1739933238792,"from_vertex":"serving-in","responses":[{"tags":null}]}

    {"id":"01951c1a-418f-74f3-a2b2-ee480d877f80","vertex":"serving-in","cb_time":1739933237779,"from_vertex":"serving-in","responses":[{"tags":null}]}
]
```

Monovertex still needs to make a callback (localhost) when Sink processing is successful. I attempted to remove this by storing a handle to the callback state within the callback handler implementation. However it resulted in generics everywhere (`start_source_forwarder`, `create_components`, etc.) since callback state is generic over store.

Currently the callbacks are stored when serving pod receives a callback from each vertex for each message. Doing this only when pod is shutting down or when all callback is received can reduce the Redis API calls. This can be implemented in the future. Currently I'm prioritizing things that will make the main branch in a working/usable state.

### DIFF
--- a/rust/serving/src/app.rs
+++ b/rust/serving/src/app.rs
@@ -19,7 +19,7 @@ use tracing::{info, info_span, Span};
 use uuid::Uuid;
 
 use self::{
-    callback::callback_handler, direct_proxy::direct_proxy, jetstream_proxy::serving_public_routes,
+    callback::callback_handler, direct_proxy::direct_proxy, jetstream_proxy::jetstream_proxy,
     message_path::get_message_path,
 };
 use crate::app::callback::store::Store;
@@ -251,7 +251,7 @@ async fn routes<T: Clone + Send + Sync + Store + 'static>(
     app_state: AppState<T>,
 ) -> crate::Result<Router> {
     let state = app_state.callback_state.clone();
-    let jetstream_proxy = serving_public_routes(app_state.clone()).await?;
+    let jetstream_proxy = jetstream_proxy(app_state.clone()).await?;
     let callback_router = callback_handler(
         app_state.settings.tid_header.clone(),
         app_state.callback_state.clone(),
--- a/rust/serving/src/app.rs
+++ b/rust/serving/src/app.rs
@@ -76,7 +76,6 @@ pub(crate) async fn router_with_auth<T>(app: AppState<T>) -> crate::Result<Route
 where
     T: Clone + Send + Sync + Store + 'static,
 {
-    let tid_header = app.settings.tid_header.clone();
     let layers = ServiceBuilder::new()
         // Add tracing to all requests
         .layer(
@@ -87,12 +86,8 @@ where
                         // We don't need request ID for these endpoints
                         return info_span!("request", method=?req.method(), path=req_path);
                     }
-                    let tid = req
-                        .headers()
-                        .get(&tid_header)
-                        .and_then(|v| v.to_str().ok())
-                        .map(|v| v.to_string())
-                        .unwrap_or_else(|| Uuid::new_v4().to_string());
+                    let uuid = Uuid::now_v7().to_string();
+                    let tid = format!("{}{}{}", &uuid[..8], &uuid[10..13], &uuid[uuid.len() - 5..]);
 
                     let matched_path = req
                         .extensions()
--- a/rust/serving/src/app.rs
+++ b/rust/serving/src/app.rs
@@ -86,6 +86,10 @@ where
                         // We don't need request ID for these endpoints
                         return info_span!("request", method=?req.method(), path=req_path);
                     }
+
+                    // Generate a tid with good enough randomness and not too long
+                    // Example of a UUID v7: 01951b72-d0f4-711e-baba-4efe03d9cb76
+                    // We use the characters representing timestamp in milliseconds (without '-'), and last 5 characters for randomness.
                     let uuid = Uuid::now_v7().to_string();
                     let tid = format!("{}{}{}", &uuid[..8], &uuid[10..13], &uuid[uuid.len() - 5..]);
 
--- a/rust/serving/src/app.rs
+++ b/rust/serving/src/app.rs
@@ -19,7 +19,7 @@ use tracing::{info, info_span, Span};
 use uuid::Uuid;
 
 use self::{
-    callback::callback_handler, direct_proxy::direct_proxy, jetstream_proxy::jetstream_proxy,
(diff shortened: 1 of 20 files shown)


rust/serving/src/app/callback/store.rs
```diff
@@ -17,13 +19,51 @@ pub(crate) enum PayloadToSave {
     },
 }
 
+#[derive(Debug, PartialEq)]
+pub(crate) enum PipelineResult {
```
yhl25, line 23: Consider a generic name that works for mvtx as well?

yhl25: Add docs..

rust/serving/src/app/callback/store/memstore.rs
```diff
@@ -72,18 +80,24 @@ impl super::Store for InMemoryStore {
...
+            ))),
         }
     }
 
     /// Retrieves data for a given id from the `HashMap`.
     /// Each piece of data is deserialized from bytes into a `String`.
-    async fn retrieve_datum(&mut self, id: &str) -> Result<Vec<Vec<u8>>, Error> {
-        let id = format!("{}_{}", id, SAVED);
+    async fn retrieve_datum(&mut self, id: &str) -> StoreResult<PipelineResult> {
+        let id = format!("{}_{}", id, "saved");
```
yhl25, line 93: Keep constant ?

rust/serving/src/app/callback/store/redisstore.rs
```diff
@@ -109,8 +109,88 @@ async fn handle_write_requests(
...
+                    }
+                    return Ok(id);
+                }
+                Err(StoreError::StoreWrite(
+                    "Could not generate a unique request id".to_string(),
+                ))
+            }
+        }
+    }
+    async fn done(&mut self, id: String) -> StoreResult<()> {
```
yhl25, line 171: Add docs

vigith:
just returning a status is not enough, we will have to return body. e.g., 409 status code is not definite enough for clients.

rust/serving/src/app.rs
```diff
@@ -87,12 +86,12 @@ where
...
-                        .headers()
-                        .get(&tid_header)
-                        .and_then(|v| v.to_str().ok())
-                        .map(|v| v.to_string())
-                        .unwrap_or_else(|| Uuid::new_v4().to_string());
+
+                    // Generate a tid with good enough randomness and not too long
+                    // Example of a UUID v7: 01951b72-d0f4-711e-baba-4efe03d9cb76
+                    // We use the characters representing timestamp in milliseconds (without '-'), and last 5 characters for randomness.
+                    let uuid = Uuid::now_v7().to_string();
```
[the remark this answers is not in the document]
vigith: what is the point of this TID? just to log?

rust/serving/src/app/callback/state.rs
```diff
@@ -43,24 +45,34 @@ where
 
     /// register a new connection
     /// The oneshot receiver will be notified when all callbacks for this connection is received from the numaflow pipeline
-    pub(crate) fn register(&mut self, id: String) -> oneshot::Receiver<Result<String, Error>> {
+    pub(crate) async fn register(
+        &mut self,
+        id: Option<String>,
```
vigith, line 51: let's make sure that the ID is no longer optional.

rust/serving/src/app/callback/store.rs
```diff
@@ -17,13 +19,52 @@ pub(crate) enum PayloadToSave {
...
+pub(crate) type Result<T> = std::result::Result<T, Error>;
+
 /// Store trait to store the callback information.
 #[trait_variant::make(Store: Send)]
 #[allow(dead_code)]
 pub(crate) trait LocalStore {
-    async fn save(&mut self, messages: Vec<PayloadToSave>) -> crate::Result<()>;
+    /// Register a request id in the store. If user provides a request id, the same should be returned
+    /// if it doesn't already exist in the store. An error should be returned if the user-specified request id
+    /// already exists in the store. If the `id` is `None`, the store should generate a new unique request id.
```
vigith, line 61: let's make sure ID is created at the top (when the request is received)

rust/serving/src/app/callback/store/redisstore.rs
```diff
@@ -109,8 +109,90 @@ async fn handle_write_requests(
...
+                            ))
+                        })?;
+
+                if !status {
+                    // The user specified request id already exists
+                    return Err(StoreError::DuplicateRequest(id));
+                }
+                Ok(id)
+            }
+            None => {
```
vigith, line 138: if ID is generated from above, we do not have to do this. This will be an over complication once we have other stores (IPS, DDB, etc.)

rust/serving/src/app.rs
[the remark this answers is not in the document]
BulkBeing: Yes, all log lines corresponding to a request will have this tid. Ideally the same should be used by the callback requests for that request too, but currently it's not propagated.

yhl25 (approved):

vigith (approved):
let's do this in follow up PR

Issue #2383, opened 2025-02-06
missing metrics for MonoVertex


vigith:
Dropped message metrics counter is missing for Transformer for MonoVertex

adarsh0728:
can i pick this up?

vigith:
> can i pick this up?

absolutely :)

vigith:
i have changed the title to include all missing metrics.

@yhl25 i think some metrics are missing in pipeline too, e.g., dropped. If yes, let's create a new issue and track metrics missing for pipeline too.

yhl25:
We will add all the missing metrics for mvtx in 1.5, will create a separate issue to tackle missing metrics in pipeline that will be done in 1.6.

PR #2389, opened 2025-02-10
fix: add missing metrics for transformer for mvtx

adarsh0728:
fix: #2383


adarsh0728:
new PR https://github.com/numaproj/numaflow/pull/2542 for the issue #2383

PR #2394, opened 2025-02-12
fix: filter out deamon pod and unncessary containers from metrics output

adarsh0728:


adarsh0728:
Screenshot 2025-02-12 at 12 21 47 PM
Screenshot 2025-02-12 at 12 22 35 PM
Screenshot 2025-02-12 at 12 22 56 PM
Screenshot 2025-02-12 at 12 23 08 PM

vigith:
@adarsh0728 what is the condition to filter out pods?

adarsh0728:
> @adarsh0728 what is the condition to filter out pods and containers? keep in mind that in future we will support builtins like kafka, sqs, etc., we cannot filter those out.

For pods: we call [pods API](https://github.com/numaproj/numaflow/blob/c88881849c8f7c68107759022b453d0c60797d2a/server/routes/routes.go#L159) and filter out only daemon.
For containers of a pod:  `pod.spec.containers` + `pod.spec.initContainers`(those with restartPolicy as Always)

we are not filtering out builtins with this

ui/src/components/common/MetricsModalWrapper/partials/MetricsModal/index.tsx
```diff
@@ -31,6 +31,7 @@ interface MetricsModalProps {
   vertexId: string;
   type: string;
   presets?: any;
+  pod?: any;
```
veds-g, line 34: you can add type `Pod` like above

ui/src/components/pages/Pipeline/partials/Graph/partials/NodeInfo/partials/Pods/partials/PodDetails/partials/Metrics/index.tsx
```diff
@@ -22,25 +22,26 @@ import {
 } from "../../../../../../../../../../../../common/SlidingSidebar/partials/VertexDetails";
 
 import "./style.css";
+import { Pod } from "../../../../../../../../../../../../../types/declarations/pods";
```
veds-g, line 25: please move this import after line 22

veds-g:
for filtering the containers, we do not have the pipeline or monovertex context is the metrics response. So in case of multiple containers with the same name, how do we ensure that we showing data for the particular container belonging to that pipeline vertex or monovertex?

ui/src/components/pages/Pipeline/partials/Graph/partials/NodeInfo/partials/Pods/partials/PodDetails/partials/Metrics/partials/LineChart/index.tsx
```diff
@@ -373,9 +374,21 @@ const LineChartComponent = ({
 
     const labels: string[] = [];
     const transformedData: Record<string, any>[] = [];
+    let filteredChartData = chartData;
     const label = groupByLabel(metricsReq?.dimension, metricsReq?.display_name);
 
-    chartData?.forEach((item) => {
+    if (JSON.stringify(label) === JSON.stringify(["container"])){
```
veds-g, line 380: lets have a stricter comparison like this - `if (Array.isArray(label) && label.length === 1 && label[0] === "container")`
stringify might be an overkill so this simple conversion.

veds-g (approved):
LGTM

PR #2398, opened 2025-02-16
fix(controller): multiple issues related to min/max of a vertex

whynowy:
1. Remove the validation of `max` can not be 0 for a pipeline vertex;
2. When resuming a pipeline, always set the the replica to `max(scale.min, 1)` no matter if it's a scalable vertex or not (e.g. HTTP), unless scaling is `disabled`;
3. Keep the original replicas as much as possible when updating, unless it's out of the range of the `[min, max]`;
4. Do not allow Vertex replicas to be out of the range of `[min, max]` when scaling is enabled - make the behavior consistent with MonoVertex.


vigith (approved):

vigith:
should we backport this to 1.4?

whynowy:
> should we backport this to 1.4?

Yes

PR #2401, opened 2025-02-18
fix: sorting metrics chart tooltip values

veds-g:
- Remove hyperlink for contextual navigation when metrics is not in config map
- Remove `default` label for pending metrics
- Sorted tooltip for metrics charts in descending order based on value
- Fixed tooltip spacing

image

image


adarsh0728 (approved):
LGTM!

vigith (approved):
Thank you 😊

PR #2409, opened 2025-02-20
feat: Redis ISB as store

BulkBeing:
Controller changes to use Redis ISB as store for Serving: #2407


vigith:
let's revisit this later based on the requirements?

PR #2416, opened 2025-02-21
chore: Controller validation for Serving source request timeout setting

BulkBeing:
```
➜  kubectl apply -f manifests/simple-pipeline.yaml
Error from server (BadRequest): error when creating "manifests/simple-pipeline.yaml": admission webhook "webhook.numaflow.numaproj.io" denied the request: pipeline.spec.lifecycle.deletionGracePeriodSeconds (value=30) must be at-least 30 seconds higher than the requestTimeoutSecs (value=25) configured for the Serving source (vertex="serving-in")
```

### DIFF
--- a/rust/numaflow-core/src/pipeline.rs
+++ b/rust/numaflow-core/src/pipeline.rs
@@ -7,6 +7,7 @@ use serving::callback::CallbackHandler;
 use tokio_util::sync::CancellationToken;
 use tracing::info;
 
+use crate::config::components::source::SourceType;
 use crate::config::is_mono_vertex;
 use crate::config::pipeline;
 use crate::config::pipeline::isb::Stream;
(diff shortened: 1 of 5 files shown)


rust/numaflow-core/src/pipeline.rs
```diff
@@ -41,6 +42,24 @@ pub(crate) async fn start_forwarder(
...
+                let serving_cln_token = CancellationToken::new();
+                tokio::spawn({
+                    let req_timeout = serving_config.drain_timeout_secs + 30;
+                    let cln_token = cln_token.clone();
+                    let serving_cln_token = serving_cln_token.clone();
+                    async move {
+                        cln_token.cancelled().await;
+                        tokio::time::sleep(Duration::from_secs(req_timeout)).await;
+                        serving_cln_token.cancel();
+                    }
```
vigith, line 58: we shouldn't do a blind sleep, we can return early if there are no requests-in-flight.

BulkBeing: The [handle](https://docs.rs/axum-server/latest/axum_server/struct.Handle.html) provides a method to know current connection count. We may have to find a way to pass this token to `serving` crate and cancel it from there.

BulkBeing:
Closing since the base branch has remove `/callback` endpoints.

Issue #2415, opened 2025-02-21
Regression (Numaflow 1.4.3 RC2) - pipeline stuck in "Deleting"


juliev0:
**Describe the bug**

The Numaplane CI includes one test of the "delete/recreate" field for isbsvc. What this does is first deletes the pipelines before it tries to delete the isbsvc.

There seems to be a regression in which most of the time when the "no strategy" test runs, it never deletes the pipeline. The Pipeline is just stuck in "Deleting" state. I ran the test locally and was able to confirm that the Pipeline was never deleted but it was in "Deleting" phase. 

Here's a reference to one failure in the CI: https://github.com/numaproj/numaplane/actions/runs/13450260803/job/37583362365?pr=590

I'll see if I can come up with a way for this to be reproducible. I wasn't easily able to reproduce it manually with a quick test.

This is the log file for Numaflow at the point of deletion:

```
{"level":"info","ts":"2025-02-21T04:59:36.300014573Z","logger":"numaflow.controller-manager","caller":"pipeline/controller.go:113","msg":"Deleting pipeline","namespace":"numaplane-system","pipeline":"test-pipeline-rollout-0"}
{"level":"info","ts":"2025-02-21T04:59:36.30918554Z","logger":"numaflow.controller-manager","caller":"pipeline/controller.go:966","msg":"Scaled vertex","namespace":"numaplane-system","pipeline":"test-pipeline-rollout-0","from":1,"to":0,"vertex":"test-pipeline-rollout-0-in"}
{"level":"info","ts":"2025-02-21T04:59:36.309249508Z","logger":"numaflow.controller-manager","caller":"pipeline/controller.go:124","msg":"Pipeline deletion is waiting to finish the unconsumed messages","namespace":"numaplane-system","pipeline":"test-pipeline-rollout-0"}
{"level":"info","ts":"2025-02-21T04:59:36.320604274Z","logger":"numaflow.controller-manager","caller":"pipeline/controller.go:113","msg":"Deleting pipeline","namespace":"numaplane-system","pipeline":"test-pipeline-rollout-0"}
{"level":"info","ts":"2025-02-21T04:59:36.325797506Z","logger":"numaflow.controller-manager","caller":"vertex/controller.go:409","msg":"Deleted Vertx pod \"test-pipeline-rollout-0-in-0-s7dr8\"","namespace":"numaplane-system","vertex":"test-pipeline-rollout-0-in","pipeline":"test-pipeline-rollout-0"}
[...]

Issue #2414, opened 2025-02-21
Introducing `lifecycle` to Vertex object spec


whynowy:
# Summary

Currently, for a vertex, there are several fields determining the replicas, which are `spec.replicas`, `spec.scale.min` and `spec.scale.max`. Together with these fields, the vertex controller also needs to look into the corresponding Pipeline status (e.g. `lifecycle.desiredPhase`) to make final decision. It makes it even more complicated when a pipeline is being paused - in that case, some of the vertices like sources might need to go to 0 replica first, while other vertices might be still running at that moment.

The complexity above makes the vetex controller implementation kind of messy, and it's very easy to generate regressions (e.g. through #2412, #2398). To solve the problem, we will need to propagate the `lifecycle.desiredPhase` into vertex spec, it will be done and controlled by the pipeline controller, and then the vertex controller does not need to get coupled with pipeline lifecycle.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

Issue #2413, opened 2025-02-21
Regression (numaflow 1.4.3 RC2): Pipeline never marked "drainedOnPause: true"


juliev0:
**Describe the bug**

Numaplane e2e is failing at the part in which we explicitly set `desiredPhase: Paused` in the PipelineRollout, which sets it in the `Pipeline`. 

Pipeline gets to "Paused" but "drainedOnPause" is not set.

I was able to reproduce this. Create this Pipeline and then set its `lifecycle.desiredPhase=Paused`:

```
apiVersion: numaflow.numaproj.io/v1alpha1
kind: Pipeline
metadata:
  labels:
    numaplane.numaproj.io/isbsvc-child-name: test-isbservice-rollout-0
    numaplane.numaproj.io/isbsvc-name: test-isbservice-rollout
    numaplane.numaproj.io/parent-rollout-name: test-pipeline-rollout
    numaplane.numaproj.io/upgrade-state: promoted
  name: test-pipeline-reproduce-0
  namespace: numaplane-system
spec:
  edges:
  - conditions: null
    from: in
    to: cat
  - conditions: null
    from: cat
    to: out
  interStepBufferServiceName: test-isbservice-rollout-0
  lifecycle:
    deletionGracePeriodSeconds: 10
  vertices:
  - name: in
    scale:
      max: 1
      min: 1
      zeroReplicaSleepSeconds: 15
    source:
      generator:
        duration: 1s
        rpu: 5
    updateStrategy: {}
  - name: cat
    scale:
      max: 1
      min: 1
      zeroReplicaSleepSeconds: 15
    udf:
      builtin:
        name: cat
      container: null
      groupBy: null
    updateStrategy: {}
  - name: out
    scale:
      max: 1
      min: 1
      zeroReplicaSleepSeconds: 15
    sink:
      log: {}
      retryStrategy: {}
    updateStrategy: {}
  watermark: {}

``` 

Reference to where it failed in Numaplane CI: https://github.com/numaproj/numaplane/actions/runs/13449066431/job/37580191564?pr=590

[numaflow-controller-9dd744855-2qzcd-controller-manager.log.zip](https://github.com/user-attachments/files/18900628/numaflow-controller-9dd744855-2qzcd-controller-manager.log.zip)

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

---

**Message from the maintainers**:

[...]

PR #2419, opened 2025-02-23
refactor: introducing lifecycle to Vertex spec

whynowy:
Introduced `spec.lifecycle.desiredPhase` to Vertex object. When pausing or resuming a pipeline, instead of patching the vertex replicas, patch its desiredPhase. With this change, 

1. `scale.min` and `scale.max` are used to define the range of scaling;
2. `replicas` is used by autoscaling, but it needs to be in the range of `min` and `max`;
3. `desiredPhase` is used for pausing and resuming, and when it happens, `replicas` does not change, in this case, when resuming a pipeline, it goes directly to its original replicas.
4. Vertex `phase` field also shows `Paused` when pipeline is paused.

Screenshot 2025-02-22 at 4 57 17 PM

Closes: #2414
Closes: #2413
Closes: #2415


vigith (approved):

api/json-schema/schema.json
```diff
@@ -22491,6 +22500,10 @@
         "interStepBufferServiceName": {
           "type": "string"
         },
+        "lifecycle": {
+          "$ref": "#/definitions/io.numaproj.numaflow.v1alpha1.VertexLifecycle",
+          "description": "Lifecycle defines the Lifecycle properties of a ertex"
```
adarsh0728, line 22505: typo: *vertex

Issue #2446, opened 2025-03-06
MonoVertex and Pipeline Status should indicate "PodsHealthy=false" if container restarted recently


juliev0:
**Describe the bug**
MonoVertex Status isn't indicating that the PodsHealthy Condition is false even though the Pod Container is continually restarting. 

If the container restarted in the last few minutes due to error, that should also warrant that Condition to be false. 

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

>For quick help and support, join our [slack](https://numaproj.slack.com/join/shared_invite/zt-19svuv47m-YKHhsQ~~KK9mBv1E7pNzfg#/shared-invite/email) channel.

PR #2447, opened 2025-03-06
chore: mark unhealthy if there is recent container restart

whynowy:
Explain what this PR does.

If there's container restart in the last 2 mins, mark it unhealthy.

Closes: #2446

### DIFF
--- a/pkg/reconciler/util.go
+++ b/pkg/reconciler/util.go
@@ -64,7 +64,7 @@ func isPodHealthy(pod *corev1.Pod) (healthy bool, reason string) {
 			}
 		}
 	}
-	// Container restart happened 2 in the last 2 mins
+	// Container restart happened in the last 2 mins
 	if !lastRestartTime.IsZero() && lastRestartTime.Add(2*time.Minute).After(time.Now()) {
 		return false, "RecentRestart"
 	}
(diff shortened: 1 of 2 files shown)


whynowy:
Be aware this will mark the pipline/mvtx unhealthy for 2 mins if there's any container restart. Maybe we should we should make it shorter.

pkg/reconciler/util.go
```diff
@@ -49,13 +50,23 @@ func CheckPodsStatus(pods *corev1.PodList) (healthy bool, reason string, message
...
 		}
+		if x := c.LastTerminationState.Terminated; x != nil && !x.FinishedAt.Time.IsZero() {
+			if lastRestartTime.IsZero() || x.FinishedAt.Time.After(lastRestartTime) {
+				lastRestartTime = x.FinishedAt.Time
+			}
+		}
+	}
+	// Container restart happened in the last 2 mins
+	if !lastRestartTime.IsZero() && lastRestartTime.Add(2*time.Minute).After(time.Now()) {
+		return false, "RecentRestart"
```
vigith, line 69: what is the app start up time is slow?

vigith (approved):

PR #2449, opened 2025-03-10
feat: error endpoints for numaflow server

### LINKED ISSUE #2455: Expose APIs for fetching error state per pod from the UI server

veds-g:
Fixes #2455


pkg/apis/proto/daemon/daemon.proto
```diff
@@ -109,6 +109,24 @@ message GetPipelineWatermarksRequest {
...
 
+message GetVertexErrorsRequest {
+  string pipeline = 1;
+  string vertex = 2;
+  string replica = 3;
+}
+
+message VertexError {
+  string container = 1;
+  string timestamp = 2;
```
whynowy, line 120: Why use `string`?

pkg/daemon/client/restful_daemon_client.go
```diff
@@ -160,3 +160,16 @@ func (rc *restfulDaemonClient) GetPipelineStatus(ctx context.Context, pipeline s
 		return res.Status, nil
 	}
 }
+
+func (rc *restfulDaemonClient) GetVertexErrors(ctx context.Context, pipeline, vertex, replica string) ([]*daemon.VertexError, error) {
+	resp, err := rc.httpClient.Get(fmt.Sprintf("%s/api/v1/pipelines/%s/vertex/%s/replicas/%s/errors", rc.hostURL, pipeline, vertex, replica))
```
whynowy, line 165: `.../pipelines/%s/vertices/%s/...`

pkg/mvtxdaemon/server/service/mvtx_service.go
```diff
@@ -33,19 +33,32 @@ import (
...
+	Details   string
+}
+
 type MonoVertexService struct {
 	mvtxdaemon.UnimplementedMonoVertexDaemonServiceServer
 	monoVtx       *v1alpha1.MonoVertex
 	httpClient    *http.Client
 	rater         raterPkg.MonoVtxRatable
 	healthChecker *HealthChecker
+	localCache    map[PodReplica][]ErrorDetails
```
whynowy, line 60: Thread safe?

veds-g: this will be moved to runtime service as part of this [PR](https://github.com/numaproj/numaflow/pull/2469). wil refactor once that is merged.

pkg/apis/proto/daemon/daemon.proto
veds-g: when we store the timestamp in json file in emptyDir we already convert it to string. Thats why when we pass it back to the numaflow server, we thought of passing it as a string.

PR #2461, opened 2025-03-12
feat: Pod changes for surfacing application errors

adarsh0728:
Feature branch `dbt` for PRs related to bubbling up application errors

1. Introduces a monitoring sidecar container(monitor)
2. Changes in headless service - addition of monitor port (2470) and `PublishNotReadyAddresses = true` in spec
3. Attaches emptyDir to pod - mount volumes for numa and monitor container
4. Write application errors to emptyDir from numa
5. Read path (GET handler) in monitor container to cater to daemon's GET calls


rust/numaflow-monitor/src/runtime_errors.rs
```diff
@@ -0,0 +1,130 @@
...
+                                ├── ts1.json
+                                └── ts2.json
+                        └── udsink/
+                                ├── ts3.json
+                                └── ts4.json
+
+*/
+pub async fn handle_runtime_app_errors() -> impl IntoResponse {
+    let runtime_info_config = RuntimeInfoConfig::default();
+    let app_err_path = Path::new(runtime_info_config.app_error_path.as_str());
```
BulkBeing, line 41: These things can be moved into the state https://docs.rs/axum/latest/axum/extract/struct.State.html
Probably `errors` Vec too (behind `Arc<Mutex<T>>>`).
We need to make sure the `errors` Vec doesn't grow forever.
Also, since the contents of a file doesn't ever change, we can skip reading it if its already in memory (list files in the directory, skip reading files that are in-memory).

rust/numaflow-monitor/src/runtime.rs
```diff
@@ -0,0 +1,339 @@
...
+
+        let timestamp = Utc::now().timestamp();
+        let file_name = format!("{}.json", timestamp);
+
+        let json_str = grpc_status_to_json(&grpc_status, container_name.as_str(), timestamp);
+
+        // create a temp file current.json, write into this file first
+        // rename it to timestamp.json once write operation is completed
+        // this is to ensure that while reading we skip this file (race condition)
+        let temp_file_path = dir_path.join("current.json");
```
yhl25, line 91: instead of temp use current?

rust/numaflow-monitor/src/runtime.rs
```diff
@@ -0,0 +1,339 @@
...
+    serde_json::to_string(&runtime_error_entry).expect("Failed to serialize runtime error message")
+}
+
+fn process_file_entry(
+    file_entry: &fs::DirEntry,
+    errors: &mut Vec<RuntimeErrorEntry>,
+) -> Result<()> {
+    let file_path = file_entry.path();
+    let file_name_to_ignore = Some(std::ffi::OsStr::new("current.json"));
+    // if file path isn't a file or file name is current.json, continue processing other files
```
yhl25, line 206: implement TryFrom on runtime error entry? to create runtime entry type from bytes

rust/numaflow-monitor/src/runtime.rs
```diff
@@ -0,0 +1,339 @@
...
+        };
+
+        // iterate over all subdirectories and its files
+        for entry in paths.flatten() {
+            // UD container will have its own directory
+            let sub_dir_path = entry.path();
+            if !sub_dir_path.is_dir() {
+                continue;
+            }
+            match fs::read_dir(&sub_dir_path) {
```
yhl25, line 130: filter out the current.json file?

rust/numaflow-monitor/src/runtime.rs
```diff
@@ -0,0 +1,339 @@
...
+    // Create a RuntimeErrorEntry instance
+    let runtime_error_entry = RuntimeErrorEntry {
+        container: container_name.to_string(),
+        timestamp: rfc3339_timestamp,
+        code,
+        message,
+        details: details_str.to_string(),
+    };
+
+    // Serialize the RuntimeErrorEntry instance to a JSON string
```
yhl25, line 196: use into implementation to convert the error to runtime entry

rust/numaflow-monitor/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+numaflow-models.workspace = true
+axum.workspace = true
+axum-server.workspace = true
+serde.workspace = true
+thiserror = "2.0.3"
+serde_json = "1.0.120"
+chrono = { version = "0.4", features = ["serde"] }
+rcgen = "0.13.1"
+http = "1.2.0"
+tempfile = "3.5"
```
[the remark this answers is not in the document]
vigith: shouldn't it be `workspace=true` ?

rust/numaflow-monitor/src/runtime.rs
```diff
@@ -0,0 +1,329 @@
...
+use std::io::Write;
+use std::path::Path;
+use std::str;
+use tonic::Status;
+use tracing::error;
+
+const CURRENT_FILE: &str = "current.json";
+
+/// Represents a single runtime error entry persisted by the application.
+#[derive(Serialize, Deserialize, Debug, Clone)]
```
vigith: why does this need a clone? i am removing it.

rust/numaflow-monitor/src/runtime.rs
```diff
@@ -0,0 +1,329 @@
...
+            }
+        }
+
+        Ok(errors)
+    }
+}
+
+///  Extracts the container name from an error message using a regular expression.
+fn extract_container_name(error_message: &str) -> String {
+    let re = Regex::new(r"\((.*?)\)").unwrap();
```
vigith, line 224: 1. please make this a static/const (defined once), we need not have to recompile the Regex all the time.
2. what `.*?`, why not `.+` ?

rust/numaflow-monitor/src/runtime.rs
```diff
@@ -0,0 +1,329 @@
...
+/// Response for managing runtime error persistence and retrieval.
+pub struct Runtime {
+    /// The root directory where error files are stored.
+    application_error_path: String,
+    /// The maximum number of error files allowed per container.
+    max_error_files_per_container: usize,
+}
+
+impl Runtime {
+    // Creates a new Runtime instance
```
vigith, line 93:
```suggestion
    /// Creates a new Runtime instance
```

rust/numaflow-monitor/src/runtime.rs
adarsh0728: i think i added it while trying to write a test, but didn't go ahead with it.

rust/numaflow-monitor/src/runtime.rs
adarsh0728: Our error_message is something like this: `"UDF_EXECUTION_ERROR(transformer): Counter reached a multiple of 10"`
`UDF_EXECUTION_ERROR(transformer)` this prefix is fixed but we append the error message to it. If by any chance there is any parenthesis  in the message, then `.+` would consider that as well while we want to stop at first closing parenthesis using  `.*?`

vigith: `.+?` then?

rust/numaflow-monitor/src/runtime.rs
vigith: please do not `Clone` without really thinking hard about it :)

adarsh0728: sure 👍

pkg/apis/numaflow/v1alpha1/container_builder.go
```diff
@@ -23,6 +23,10 @@ import (
 
 type containerBuilder corev1.Container
 
+func createMonitorContainer(req getContainerReq) corev1.Container {
```
whynowy, line 26: Should not put it here.

pkg/apis/numaflow/v1alpha1/mono_vertex_types.go
```diff
@@ -104,11 +104,19 @@ func (mv MonoVertex) GetHeadlessServiceName() string {
 }
 
 func (mv MonoVertex) GetServiceObjs() []*corev1.Service {
-	svcs := []*corev1.Service{mv.getServiceObj(mv.GetHeadlessServiceName(), true, MonoVertexMetricsPort, MonoVertexMetricsPortName)}
+	svcs := []*corev1.Service{mv.getServiceObj(mv.GetHeadlessServiceName(), true, []int32{MonoVertexMetricsPort, MonoVertexMonitorPort}, []string{MonoVertexMetricsPortName, MonoVertexMonitorPortName})}
 	return svcs
 }
 
-func (mv MonoVertex) getServiceObj(name string, headless bool, port int32, servicePortName string) *corev1.Service {
+func (mv MonoVertex) getServiceObj(name string, headless bool, ports []int32, servicePortNames []string) *corev1.Service {
```
whynowy, line 111: Use a map for port and name.

pkg/apis/numaflow/v1alpha1/sink.go
```diff
@@ -53,6 +53,8 @@ func (s Sink) getContainers(req getContainerReq) ([]corev1.Container, []corev1.C
 		s.getMainContainer(req),
 	}
 	sidecarContainers := []corev1.Container{}
+	monitorContainer := createMonitorContainer(req)
```
whynowy, line 56: Instead of doing it in each type of vertex, can we do it in vertex level?

adarsh0728: will check

pkg/apis/numaflow/v1alpha1/container_builder.go
adarsh0728: thought of this, let me see what could be the appropriate place

rust/numaflow-monitor/src/runtime.rs
```diff
@@ -14,6 +15,8 @@ use tonic::Status;
 use tracing::error;
 
 const CURRENT_FILE: &str = "current.json";
+/// A static regex pattern used to extract the container name from an error message.
+static CONTAINER_NAME_REGEX: Lazy<Regex> = Lazy::new(|| Regex::new(r"\((.+?)\)").unwrap());
```
yhl25, line 19: use `std::sync::OnceLock`?

pkg/apis/numaflow/v1alpha1/container_builder.go
adarsh0728: should I create a new file `monitor_container.go` ..assuming i have moved the logic to create monitor container at vertex level in `GetPodSpec` (vertex_types.go)

rust/numaflow-monitor/src/runtime.rs
adarsh0728: something like this would do..

```
fn extract_container_name(s: &str) -> String {
    if let Some(start) = s.find('(') {
        if let Some(end) = s[start + 1..].find(')') {
            return s[start + 1..start + 1 + end].to_string();
        }
    }
    String::new()
}
```

vigith: yes

pkg/apis/numaflow/v1alpha1/mono_vertex_types.go
```diff
@@ -404,6 +419,11 @@ func (mv MonoVertex) GetPodSpec(req GetMonoVertexPodSpecReq) (*corev1.PodSpec, e
 		MountPath: RuntimeDirMountPath,
 	})
 
+	sidecarContainers[0].VolumeMounts = append(sidecarContainers[0].VolumeMounts, corev1.VolumeMount{
```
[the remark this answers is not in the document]
whynowy: Similarly, move line  417 in `buildContainers()`?

pkg/apis/numaflow/v1alpha1/vertex_types.go
```diff
@@ -354,6 +368,11 @@ func (v Vertex) GetPodSpec(req GetVertexPodSpecReq) (*corev1.PodSpec, error) {
 		MountPath: RuntimeDirMountPath,
 	})
 
+	sidecarContainers[0].VolumeMounts = append(sidecarContainers[0].VolumeMounts, corev1.VolumeMount{
```
whynowy, line 371: Though about it again, doing it in each different vertex type might be better.

Also, to be consistent, add the volume mount to numa container as well.

pkg/apis/numaflow/v1alpha1/monitor_container.go
```diff
@@ -0,0 +1,22 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package v1alpha1
+
+import corev1 "k8s.io/api/core/v1"
+
+func createMonitorContainer(req getContainerReq) corev1.Container {
+	return containerBuilder{}.init(req).name(CtrMonitor).asSidecar().command(NumaflowRustBinary).args("--monitor").build()
```
whynowy, line 21: Remove volumeMount from this function, because the volumeMounts contains one item is only supposed to be for the numa container.


```
volumeMounts := []corev1.VolumeMount{{Name: varVolumeName, MountPath: PathVarRun}}
```

pkg/apis/numaflow/v1alpha1/monitor_container.go
```diff
@@ -0,0 +1,22 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package v1alpha1
+
+import corev1 "k8s.io/api/core/v1"
+
+func createMonitorContainer(req getContainerReq) corev1.Container {
```
whynowy, line 20: buildMonitorContainer?

rust/numaflow-monitor/src/app.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+/// Handler for the /runtime/errors route to get application errors.
+async fn handle_runtime_app_errors(State(state): State<Arc<AppState>>) -> impl IntoResponse {
+    let runtime = state.runtime.as_ref();
+
+    // Get the application errors from the runtime.
+    match runtime.get_application_errors() {
+        Ok(errors) => (
+            StatusCode::OK,
```
yhl25, line 66: we need to add unit tests for handlers, we can do it in a follow-up PR.

yhl25 (approved):

PR #2473, opened 2025-03-17
feat: errors tab for application errors

veds-g:
image

### DIFF
--- a/ui/src/components/pages/Pipeline/partials/Graph/partials/NodeInfo/partials/Pods/partials/PodDetails/index.tsx
+++ b/ui/src/components/pages/Pipeline/partials/Graph/partials/NodeInfo/partials/Pods/partials/PodDetails/index.tsx
@@ -4,6 +4,7 @@ import Tab from "@mui/material/Tab";
 import Tabs from "@mui/material/Tabs";
 import { Metrics } from "./partials/Metrics";
 import { PodLogs } from "./partials/PodLogs";
+import { Errors } from "./partials/Errors";
 import { PodDetailProps } from "../../../../../../../../../../../types/declarations/pods";
 import { AppContextProps } from "../../../../../../../../../../../types/declarations/app";
 import { AppContext } from "../../../../../../../../../../../App";
@@ -22,7 +23,8 @@ const headerSx = {
 };
 
 const LOGS_TAB_INDEX = 0;
-const METRICS_TAB_INDEX = 1;
+const ERRORS_TAB_INDEX = 1;
+const METRICS_TAB_INDEX = 2;
 
 export function PodDetail({
   namespaceId,
@@ -30,7 +32,7 @@ export function PodDetail({
   type,
   containerName,
   pod,
-  vertexId
+  vertexId,
 }: PodDetailProps) {
   if (!pod) return null;
 
@@ -65,6 +67,15 @@ export function PodDetail({
           label="Logs"
           data-testid="logs-tab"
         />
+        <Tab
+          className={
+            podsViewTab === ERRORS_TAB_INDEX
+              ? "vertex-details-tab-selected"
+              : "vertex-details-tab"
+          }
+          label="Errors"
+          data-testid="errors-tab"
+        />
         {!disableMetricsCharts && (
           <Tab
             className={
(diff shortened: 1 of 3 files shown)


ui/src/components/common/SlidingSidebar/partials/VertexDetails/index.tsx
```diff
@@ -224,6 +239,58 @@ export function VertexDetails({
     [setModalOnClose]
   );
 
+  const { data: errorsDetailsData } = useErrorsFetch({
```
adarsh0728, line 242: should we handle loading and errors from result of useErrorsFetch?

ui/src/components/common/SlidingSidebar/partials/VertexDetails/partials/Errors/index.tsx
```diff
@@ -0,0 +1,212 @@
...
+    updateLists(details);
+  }, [details, updateLists]);
+
+  const filterDetails = useCallback(
+    (
+      selectedPod: string,
+      selectedContainer: string,
+      details: (ContainerError & { pod: string })[]
+    ) => {
+      if (details.length === 0) return [];
```
adarsh0728, line 53: this can be simplified

```
    if (details.length === 0) return [];
    
    return details.filter(d => 
      (selectedPod === "All" || d.pod === selectedPod) &&
      (selectedContainer === "All" || d.container === selectedContainer)
    );
```

ui/src/components/common/SlidingSidebar/partials/VertexDetails/partials/Errors/index.tsx
```diff
@@ -0,0 +1,212 @@
...
+        const pods = extractUniqueItems(details.map((d) => d.pod));
+        const containers = extractUniqueItems(details.map((d) => d.container));
+        setPodList(["All", ...pods]);
+        setContainerList(["All", ...containers]);
+      } else {
+        setPodList(["All"]);
+        setContainerList(["All"]);
+      }
+    },
+    []
```
adarsh0728, line 40: should we add [setPodList, setContainerList] as dependencies in callback?

ui/src/components/common/SlidingSidebar/partials/VertexDetails/partials/Errors/index.tsx
```diff
@@ -0,0 +1,212 @@
...
+  );
+
+  useEffect(() => {
+    const filtered = filterDetails(selectedPod, selectedContainer, details);
+    // sort by timestamp
+    const sorted = [...filtered].sort(
+      (a, b) =>
+        new Date(b.timestamp).getTime() - new Date(a.timestamp).getTime()
```
adarsh0728, line 74: unix timestamp would be in seconds (write path).. should not affect sorting.. but we need to ensure `Date` works fine with seconds as well.

ui/src/components/common/SlidingSidebar/partials/VertexDetails/index.tsx
```diff
@@ -224,6 +239,58 @@ export function VertexDetails({
...
+  }, [errorsDetailsData, constructDetails]);
+
+  const filterErrorsWithinLast24Hours = useCallback(
+    (containerErrorsData: (ContainerError & { pod: string })[]) => {
+      const currentTime = new Date().getTime();
+      const twentyFourHoursAgo = currentTime - 24 * 60 * 60 * 1000; // 24 hours in milliseconds
+      return containerErrorsData.filter((errorData) => {
+        return (
+          errorData.timestamp &&
+          new Date(errorData.timestamp).getTime() >= twentyFourHoursAgo
```
adarsh0728, line 281: same..ensure it is correct (seconds/milliseconds)

ui/src/components/common/SlidingSidebar/partials/VertexDetails/index.tsx
veds-g: this is at the the top level along with other Tabs logic, so loading will prevent rendering other tabs. That is why i am handling loading and errors using `undefined` value of data.

ui/src/components/common/SlidingSidebar/partials/VertexDetails/partials/Errors/index.tsx
veds-g: they are useState function which are stable and do not change between renders, hence not needed in dependency array.

ui/src/components/common/SlidingSidebar/partials/VertexDetails/partials/Errors/index.tsx
veds-g: using `google.protobuf.Timestamp timestamp` in proto it is converted to a RFC 3339 date string when converted to JSON mapping

adarsh0728 (approved):

yhl25 (approved):

Issue #2470, opened 2025-03-17
Implement Accumulator in Numaflow Core


vigith:
- [ ] implement a stateful accumulator as an extension of Unaligned Window
- [ ] use unaligned WAL to garbage collect the accumulated state
- [ ] Controller changes to add a new reduce type called Accumulator

PR #2475, opened 2025-03-18
feat: Accumulator

yhl25:
fixes https://github.com/numaproj/numaflow/issues/2470


pkg/reduce/pnf/pnf.go
```diff
@@ -206,6 +209,20 @@ forwardLoop:
...
+					if win.EndTime().Before(response.Window.EndTime()) {
+						if err := pf.handleEOFWindow(ctx, response.Window); err != nil {
+							return
+						}
+						pf.lastSeenWindow[winKey] = response.Window
+					}
+				} else {
+					pf.lastSeenWindow[winKey] = response.Window
+				}
+			}
```
vigith, line 224: now that we return WM as Window.Endtime, this can be simplified.

pkg/sdkclient/serverinfo/types.go
```diff
@@ -97,6 +98,7 @@ var minimumSupportedSDKVersions = sdkConstraints{
 		ContainerTypeSessionreducer:    "0.9.0rc100",
 		ContainerTypeSideinput:         "0.9.0rc100",
 		ContainerTypeFbsinker:          "0.9.0rc100",
+		ContainerTypeAccumulator:       "0.9.0-z",
```
vigith, line 101: this is wrong, it should be 0.10.0-z

pkg/window/strategy/accumulate/accumulate.go
```diff
@@ -0,0 +1,230 @@
...
+
+func (w *accumulatorWindow) Expand(time.Time) {
+	// No expand operation for Accumulator window
+}
+
+// windowState maintains the state of the window, consisting of the window and the message timestamps.
+type windowState struct {
+	window            window.TimedWindow
+	messageTimestamps []time.Time
+	latestEventTime   time.Time
```
vigith, line 77: do we need this?

rust/Cargo.toml
```diff
@@ -41,8 +41,8 @@ verbose_file_reads = "warn"
 # This profile optimizes for runtime performance and small binary size at the expense of longer build times.
 # Compared to default release profile, this profile reduced binary size from 29MB to 21MB
 # and increased build time (with only one line change in code) from 12 seconds to 133 seconds (tested on Mac M2 Max).
-[profile.release]
-lto = "fat"
+#[profile.release]
+#lto = "fat"
```
vigith, line 45: revert

pkg/window/strategy/accumulate/accumulate.go
yhl25: follow-up PR.

pkg/reduce/pnf/pnf.go
yhl25: This is more of an optimisation to persist the gc event only when the window end time changes. Added a comment.

vigith (approved):

PR #2486, opened 2025-03-20
feat: Support for pending count with Adaptive lookback 

RohanAshar:
Continuation of https://github.com/numaproj/numaflow/pull/2373

This adds the pending calculation in the daemon and emits the metric. Timeline is now maintained in the daemon. Currently uses the 1m value of pending metric. This will change once we have the monovertex starts sending the lag metric without the timeline calculation.

Follow Up Work:
- Monovertex to send pending lag without the timeline calculation to daemon
- Monovertex to call pending for only pod 0


kohlisid:
https://github.com/numaproj/numaflow/pull/2487
PR for the lag reader changes! Adds the new impl but triggers it only for monovertex right now.
Once similar changes are made for pipeline, then we can switch both monovertex and pipeline to the new impl.
Requires clean up! 
We can discuss if we want to break this down in parts and merge, or only merge when both are done!

pkg/mvtxdaemon/server/service/rater/helper.go
```diff
@@ -79,6 +81,34 @@ func CalculateRate(q *sharedqueue.OverflowQueue[*TimestampedCounts], lookbackSec
...
+		currentPending := counts[i].PodCountSnapshot()
+		for _, pendingCount := range currentPending {
+			delta += int64(pendingCount)
+			num++
+		}
+	}
+	if num > 0 {
+		return delta / num
+	}
+	return delta
```
yhl25, line 109:
```suggestion
        if num == 0 {
            return pendingNotAvailable
        }
        return delta / num
```

When num is zero we should return -1 right?

pkg/mvtxdaemon/server/service/rater/rater.go
```diff
@@ -148,24 +155,36 @@ func (r *Rater) monitorOnePod(ctx context.Context, key string, worker int) error
...
-		podReadCount = nil
+	} else {
+		podMetrics := r.getPodMetrics(pInfo.podName)
+		if podMetrics == nil {
+			log.Debugf("No metrics available for pod %s", pInfo.podName)
+		} else {
+			podReadCount = r.getPodReadCounts(pInfo.podName, podMetrics)
+			if podReadCount == nil {
+				log.Debugf("Failed retrieving read counts for pod %s", pInfo.podName)
+			}
```
yhl25, line 171: Do we need to continue when pod read count is nil?

pkg/mvtxdaemon/server/service/rater/rater.go
```diff
@@ -183,22 +202,53 @@ func (r *Rater) getPodReadCounts(podName string) *PodReadCount {
...
+func (r *Rater) getPodPendingCounts(podName string, result map[string]*dto.MetricFamily) *PodMetricsCount {
+	if value, ok := result[monoVtxPendingRawMetric]; ok && value != nil && len(value.GetMetric()) > 0 {
+		metricsList := value.GetMetric()
+		podPendingCount := &PodMetricsCount{podName, metricsList[0].Gauge.GetValue()}
+		return podPendingCount
+	} else {
+		r.log.Infof("[Pod name %s]: Metric %q is unavailable, the pod might haven't started processing data", podName, monoVtxPendingRawMetric)
+	}
+	return nil
+}
```
yhl25, line 235: return the negation case early to make it more readable?

pkg/mvtxdaemon/server/service/rater/helper.go
```diff
@@ -79,6 +81,34 @@ func CalculateRate(q *sharedqueue.OverflowQueue[*TimestampedCounts], lookbackSec
...
+// CalculatePending calculates the pending of a MonoVertex for a given lookback period.
+func CalculatePending(q *sharedqueue.OverflowQueue[*TimestampedCounts], lookbackSeconds int64) int64 {
+	counts := q.Items()
+	if len(counts) <= 1 {
+		return pendingNotAvailable
+	}
+	startIndex := findStartIndex(lookbackSeconds, counts)
+	// we consider the last but one element as the end index because the last element might be incomplete
+	// we can be sure that the last but one element in the queue is complete.
+	endIndex := len(counts) - 1
```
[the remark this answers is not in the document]
RohanAshar: I think the comment is misleading here. We can check the last element as this would be the raw pending metric and not the calculated rate. Fixed the comment.

pkg/mvtxdaemon/server/service/rater/helper.go
RohanAshar: Yes you are correct. Good catch.

pkg/mvtxdaemon/server/service/rater/rater.go
RohanAshar: Pending count can still be fetched even if the read is not yet ready.

pkg/mvtxdaemon/server/service/rater/rater.go
RohanAshar: This is symmetric with getPodReadCounts 
If needed we can change both

pkg/mvtxdaemon/server/service/rater/rater.go
```diff
@@ -71,27 +75,29 @@ type Rater struct {
...
 	// lookBackSeconds is the lookback time window used for scaling calculations
 	// this can be updated dynamically, defaults to user-specified value in the spec
 	lookBackSeconds *atomic.Float64
 	options         *options
 }
 
-// PodReadCount is a struct to maintain count of messages read by a pod of MonoVertex
-type PodReadCount struct {
+// PodMetricsCount is a struct to maintain count of messages read by a pod of MonoVertex
+type PodMetricsCount struct {
```
whynowy, line 87: The new name seems to be worse than the old one :)

Does this struct need to be public?

pkg/mvtxdaemon/server/service/rater/rater.go
```diff
@@ -148,24 +155,36 @@ func (r *Rater) monitorOnePod(ctx context.Context, key string, worker int) error
...
 	}
-	now := time.Now().Add(CountWindow).Truncate(CountWindow).Unix()
 	UpdateCount(r.timestampedPodCounts, now, podReadCount)
 	return nil
 }
 
-// getPodReadCounts returns the total number of messages read by the pod
-// It fetches the total pod read counts from the Prometheus metrics endpoint.
-func (r *Rater) getPodReadCounts(podName string) *PodReadCount {
+// getPodMetrics Fetches the total pod read counts from the Prometheus metrics endpoint.
```
whynowy, line 186: Is the doc correct?

pkg/mvtxdaemon/server/service/rater/rater.go
```diff
@@ -37,6 +39,7 @@ import (
 
 const CountWindow = time.Second * 10
 const monoVtxReadMetricName = "monovtx_read_total"
+const monoVtxPendingRawMetric = "monovtx_pending_raw"
```
whynowy, line 42: Is this the one to be updated?

Having this PR merged will break anything?

pkg/mvtxdaemon/server/service/rater/rater.go
```diff
@@ -209,7 +259,7 @@ func (r *Rater) GetRates() map[string]*wrapperspb.DoubleValue {
 		rate := CalculateRate(r.timestampedPodCounts, i)
 		result[n] = wrapperspb.Double(rate)
 	}
-	r.log.Debugf("Got rates for MonoVertex %s: %v", r.monoVertex.Name, result)
+	r.log.Infof("Got rates for MonoVertex %s: %v", r.monoVertex.Name, result)
```
whynowy, line 262: Still use debug?

pkg/mvtxdaemon/server/service/rater/rater.go
RohanAshar: Open to a new name :) As this now holds pending count as well so thought of changing it to MetricsCount to be generic

pkg/mvtxdaemon/server/service/rater/rater.go
RohanAshar: This PR should go in With https://github.com/numaproj/numaflow/pull/2487

pkg/mvtxdaemon/server/service/rater/rater.go
RohanAshar: Done

pkg/mvtxdaemon/server/service/rater/rater.go
RohanAshar: The above PR is merged so should be good to go

pkg/mvtxdaemon/server/service/rater/helper.go
```diff
@@ -29,10 +29,12 @@ const (
 	// rateNotAvailable is returned when the processing rate cannot be derived from the currently
 	// available pod data, a negative min is returned to indicate this.
 	rateNotAvailable = float64(math.MinInt)
+	// pendingNotAvailable default value returned when Pending not available
+	pendingNotAvailable = int64(-1)
```
whynowy, line 33: There's also `isb.PendingNotAvailable`, which is different from this one, and used in multiple places. Please check if this will break any existing feature.

pkg/metrics/metrics.go
```diff
@@ -32,6 +32,7 @@ const (
 	LabelVertexType         = "vertex_type"
 	LabelPartitionName      = "partition_name"
 	LabelMonoVertexName     = "mvtx_name"
+	LabelPendingPeriod      = "period"
```
whynowy, line 35: There's also `LabelPeriod` existing in metrics_server.go, merge them?

RohanAshar: Done

pkg/mvtxdaemon/server/service/rater/helper.go
RohanAshar: I dont think it breaks anything as we check for < 0 in our autoscaler. Having said that, I have consolidated both. It does make sense to use just one. Thanks

pkg/mvtxdaemon/server/service/rater/rater.go
```diff
@@ -183,22 +202,53 @@ func (r *Rater) getPodReadCounts(podName string) *PodReadCount {
...
+// GetPending returns the pending count for the mono vertex
+func (r *Rater) GetPending() map[string]*wrapperspb.Int64Value {
+	r.log.Debugf("Current timestampedPodCounts for MonoVertex %s is: %v", r.monoVertex.Name, r.timestampedPendingCount)
+	var result = make(map[string]*wrapperspb.Int64Value)
+	// calculate pending for each lookback seconds
+	for n, i := range r.buildLookbackSecondsMap() {
+		pending := CalculatePending(r.timestampedPendingCount, i)
+		result[n] = wrapperspb.Int64(pending)
+		// Expose the metric for pending
+		metrics.MonoVertexPendingMessages.WithLabelValues(r.monoVertex.Name, n).Set(float64(pending))
```
whynowy, line 246: When the pending is still building up, do not expose the value, otherwise following values will be seen.

```
# TYPE monovtx_pending gauge
monovtx_pending{mvtx_name="simple-mono-vertex",period="15m"} -9.223372036854776e+18
monovtx_pending{mvtx_name="simple-mono-vertex",period="1m"} -9.223372036854776e+18
monovtx_pending{mvtx_name="simple-mono-vertex",period="5m"} -9.223372036854776e+18
monovtx_pending{mvtx_name="simple-mono-vertex",period="default"} -9.223372036854776e+18
```

whynowy (approved):

PR #2487, opened 2025-03-20
chore: refactor lag calculator and source pending only for replica 0

kohlisid:


pkg/metrics/metrics_server.go
```diff
@@ -281,10 +282,22 @@ func (ms *metricsServer) Start(ctx context.Context) (func(ctx context.Context) e
 		Handler:   mux,
 		TLSConfig: &tls.Config{Certificates: []tls.Certificate{*cer}, MinVersion: tls.VersionTLS12},
 	}
+	replicaStr, defined := os.LookupEnv(dfv1.EnvReplica)
+	if !defined {
+		return nil, fmt.Errorf("required environment variable '%s' not defined", dfv1.EnvReplica)
+	}
+	replica, err := strconv.Atoi(replicaStr)
+	if err != nil {
+		return nil, fmt.Errorf("invalid replica %q", replicaStr)
```
yhl25, line 291:
```suggestion
                return nil, fmt.Errorf("invalid replica value %q: %w", replicaStr, err)
```

pkg/metrics/metrics_server.go
```diff
@@ -281,10 +282,22 @@ func (ms *metricsServer) Start(ctx context.Context) (func(ctx context.Context) e
...
 	// Buildup pending information
-	go ms.buildupPendingInfo(ctx)
-	// Expose pending metrics
-	go ms.exposePendingMetrics(ctx)
+	// If the vertex is a source, then only trigger for pod replica id=0
+	if !(ms.vertex.IsASource() && replica != 0) {
+		go ms.buildupPendingInfo(ctx)
+		// Expose pending metrics
+		go ms.exposePendingMetrics(ctx)
+	}
```
yhl25, line 299: only for source?

rust/numaflow-core/src/metrics.rs
```diff
@@ -793,6 +847,89 @@ impl Drop for PendingReaderTasks {
...
+    loop {
+        ticker.tick().await;
+
+        match &mut lag_reader {
+            LagReader::Source(source) => match fetch_source_pending(source).await {
+                Ok(pending) => {
+                    if pending != -1 {
+                        info!("Pending messages {:?}", pending);
+                        if is_mono_vertex {
+                            let metric_labels = mvtx_forward_metric_labels().clone();
```
yhl25, line 882: Try to see if we can avoid level 2 branching.

pkg/metrics/metrics_server.go
```diff
@@ -281,10 +282,22 @@ func (ms *metricsServer) Start(ctx context.Context) (func(ctx context.Context) e
...
+	replica, err := strconv.Atoi(replicaStr)
+	if err != nil {
+		return nil, fmt.Errorf("invalid replica %q", replicaStr)
+	}
 	// Buildup pending information
-	go ms.buildupPendingInfo(ctx)
-	// Expose pending metrics
-	go ms.exposePendingMetrics(ctx)
+	// If the vertex is a source, then only trigger for pod replica id=0
+	if !(ms.vertex.IsASource() && replica != 0) {
```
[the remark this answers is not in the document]
kohlisid: We have an not inversion outside, 
So do not enter if `vertex is source and replica!=0`

pkg/metrics/metrics_server.go
kohlisid: I had thought will turn off for other vertices after end to end tested with the new changes for pipeline.
We can change now as well if needed.

pkg/metrics/metrics_server.go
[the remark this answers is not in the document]
kohlisid: Equivalent to `if (!ms.vertex.IsASource() || replica == 0) `
Can use whichever preferred.

rust/numaflow-core/src/metrics.rs
kohlisid: Can do refactor in a separate PR if you're fine with this right now!

rust/numaflow-core/src/pipeline.rs
```diff
@@ -179,12 +181,16 @@ async fn start_source_forwarder(
...
     .await?;
 
-    let pending_reader = shared::metrics::create_pending_reader(
-        &config.metrics_config,
-        LagReader::Source(source.clone()),
-    )
-    .await;
-    let _pending_reader_handle = pending_reader.start(is_mono_vertex()).await;
+    let mut _pending_reader_handle: Option<PendingReaderTasks> = None;
+    // only check the pending and lag for source for pod_id = 0
```
yhl25, line 185: No need for `mut` here, can just do it inline with an `if`, same in MonoVertex code.

yhl25 (approved):

pkg/metrics/metrics_server.go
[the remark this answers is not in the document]
whynowy: ha. 👍

PR #2491, opened 2025-03-21
chore: use slices.Contains instead of in-house built function

whynowy:
Explain what this PR does.


pkg/apis/numaflow/v1alpha1/udf.go
```diff
@@ -190,7 +190,7 @@ type SessionWindow struct {
 // NOTE: Quite powerful, should not be abused; it can cause stalling of pipelines and leaks.
 type AccumulatorWindow struct {
 	// Timeout is the duration of inactivity after which the state of the accumulator is removed.
-	Timeout *metav1.Duration `json:"ttl,omitempty" protobuf:"bytes,1,opt,name=timeout"`
+	Timeout *metav1.Duration `json:"ttl,omitempty" protobuf:"bytes,1,opt,name=ttl"`
```
vigith, line 193:
```suggestion
	Timeout *metav1.Duration `json:"ttl,omitempty" protobuf:"bytes,1,opt,name=timeout"`
```

vigith (approved):

pkg/apis/numaflow/v1alpha1/udf.go
vigith: it is in aligned meant with session window terminology.

whynowy: Then why don't use `timeout` for all of them?

vigith: session uses timeout, doesn't it?

whynowy: I meant, in `json`, it uses `ttl`.

vigith: let me fix this in a follow up?

whynowy: Sounds good, reverted.

PR #2494, opened 2025-03-21
fix: typo, use timeout instead of ttl

vigith:
Explain what this PR does.


vigith:
@yhl25 any other reference we have since we refactored from ttl to timeout

whynowy (approved):
update the example.

yhl25 (approved):

PR #2499, opened 2025-03-24
feat: surface application errors

### LINKED ISSUE #2404: Persist error state per pod in emptyDir
- [ ] Controller changes to attach emptyDir for each pod
- [ ] Monovertex and Pipeline changes to write the error to emptyDir 
- [ ] expose endpoints to fetch error details in pod's metrics server

### LINKED ISSUE #2405: Daemon server changes to cache error state of pods
- [ ] Background task to fetch error state for each pod and store it using in-memory cache
- [ ] Proto changes to add new api endpoint for daemon server
- [ ] API implementation to get the pod level error details (gRPC and Rest)

adarsh0728:
fixes #2405 fixes #2404 
Pod and daemon server changes to surface application errors


pkg/apis/numaflow/v1alpha1/source.go
```diff
@@ -61,6 +62,11 @@ func (s Source) getContainers(req getContainerReq) ([]corev1.Container, []corev1
 	if req.servingStore != nil && req.servingStore.Container != nil {
 		sidecarContainers = append(sidecarContainers, req.servingStore.getUDStoreContainer(req))
 	}
+	// volume mount to the runtime path
+	containers[0].VolumeMounts = append(containers[0].VolumeMounts, corev1.VolumeMount{
```
whynowy, line 66: Move this to `getMainContainer`?

pkg/apis/numaflow/v1alpha1/vertex_types.go
```diff
@@ -382,9 +388,12 @@ func (v Vertex) GetPodSpec(req GetVertexPodSpecReq) (*corev1.PodSpec, error) {
 		}
 		sideInputsWatcher.VolumeMounts = append(sideInputsWatcher.VolumeMounts, corev1.VolumeMount{Name: sideInputsVolName, MountPath: PathSideInputsMount})
 		containers = append(containers, sideInputsWatcher)
-		for i := 0; i < len(sidecarContainers); i++ {
-			// Readonly mount for user-defined containers
-			sidecarContainers[i].VolumeMounts = append(sidecarContainers[i].VolumeMounts, corev1.VolumeMount{Name: sideInputsVolName, MountPath: PathSideInputsMount, ReadOnly: true})
+		// do not add for monitor container
+		if len(sidecarContainers) > 1 {
```
whynowy, line 392: Checking the container name in the for loop would be better.

pkg/daemon/server/service/runtime/runtime.go
```diff
@@ -0,0 +1,207 @@
...
+
+	cacheKey := PodReplica(fmt.Sprintf("%s-%s-%v", r.pipeline.Name, vtxName, podIndex))
+
+	// Lock the cache before updating
+	r.cacheMutex.Lock()
+	defer r.cacheMutex.Unlock()
+
+	// overwrite the errors
+	r.localCache[cacheKey] = apiResponse.Data
+	r.log.Infof("Persisting error in local cache for: %s", cacheKey)
```
whynowy, line 192: Will this log be too much? It seems we log it everytime the fetch happens.

adarsh0728: no. if errors are not written, there would be no `application-errors `directory. In that case, we will return before

```
	if apiResponse.ErrMessage != "" {
		return
	}
```

adarsh0728: the above case is for those scenarios where there are no errors, but if there is even error we would be logging this every 60 seconds.. so thinking of making it a DEBUG log

yhl25 (approved):

pkg/apis/numaflow/v1alpha1/container_builder.go
```diff
@@ -83,6 +83,11 @@ func (b containerBuilder) appendVolumeMounts(x ...corev1.VolumeMount) containerB
 	return b
 }
 
+func (b containerBuilder) setVolumeMounts(x ...corev1.VolumeMount) containerBuilder {
```
whynowy, line 86: nit: `volumeMounts()`

pkg/daemon/server/service/runtime/runtime.go
```diff
@@ -0,0 +1,207 @@
...
+type PipelineRuntimeCache interface {
+	// StartCacheRefresher starts the cache refresher to update the local cache with the runtime errors.
+	StartCacheRefresher(ctx context.Context) error
+	// GetLocalCache returns the local cache of runtime errors.
+	GetLocalCache() map[PodReplica][]ErrorDetails
+}
+
+var _ PipelineRuntimeCache = (*pipelineRuntimeCache)(nil)
+
+// TODO: use a separate struct instead of concatenating strings
```
whynowy, line 51: Does this comment still apply?

pkg/daemon/server/daemon_server.go
```diff
@@ -154,6 +157,13 @@ func (ds *daemonServer) Run(ctx context.Context) error {
 		}
 	}()
 
+	// Start the pipelineRuntimeCache
+	go func() {
+		if err := pipelineRuntimeCache.StartCacheRefresher(ctx); err != nil {
```
whynowy, line 162: `StartCacheRefresher()` doesn't seem to be a blocking call, do we need a goroutine to start it?

pkg/mvtxdaemon/server/daemon_server.go
```diff
@@ -108,6 +111,13 @@ func (ds *daemonServer) Run(ctx context.Context) error {
 		}
 	}()
 
+	// Start the monoVertexRuntimeCache
+	go func() {
+		if err := monoVertexRuntimeCache.StartCacheRefresher(ctx); err != nil {
```
whynowy, line 116: Same here.

whynowy (approved):

pkg/daemon/server/daemon_server.go
adarsh0728: checking this cc @veds-g

pkg/daemon/server/service/runtime/runtime.go
adarsh0728: not for this PR. If required, this will be taken up in numaflow server API changes as per UI design

pkg/daemon/server/daemon_server.go
adarsh0728: yes, not a blocking call as we return after spinning up go routines.

PR #2511, opened 2025-03-30
chore: e2e test for accumulator

yhl25:


test/reduce-two-e2e/reduce_two_test.go
```diff
@@ -1,5 +1,3 @@
-//go:build test
```
vigith, line 1: accidental?

test/reduce-two-e2e/testdata/accumulator/stream-sorter-go.yaml
```diff
@@ -0,0 +1,56 @@
...
+        http: {}
+    - name: input-two
+      scale:
+        min: 1
+      source:
+        http: {}
+    - name: accum
+      udf:
+        container:
+          # compute the sum
```
vigith, line 22:
```suggestion
          # sort the incoming stream 
```

vigith (approved):

PR #2518, opened 2025-04-01
chore: append numa to filename being persisted for runtime errors

adarsh0728:
We are writing error files from ud containers as well. To avoid race conditions on files, we are appending `-udf` to timestamp in udf. Similarly, `-numa` when writing from numa container.


yhl25 (commented):
add -numa to current file as well?

vigith (approved):

vigith:
Since we append for all cases except "numa", do we really need the prefix? Isn't it implicit?

adarsh0728:
> Since we append for all cases except "numa", do we really need the prefix? Isn't it implicit?

Yes it's implicit, but we wanted to remain consistent with the file naming. This would also help in sorting the files easily with `'-'` as the separator.

PR #2519, opened 2025-04-01
feat: vertex error endpoints

veds-g:


pkg/apis/proto/daemon/daemon.proto
```diff
@@ -109,6 +109,28 @@ message GetPipelineWatermarksRequest {
...
 }
 
+message GetVertexErrorsRequest {
+  string pipeline = 1;
+  string vertex = 2;
+}
+
+message ContainerError {
+  string container = 1;
+  string timestamp = 2;
```
adarsh0728, line 119: thinking if we should use `google.protobuf.Timestamp` for timestamp. While writing to file, we can use `let timestamp = Utc::now().timestamp()`; directly. 
And then

server/apis/v1/handler.go
```diff
@@ -1262,6 +1262,42 @@ func (h *handler) GetMonoVertexHealth(c *gin.Context) {
 	c.JSON(http.StatusOK, NewNumaflowAPIResponse(nil, response))
 }
 
+func (h *handler) GetVertexErrors(c *gin.Context) {
```
adarsh0728, line 1265: add some comments/docs

pkg/apis/proto/daemon/daemon.proto
```diff
@@ -109,6 +110,28 @@ message GetPipelineWatermarksRequest {
   string pipeline = 1;
 }
 
+message GetVertexErrorsRequest {
```
yhl25, line 113: add some code comments

yhl25 (approved):

adarsh0728 (approved):

pkg/daemon/server/service/runtime/runtime.go
```diff
@@ -38,32 +38,34 @@ const (
...
-type ErrorDetails struct {
-	Container string `json:"container"`
-	Timestamp string `json:"timestamp"`
-	Code      string `json:"code"`
-	Message   string `json:"message"`
-	Details   string `json:"details"`
+
+type ReplicaErrors struct {
+	Replica         string         `json:"replica"`
+	ContainerErrors []ErrorDetails `json:"container_errors"`
```
whynowy, line 56: json - `containerErrors`

whynowy (approved):

PR #2521, opened 2025-04-03
fix: create a new bidirectional stream for every partition to avoid race condition

yhl25:


pkg/sinks/sink.go
```diff
@@ -156,26 +156,34 @@ func (u *SinkProcessor) Start(ctx context.Context) error {
...
+			// Readiness check
+			if err := sinkHandler.WaitUntilReady(ctx); err != nil {
+				return fmt.Errorf("failed on UDSink readiness check, %w", err)
+			}
 
-		defer func() {
-			err = sdkClient.CloseConn(ctx)
-			if err != nil {
-				log.Warnw("Failed to close sdk client", zap.Error(err))
+			// add sinkHandler to healthCheckers
```
vigith, line 173:
```suggestion
			// since it is the same UD Container, we need to do health check only once.
```

vigith (approved):

PR #2525, opened 2025-04-04
chore: Adding QualiBooth to Users list

yhl25:
Add QualiBooth to the users.md file.

They use Numaflow to process real-time data related to web accessibility.
Permission to add was confirmed by an engineer from the company.

Submitted by @borkox.


borkox:
Thank you

PR #2529, opened 2025-04-07
chore: update grpc go to v1.71.1

yhl25:
### Summary

This PR updates the `grpc-go` dependency from `v1.66` to `v1.71` to fix a [long-running stream bug](https://github.com/grpc/grpc-go/issues/7641).

Starting with `v1.67`, grpc-go enforces ALPN during TLS handshakes for HTTP/2 compliance. Without ALPN, connections fail.

Since we use cmux to run both HTTP and gRPC servers on the same port (using HTTP/1.1), we have to disable ALPN enforcement using the `GRPC_ENFORCE_ALPN_ENABLED` env var.

An alternative is the [experimental credentials package](https://github.com/grpc/grpc-go/commit/54b3eb97dbf7400efb5750f26084c2d3b2eff120) introduced in v1.70, but we’ll stick with the env approach for now. We can revert this once official support to disable ALPN is [available](https://github.com/grpc/grpc-go/issues/434).


pkg/reconciler/pipeline/controller.go
```diff
@@ -448,6 +448,7 @@ func (r *pipelineReconciler) createOrUpdateDaemonDeployment(ctx context.Context,
 	log := logging.FromContext(ctx)
 	isbSvcType, envs := sharedutil.GetIsbSvcEnvVars(isbSvcConfig)
 	envs = append(envs, corev1.EnvVar{Name: dfv1.EnvPipelineName, Value: pl.Name})
+	envs = append(envs, corev1.EnvVar{Name: dfv1.EnvGrpcAlpnEnabled, Value: "false"})
```
whynowy, line 451: Why don't go with the approach of `setenv` in the source code? That will be easier to clean up in the future.

config/base/numaflow-server/numaflow-server-deployment.yaml
```diff
@@ -149,6 +149,8 @@ spec:
                 name: numaflow-cmd-params-config
                 key: server.daemon.client.protocol
                 optional: true
+          - name: GRPC_ENFORCE_ALPN_ENABLED
```
whynowy, line 152: Same question.

config/base/controller-manager/controller-manager-deployment.yaml
```diff
@@ -69,6 +69,8 @@ spec:
                   name: numaflow-cmd-params-config
                   key: controller.leader.election.lease.renew.period
                   optional: true
+            - name: GRPC_ENFORCE_ALPN_ENABLED
```
whynowy, line 72: same question.

pkg/reconciler/pipeline/controller.go
yhl25: https://github.com/grpc/grpc-go/blob/b368379ef8f6fbe8b3b856d38dbb34a50fbfbe9a/internal/envconfig/envconfig.go#L48
It is fetched during init

vigith:
ALPN is a standard from 2014, should we create an issue to track to support/enable ALPN?

pkg/reconciler/pipeline/controller.go
whynowy: Should it be okay if we add it in the place of code start? Adding it to the manifests, especially the server and controller manifests is nasty, and not able to clean up if ppl do `kubectl apply` during upgrade.

Makefile
```diff
@@ -135,7 +135,7 @@ endif
 	$(MAKE) restart-control-plane-components
 	cat test/manifests/e2e-api-pod.yaml | sed 's@quay.io/numaproj/@$(IMAGE_NAMESPACE)/@' | sed 's/:latest/:$(VERSION)/' | kubectl -n numaflow-system apply -f -
 	go generate $(shell find ./test/$* -name '*.go')
-	go test -v -timeout 15m -count 1 --tags test -p 1 ./test/$*
+	GRPC_ENFORCE_ALPN_ENABLED=false go test -v -timeout 15m -count 1 --tags test -p 1 ./test/$*
```
whynowy, line 138: Similarly, will adding it to `SetupSuite()`?

BTW - it should use `-go test`.

whynowy (approved):

PR #2532, opened 2025-04-08
chore: ability to use rust runtime using env, fix serving store

yhl25:


entrypoint.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+set -e
```
BulkBeing, line 2: Shell scripting best practice: `set -euo pipefail` 
https://gist.github.com/mohanpedala/1e2ff5661761d3abd0385e8223e16425

yhl25: Tried it got a warning, apparently it doesn't work if we don't use [bash](https://github.com/koalaman/shellcheck/wiki/SC3040). Also since our script is very minimal I don't think we really need it.

BulkBeing (approved):

vigith (approved):

examples/31-serving-custom-store.yaml
```diff
@@ -1,42 +1,44 @@
...
+      container:
+        image: quay.io/numaio/numaflow-go/serving-redis-store:stable
+  pipeline:
+    vertices:
+      - name: serving-in
+        scale:
+          min: 1
+        source:
+          serving:
+            service: true
```
whynowy, line 19: The examples need to be cleaned up later on.

whynowy (approved):

examples/31-serving-custom-store.yaml
vigith: @yhl25 you want to tackle this in this PR?

PR #2538, opened 2025-04-09
feat: [experimental] start golang vs. rust image based on env (#2537)

vigith:
This is an experimental feature to use the newer Async data-movement for all vertices except Reduce.


rust/Cargo.toml
```diff
@@ -42,8 +42,8 @@ verbose_file_reads = "warn"
 # This profile optimizes for runtime performance and small binary size at the expense of longer build times.
 # Compared to default release profile, this profile reduced binary size from 29MB to 21MB
 # and increased build time (with only one line change in code) from 12 seconds to 133 seconds (tested on Mac M2 Max).
-[profile.release]
-lto = "fat"
+#[profile.release]
```
yhl25, line 45: revert?

vigith: done

BulkBeing (approved):

PR #2539, opened 2025-04-09
feat: persist error in case of partial/no responses from udf

### LINKED ISSUE #2458: Bubble up errors in UI for no/partial response from UDF
We should be able to surface errors in the UI for better debuggability in cases where there is no response or partial response in the UDF. For example, user code returns `null` instead of `Message.toDrop` when the message has to be dropped.
 
Log in Case of udsink:

`2025-03-07T07:07:39.632405Z ERROR numaflow_core::sink::user_defined: received EOT message before all responses are received, we will wait indefinitely for the remaining responses`

adarsh0728:
fixes #2458 

### Changes

- `persist_application_error` can be executed only once, this is okay as we are persisting either exceptions (which trigger restarts) or cases where we go to an indefinite wait state. Also, made it independent of runtime instance.

-   call `persist_application_error` from sink and transformer in case of partial responses.

### Observed Behaviour

- **Sink**: if we receive `EOT` before all responses, we get stuck in loop waiting for all responses. A single error is persisted. Also, **udsink** receives `SIGTERM` after around 5 minutes and shuts down, triggering restarts (this will be fixed https://github.com/numaproj/numaflow/pull/2541 ). 
- **Transformer**: if we receive empty response, we sleep indefinitely. A single error is persisted. We do not see restarts in this case.

### File Structure 
```
root@simple-mono-vertex-mv-23:/var/numaflow/runtime/application-errors/transformer# ls
1744266107-numa.json
root@simple-mono-vertex-mv-23:/var/numaflow/runtime/application-errors/transformer# cat 1744266107-numa.json 
{"container":"transformer","timestamp":1744266107,"code":"Internal error","message":"UDF_PARTIAL_RESPONSE(transformer)","details":"received empty response from server (transformer), we will wait indefinitely"}root@simple-mono-vertex-mv-23:/var/numaflow/runtime/application-errors/transformer#
```

### API response for UI

```
{
    "data": [
        {
            "replica": "simple-mono-vertex-mv-0",
            "containerErrors": [
                {
                    "container": "transformer",
                    "timestamp": "2025-04-10T06:04:52Z",
                    "code": "Internal error",
                    "message": "UDF_PARTIAL_RESPONSE(transformer)",
                    "details": "received empty response from server (transformer), we will wait indefinitely"
                }
            ]
        },
        {
            "replica": "simple-mono-vertex-mv-2",
            "containerErrors": [
                {
                    "container": "transformer",
                    "timestamp": "2025-04-10T06:05:17Z",
                    "code": "Internal error",
                    "message": "UDF_PARTIAL_RESPONSE(transformer)",
                    "details": "received empty response from server (transformer), we will wait indefinitely"
                }
            ]
        },
        {
            "replica": "simple-mono-vertex-mv-1",
            "containerErrors": [
                {
                    "container": "transformer",
                    "timestamp": "2025-04-10T06:05:17Z",
                    "code": "Internal error",
                    "message": "UDF_PARTIAL_RESPONSE(transformer)",
                    "details": "received empty response from server (transformer), we will wait indefinitely"
                }
            ]
        },
```

### UI

image

### DIFF
--- a/rust/numaflow-core/src/sink/user_defined.rs
+++ b/rust/numaflow-core/src/sink/user_defined.rs
@@ -132,7 +132,7 @@ impl Sink for UserDefinedSink {
                         Code::Internal,
                         "UDF_PARTIAL_RESPONSE(udsink)",
                         Bytes::from_static(
-                            b"received End-Of-Transmission (EOT) before all responses are received from ud sink,\
+                            b"received End-Of-Transmission (EOT) before all responses are received from the ud-sink,\
                             we will wait indefinitely for the remaining responses. This indicates that there is a bug\
                             in the user-code. Please check whether you are accidentally skipping the messages.",
                         ),
--- a/rust/numaflow-core/src/transformer.rs
+++ b/rust/numaflow-core/src/transformer.rs
@@ -131,7 +131,7 @@ impl Transformer {
                 Code::Internal,
                 "UDF_PARTIAL_RESPONSE(transformer)",
                 Bytes::from_static(
-                    b"received End-Of-Transmission (EOT) before all responses are received from ud sink,\
+                    b"received End-Of-Transmission (EOT) before all responses are received from the transformer,\
                             we will wait indefinitely for the remaining responses. This indicates that there is a bug\
                             in the user-code. Please check whether you are accidentally skipping the messages.",
                 ),


rust/numaflow-monitor/src/runtime.rs
```diff
@@ -7,12 +7,20 @@ use std::convert::TryFrom;
...
 use tonic::Status;
 use tracing::error;
+use uuid::Uuid;
 
-const CURRENT_FILE: &str = "current-numa.json";
+static PERSIST_ERROR_LOCK: OnceLock<Arc<Mutex<()>>> = OnceLock::new();
+
+fn get_persist_error_lock() -> &'static Arc<Mutex<()>> {
+    PERSIST_ERROR_LOCK.get_or_init(|| Arc::new(Mutex::new(())))
+}
```
yhl25, line 23: We don't need a lock, we only surface critical errors that caused the pod to crash or go to container state, we could make the `persist_application_error` method `FnOnce`.

yhl25 (approved):

adarsh0728:
Merging this. Will take up persistence of non-gRPC errors in a follow up PR

PR #2547, opened 2025-04-13
fix: propogate health status for pipeline

### LINKED ISSUE #2070: UI: Unhealthy pipeline does not propagate to namespace overview
**Describe the bug**
When one (1) of multiple pipelines (4) are unhealthy (with red icon), the health status does not reflect in the namespace summary.

The summary still has a green heart-icon and ": 4", i.e. all pipelines healthy.

image

- Numaflow: [e.g. v1.2.1]

---

veds-g:
Fixes #2070 

- updated cluster summary api to return health status based on health checker to return critical and warning status
- TODO: cover the same status for mono vertex, also the monoVertex health is hardcoded to healthy in both backend and UI.


server/apis/v1/handler.go
```diff
@@ -269,7 +269,29 @@ func (h *handler) GetClusterSummary(c *gin.Context) {
 		if value, ok := namespaceSummaryMap[pipeline.Namespace]; ok {
 			summary = value
 		}
-		status, err := getPipelineStatus(&pipeline)
+
```
whynowy, line 272: How bad is the performance (e.g.with 20, 50 pipelines)?

PR #2548, opened 2025-04-13
fix: format stacktrace

veds-g:
- Making the stacktrace more readable when viewing the error details
image
image


vigith (commented):
Java does not include path in the stack trace.

veds-g:
> Java does not include path in the stack trace.

should look like this for Java stacktrace
image

ui/src/components/common/SlidingSidebar/partials/VertexDetails/partials/Errors/partials/CollapsableError/index.tsx
```diff
@@ -82,7 +106,9 @@ export const CollapsableError = ({ detail }: CollapsableErrorProps) => {
           </Box>
           <Divider orientation="vertical" flexItem color={"#878789"} />
           <Box className={"collapsable-error-accordion-details-title-content"}>
-            {detail.details}
+            <pre style={{ whiteSpace: "pre-wrap", wordWrap: "break-word" }}>
+              {highlightFilePaths(detail.details)}
```
adarsh0728, line 110: should we add a null check/optional chaining here ? similarly for other fields of detail..

adarsh0728 (approved):
approving with a minor comment

vigith (approved):
awesome, thank you.

PR #2551, opened 2025-04-14
feat: Builtin serve sink

BulkBeing:
With `serve` sink:
```
➜  kubectl apply -f numaflow/examples/1-simple-pipeline.yaml
Error from server (BadRequest): error when creating "numaflow/examples/1-simple-pipeline.yaml": admission webhook "webhook.numaflow.numaproj.io" denied the request: builtin 'serve' sink used in "out" vertex is only allowed with ServingPipeline
```


vigith (commented):
can you add a validation in the validator to make sure "serve" sink is used only while using `ServingPipeline` CRD?

vigith (approved):

PR #2554, opened 2025-04-15
feat: Serving - option to specify container template

BulkBeing:
Example:
```yaml
kind: ServingPipeline
metadata:
  name: serving-pipeline
spec:
  serving:
    service: true
    msgIDHeaderKey: "X-Numaflow-Id"
    containerTemplate:
      resources:
        requests:
          cpu: 1
          memory: 1Gi
  pipeline:
    vertices:
      - name: in
        source:
          serving: {}
      - name: cat
        udf:
          container:
            image: quay.io/numaio/numaflow-go/map-forward-message:stable
      - name: out
        sink:
          serve: {}
    edges:
      - from: in
        to: cat
      - from: cat
        to: out
```

- If the `scale.min` is not set for each vertex, the default value is 1.
- The `replicas` field will be directly applied to the deployment object created for serving server. If user uses an HPA to manage this deployment, omit the `replicas` field (same behavior as the `replicas` field of the native deployment object).
- `spec.Serving.requestTimeoutSeconds` + 10 will be set as the `terminationGracePeriodSecs` on the deployment object.


pkg/apis/numaflow/v1alpha1/serving_pipeline_types.go
```diff
@@ -210,12 +214,22 @@ func (sp ServingPipeline) GetServingDeploymentObj(req GetServingPipelineResource
...
 		Image:           req.Image,
 		ImagePullPolicy: req.PullPolicy,
-		Resources:       req.DefaultResources, // TODO: need to have a way to override.
+		Resources:       req.DefaultResources,
 		Env:             envVars,
 		Command:         []string{NumaflowRustBinary},
 		Args:            []string{"--serving"},
 		VolumeMounts:    volumeMounts,
 	}
+	if ct := sp.Spec.Serving.ContainerTemplate; ct != nil {
```
whynowy, line 223: Use `ContainerTemplate.ApplyToContainer()`?

config/cluster-install/rbac/controller-manager/numaflow-cluster-role.yaml
```diff
@@ -107,3 +107,15 @@ rules:
       - update
       - patch
       - delete
+  - apiGroups:
```
whynowy, line 110: Why adding this?

pkg/reconciler/servingpipeline/controller.go
```diff
@@ -189,6 +191,13 @@ func (r *servingPipelineReconciler) reconcileFixedResources(ctx context.Context,
 		r.recorder.Eventf(spl, corev1.EventTypeWarning, "CreateOrUpdateServingServerFailed", "Failed to create or update serving server: %w", err.Error())
 		return fmt.Errorf("failed to create or update serving server: %w", err)
 	}
+
+	if err := r.createOrUpdateServingHPA(ctx, spl); err != nil {
```
whynowy, line 195: Please do NOT create HPA on behalf.

pkg/reconciler/servingpipeline/controller.go
```diff
@@ -83,6 +84,7 @@ func (r *servingPipelineReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	result, reconcileErr := r.reconcile(ctx, splCopy)
 	if reconcileErr != nil {
 		log.Errorw("Reconcile error", zap.Error(reconcileErr))
+		return ctrl.Result{}, reconcileErr
```
whynowy, line 87: Do not return here. The status needs to be updated.

pkg/reconciler/servingpipeline/controller.go
whynowy: Let the user manage the autoscaling strategy by themselves - could use HPA, or something else. Here, we might introduce a field `replicas` to control the initial replicas.

BulkBeing: Should we use the `spec.serving.scale.min` to set the replicas (`max` will be ignored) or should we introduce a `spec.serving.replicas` field ?
```yaml
spec:
  serving:
    service: true
    msgIDHeaderKey: "X-Numaflow-Id"
    containerTemplate:
      resources:
        requests:
          cpu: 1
          memory: 1Gi
    scale:
      min: 2
      max: 10
```

BulkBeing: Going to add `spec.serving.replicas` field. It will only have an effect when deployment is created for the first time. Since there could be a user managed HPA, I don't think we should update the deployment objects replicas other than during initial creation.

whynowy: Right. Exclude the `replicas` when generating the hash.

BulkBeing: Does this work:
```go
	// Number of replicas. If an HPA is used to manage the deployment object, do not set this field.
	// +optional
	Replicas *int32 `json:"replicas,omitempty" protobuf:"varint,7,opt,name=replicas"`
```
And then simply use the `deploymentObject.Replicas = spl.Spec.Serving.Replicas`

BulkBeing: This way, if the user doesn't have an HPA, and wanted to scale down to a lower value, it works

whynowy: hmmm.. take it back - `Exclude the replicas when generating the hash.` I think we still need to include replicas in the hash.

pkg/apis/numaflow/v1alpha1/get_spec_req.go
```diff
@@ -113,9 +113,11 @@ type GetMonoVertexPodSpecReq struct {
...
-	Image            string                      `protobuf:"bytes,2,opt,name=image"`
-	PullPolicy       corev1.PullPolicy           `protobuf:"bytes,3,opt,name=pullPolicy,casttype=k8s.io/api/core/v1.PullPolicy"`
-	Env              []corev1.EnvVar             `protobuf:"bytes,4,rep,name=env"`
-	DefaultResources corev1.ResourceRequirements `protobuf:"bytes,5,opt,name=defaultResources"`
+	ISBSvcConfig                  BufferServiceConfig         `protobuf:"bytes,1,opt,name=isbSvcConfig"`
+	Image                         string                      `protobuf:"bytes,2,opt,name=image"`
+	PullPolicy                    corev1.PullPolicy           `protobuf:"bytes,3,opt,name=pullPolicy,casttype=k8s.io/api/core/v1.PullPolicy"`
+	Env                           []corev1.EnvVar             `protobuf:"bytes,4,rep,name=env"`
+	DefaultResources              corev1.ResourceRequirements `protobuf:"bytes,5,opt,name=defaultResources"`
+	Replicas                      *int32                      `protobuf:"varint,6,opt,name=replicas"`
```
whynowy, line 121: nit: Can get these 2 directly from the `ServingSpec`, do not need to be set in the Req.

whynowy (approved):
Conditional approval.

pkg/apis/numaflow/v1alpha1/get_spec_req.go
BulkBeing: Done

PR #2559, opened 2025-04-18
chore: clean up unused property

whynowy:
Explain what this PR does.


pkg/apis/numaflow/v1alpha1/serving_source.go
```diff
@@ -1,5 +1,5 @@
 package v1alpha1
 
-// ServingSource is the HTTP endpoint for Numaflow.
+// ServingSource is a placeholder for ServingPipeline source vertex.
```
vigith, line 3:
```suggestion
// ServingSource is the source vertex for ServingPipeline and should be used only with ServingPipeline.
```

WDYT?

whynowy: updated.

vigith (approved):

PR #2571, opened 2025-04-25
feat: generic Write Ahead Log framework

vigith:
A generic Write Ahead Log to combine both Aligned and Unaligned, currently we have two different WAL implementation.


yhl25 (approved):
LGTM!

PR #2572, opened 2025-04-25
chore: upgrade k8s api version and some other fixes

whynowy:
Explain what this PR does.


hack/generate-proto.sh
```diff
@@ -11,7 +11,7 @@ ensure_vendor
 
 make_fake_paths
 # This overrides the existing trap
-trap 'cd "${FAKE_GOPATH}" && go clean -modcache && cd /tmp && rm -rf "${FAKE_GOPATH}"' EXIT
+trap 'cd "${FAKE_GOPATH}" && go clean -modcache && rm -rf "${FAKE_GOPATH}"' EXIT
```
whynowy, line 14: @tmenjo - Can not think of a reason why it fails in your env, but your approach is harmless.

vigith (commented):
upgrade https://github.com/numaproj/numaflow/blob/main/rust/numaflow-models/Cargo.toml#L10 too?

vigith (approved):

PR #2587, opened 2025-05-01
chore: extend compactor to write to tx on replay

vigith:
To avoid an extra read after replay on bootup, we can replay while compacting.

- [x] create a new method called compact_with_replay and take an Option<tx>
- [x] make replay_wal private
- [x] make Offset bytes

### DIFF
--- a/rust/numaflow-core/src/reduce/wal/segment/append.rs
+++ b/rust/numaflow-core/src/reduce/wal/segment/append.rs
@@ -463,13 +463,17 @@ mod tests {
 
         // We can't sort Offset values directly, so we'll just check that we received the same number of IDs
         assert_eq!(
-            received_ids.len(), expected_ids.len(),
+            received_ids.len(),
+            expected_ids.len(),
             "Mismatch between expected and received successful IDs count"
         );
 
         // Check that each expected ID is in the received IDs
         for expected_id in expected_ids {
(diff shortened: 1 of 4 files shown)


rust/numaflow-core/src/reduce/wal/segment/compactor.rs
```diff
@@ -215,11 +227,22 @@ impl Compactor {
...
                                 format!("Failed to send message to compaction WAL: {}", e)
                             })?;
+
+                        // if replay_tx is provided, send the message to it.
+                        // This is used to replay the compacted data during boot up
+                        if let Some(tx) = replay_tx.clone() {
+                            tx.send(data).await.map_err(|e| {
+                                format!("Failed to send message to replay channel: {}", e)
+                            })?;
+                        }
```
yhl25, line 245: Can we avoid cloning tx by using ref?

rust/numaflow-core/src/reduce/wal/segment/append.rs
```diff
@@ -24,7 +25,7 @@ pub(crate) enum SegmentWriteMessage {
     /// Writes the given payload to the WAL.
     WriteData {
         /// Unique ID of the payload. Useful to detect write failures.
-        id: Option<String>,
+        id: Option<Offset>,
```
yhl25, line 28: variable name can be changed to offset?

yhl25 (approved):

PR #2588, opened 2025-05-02
fix: map udf metrics

tmenjo:
## fix: latency metrics for UDF processing time in map UDF

This PR has map UDF...

- Visit forwarder_udf_processing_time (UDFProcessingTime) in **non stream map mode.**
- Visit forwarder_**concurrent**_udf_processing_time (ConcurrentUDFProcessingTime) instead of forwarder_udf_processing_time in **stream map mode.** I think it's better because streaming map is concurrrent processing.

## doc: add forwarder_concurrent_udf_processing_time

This and the following commit in this PR updates the [Metrics](https://numaflow.numaproj.io/operations/metrics/metrics/) document. This one adds forwarder_concurrent_udf_processing_time to the document.

## doc: relationship between UDF and write processing time in Map UDF

This clarifies relationship between forwarder_udf_processing_time/forwarder_concurrent_udf_processing_time and forwarder_write_processing_time. I'd say it's helpful for pipeline developers to analyze metrics.

## doc: upstream or downstream partition for each LET metric

This adds a new column *Which partition* to the latency, traffic, and error metrics tables in the document to clarify what `partition_name=<partition-name>` means. It would be certainly helpful for Numaflow users to understand each metric correctly.


docs/operations/metrics/metrics.md
```diff
@@ -10,64 +10,65 @@ These metrics in combination can be used to determine the overall health of your
...
-| `forwarder_udf_read_total`                 | Counter     | `pipeline=<pipeline-name>` <br> `vertex=<vertex-name>` <br> `vertex_type=<vertex-type>` <br> `replica=<replica-index>` <br> `partition_name=<partition-name>` | Provides the total number of messages read by UDF                                                               |
-| `forwarder_udf_write_total`                | Counter     | `pipeline=<pipeline-name>` <br> `vertex=<vertex-name>` <br> `vertex_type=<vertex-type>` <br> `replica=<replica-index>` <br> `partition_name=<partition-name>` | Provides the total number of messages written by UDF                                                            |
+| Metric name                                | Metric type | Labels                                                                                                                                                        | Which partition | Description                                                                                                     |
```
whynowy, line 13: What does `Which partition` mean? Is it the explanation of the `partition_name`?

pkg/udf/forward/forward.go
```diff
@@ -476,7 +478,7 @@ func (isdf *InterStepDataForward) streamMessage(ctx context.Context, dataMessage
 		return nil, fmt.Errorf("failed to applyUDF, error: %w", err)
 	}
 
-	metrics.UDFProcessingTime.With(metricLabels).Observe(float64(time.Since(start).Microseconds()))
+	metrics.ConcurrentUDFProcessingTime.With(metricLabels).Observe(float64(time.Since(start).Microseconds()))
```
whynowy, line 481: Neither of these metrics makes sense to stream mode, since we don't have a way to exclude the buffer writing time.

pkg/udf/forward/forward.go
```diff
@@ -297,13 +297,15 @@ func (isdf *InterStepDataForward) forwardAChunk(ctx context.Context) error {
...
 		// This will be a blocking call until the all the UDF results for the batch are received.
+		udfStart := time.Now()
 		udfResults, err = isdf.applyUDF(ctx, dataMessages)
 		if err != nil {
 			isdf.opts.logger.Errorw("failed to applyUDF", zap.Error(err))
 			// As there's no partial failure, non-ack all the readOffsets
 			isdf.fromBufferPartition.NoAck(ctx, readOffsets)
 			return err
 		}
+		metrics.UDFProcessingTime.With(metricLabels).Observe(float64(time.Since(udfStart).Microseconds()))
```
whynowy, line 308: should this be `ConcurrentUDFProcessingTime`?

tmenjo: It would be in batch mode, but how about in unary mode? In my understanding, calling applyUDF() will blocks until all the UDF results are received.

Is it better that we visit either concurrent or non-concurrent metric, depending on the mode?

pkg/udf/forward/forward.go
tmenjo: > we don't have a way to exclude the buffer writing time.

Yes, I know that. My idea is: how about redefining ConcurrentUDFProcessingTime as it includes both udf processing time and writing time to buffers. I described that in the metrics document, in the third commit (doc: relationship between...).

docs/operations/metrics/metrics.md
tmenjo: Here "which" means upstream or downstream, not a partition name. Sorry for the confusion.

I believe this column helps operators to understand each metric more correcly. I would choose a more appropriate word.

tmenjo:
Hello commiters, could you review again this pull request? Otherwise, I'd like to hear whether this is put off like #2624.

docs/operations/metrics/metrics.md
adarsh0728: Mostly, based on the metrics definition `which partition` (downstream/upstream) can be inferred.
For eg: [fromBufferPartition](https://github.com/numaproj/numaflow/blob/fd897c8dc9faf7db821580d3058d8fdc70c2c01b/pkg/sinks/forward/forward.go#L206) in case of read metric and [toBufferPartition](https://github.com/numaproj/numaflow/blob/fd897c8dc9faf7db821580d3058d8fdc70c2c01b/pkg/sources/forward/data_forward.go#L533) in case of write metric.

If we really want to go with this, can we remove Source from `Upstream/Source` and similarly `Sink` as those are vertices and can be confusing.

docs/operations/metrics/metrics.md
```diff
@@ -10,64 +10,65 @@ These metrics in combination can be used to determine the overall health of your
...
+| `forwarder_drop_bytes_total`               | Counter     | `pipeline=<pipeline-name>` <br> `vertex=<vertex-name>` <br> `vertex_type=<vertex-type>` <br> `replica=<replica-index>` <br> `partition_name=<partition-name>` | Downstream/Sink | Provides the total number of bytes dropped by a given Vertex due to a full Inter-Step Buffer Partition          |
+| `forwarder_udf_read_total`                 | Counter     | `pipeline=<pipeline-name>` <br> `vertex=<vertex-name>` <br> `vertex_type=<vertex-type>` <br> `replica=<replica-index>` <br> `partition_name=<partition-name>` | Upstream        | Provides the total number of messages read by UDF                                                               |
+| `forwarder_udf_write_total`                | Counter     | `pipeline=<pipeline-name>` <br> `vertex=<vertex-name>` <br> `vertex_type=<vertex-type>` <br> `replica=<replica-index>` <br> `partition_name=<partition-name>` | Upstream        | Provides the total number of messages written by UDF                                                            |
```
adarsh0728: should be Downstream?

docs/operations/metrics/metrics.md
```diff
@@ -10,64 +10,65 @@ These metrics in combination can be used to determine the overall health of your
...
+| `forwarder_ack_processing_time`                | Histogram   | `pipeline=<pipeline-name>` <br> `vertex=<vertex-name>` <br> `vertex_type=<vertex-type>` <br> `replica=<replica-index>` <br> `partition_name=<partition-name>` | Upstream/Source | Provides the histogram distribution of the processing times of ack operations                                                                                                   |
+| `forwarder_fbsink_write_processing_time`       | Histogram   | `pipeline=<pipeline-name>` <br> `vertex=<vertex-name>` <br> `vertex_type=<vertex-type>` <br> `replica=<replica-index>` <br> `partition_name=<partition-name>` | Sink            | Provides the histogram distribution of the processing times of write operations to a fallback sink                                                                              |
```
adarsh0728: `Sink` here doesn't seem to be consistent with `which partition` column discussed above

pkg/udf/forward/forward.go
adarsh0728: curious if any conclusion was reached here. But in my opinion, to avoid confusion can we use just one metric instead of two? Modes are exclusive. JFYI, In Rust we are only using `udf_processing_time`.

tmenjo: > In Rust we are only using udf_processing_time.

I never knew that, thanks. I rethink it is better to use `UDFProcessingTime` in Golang. I'll update this PR.

tmenjo:
@adarsh0728 Thank you for your comment. I reply to you all at once here.

I become aware that my commits related to `which partition` immature. So I'm removing (reverting) the commits to have this PR be more acceptable. Possibly I make another PR for the metric document some other time.

adarsh0728:
> @adarsh0728 Thank you for your comment. I reply to you all at once here.
> 
> 
> 
> I become aware that my commits related to `which partition` immature. So I'm removing (reverting) the commits to have this PR be more acceptable. Possibly I make another PR for the metric document some other time.

Thank you. Sounds good 👍

tmenjo:
Updated.

PR #2614, opened 2025-05-06
feat: exponential backoff retry strategy for sink

### LINKED ISSUE #2356: Support for exponential retries
# Summary

Numaflow currently supports retries based on fixed time intervals but does not provide the ability to configure retries with varying time windows, such as exponential backoff retries

adarsh0728:
fixes #2356


pkg/apis/numaflow/v1alpha1/retry_strategy.go
```diff
@@ -31,9 +31,9 @@ const (
 )
 
 // RetryStrategy struct encapsulates the settings for retrying operations in the event of failures.
-// It includes a BackOff strategy to manage the timing of retries and defines the action to take upon failure.
+// It includes a Exponential BackOff strategy to manage the timing of retries and defines the actions to be take upon failure.
```
whynowy, line 34: Is `be take` grammatically correct?

vigith, line 34:
```suggestion
// It includes a Exponential BackOff strategy to manage the timing of retries and defines the actions to be taken upon failure.
```

vigith: to be taken

adarsh0728: will fix this 👍

pkg/apis/numaflow/v1alpha1/const.go
```diff
@@ -259,11 +259,14 @@ const (
 
 	// Retry Strategy
 
-	// DefaultRetryInterval specifies the default time interval between retry attempts.
+	// DefaultRetryInterval specifies the default initial retry duration.
```
yhl25, line 262: `Default initial ` incase of exponential and `time interval between` in case of fixed right?

rust/numaflow-core/src/sink.rs
```diff
@@ -554,16 +558,15 @@ impl SinkWriter {
...
                     )
                     .await;
                 match status {
                     Ok(true) => break,
                     Ok(false) => {
-                        attempts += 1;
                         warn!(
                             "Retry attempt {} due to retryable error. Errors: {:?}",
-                            attempts, error_map
+                            attempts + 1,
```
yhl25, line 568: initialize attempts to 1?

rust/numaflow-core/src/sink.rs
```diff
@@ -925,6 +928,30 @@ impl SinkWriter {
                 .observe(fallback_sink_start.elapsed().as_micros() as f64);
         }
     }
+
+    fn calculate_exponential_delay<R: Rng>(
```
yhl25, line 932: Why do we need generics here? Do you have multiple implementation of Rng?

pkg/apis/numaflow/v1alpha1/const.go
adarsh0728: yes, updated the comment

yhl25 (approved):

PR #2627, opened 2025-05-12
fix: sink instantiation

cosmic-chichu:
Fix sink instantiation and add tests


rust/numaflow-core/src/config/components.rs
```diff
@@ -453,7 +454,8 @@ pub(crate) mod sink {
                             .as_ref()
                             .map(|_| Ok(SinkType::Blackhole(BlackholeConfig::default())))
                     })
-                    .or_else(|| sink.serve.as_ref().map(|_| Ok(SinkType::Serve)))
```
cosmic-chichu, line 456: im not sure about this change

cosmic-chichu: reverted this

vigith (approved):

PR #2664, opened 2025-05-20
chore: handle late data in reduce async data plane

yhl25:
Explain what this PR does.


rust/numaflow-core/src/transformer.rs
```diff
@@ -374,6 +375,7 @@ mod tests {
                 },
                 headers: Default::default(),
                 metadata: None,
+                is_late: false,
```
vigith, line 378:
```suggestion
                ..Default::default(),
```

vigith (approved):

Issue #2652, opened 2025-05-19
Serving request with X-Numaflow-Id with "." never returns


cosmic-chichu:
**Describe the bug**
A serving request with the numaflow id set to "1747681939.045399" never returns.

PR #2669, opened 2025-05-23
fix: return 400 if serving ID header contains '.'

vigith:
fixes: #2652


rust/serving/src/app.rs
```diff
@@ -199,8 +201,33 @@ async fn graceful_shutdown(handle: Handle, duration_secs: u64) {
...
+            // make sure value does not contain "."
+            if header
+                .to_str()
+                .expect("header should be a string")
+                .contains(".")
+            {
+                return Response::builder()
+                    .status(StatusCode::BAD_REQUEST)
+                    .body(Body::from(format!(
+                        "Header-ID should not container '.', found {}",
```
yhl25, line 218:
```suggestion
                        "Header-ID should not contain '.', found {}",
```

yhl25 (approved):

Issue #2675, opened 2025-05-27
Label `replica` is missing in `vertex_pending_messages`


whynowy:
**Describe the bug**

In golang code, `vertex_pending_messages` is exposed on each every replica, and does not have `replica` label, this makes no way to aggregate the correct pending messages when there are multiple partitions.

```golang
	pending = promauto.NewGaugeVec(prometheus.GaugeOpts{
		Name: VertexPendingMessages,
		Help: "Average pending messages in the last period of seconds. It is the pending messages of a vertex, not a pod.",
	}, []string{LabelPipeline, LabelVertex, LabelPeriod, LabelPartitionName})
```

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

>For quick help and support, join our [slack](https://numaproj.slack.com/join/shared_invite/zt-19svuv47m-YKHhsQ~~KK9mBv1E7pNzfg#/shared-invite/email) channel.

PR #2676, opened 2025-05-27
chore: remove pod as dimension for `vertex_pending_messages` metric

adarsh0728:
**Merge this PR only after #2675 is closed.**

- pending is calculated only for replica 0. Added replica as required filter for `vertex_pending_messages` metric
- removed `pod` as dimension


adarsh0728:
Since we do not have replica label in Golang based code, closing this for now. Can be taken up as config changes.

PR #2687, opened 2025-05-28
feat: extension for http source

vigith:
fixes: #2683


rust/extns/numaflow-http/src/lib.rs
```diff
@@ -0,0 +1,528 @@
...
+            header_map.insert(key.to_string(), value_str.to_string());
+        } else {
+            warn!(?key, "Skipping header with invalid UTF-8");
+        }
+    }
+
+    // Generate or extract X-Numaflow-Id
+    let id = header_map
+        .get("x-numaflow-id")
+        .or_else(|| header_map.get("X-Numaflow-Id"))
```
yhl25, line 260: can be removed? since its case insensitive.

yhl25 (approved):

Issue #2643, opened 2025-05-16
Allow setting kafka options using config section


BulkBeing:
# Summary

The `config` section in [Kafka source spec](https://numaflow.numaproj.io/user-guide/sources/kafka/#option-2-built-in-kafka-source) allows specifying kafka options supported by Golang's Sarama library. The equivalent ones should be supported in the Rust implementation.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #2689, opened 2025-05-30
feat: Allow specifying JVM Kafka opts directly

BulkBeing:
closes https://github.com/numaproj/numaflow/issues/2643

### DIFF
--- a/rust/extns/numaflow-kafka/src/source.rs
+++ b/rust/extns/numaflow-kafka/src/source.rs
@@ -29,7 +29,7 @@ pub struct KafkaSourceConfig {
     pub tls: Option<TlsConfig>,
     /// Any supported kafka client configuration options from
     /// https://docs.confluent.io/platform/current/clients/librdkafka/html/md_CONFIGURATION.html
-    pub kafka_raw_config: Option<HashMap<String, String>>,
+    pub kafka_raw_config: HashMap<String, String>,
 }
 
 /// Message represents a message received from Kafka which can be converted to Numaflow Message.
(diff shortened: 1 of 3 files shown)


rust/extns/numaflow-kafka/src/sink.rs
```diff
@@ -29,6 +30,9 @@ pub struct KafkaSinkConfig {
     /// which is useful for maintaining message ordering for related messages.
     /// When set to false, messages will be distributed across partitions randomly.
     pub set_partition_key: bool,
+    /// Any supported kafka client configuration options from
+    /// https://docs.confluent.io/platform/current/clients/librdkafka/html/md_CONFIGURATION.html
+    pub kafka_raw_config: Option<HashMap<String, String>>,
```
vigith, line 35: I don't think this need to be `Option`, we can iterate on an empty `HashMap`?

rust/extns/numaflow-kafka/src/sink.rs
```diff
@@ -63,9 +67,23 @@ pub struct KafkaSinkMessage {
 pub fn new_sink(config: KafkaSinkConfig) -> crate::Result<KafkaSink> {
     let mut client_config = ClientConfig::new();
     client_config
-        .set("bootstrap.servers", config.brokers.join(","))
         .set("message.timeout.ms", "5000")
-        .set("client.id", "numaflow-kafka-sink")
+        .set("client.id", "numaflow-kafka-sink");
+    if let Some(kafka_raw_config) = config.kafka_raw_config {
```
vigith, line 72: we really don't require `if let Some()...`, iterate on empty is fine

rust/extns/numaflow-kafka/src/source.rs
```diff
@@ -27,6 +27,9 @@ pub struct KafkaSourceConfig {
     pub auth: Option<KafkaSaslAuth>,
     /// The TLS configuration for the Kafka consumer.
     pub tls: Option<TlsConfig>,
+    /// Any supported kafka client configuration options from
+    /// https://docs.confluent.io/platform/current/clients/librdkafka/html/md_CONFIGURATION.html
+    pub kafka_raw_config: Option<HashMap<String, String>>,
```
vigith, line 32: no need of `Option`

BulkBeing: Updated

vigith (approved):

PR #2700, opened 2025-06-04
fix: Rust HTTP source - More validations on HTTP headers

BulkBeing:
- Return error if `x-numaflow-id` or `x-numaflow-event-time` is set by user and is not a valid value. Else generate default values.
- Do not forward `x-numaflow-event-time` ([current Go implementation](https://github.com/numaproj/numaflow/blob/2cab60c2a1ddde0f0272b6570144071a49c4e94b/pkg/sources/http/http.go#L146))  and `Authorization` header.
- Fix `x-numaflow-event-time`. The time will be specified by user as epoch milliseconds https://numaflow.numaproj.io/user-guide/sources/http/#x-numaflow-event-time


vigith (approved):

vigith:
@BulkBeing The codecov is dropping for this PR

rust/extns/numaflow-http/src/lib.rs
```diff
@@ -413,39 +413,54 @@ async fn health_handler() -> impl IntoResponse {
 /// Data ingestion endpoint handler
 async fn data_handler(
     State(http_source): State<HttpState>,
-    headers: HeaderMap,
+    mut headers: HeaderMap,
     body: Bytes,
 ) -> impl IntoResponse {
+    // Generate or extract X-Numaflow-Id
+    let id = match headers.get("x-numaflow-id") {
```
adarsh0728, line 420: have we considered the case where a key can have multiple values? ( comma separated)

BulkBeing: Client can specify multiple `x-numaflow-id` headers. We only use the first value.

rust/extns/numaflow-http/src/lib.rs
```diff
@@ -413,39 +413,54 @@ async fn health_handler() -> impl IntoResponse {
...
     let mut header_map = HashMap::new();
+    // Ensure X-Numaflow-Id is in the headers
+    header_map.insert("X-Numaflow-Id".to_string(), id.clone());
 
     for (key, value) in headers.iter() {
         if let Ok(value_str) = value.to_str() {
             header_map.insert(key.to_string(), value_str.to_string());
         } else {
-            warn!(?key, "Skipping header with invalid UTF-8");
+            warn!(?key, "Skipping header with invalid ASCII characters");
```
adarsh0728: nit: should we log key and value both here?

rust/extns/numaflow-http/src/lib.rs
```diff
@@ -494,6 +509,69 @@ async fn data_handler(
...
+            axum::Json(serde_json::json!({
+                "error": "Value of 'x-numaflow-id' header contains non-printable ASCII characters",
+            })),
+        )
+    })?;
+    Ok(id.to_string())
+}
+
+fn parse_event_time_from_header(
+    eventtime_header_value: &HeaderValue,
```
adarsh0728: nit: `event_time_header_value` more apt, in sync with other variables with event time like `event_time_millis`

adarsh0728 (approved):

PR #2701, opened 2025-06-04
chore: infinite retry in source if ack fails

yhl25:
* implements infinite retry while doing source ack

### DIFF
--- a/rust/numaflow-core/src/pipeline/isb/jetstream/reader.rs
+++ b/rust/numaflow-core/src/pipeline/isb/jetstream/reader.rs
@@ -306,10 +306,10 @@ impl JetStreamReader {
         Ok((ReceiverStream::new(messages_rx), handle))
     }
 
-    // Intended to be run as background task which waits for the ack and continuously sends wip acks
-    // until the final ack/nak is received. We will continuously retry if there is an error in acknowledging
-    // the message as work-in-progress. If the sender end of the ack_rx channel was dropped before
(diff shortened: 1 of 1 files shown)


rust/numaflow-core/src/pipeline/isb/jetstream/reader.rs
```diff
@@ -306,10 +306,11 @@ impl JetStreamReader {
...
     }
 
-    // Intended to be run as background task which will continuously send InProgress acks to Jetstream.
-    // We will continuously retry if there is an error in acknowledging the message as work-in-progress.
-    // If the sender end of the ack_rx channel was dropped before sending a final Ack or Nak (due to some unhandled/unknown failure), we will send a Nak to Jetstream.
-    async fn start_work_in_progress(
+    // Intended to be run as background task which waits for the ack and continuously sends wip acks
+    // until the final ack/nak is received. We will continuously retry if there is an error in acknowledging
+    // the message as work-in-progress. If the sender end of the ack_rx channel was dropped before
+    // sending a final Ack or Nak (due to some unhandled/unknown failure), we will send a Nak to Jetstream.
```
vigith:
```suggestion
    /// A background task which waits for the `Ack`, meanwhile it continuously sends `WIP` acks
    /// until the final ack/nak is received. This will continuously retry if there is an error in acknowledging.
    /// If the sender's end of the ack_rx channel was dropped before
    /// sending the final `Ack` or `Nak` (due to some unhandled/unknown failure), we will send `Nak` to Jetstream.
```

vigith (approved):

PR #2717, opened 2025-06-10
[WIP] feat: Kerberos keytab/password auth

BulkBeing:
Explain what this PR does.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -74,6 +74,8 @@ COPY --from=base /bin/numaflow /bin/numaflow
 COPY --from=base /bin/numaflow-rs /bin/numaflow-rs
 COPY ui/build /ui/build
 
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -74,6 +74,8 @@ COPY --from=base /bin/numaflow /bin/numaflow
 COPY --from=base /bin/numaflow-rs /bin/numaflow-rs
 COPY ui/build /ui/build
 
+RUN apt update && apt install -y krb5-user libpam-krb5
```
vigith, line 77: this might bring in glibc which will create a lot of CVEs. perhaps, we can also give a non-scratch image too?

whynowy: Think about if we should run them in one container with the same image?

PR #2728, opened 2025-06-16
chore: upgrade nats server to 2.10.29

vigith:
upgrade nats-server to 2.10.29 (from 2.10.17)


KeranYang (approved):

KeranYang:
I took a look at our previous upgrade, it usually followed with a go.mod change to upgrade client. Not sure if we want to do that as well. @vigith

PR #2750, opened 2025-06-24
chore: deprecate transformer builtins

adarsh0728:
- controller changes
- doc changes
- removed e2e tests specific to transformer builtins
- removed builtin files


vigith (commented):
Can you fix the examples too?

adarsh0728:
> Can you fix the examples too?

I don't think any example is using builtin transformer. I have removed e2e tests specific to transformer builtins. There is one idle source kafka e2e test (TestIdleKeyedReducePipelineWithKafkaSource)which uses builtin and it fails, but shows as passed in CI, will check that and update.

vigith (approved):

yhl25 (approved):

PR #2752, opened 2025-06-25
chore: deprecate udf(map) builtins

adarsh0728:
- controller changes to remove udf builtins
-  docs changes
- change in examples
   - replaced builtin cat with image `quay.io/numaio/numaflow-go/map-cat:stable`, implemented here https://github.com/numaproj/numaflow-go/tree/main/pkg/mapper/examples/cat
   - replaced builtin `filter (int(json(payload).id) < 100)` with `quay.io/numaio/numaflow-go/map-filter-id:stable`
- removed builtin files
- fixed unit tests
- fixed e2e tests
   - replaced `builtin filter (expression: int(json(payload).id) < 100 && json(payload).msg == 'hello' && sprig.contains('good', string(json(payload).desc)))` with `quay.io/numaio/numaflow-go/map-filter:stable`, implemented here https://github.com/numaproj/numaflow-go/tree/main/pkg/mapper/examples/filter


yhl25 (approved):
LGTM!

PR #2756, opened 2025-06-26
fix: set childrenHealthStatus of external redis isbsvc

saty9:
External redis isbsvc resources had condition 
```
Type: childrenHealthStatus
Status: Unknown
```
this made it impossible to use it as pipelines would fail to deploy saying the isbsvc was unhealthy. 

This marks it's children as healthy when checked.


vigith:
We are planning to decommission redis ISB from 1.6 release. We started with Redis Streams as ISB and it gave us lots of problems (data inconsistency issues). OTOH, Jetstream has been quite stable for last two years (if provisioned correctly). We do have plans to revisit ISB in the future.

For now, I am closing this issue.

PR #2761, opened 2025-06-29
fix(UI): health filter & support for mono-vertex health on pipeline listing page

adarsh0728:


veds-g:
instead of calling the health apis for all the pipelines, you can use the existing logic.
Create a state in parent component like `const [plHealthMap, setPlHealthMap]` in `PipelineListing.tsx` and pass it to the child component (`PipelineCard.tsx`), which updates the state based on the `usePipelineHealthFetch` hook.
This way we only call the health status for the pipelines currently in view.

veds-g (approved):

PR #2768, opened 2025-07-02
chore: Upgrade to Rust 1.88

BulkBeing:
- Rust 1.88 upgrade
- Move common dependencies to workspace `Cargo.toml` file
- Upgrade minor/patch versions for dependencies
- Updated code with if-let chains

I didn't find any `while let` followed by `if` statement that we could replace.

Note that the conditional chaining breaks the loop. So below two are NOT equivalent:
```rust
let v = &[Some(1), Some(20), None, Some(3), None, Some(300)];
let mut iter = v.iter();

// Iterates over all items and prints: 1, 20, 3
while let Some(item) = iter.next() {
    if let Some(n) = item {
        if *n < 250 {
            println!("{n}");
        }
    }
}

// Stops at first None (due to second condition here), only prints 1, 20
while let Some(item) = iter.next()
    && let Some(n) = item
    && *n < 250
{
    println!("{n}");
}
```


rust/extns/numaflow-sqs/src/lib.rs
```diff
@@ -92,10 +92,10 @@ pub async fn create_sqs_client(config: SqsConfig) -> Result<Client, Error> {
...
 
     // Apply endpoint URL if configured for source
-    if let SqsConfig::Source(cfg) = &config {
-        if let Some(endpoint_url) = &cfg.endpoint_url {
-            config_builder = config_builder.endpoint_url(endpoint_url);
-        }
+    if let SqsConfig::Source(cfg) = &config
+        && let Some(endpoint_url) = &cfg.endpoint_url
+    {
+        config_builder = config_builder.endpoint_url(endpoint_url);
```
yhl25, line 98: `If let` chaining is nice :).

Issue #2740, opened 2025-06-19
Implement SideInput Manager


vigith:
1. Connects to NATS JetStream for storage
2. Establishes gRPC connection to side input server
3. Sets up cron scheduler based on trigger configuration
4. Periodically calls the side input server to retrieve data
5. Stores the data in the KV store (unless NoBroadcast flag is set)
6. Makes the data available to other pipeline components

vigith:
https://docs.rs/cron/latest/cron/ seems like a decent simple module, but it doesn't support 5 `*` cron format.

https://www.quartz-scheduler.org/documentation/quartz-2.3.0/tutorials/crontrigger.html

PR #2769, opened 2025-07-02
chore: write side-input manager in Rust

vigith:
Fixes: #2740


rust/numaflow-sideinput/src/manager.rs
```diff
@@ -0,0 +1,282 @@
...
+        &mut self,
+        bucket: jetstream::kv::Store,
+        side_input_trigger: SideInputTrigger,
+    ) {
+        // do the first run before running the schedule
+        info!("Running first side-input generation");
+        if let Err(e) = self.generate_side_input(&bucket).await {
+            error!(?e, "Failed to generate the first, initial side input");
+        }
+
```
yhl25, line 115: Do we need to differentiate between the first and subsequent generations?

vigith: if this is a daily cron, then it might not kickoff, hence does an initial run.

PR #2776, opened 2025-07-06
fix: allow subcommands other than monitor and serving in cli

adarsh0728:


rust/numaflow/src/cmdline.rs
```diff
@@ -8,6 +8,7 @@ pub(super) fn root_cli() -> Command {
         .arg_required_else_help(true)
         .subcommand(add_monitor_subcommand())
         .subcommand(add_serving_subcommand())
+        .allow_external_subcommands(true)
```
vigith, line 12:
```suggestion
        // TODO: remove after moving e2e to Rust, this is a hack
        .allow_external_subcommands(true)
```

vigith (approved):

PR #2786, opened 2025-07-07
chore: side-input cli parsing and config (#2785)

vigith:
CLI and Config parsing


rust/numaflow/src/main.rs
```diff
@@ -31,6 +36,17 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+    let shutdown_cln_token = cln_token.clone();
+
+    // wait for SIG{INT,TERM} and invoke cancellation token.
+    let shutdown_handle: JoinHandle<Result<(), Box<dyn Send + Sync + Error>>> =
+        tokio::spawn(async move {
+            shutdown_signal().await;
+            shutdown_cln_token.cancel();
+            Ok(())
+        });
+
```
yhl25, line 49: if we want listen on the sigterm here, we need to follow the same for others. Otherwise there will be duplicate shutdown handles.

rust/numaflow/src/cmdline/sideinput.rs
```diff
@@ -0,0 +1,192 @@
...
+                .collect();
+            let side_input_store = args
+                .get_one::<String>("side-inputs-store")
+                .expect("side-inputs-store is required");
+            let side_input_store = Box::leak(side_input_store.clone().into_boxed_str());
+
+            let mode = SideInputMode::Synchronizer {
+                side_inputs,
+                side_input_store,
+            };
```
yhl25: Do we need to have separate modes for `Synchronizer` and `Initializer`? Can it be merged by using an `execute_once` flag inside the mode enum?

vigith: good idea

rust/numaflow/src/main.rs
vigith: what do you mean?

PR #2797, opened 2025-07-11
fix: kafka writes messages to same partition when setKey is not set

yhl25:


pkg/sinks/kafka/kafka.go
```diff
@@ -180,7 +180,7 @@ func (tk *ToKafka) Write(_ context.Context, messages []isb.Message) ([]isb.Offse
 			}
 		}
 
-		var kafkaKey sarama.StringEncoder
+		var kafkaKey sarama.Encoder
```
vigith, line 184:
```suggestion
		// all that ends well with `'er` are not interfaces :-) 
		var kafkaKey sarama.Encoder
```

vigith (approved):

PR #2804, opened 2025-07-14
feat: nats source in rust

adarsh0728:
- [x] Nats source implementation in rust
- [x] Nats E2E test migration to rust
- [x] Unit tests
- [x] Refactor extn `numaflow-jetstream`

### DIFF
--- a/rust/extns/numaflow-nats/src/nats.rs
+++ b/rust/extns/numaflow-nats/src/nats.rs
@@ -8,25 +8,6 @@ use tokio_stream::StreamExt;
 use tracing::debug;
 use uuid::Uuid;
 
-#[cfg(test)]
-fn make_conn_opts() -> ConnectOptions {
-    ConnectOptions::new()
-        .max_reconnects(1)
-        .reconnect_delay_callback(|_| std::time::Duration::from_millis(10))
-        .ping_interval(Duration::from_secs(1))
-}
-
-#[cfg(not(test))]
-fn make_conn_opts() -> ConnectOptions {
-    ConnectOptions::new()
-        .max_reconnects(None) // unlimited reconnects
-        .reconnect_delay_callback(|attempts| {
-            std::time::Duration::from_millis(std::cmp::min((attempts * 10) as u64, 1000))
-        })
-        .ping_interval(Duration::from_secs(3))
-        .retry_on_initial_connect() // run in background
-}
-
 #[derive(Debug, Clone, PartialEq)]
 pub struct NatsSourceConfig {
     pub addr: String,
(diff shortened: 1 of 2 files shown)


rust/numaflow-core/src/config/components.rs
```diff
@@ -354,6 +357,139 @@ pub(crate) mod source {
...
+        ) -> std::result::Result<Self, Self::Error> {
+            let auth: Option<NatsAuth> = match value.auth {
+                Some(auth) => {
+                    if let Some(basic_auth) = auth.basic {
+                        let user_secret_selector = &basic_auth.user.ok_or_else(|| {
+                            Error::Config("Username can not be empty for basic auth".into())
+                        })?;
+                        let username = crate::shared::create_components::get_secret_from_volume(
+                            &user_secret_selector.name,
+                            &user_secret_selector.key,
```
yhl25: refactor this to avoid duplicate in jetstream and nats?

rust/numaflow-core/src/source/nats.rs
```diff
@@ -0,0 +1,78 @@
...
+impl From<NatsMessage> for Message {
+    fn from(message: NatsMessage) -> Self {
+        let offset = Offset::String(StringOffset::new(message.id.clone(), *get_vertex_replica()));
+        Message {
+            typ: Default::default(),
+            keys: Arc::from(vec![]),
+            tags: None,
+            value: message.value,
+            offset: offset.clone(),
+            event_time: message.event_time,
```
yhl25, line 29: check if the event time can be extracted from the nats payload

yhl25 (approved):

rust/numaflow-core/src/source/nats.rs
adarsh0728: checked, unlike jetstream's info() method which has the metadata, core NATS messages do not include publish time or metadata. They only have the payload, subject, headers, etc.

yhl25:
Make sure the code coverage is > 85%

adarsh0728:
> Make sure the code coverage is > 85%

yeah, added some more tests. The coverage is not yet updated. Will look

rust/extns/numaflow-nats/src/nats.rs
```diff
@@ -0,0 +1,277 @@
...
+                break;
+            }
+            tokio::select! {
+                biased;
+
+                _ = &mut timeout => {
+                    tracing::debug!(msg_count = messages.len(), "Timed out waiting for NATS messages");
+                    break;
+                }
+                maybe_msg = self.sub.next() => {
```
vigith, line 134: should we do batch read? even in https://github.com/numaproj/numaflow/blob/3a82110b1ad4a5e50e1bb5bddfdfeb7e7c9de43f/rust/extns/numaflow-nats/src/jetstream.rs#L243 this could cause duplicates if there is backpressure and writes and not going through. 

This is because async-nats lib is will read ahead and the ack timeout will be triggered causing replay.

vigith (approved):

Issue #2802, opened 2025-07-14
aws-smithy-mocks-experimental is deprecated, sqs extension should use the stable aws-smithy-mocks


yhl25:
```
warning: use of deprecated method `aws_smithy_mocks_experimental::MockResponseInterceptor::with_rule`: The `aws-smithy-mocks-experimental` crate is now deprecated and is replaced by the `aws-smithy-mocks` crate. Please migrate to the non-experimental crate.
Warning:    --> extns/numaflow-sqs/src/source.rs:622:14
```

Issue #2722, opened 2025-06-12
Configurability of Jetstream Consumer


Koalk:
# Summary

Expose some configuration properties of the Jetstream consumer created_or_updated by the Jetstream source. It could be just the name of the consumer/durable so you can create it separatedly and just point the source to it. The other option would be making the name of the consumer deterministic and documented but that could be more cumbersome.

# Use Cases

A dead letter process for the streaming into numaflow needs some non-default configurations.
Filtering of the messages in the stream needs extra configuration of the consumer.
Fine tuning based on your throughput.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

nfrushQH:
Another good option to expose would be `FilterSubjects` for the stream you are connecting the consumer to, since some folks may have a bigger stream but only want to operate on a subset of subjects.

vigith:
> Another good option to expose would be `FilterSubjects` for the stream you are connecting the consumer to, since some folks may have a bigger stream but only want to operate on a subset of subjects.

@nfrush  I have opened #2819 to track your ask. We already have #2805 for the original ask.

PR #2805, opened 2025-07-14
fix: Jetstream source - Option to specify consumer name

BulkBeing:
Fixes: #2722 

- Currently there is no option to specify a consumer name. This PR adds that option
- Golang implementation creates a consumer with the naming format: `numaflow-<pipeline_name>-<vertex_name>-<stream_name>`
- The Rust implementation assumed that the consumer will be created by the user. So e2e tests were failing.

Also closes https://github.com/numaproj/numaflow/issues/2802

---
UPDATE: Added user option to specify a consumer name. If not specified, the default naming convention will be used to generate one. In both cases, first we will check if the consumer exists on the stream, if not, we will attempt to create it.


whynowy:
Why don't do the same logic as golang in rust? The logic of `create_or_update_consumer` in both languages will update the consumer settings if there's any difference.

BulkBeing:
> Why don't do the same logic as golang in rust? The logic of `create_or_update_consumer` in both languages will update the consumer settings if there's any difference.

I was not sure what approach to take since we don't create the consumer for other sources like Kafka. I will update the code to do the same as Golang - remove the option to specify consumer name (added in this PR), create a consumer in Rust with the naming format `<pipeline_name>-<vertex_name>-<stream_name>`

vigith:
> I was not sure what approach to take since we don't create the consumer for other sources like Kafka. I will update the code to do the same as Golang - remove the option to specify consumer name (added in this PR), create a consumer in Rust with the naming format `<pipeline_name>-<vertex_name>-<stream_name>`

What if the user wants to change the consumer-name, the way we do today in Kafka? So I think it is okay to keep it I, if user doesn't give one, use the generate one?

vigith (approved):
looks like this will close https://github.com/numaproj/numaflow/issues/2802 too.. if yes, please link it.

rust/extns/numaflow-jetstream/src/lib.rs
```diff
@@ -59,11 +60,32 @@ pub struct TlsClientAuthCerts {
...
 
+/// Consumer name field is optional in the spec.
+/// If not provided, we will use a default name in the format of:
+///     - `numaflow-{pipeline_name}-{vertex_name}-{stream_name}` for pipeline
+///     - `numaflow-{pipeline_name}mvtx-{stream_name}` for monovertex
+/// If the user specifies a consumer name, we expect the consumer to already exist in the stream.
+/// If the user does not specify a consumer name, we will create a new consumer on the stream.
+#[derive(Debug, Clone, PartialEq)]
+pub enum JetstreamConsumerName {
+    UserSpecified(String),
```
yhl25: use `&'static str`?

yhl25 (approved):

rust/extns/numaflow-jetstream/src/lib.rs
yhl25: Is it a good idea to use an enum with a value `Default`? Is it better to use a struct with default implementaiton?

BulkBeing: Making `&'static str` won't have much impact, these are used once and passed it to the function in `async-nats` for establishing connection.
We need the info whether the user set the consumer name or not in the background actor that connects to Nats. This is used to decide between whether we should call the `create_or_update_consumer_on_stream` API or just `get_consumer_from_stream`.

yhl25 (changes requested):
Lets keep it simple:
* If the consumer name is empty, use a default generated one.
* If the consumer is not found, create a new one.

whynowy:
> Lets keep it simple:
> 
> * If the consumer name is empty, use a default generated one.
> * If the consumer is not found, create a new one.

This is not good, it covers the real problem.

yhl25:
> > Lets keep it simple:
> > 
> > * If the consumer name is empty, use a default generated one.
> > * If the consumer is not found, create a new one.
> 
> This is not good, it covers the real problem.

It is similar to `create_or_update_consumer` method right?

whynowy:
> > Lets keep it simple:
> 
> > 
> 
> > * If the consumer name is empty, use a default generated one.
> 
> > * If the consumer is not found, create a new one.
> 
> 
> 
> This is not good, it covers the real problem.

Ignore it.

hasmukhmparmar-27:
Do we require customers to follow the same naming convention, "numaflow-{pipeline_name}-{vertex_name}-{}"?

BulkBeing:
> Do we require customers to follow the same naming convention, "numaflow-{pipeline_name}-{vertex_name}-{}"?

No. User may specify any name. The field is optional. If they do not specify one, we generate a name in this format.

hasmukhmparmar-27:
> No. User may specify any name. The field is optional. If they do not specify one, we generate a name in this format.

It is good practice to follow same naming conventions to avoid confusion and help with debugging.

BulkBeing:
> It is good practice to follow same naming conventions to avoid confusion and help with debugging.

In this case, the name format really doesn't matter since we attempt to create it if it doesn't exist (and its a one time action). 
We use this naming format so that the name is some what unique (user may have multiple pipelines listening to same stream, or multiple jetstream source vertices in same pipeline).

rust/numaflow-core/src/config/monovertex.rs
```diff
@@ -122,14 +122,21 @@ impl MonovertexConfig {
             .clone()
             .ok_or_else(|| Error::Config("Source not found".to_string()))?;
 
+        let mut source_type: SourceType = source.try_into()?;
+        if let SourceType::Jetstream(ref mut js_config) = source_type {
+            if js_config.consumer.is_empty() {
+                js_config.consumer = format!("numaflow-{}-{}", mono_vertex_name, &js_config.stream);
+            }
+        }
```
yhl25: Can this be avoided?

rust/numaflow-core/src/config/pipeline.rs
```diff
@@ -401,14 +402,24 @@ impl PipelineConfig {
...
+            let mut source_type: SourceType = source.try_into()?;
+            if let SourceType::Jetstream(ref mut js_config) = source_type {
+                if js_config.consumer.is_empty() {
+                    js_config.consumer = format!(
+                        "numaflow-{pipeline_name}-{vertex_name}-{}",
+                        js_config.stream
+                    );
+                }
+            }
+
```
yhl25: same comment.

PR #2807, opened 2025-07-15
chore: move server_info from core to a new crate

vigith:
Side-Input supports UDFs thus have a dependency on SDK. To validate the SDK, we need to parse the server-info file. Since side-input is a separate crate, we need to move this server-info parser out.

In future we can add other shared code in here, e.g., parsing ISB config, which is required by `serving`, `core`, and `side-input`.


rust/numaflow-core/src/error.rs
```diff
@@ -75,3 +75,11 @@ pub enum Error {
     #[allow(clippy::upper_case_acronyms)]
     WAL(String),
 }
+
+impl From<numaflow_shared::error::Error> for Error {
+    fn from(value: numaflow_shared::error::Error) -> Self {
+        match value {
+            numaflow_shared::error::Error::ServerInfo(e) => Error::ServerInfo(e),
```
yhl25: Error could be named `shared`?

vigith: will do

yhl25 (approved):

PR #2827, opened 2025-07-21
fix: aligned and unaligned reduce in Rust runtime

yhl25:
* Update last publish offset to larger value even when the watermark is -1 to avoid watermark regression
* Fix session merge edge cases
* Add keys to message id to avoid duplicates in session windows
* Refactored aligned and unaligned reducer to make it more readable
* Close windows when there is no input data(idling)

### DIFF
--- a/rust/numaflow-core/src/reduce/reducer/unaligned/windower/session.rs
+++ b/rust/numaflow-core/src/reduce/reducer/unaligned/windower/session.rs
@@ -122,7 +122,7 @@ impl SessionWindowManager {
     /// Windows must overlap first (considering end time is exclusive) before they can be expanded.
     fn expand_window_if_needed(new_window: &Window, existing_window: &Window) -> Option<Window> {
         // First check if windows overlap (end time is exclusive, start time is inclusive)
-        // [a,b) and [c,d) overlap if a < c and c < b
+        // [a,b) and [c,d) overlap if a < d and c < b
         let windows_overlap = new_window.start_time < existing_window.end_time
             && existing_window.start_time < new_window.end_time;
 
@@ -130,7 +130,16 @@ impl SessionWindowManager {
             return None;
         }
 
-        // If windows overlap, always expand to accommodate both windows
+        // Check if the new window is completely contained within the existing window
+        // If so, this should be an append operation, not an expand operation
+        let new_window_contained = new_window.start_time >= existing_window.start_time
+            && new_window.end_time <= existing_window.end_time;
+
+        if new_window_contained {
+            return None; // No expansion needed, should append to existing window
+        }
+
+        // If windows overlap and expansion is needed, expand to accommodate both windows
         // Use min of start times and max of end times
         Some(Window::new(
             new_window.start_time.min(existing_window.start_time),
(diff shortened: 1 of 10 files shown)


rust/numaflow-core/src/watermark/isb/wm_publisher.rs
```diff
@@ -184,8 +185,8 @@ impl ISBWatermarkPublisher {
         // Supposed publish watermark offset=3605637 watermark=1750758998480 last_published_offset=3605147 last_published_watermark=1750758997480
         // Actual published watermark offset=3605637 watermark=1750758998480
         // We should've published watermark for offset 3605646 and skipped publishing for offset 3605637
-        if watermark == last_state.watermark {
-            last_state.offset = offset;
+        if watermark == -1 || watermark == last_state.watermark {
```
vigith, line 189:
```suggestion
        // if watermark cannot be computed, still we should publish the last known valid WM for the latest offset
        if watermark == -1 || watermark == last_state.watermark {
```

rust/numaflow-core/src/reduce/reducer/unaligned/windower/session.rs
```diff
@@ -118,27 +116,41 @@ impl SessionWindowManager {
...
-            || new_window.end_time > existing_window.end_time)
-            .then(|| {
-                Window::new(
-                    new_window.start_time.min(existing_window.start_time),
-                    new_window.end_time.max(existing_window.end_time),
-                    Arc::clone(&new_window.keys),
-                )
-            })
+        // First check if windows overlap (end time is exclusive, start time is inclusive)
+        // [a,b) and [c,d) overlap if a < d and c < b
```
vigith, line 125:
```suggestion
        // [a,b) and [c,d) overlap if a < c and c < b
```

let's c < d, this is given

rust/numaflow-core/src/reduce/reducer/unaligned/windower/session.rs
```diff
@@ -118,27 +116,41 @@ impl SessionWindowManager {
...
     fn find_window_to_merge(window_set: &BTreeSet<Window>, window: &Window) -> Option<Window> {
+        // TODO: this is not efficient, since its sorted we should be able to find the window in O(log n) time.
         window_set
             .iter()
             .find(|existing_window| {
-                // Windows overlap if they intersect
-                window.start_time <= existing_window.end_time
-                    && existing_window.start_time <= window.end_time
+                // Windows overlap if they intersect (end time is exclusive, start time is inclusive)
+                // [a,b) and [c,d) overlap if a < d and c < b
```
vigith, line 160: same

rust/numaflow-core/src/reduce/reducer/unaligned/windower/session.rs
```diff
@@ -211,42 +224,86 @@ impl SessionWindowManager {
         &self,
         key: &str,
         closing_group: Vec<Window>,
-    ) -> Option<UnalignedWindowMessage> {
-        if closing_group.is_empty() {
-            return None;
+    ) -> Option<Vec<UnalignedWindowMessage>> {
+        match closing_group.len() {
+            0 => None,
+            1 => self.process_single_closing_window(key, closing_group.into_iter().next().unwrap()),
```
vigith: merge or comment why it should be handled separately.

vigith (approved):

rust/numaflow-core/src/reduce/reducer/unaligned/windower/session.rs
yhl25: merged them

PR #2832, opened 2025-07-23
fix: handle stream already existing when making consumer

saty9:
We had been experiencing issues where sometimes the ISBSvc was not initialised correctly for a pipeline.

This changes initialisation to not assume consumers exist if streams exist.


saty9:
@vigith are you able to merge this now?

vigith:
> @vigith are you able to merge this now?

done. thank you for the contribution 🎉

PR #2833, opened 2025-07-23
feat: enable rust runtime by default

yhl25:
* Moved watermark publishing to tracker
* Watermark computation logic is moved from the handler to the actor to avoid race condition
* Http source default timeout is reduced from 1s to 5ms.
* Also this makes sure that the watermark never regresses inside a vertex (added a warn log to capture these cases).


test/diamond-e2e/testdata/cycle-backward.yaml
```diff
@@ -5,44 +5,28 @@ metadata:
...
     - name: cat
-      containerTemplate:
-        env:
-          - name: NUMAFLOW_RUNTIME
-            value: "rust"
       udf:
         container:
           image: quay.io/numaio/numaflow-go/map-cat:stable # A UDF which simply cats the message
-          imagePullPolicy: Always
+          imagePullPolicy: IfNotPresent
```
whynowy, line 14: Why update them?

yhl25: To avoid downloading some images again and again, I feel like the github action has already reached the limit.

pkg/apis/numaflow/v1alpha1/udf.go
```diff
@@ -38,6 +38,8 @@ func (in UDF) getContainers(req getContainerReq) ([]corev1.Container, []corev1.C
 }
 
 func (in UDF) getMainContainer(req getContainerReq) corev1.Container {
+	// TODO: Default runtime is rust in 1.6, we will remove this env in 1.7
+	req.env = append(req.env, corev1.EnvVar{Name: EnvNumaflowRuntime, Value: "rust"})
```
vigith, line 42: what will be the procedure to switch to golang binary at runtime?

rust/numaflow-core/src/pipeline/isb/jetstream/writer.rs
```diff
@@ -602,16 +602,11 @@ impl JetstreamWriter {
...
 
-            if let Some(watermark_handle) = this.watermark_handle.as_mut() {
-                // now the pafs have resolved, lets use the offsets to send watermark
-                for (stream, offset) in offsets {
-                    JetstreamWriter::publish_watermark(watermark_handle, stream, offset, &message)
-                        .await;
-                }
-            }
+            // Delete from tracker with optional watermark publishing
+            let watermark_info = this.watermark_handle.as_ref().map(|_| offsets);
```
vigith, line 607: this can be optional because of Watermark being turned off, right?

rust/numaflow-core/src/watermark/source/source_wm_publisher.rs
```diff
@@ -66,6 +66,12 @@ impl SourceWatermarkPublisher {
             self.publishers.insert(processor_name.clone(), publisher);
         }
 
+        // subtract the max delay from the watermark, since we are publishing from source itself
+        // if the watermark is not idle.
```
vigith, line 70: this is because idle publishing happens independently

test/diamond-e2e/testdata/cycle-backward.yaml
vigith: e2e tests are more stable now

vigith (approved):

rust/numaflow-core/src/pipeline/isb/jetstream/writer.rs
yhl25: yes

pkg/apis/numaflow/v1alpha1/udf.go
yhl25: if the users set that env to golang, it will override it.

rust/numaflow-core/src/watermark/source/source_wm_publisher.rs
yhl25: yes.

PR #2835, opened 2025-07-23
fix: add serving pipeline transformer config

cosmic-chichu:
Add transformer configs for ServingPIpeline


docs/user-guide/reference/kustomize/numaflow-transformer-config.yaml
```diff
@@ -22,6 +22,22 @@ images:
     kind: MonoVertex
   - path: spec/sink/fallback/udsink/container/image
     kind: MonoVertex
+  - path: spec/pipeline/vertices/udf/container/image
```
whynowy, line 25: All the pipeline related config needs to be copied.

whynowy (approved):

PR #2838, opened 2025-07-27
chore: make watermark operations blocking

yhl25:
* Avoids tracking of the offsets in the watermark code, uses tracker for that.
* Tracker also tracks the watermark along with the offsets, it also exposes method to fetch the lowest watermark for a given vertex.
* Watermark calls are made blocking since they are in memory operations and ideally should be very fast, with fire and forget chances of inconsistent state is high.
* Avoids performing hash operation to determine the partition to write to when we only have single partition.
* Sliding replay bug fix: sliding replay was not working because we were not updating the append operation to open operation.


hasmukhmparmar-27:
Please add PR details like why are you making watermark operations blocking?.

vigith (approved):

PR #2845, opened 2025-07-30
fix: graceful shutdown during reduce udf panic

yhl25:
* Introducing join handle abort macro, to abort the task during shutdown.
* Truncating session windows at millisecond granularity to avoid too many expand operations.
* renaming `PAF_CONCURRENCY` to `WRITE_CONCURRENCY`.
* Fixing error types for reduce grpc errors.

### DIFF
--- a/pkg/apis/proto/sessionreduce/v1/sessionreduce.proto
+++ b/pkg/apis/proto/sessionreduce/v1/sessionreduce.proto
@@ -42,7 +42,7 @@ message KeyedWindow {
   repeated string keys = 4;
 }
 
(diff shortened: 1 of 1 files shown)


pkg/apis/proto/sessionreduce/v1/sessionreduce.proto
```diff
@@ -42,7 +42,7 @@ message KeyedWindow {
   repeated string keys = 4;
 }
 
-// SessionReduceRequest represents a request element.
+//SessionReduceRequest represents a request element.
```
whynowy: Mino: space?

PR #2850, opened 2025-08-01
chore: add drop metrics for duplicates

adarsh0728:
PR adds drop metrics for duplicates with reason "duplicate-id"

### DIFF
--- a/rust/numaflow-core/src/pipeline/isb/jetstream/writer.rs
+++ b/rust/numaflow-core/src/pipeline/isb/jetstream/writer.rs
@@ -449,7 +449,7 @@ impl JetstreamWriter {
                                 .get_or_create(&pipeline_drop_metric_labels(
                                     self.vertex_type.as_str(),
                                     stream.name,
-                                    "Buffer full",
+                                    "buffer-full",
                                 ))
                                 .inc();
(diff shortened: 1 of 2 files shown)


rust/numaflow-core/src/pipeline/isb/jetstream/writer.rs
```diff
@@ -584,6 +584,25 @@ impl JetstreamWriter {
...
                                 "Duplicate message detected"
                             );
+                            // Increment drop metric for duplicate messages
+                            pipeline_metrics()
+                                .forwarder
+                                .drop_total
+                                .get_or_create(&pipeline_drop_metric_labels(
+                                    this.vertex_type.as_str(),
+                                    stream.name,
+                                    "Duplicate ID",
```
vigith:
```suggestion
                                    "duplicate-id",
```

This is more cleaner?

vigith (approved):
I know it changes the behavior, but i wanted it to look as if the cardinality is low. Currently it looks like text.

adarsh0728:
> I know it changes the behavior, but i wanted it to look as if the cardinality is low. Currently it looks like text.

got your point! making change for Buffer Full txt as well --> "buffer-full"

PR #2851, opened 2025-08-01
chore: option to disable read ahead in udf 

yhl25:


rust/numaflow-core/src/pipeline/isb/jetstream/reader.rs
```diff
@@ -230,8 +235,19 @@ impl JetStreamReader {
         mut self,
         cancel_token: CancellationToken,
     ) -> Result<(ReceiverStream<Message>, JoinHandle<Result<()>>)> {
-        let batch_size = self.batch_size;
         let read_timeout = self.read_timeout;
+        let max_ack_pending = self.config.max_ack_pending;
+
+        let batch_size = if max_ack_pending < self.batch_size {
```
vigith, line 241: this is a local override, will there be anywhere in the code where we still use `self.batch_size` which might conflict with this local override?

vigith (approved):

PR #2868, opened 2025-08-08
doc: update compatibility

vigith:
Explain what this PR does.


th0ger:
Suggests just use filled bullets for entire tables.

vigith:
> Suggests just use filled bullets for entire tables.

what do you mean? an example please. Please feel free to create a PR and I can close mine.

th0ger:
When I created the compatibility matrices originally I used open bullets (○) to indicate future (next) releases. But since all in the matrix is now released, just use filled (●) for everything.

Another comment: You mark "v0.9.0" compatible with "v1.6.0". But [release notes v1.6.0](https://github.com/numaproj/numaflow/releases/tag/v1.6.0) says "Please use 0.10 and above". So which one is correct? Is "v0.9.0" compatible with "v1.6.0" or not?

vigith:
Accumulator does not work with 0.9, that is all.

th0ger:
Ok I would mark that not compatible then. The matrix should only mark full compatibility. 
(Many versions can be regarded as partially compatible if one disregards certain features.)

vigith:
My definition of compatibility is, whether the previously compiled UDF can run on the newer version of Numaflow.

vigith:
Perhaps we should call this backward compatibility matrix?

th0ger:
When I see this:
| Python \ Numaflow | v1.6.0 | v1.5.0 | 
|:------------------|:------:|:------:|
| __v0.10.0__       |   ●    |   ●    |
| __v0.9.0__        |   ●    |   ●    |

It's not longer clear to understand how we should upgrade from (v1.5+v0.9), i.e. upgrade SDK or numaflow first.

th0ger:
Yeah maybe it makes sense to mark backward/forward compatibility with different "markers" in the table. But make sure to write clearly in the text what those two cases mean.

vigith:
I have improved the wording.

PR #2871, opened 2025-08-11
chore: update roadmap

vigith:
Explain what this PR does.


vigith:
perhaps we need to add reliable "Source Shutdown" too.

PR #2874, opened 2025-08-12
chore: improvements for js source 

yhl25:
### Problem
* We will see potential duplicate messages in JetStream source if the pod restarts.
* Infinite retries in JetStream source didn't honor the cancellation signal.

### Fix
* Use `double_ack()` instead of `ack()` when acknowledging the messages back to JetStream source to avoid duplicates.
* Added a conditional to retry backoff to stop when the cancellation token is cancelled.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2582,6 +2582,7 @@ dependencies = [
  "thiserror 2.0.12",
  "tokio",
  "tokio-stream",
+ "tokio-util",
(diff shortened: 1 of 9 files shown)


vigith (approved):

hasmukhmparmar-27:
Why does using double ACK resolve the duplicate message issue in JetStream? Is this approach recommended by JetStream?

PR #2879, opened 2025-08-12
ci: enforce PR description template and add specialized templates

hasmukhmparmar-27:
### Summary
Introduce a consistent PR description structure and enforce it via CI:
- Update default PR template to include clear sections (problem/solution, links, type, testing).
- Add specialized templates for feature, bugfix, docs, and chore PRs to improve authoring quality.
- Add a workflow that checks PR descriptions contain required sections (“Summary” and a testing section), accepting “How has this been tested?”, “Testing”, or “Testing/Preview”.

### Linked Issues
N/A

### Type of change
- [x] Docs
- [x] CI/CD
- [x] Chore/Refactor

### Breaking change?
- [x] No

### How has this been tested?
- Testing will be done after this PR merge as template is selected from mainline branch


whynowy:
1. The PR content enforcement will block auto generated PRs, such as security fix, changelog, etc.
2. As an open sourced project, we would encourage more contribution, instead of raising the bar. A refined `pull_request_template.md`, plus PR title check might be good enough. Examples such as https://github.com/kubernetes/kubernetes/blob/master/.github/PULL_REQUEST_TEMPLATE.md?plain=1, https://github.com/argoproj/argo-cd/blob/master/.github/pull_request_template.md

Thoughts? @vigith

hasmukhmparmar-27:
> 1. The PR content enforcement will block auto generated PRs, such as security fix, changelog, etc.
> 
>     2. As an open sourced project, we would encourage more contribution, instead of raising the bar. A refined `pull_request_template.md`, plus PR title check might be good enough. Examples such as https://github.com/kubernetes/kubernetes/blob/master/.github/PULL_REQUEST_TEMPLATE.md?plain=1, https://github.com/argoproj/argo-cd/blob/master/.github/pull_request_template.md
> 
> 
> Thoughts? @vigith

Thanks for raising this. Agreed—we shouldn’t block auto-generated. However, well-designed PR templates can actually empower contribution rather than hinder it. They make it easier for contributors (especially newcomers) to know what information is expected, reducing back-and-forth and helping reviewers act faster and more effectively. No template results in the risk of poor documentation, slow reviews, confusion.

``Proposal:``
* Keep the refined PR template (aligned with the Kubernetes PR template as mentioned above for the reference) for guidance.
* Remove the hard-fail body check; replace with a non-blocking reminder comment when sections are missing.
* Add a PR title check (semantic/conventional) to keep release notes clean.
* Exempt bot and auto-generated PRs (dependabot/github-actions), release/changelog-only PRs, or PRs with a skip label.

``If there’s no objection, I’ll:``
* Swap the current workflow to warning-only (comment instead of fail).
* Add a semantic PR title check (e.g., amannn/action-semantic-pull-request).
* Keep the template as the main nudge for contributors who rarely add descriptions.

whynowy:
> * Swap the current workflow to warning-only (comment instead of fail).

What will `warning-only` reflect on the check?

hasmukhmparmar-27:
> What will `warning-only` reflect on the check?

The check will pass (green). It won’t block merging. The workflow posts a comment on the PR with the reminder. No red X. In the job logs you’ll see the step run; unless we log warnings, there’s no yellow badge.

.github/workflows/pr-title-check.yml
```diff
@@ -0,0 +1,25 @@
...
+  contents: read
+  
+on:
+  pull_request_target:
+    types: [opened, edited, synchronize, reopened]
+
+jobs:
+  semantic-title:
+    if: >-
+      !contains(github.event.pull_request.title, 'chore(release)') &&
```
whynowy: Don't worry about these, even those auto generated RR titles have correct format.

.github/pull_request_template.md
```diff
@@ -1,4 +1,14 @@
-Explain what this PR does.
+### What this PR does / why we need it
+Briefly describe the problem and the solution.
+
+### Related issues
+Fixes #
```
whynowy, line 5: Use `Closes` might be better, which would be more appropriate for both issues and features.

.github/workflows/pr-template-check.yml
```diff
@@ -0,0 +1,57 @@
+name: PR description reminder
```
whynowy: Prefer not to have this check - we already put those template in the PR, if the contributor still removes them, then there's no point of giving a warning comment, which still could be ignored.

hasmukhmparmar-27: > Prefer not to have this check - we already put those template in the PR, if the contributor still removes them, then there's no point of giving a warning comment, which still could be ignored.

I will remove it for now. I would prefer to have a blocking check in case of missing description.

.github/pull_request_template.md
hasmukhmparmar-27: > Use `Closes` might be better, which would be more appropriate for both issues and features.

I would prefer fixes as it seems relevant.

Issue #2811, opened 2025-07-16
Propagate Platform generated and User Generated context/metadata from Source to Sink


vigith:
# Summary

There is need for passing some context between "processors" (Source, Map, Sink, Transformer, and Reduce) so that these meta information can be used by subsequent vertices (including during joins and cycles). These have to be an additive operations as more information can be injected per vertex.

## Metadata Format
A vertex could create one or more set of key-value groups of per group-name. This is required because a vertex could forward a message to a Kafka sink with Kafka headers, and a metrics Sink with some key/value points.

There will be two kinds on metadata, 
  1. system - passing system generated metadata with limited R_ONLY access
  1. user - user generated with RW access

```text
                         +-> [group-m] -> {k1, v1 ... }
                         |
           +-> [user] -> +-> [group-n] -> {k1, v1 ... }
           |             |
           |             +-> [group-o] -> {k1, v1 ... }
           |
[vertex-N] +
           |
           |            +-> [group-h] -> {k1, v1 ... }
           |            |
           +-> [sys]--> +-> [group-i] -> {k1, v1 ... }
                        |
                        +-> [group-k] -> {k1, v1 ... }

```

## SDK Interface

```python
# user defined metadata
metadata: dict({str: {str: str}}) = datum.metadata()

# platform metadata
trace = datum.OtelHeaders()

# inserting into metadata
messages = new Messages()
message.append(val).metadata({"group-a": {"k1": "v1 }})
```

# Use Cases

- [ ] #2255 
- [ ] Out-of-the-box support for writing to DLQs in "standardized sinks" (if transformer fails, sent to Fallback Sink for MonoVertex)
- [ ] Distributed Tracing

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #2885, opened 2025-08-17
doc: document compression

vigith:
Enable compression if users have to increase max-message size.

Document for https://github.com/numaproj/numaflow/pull/2709

### DIFF
--- a/docs/user-guide/reference/configuration/max-message-size.md
+++ b/docs/user-guide/reference/configuration/max-message-size.md
@@ -1,13 +1,15 @@
 # Maximum Message Size
 
-The default maximum message size is `1MB`. There's a way to increase this limit in case you want to, but please think it through before doing so.
+The default maximum message size is `1MB`. There's a way to increase this limit in case you want to, but please think it
+through before doing so.
 
 The max message size is determined by:
 
 - Max messages size supported by gRPC (default value is `64MB` in Numaflow).
 - Max messages size supported by the Inter-Step Buffer implementation.
 
-If `JetStream` is used as the Inter-Step Buffer implementation, the default max message size for it is configured as `1MB`. You can change it by setting the `spec.jetstream.settings` in the `InterStepBufferService` specification.
+If `JetStream` is used as the Inter-Step Buffer implementation, the default max message size for it is configured as `1MB`.
+You can change it by setting the `spec.jetstream.settings` in the `InterStepBufferService` specification.
 
 ```yaml
 apiVersion: numaflow.numaproj.io/v1alpha1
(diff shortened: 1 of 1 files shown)


docs/user-guide/reference/configuration/max-message-size.md
```diff
@@ -22,4 +24,54 @@ spec:
...
+| ZSTD        | 900                | 4.5 ~ 4.7              |
+| LZ4         | 1000               | 2.8 ~ 3                |
+
+Clearly the best compression (least disk usage) is `gzip`, but it has the lowest throughput. `lz4` has the best
+throughput and `zstd` is in the middle. If you want to use `gzip`, you might need to increase the CPU of `numa` container
+to get better performance.
+
+### Configuration
+
+You can enable it by setting the `compression` field in the `Pipeline` specification.
```
KeranYang, line 64: I see https://numaflow.numaproj.io/APIs/#numaflow.numaproj.io/v1alpha1.VertexSpec, seems we can also enabling at vertex level. What does it mean to enable on vertex level? Is it applied to the in buffer or out?

vigith: That is JetStream's compression featire, it does compression only while writing to disk. It won't help in reducing the max-message problem and will need way more CPU on the ISB.

vigith: I am wrong, I am not sure why [`VertexSpec`](https://numaflow.numaproj.io/APIs/#numaflow.numaproj.io/v1alpha1.VertexSpec) is showing [`InterStepBuffer`](https://numaflow.numaproj.io/APIs/#numaflow.numaproj.io/v1alpha1.InterStepBuffer)

vigith: I clearly know that compression cannot be enabled at Vertex level because we do not inspect the payload to see whether it is compressed or not, to implement that we need #2811

vigith: Ah, it could be because the pipeline level specs are copied over to each Vertex.

KeranYang: I think we might put InterStepBuffer under VertexSpec by mistake - https://github.com/numaproj/numaflow/blob/53f8c182f42a7c34ce7133991e7764bb2eae5921/pkg/apis/numaflow/v1alpha1/vertex_types.go#L532. Should I make a separate PR to remove?

vigith: Yes please

vigith: I don't think we can avoid that because it needs to be propagated to each vertex too. @KeranYang please feel free to see whether we can remove it.

KeranYang: Will do. Thanks!

PR #2892, opened 2025-08-21
feat: framework for distributed throttling

### LINKED ISSUE #2887: high level design for distributed throttling

### LINKED ISSUE #2888: extensible framework for distributed throttling
- [ ] support standalone rate-limiter
- [ ] distributed state management with pluggable stores
- [ ] implement in-memory and redis stores
- [ ] well defined traits for extendability

vigith:
Closes #2887 
Closes #2888

[Design](https://github.com/numaproj/numaflow/blob/759d152f1e1ed703646093aa3bb6c504ac55f7ee/rust/numaflow-throttling/DESIGN.md)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -70,6 +70,9 @@ jobs:
       - name: Start Kafka container
         run: docker run -d -p 9092:9092 --name kafka-broker apache/kafka:4.0.0
 
+      - name: Start Redis standalone container
+        run: docker run -d -p 6379:6379 --name redis-server redis:8-alpine
+
       - name: Set up Go 1.x
         uses: actions/setup-go@v5
         with:
--- a/redis-deployment.yaml
+++ b/redis-deployment.yaml
@@ -1,0 +1,35 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: redis
+  namespace: default
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: redis
+  template:
+    metadata:
+      labels:
+        app: redis
+    spec:
+      containers:
+      - name: redis
+        image: redis:7-alpine
+        ports:
+        - containerPort: 6379
+        command: ["redis-server"]
+        args: ["--appendonly", "yes"]
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: redis-service
+  namespace: default
+spec:
+  selector:
+    app: redis
+  ports:
+  - port: 6379
+    targetPort: 6379
+  type: ClusterIP
--- a/rust/numaflow-throttling/DESIGN.md
+++ b/rust/numaflow-throttling/DESIGN.md
@@ -17,9 +17,7 @@ level.
 * `to_drop` need not (optional) affect the throttle limit (rationale: to_drop might not hit the critical resource)
 * Throughput should not be affected if DT is not enabled
 * Should support external store (e.g., Redis, etc.)
-* Users might combine DT with libraries
-  like [bucket4j](https://github.com/bucket4j/bucket4j), [pyratelimiter](https://pyratelimiter.readthedocs.io/en/latest/),
-  etc.
+* Users might combine DT with libraries like [bucket4j](https://github.com/bucket4j/bucket4j), [pyratelimiter](https://pyratelimiter.readthedocs.io/en/latest/), etc.
 * Avoid querying external store for every message (use background task).
 * Optionally remove external store (use pod count to throttle)
 
@@ -31,18 +29,19 @@ level.
 
 ## Assumptions
 
+* Since Numaflow throttling is a client-side throttler, we do not have to implement realtime reconfiguration
 * UDF is considered as a black-box and as a single critical resource
 * All UDFs processes at the same rate, tokens are evenly distributed across all the pods in the vertex
 
 ## Open Issues
 
 * What should happen if the external store is unavailable?
-* Total tokens are distributed across all the pods in the vertex, how do we detect dead pods?
 * How to distribute the initial burst tokens?
 
 ## Closed Issues
 
-* Use Background thread to update the current tokens.
+* Use Background thread to update the current token which is dependent on the pool-size
+* Total tokens are distributed across all the pods in the vertex, to detect dead pods, we use stale heartbeats
 
 # Design Details
 
@@ -50,11 +49,12 @@ To achieve high TPS, during happy path the decisions are made locally and all th
 
 ## High Level Design
 
-* Local bucket will have enough information to decide if the request should be throttled or not.
-* Local bucket if not updated by the background thread will be considered as unavailable and will be forced to query the
+* The distributed state stores only the pool-size and tokens are calculated locally based on the pool-size.
+* Local store (using pool-size) will have enough information to decide if the request should be throttled or not.
+* Local store if not updated by the background thread will be considered as unavailable and will be forced to query the
   external store. This should not happen during happy path.
-* Background thread will update the local bucket every X ms and there will be at least Y seconds worth of data in the
-  local bucket.
+* Background thread will update the local store (from the external store) every X ms and there will be at least Y seconds
+* worth of data in the local store.
 
 ### Consensus on Active Candidates (consumers of tokens)
 
--- a/rust/numaflow-throttling/DESIGN.md
+++ b/rust/numaflow-throttling/DESIGN.md
@@ -1,0 +1,91 @@
+# Distributed Throttling (DT) in Numaflow
+
+DT in Numaflow is a feature that allows Numaflow to limit number of message concurrently processed by a pipeline at any
+UDF. Users can configure DT at pipeline level and it will be applied to all the UDFs in the pipeline or at each vertex
+level.
+
+## Use Cases
+
+* Limit number of messages read from Source (useful
+  for [MonoVertex](https://numaflow.numaproj.io/core-concepts/monovertex/))
+* Limit number of messages processed by a UDF (e.g., calling a critical resource like LLM)
+* Limit number of messages written to Sink (e.g., calling external endpoints that does not have rate limiting)
+
+## Requirements
+
+* Should be able to support ramp up (say from X to Y in t seconds)
+* `to_drop` need not (optional) affect the throttle limit (rationale: to_drop might not hit the critical resource)
+* Throughput should not be affected if DT is not enabled
+* Should support external store (e.g., Redis, etc.)
+* Users might combine DT with libraries
+  like [bucket4j](https://github.com/bucket4j/bucket4j), [pyratelimiter](https://pyratelimiter.readthedocs.io/en/latest/),
+  etc.
+* Avoid querying external store for every message (use background task).
+* Optionally remove external store (use pod count to throttle)
+
+## Non-Requirements
+
+* Key space is limited by the number of vertices in the pipeline, need not have to solve too many keys
+* Retrying is not required to be solved by DT
+* Managing external store is not in scope of DT
+
+## Assumptions
+
+* UDF is considered as a black-box and as a single critical resource
+* All UDFs processes at the same rate, tokens are evenly distributed across all the pods in the vertex
+
+## Open Issues
+
+* What should happen if the external store is unavailable?
+* Total tokens are distributed across all the pods in the vertex, how do we detect dead pods?
+* How to distribute the initial burst tokens?
+
+## Closed Issues
+
+* Use Background thread to update the current tokens.
+
+# Design Details
+
+To achieve high TPS, during happy path the decisions are made locally and all the information is available locally.
+
+## High Level Design
+
+* Local bucket will have enough information to decide if the request should be throttled or not.
+* Local bucket if not updated by the background thread will be considered as unavailable and will be forced to query the
+  external store. This should not happen during happy path.
+* Background thread will update the local bucket every X ms and there will be at least Y seconds worth of data in the
+  local bucket.
+
+### Consensus on Active Candidates (consumers of tokens)
+
+When a new Candidate (aka Consumer/Pod) is brought up, we have to make sure that the token pool is equally distributed
+across all the active candidates. The **tricky bit** is to make sure that all the candidates agree on the active
+candidates.
+
+The algorithm's main goal is to track a group of processors and determine if they all agree on a shared value (the pool
+size), while automatically removing any processors that have gone silent.
+
+#### On the Store
+
+* Update Status: When a processor checks in, the algorithm updates two lists. The first list tracks the processor's
+  heartbeat with the current time, and the second tracks its reported pool size.
+* Clean Up Stale Processors: It then checks the heartbeat list for any processors that haven't sent an update in the
+  last X seconds. Any processor that is considered "stale" is removed from both the heartbeat list and the pool size
+  list.
+* Check for Agreement: Finally, it looks at the remaining active processors.
+    * If the smallest reported pool size is the same as the largest reported pool size, it means everyone agrees.
+      The algorithm returns **AGREE** with the consensus value (provided it matches with active processor count).
+    * If the sizes don't match, it returns **DISAGREE** along with the largest size reported.
+
+#### On the Consumer
+
+* If there is AGREEMENT, the consumer can use the agreed upon pool size.
+* If there is DISAGREEMENT
+    * if the pool size is smaller, then it goes ahead with larger pool size it had but will report back the smaller pool
+      size.
+    * if the pool size is larger, then it goes ahead with larger pool size it had and reports back the smaller pool
+      size.
+
+The reason for sending the fake pool size is to be on the conservative side but wants everyone to make progress and
+AGREE.
+ 
--- a/rust/numaflow-throttling/src/error.rs
+++ b/rust/numaflow-throttling/src/error.rs
@@ -1,0 +1,18 @@
+use thiserror::Error;
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Error, Debug, Clone)]
+pub enum Error {
+    #[error("Store Error - {0}")]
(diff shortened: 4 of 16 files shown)


.github/workflows/ci.yaml
```diff
@@ -70,6 +70,9 @@ jobs:
       - name: Start Kafka container
         run: docker run -d -p 9092:9092 --name kafka-broker apache/kafka:4.0.0
 
+      - name: Start Redis standalone container
+        run: docker run -d -p 6379:6379 --name redis-server redis:8-alpine
```
hasmukhmparmar-27, line 74: Avoid version mismatch between environments - CI workflow 8-alphine, K8 menifest Redis 7 alphine.

hasmukhmparmar-27: No check that Redis is ready before tests start!

redis-deployment.yaml
```diff
@@ -0,0 +1,35 @@
...
+    matchLabels:
+      app: redis
+  template:
+    metadata:
+      labels:
+        app: redis
+    spec:
+      containers:
+      - name: redis
+        image: redis:7-alpine
```
hasmukhmparmar-27, line 18: Version mismatch with CI workflow as mentioned above!

redis-deployment.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: apps/v1
```
hasmukhmparmar-27, line 1: Missing probes for the readiness or healthy!

rust/numaflow-throttling/src/error.rs
```diff
@@ -0,0 +1,18 @@
+use thiserror::Error;
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Error, Debug, Clone)]
+pub enum Error {
```
hasmukhmparmar-27, line 6: Shall we consider retryable vs non-retryable errors distinction?

rust/numaflow-throttling/src/lib.rs
```diff
@@ -0,0 +1,977 @@
...
+pub(crate) mod error;
+
+pub(crate) use error::Error;
+pub(crate) use error::Result;
+
+/// State module contains the implementation of the [RateLimiterDistributedState] which has the
+/// holds distributed consensus state required to compute the available tokens.
+pub mod state;
+
+#[derive(Clone)]
```
hasmukhmparmar-27, line 19: lib.rs is a big code file. Better to break it down!

rust/numaflow-throttling/src/lib.rs
```diff
@@ -0,0 +1,977 @@
...
+pub struct WithoutDistributedState;
```
hasmukhmparmar-27, line 20: nit - Rename to Local!

rust/numaflow-throttling/src/lib.rs
```diff
@@ -0,0 +1,977 @@
...
+
+#[derive(Clone)]
+pub struct WithDistributedState<S>(RateLimiterDistributedState<S>);
```
hasmukhmparmar-27, line 23: nit - rename to ``Distributed<S>``

rust/numaflow-throttling/src/lib.rs
```diff
@@ -0,0 +1,977 @@
...
+
+/// RateLimiter will expose methods to query the tokens available per unit time
+#[trait_variant::make(Send)]
+pub trait RateLimiter {
+    /// GetN returns the number of tokens available for the next n time windows or blocks till then.
+    /// If timeout is None, it will block till the tokens are available, else returns whatever is
+    /// available. If `n` is provided, it will try to acquire `n` tokens else it will acquire all the tokens.
```
hasmukhmparmar-27, line 30: Why do we try to acquire partially? It seems deviating from standard patterns across rate limiter libraries!

It can introduce the retry storm! Also, metrics inconsistency!

There could be starvation without queuing or virtual time based reservation like ``nextFreeTicketMicros`` in the Guava

rust/numaflow-throttling/src/lib.rs
```diff
@@ -0,0 +1,977 @@
...
+    async fn acquire_n(&self, n: Option<usize>, timeout: Option<Duration>) -> usize;
```
hasmukhmparmar-27, line 31: This also introduces ambiguity for 0 - timeout? none available?

hasmukhmparmar-27: why did we deviate from the contracts available in the rate limiter libraries?

rust/numaflow-throttling/src/lib.rs
```diff
@@ -0,0 +1,977 @@
...
+    /// Max number of tokens ever filled till now. The next refill relies on the current max allowed
+    /// value. For the first run this will be set to `burst` and will be slowly incremented to the
+    /// `max` value.
+    max_ever_filled: Arc<AtomicUsize>,
+    /// Last time the token was queried. The tokens are replenished based on the time elapsed since
+    /// the last query.
+    last_queried_epoch: Arc<AtomicU64>,
+    /// Optional [RateLimiterDistributedState] to query for the pool-size in a distributed setting.
+    state: W,
+}
```
hasmukhmparmar-27, line 51: RateLimiterBuilder::new(limits/config).with_backend(backend).build();

Did we consider ergonomic public API design pattern where backend can be injected! It's trait driven approach! Testing is first class! Same API for the testing and production!

rust/numaflow-throttling/src/lib.rs
```diff
@@ -0,0 +1,977 @@
...
+    pub(crate) fn compute_refill(&self) -> usize {
+        let max_ever_filled = self
+            .max_ever_filled
+            .load(std::sync::atomic::Ordering::Relaxed);
+
+        // let's make sure we do not go beyond the max
+        if max_ever_filled >= self.token_calc_bounds.max {
+            self.token_calc_bounds.max
+        } else {
+            let refill = max_ever_filled + (self.token_calc_bounds.slope as usize);
```
hasmukhmparmar-27, line 90: Ramp up is stalled when slope is less than 1.0!

rust/numaflow-throttling/src/lib.rs
```diff
@@ -0,0 +1,977 @@
...
+    /// Tokens are not available for this epoch, there is no need to recompute this we have exhausted
+    /// all the tokens for this epoch.
+    Exhausted,
+    /// Recompute the number of tokens available since we are in a new epoch.
+    Recompute,
+}
+
+impl<W> RateLimit<W> {
+    /// Computes the number of tokens to be refilled for the next epoch.
+    pub(crate) fn compute_refill(&self) -> usize {
```
hasmukhmparmar-27, line 81: Why is ``duration`` ignored for the refill? The code only refills one second’s tokens even if more than one second has elapsed!

rust/numaflow-throttling/src/lib.rs
```diff
@@ -0,0 +1,977 @@
...
+                TokenAvailability::Available(acquired)
+            }
+            Err(_) => TokenAvailability::Exhausted,
+        }
+    }
+}
+
+impl RateLimiter for RateLimit<WithoutDistributedState> {
+    /// Acquire `n` tokens. If `n` is not provided, it will acquire all the tokens. For non-distributed
+    /// state, `timeout` is not used since there are no blocking calls.
```
hasmukhmparmar-27, line 163: Why? This creates gaps in the testing!

rust/numaflow-throttling/src/lib.rs
```diff
@@ -0,0 +1,977 @@
...
+        }
+    }
+}
+
+impl TokenCalcBounds {
+    /// `Maximum` number of tokens that can be added in a unit of time with an initial `burst`.
+    /// The `duration` (in seconds) at with we can add `max` tokens.
+    pub fn new(max: usize, min: usize, duration: Duration) -> Self {
+        TokenCalcBounds {
+            slope: (max - min) as f32 / duration.as_secs_f32(),
```
hasmukhmparmar-27, line 370: Validate duration!

rust/numaflow-throttling/src/lib.rs
```diff
@@ -0,0 +1,977 @@
...
+    /// Create a new [RateLimit] without a distributed state.
+    pub fn new(token_calc_bounds: TokenCalcBounds) -> Result<Self> {
+        let burst = token_calc_bounds.min;
+        Ok(RateLimit {
+            token_calc_bounds,
+            token: Arc::new(AtomicUsize::new(burst)),
+            max_ever_filled: Arc::new(AtomicUsize::new(burst)),
+            last_queried_epoch: Arc::new(AtomicU64::new(
+                std::time::SystemTime::now()
+                    .duration_since(std::time::UNIX_EPOCH)
```
hasmukhmparmar-27, line 109: Use Instant!

hasmukhmparmar-27 (commented):
Review in-progress!

rust/numaflow-throttling/src/lib.rs
yhl25: Validation will be done at the controller side

rust/numaflow-throttling/src/lib.rs
yhl25: Why?

rust/numaflow-throttling/src/lib.rs
yhl25: We can add

rust/numaflow-throttling/src/lib.rs
yhl25: The tokens of the previous second is not carried over, we only care about the current second.

rust/numaflow-throttling/src/lib.rs
yhl25: The slope can be less than 1 only if the duration (in seconds) is greater than the number of tokens. Its config mistake from the user.

rust/numaflow-throttling/src/lib.rs
yhl25: Starvation is not a concern here because the api is designed to be accessed by the Numaflow Pod, as long as we have an instance per partition(or distribute the tokens among the partition) we should good.

redis-deployment.yaml
yhl25: This file got added by mistake, its not used anywhere, will remove it.

redis-deployment.yaml
yhl25: not used.

.github/workflows/ci.yaml
yhl25: we can add a check

rust/numaflow-throttling/src/lib.rs
yhl25: Felt like the builder pattern would be overkill here since all the fields are mandatory. Since we accept the store trait, its trait driven approach that is how we were able to test it using in-memory store.

rust/numaflow-throttling/src/error.rs
yhl25: Not sure what errors can be retryable here.

hasmukhmparmar-27: For example, timeout!

rust/numaflow-throttling/src/lib.rs
hasmukhmparmar-27: SystemTime can jump or go backwards. For refill deltas, prefer monotonic time. There are multiple system now calls in the acquire call paths! Please verify what could happen if any of these call crosses second boundary!

rust/numaflow-throttling/src/lib.rs
hasmukhmparmar-27: Discussed with Vigith - ramp up must consider elapsed duration! otherwise, it's bug!

rust/numaflow-throttling/src/lib.rs
hasmukhmparmar-27: Generics are great for zero-cost abstraction, but it doesn’t let callers pick the backend at runtime without plumbing generics through every layer. Recommendation is to use Rust traits with the dynamic dispatch!

vigith: What is wrong with adding generics across?

rust/numaflow-throttling/src/lib.rs
hasmukhmparmar-27: Is it a single-tenant, single-caller use case?

rust/numaflow-throttling/src/lib.rs
hasmukhmparmar-27: Ensure that we validate the fix with tests! It seems there are gaps in the tests!

vigith: This is a known issue, there are few more to cover. The main idea of this PR is to get the end to end framework up. This is not feature complete in any means.

hasmukhmparmar-27: Please post TODO comments with issue in case of know issue and planned fix!

rust/numaflow-throttling/src/lib.rs
hasmukhmparmar-27: Generics are great :-) 
Challenge is forcing downstream code to carry the type parameter everywhere.

It increases API surface. Recommendation is to use facades with traits or enums with generics internally if required!

vigith: We are trying to reduce the type parameter carryover using associated types. Let's see how complex it can get, unless you try you never know :)

hasmukhmparmar-27: Associated types are refinement over traits! I am good with it! I would suggest to start simple (``Arc<dyn RateLimiter>``). It is easy to extend with associated types!

rust/numaflow-throttling/DESIGN.md
```diff
@@ -0,0 +1,91 @@
...
+## Closed Issues
+
+* Use Background thread to update the current token which is dependent on the pool-size
+* Total tokens are distributed across all the pods in the vertex, to detect dead pods, we use stale heartbeats
+
+# Design Details
+
+To achieve high TPS, during happy path the decisions are made locally and all the information is available locally.
+
+## High Level Design
```
hasmukhmparmar-27, line 50: Let's discuss the design during our Thursday review.

* What alternatives did we considered?
* Redis - Is it per pipeline? or Central?
* Redis - Single pod or Multi-AZ? Does Redis guarantees transactional properties (serialization) with multi-az deployment? Is single pod sufficient for the availability guarantees?
* Will it scale with the large number of pods?

rust/numaflow-throttling/src/state.rs
```diff
@@ -0,0 +1,191 @@
...
+}
+
+/// The state the [RateLimiter] always relies on to compute the available tokens.
+#[derive(Clone, Debug)]
+pub struct RateLimiterDistributedState<S> {
+    /// Timestamp (epoch) for which we have the pool size computed and beyond this timestamp the state is
+    /// invalid. This is a moving timestamp and will be updated as we get updated data from the
+    /// external store.
+    // FIXME: use me
+    valid_till_epoch: Arc<AtomicU64>,
```
hasmukhmparmar-27, line 27: Did we missed to use it?

rust/numaflow-throttling/src/state.rs
```diff
@@ -0,0 +1,191 @@
...
+    /// early, we might end up with wrong pool size, and we might end up consuming more tokens than
+    /// what is available.
+    async fn wait_for_consensus(
+        s: &S,
+        processor_id: &str,
+        initial_pool_size: usize,
+        cancel: &CancellationToken,
+    ) -> crate::Result<usize> {
+        let mut desired_pool_size = initial_pool_size;
+        loop {
```
hasmukhmparmar-27, line 132: Shall we consider timeout with alert in case of longer disagree?

rust/numaflow-throttling/DESIGN.md
```diff
@@ -0,0 +1,91 @@
...
+
+* The distributed state stores only the pool-size and tokens are calculated locally based on the pool-size.
+* Local store (using pool-size) will have enough information to decide if the request should be throttled or not.
+* Local store if not updated by the background thread will be considered as unavailable and will be forced to query the
+  external store. This should not happen during happy path.
+* Background thread will update the local store (from the external store) every X ms and there will be at least Y seconds
+* worth of data in the local store.
```
hasmukhmparmar-27, line 57: Note: It's approximate fairness with fast processor cannot misuse! Equal distribution distribution across pods results in the under-utilization as reminder is ignored (May not be much!)!

rust/numaflow-throttling/DESIGN.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Consensus on Active Candidates (consumers of tokens)
```
hasmukhmparmar-27, line 59: Heuristic may be correct term rather than Consensus!

rust/numaflow-throttling/src/state/store/redis_store.rs
```diff
@@ -0,0 +1,262 @@
...
+        let result: Vec<String> = self
+            .exec_lua_script(
+                &self.sync_pool_size_script,
+                &[self.key_prefix],
+                &[processor_id, &timestamp, &pool_size_str],
+            )
+            .await
+            .map_err(|e| crate::Error::Redis(e.to_string()))?;
+
+        if result.len() != 3 {
```
hasmukhmparmar-27, line 198: why 3? We are using only 0 and 1 index!

rust/numaflow-throttling/src/state.rs
vigith: dude, didn't you read the line above. Please make meaningful comments!

hasmukhmparmar-27: Oh! I did search for the variable based on update but missed to read comment!

PR #2906, opened 2025-08-26
fix: Minimized CRDs should define fields which are referenced elsewhere in the CRD

juliev0:

### Related issues
Fixes #

### Testing

### DIFF
--- a/config/advanced-install/minimal-crds.yaml
+++ b/config/advanced-install/minimal-crds.yaml
@@ -98,6 +98,11 @@ spec:
           metadata:
             type: object
           spec:
+            properties:
+              replicas:
+                default: 1
+                format: int32
+                type: integer
             type: object
             x-kubernetes-preserve-unknown-fields: true
           status:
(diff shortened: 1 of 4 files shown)


config/advanced-install/minimal-crds.yaml
```diff
@@ -98,9 +98,20 @@ spec:
...
           spec:
+            properties:
+              replicas:
+                default: 1
+                format: int32
+                type: integer
             type: object
             x-kubernetes-preserve-unknown-fields: true
           status:
+            properties:
```
whynowy, line 109: Is this needed?

juliev0: It may not be. I was just trying to be consistent within the issue with the idea "if it's referenced elsewhere in the CRD, then keep it", but if we want to instead consider whether including a given field has some use case before adding it and only adding the minimum set of fields (probably just `spec.replicas`), we could do that. I'll verify HPA doesn't issue errors when the status fields aren't specified.

juliev0: this is the new PR with just "spec" changed: https://github.com/numaproj/numaflow/pull/2910

PR #2910, opened 2025-08-27
fix: Add `spec.replicas` field to minimal CRDs

juliev0:

This is a follow up from this other PR: https://github.com/numaproj/numaflow/pull/2906

Except that it only modifies `spec.replicas` and not `status` fields. 

### Testing
I ran this through all of the numaplane e2e tests (https://github.com/numaproj/numaplane/pull/852) and it passed.

I also tried temporarily changing this PR such that it runs minimal CRD during e2e tests and that passed.

This proves that the `x-kubernetes-preserve-unknown-fields: true` set on spec is still behaving correctly: making sure that fields are not pruned by Kubernetes. Just to be sure, I downloaded from one of the longer runs the numaflow log that was produced and didn't see any errors in there that could be related.


config/base/kustomization.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 resources:
-  - crds/full
+  - crds/minimal
```
juliev0: todo: remove this (just for testing)

juliev0:
FYI @whynowy - this is my new PR. 

I think there may be a couple of flakey e2e tests? 2 tests failed but I believe they are using full CRDs so shouldn't be related

whynowy (approved):
LGTM, thanks!

PR #2917, opened 2025-08-28
feat: Redis Sentinel Mode Support For Rate Limiter

yhl25:
* Integrated Numaflow Throttling to Numaflow Core to support out of the box Rate limiting support for Numaflow Pipeline and MonoVertex

TODO: (follow-up PR)
* Disable rate limit for reduce vertex.
* Don't apply pipeline limits to source vertex.


rust/numaflow-core/Cargo.toml
```diff
@@ -61,6 +61,7 @@ flate2 = "1.1.2"
 zstd = "0.13.3"
 lz4 = "1.28.1"
 async-trait = "0.1.88"
+redis = { version = "0.32.5", features = ["sentinel", "tokio-comp", "connection-manager", "script"] }
```
vigith: damn, we need this in numaflow-core! Let's avoid this by passing minimal info, there will be some duplicates.

rust/numaflow-core/src/typ.rs
```diff
@@ -133,3 +133,87 @@ where
     .await
     .map_err(|e| Error::Config(format!("Failed to create rate limiter: {}", e)))
 }
+
+/// Create RedisMode from the rate limiter Redis store configuration using builder pattern
+fn create_redis_mode(
```
vigith: let's move this to `numaflow-throttling`, that will remove the Redis dependency

vigith (approved):

Issue #2918, opened 2025-08-29
Slow handling of node shutdown


saty9:
**Describe the bug**
We have been seeing an issue most noticeable when a vertex is scaled at 1.

When a node shuts down it terminates all the pods running on it and we have seen the pod for our vertex stop processing messages and exit but a replacement doesnt get created for at least another minute. 

We think this is because the pod doesn't change readiness but does change phase "Running" -> "Suceeded" and then sits until the node finishes shutting down. Once that happens the pod stops existing and the controller immediately creates a replacement. This leaves us a large gap where we have nothing running and the controller doesn't create any replacements.

**To Reproduce**

1. scale vertex to 1
2. shutdown/restart node the vertex is located on
3. observe gap between pod containers exiting and new pod being created

**Expected behavior**
A replacement pod is scheduled as soon as the previous pod finishes shutting down (an option for as soon as it starts terminating would be nice).

**Environment (please complete the following information):**

- Kubernetes: 1.33.2-gke.1111000
- Numaflow: 1.6.0

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

>For quick help and support, join our [slack](https://numaproj.slack.com/join/shared_invite/zt-19svuv47m-YKHhsQ~~KK9mBv1E7pNzfg#/shared-invite/email) channel.

whynowy:
Not sure you are doing what kind of node shutdown, if it's a sudden shutdown (e.g. power loss), nothing helps before the kubelet heartbeat loss is detected by the control plane. But if it's a scheduled shutdown, you should drain the node beforehand, which will evict all the existing pods.

I noticed you are using GKE, if you are using preemptible instances, I believe the [Graceful Shutdown](https://cloud.google.com/kubernetes-engine/docs/how-to/preemptible-vms#graceful-shutdown) feature is enabled by default, that shouldn't have any problem.

saty9:
I believe in this case the node was restarting and it was doing a graceful shutdown but seems it doesnt remove the old pods until it finishes the shutdown instead of as each pod terminates. Is there a reason the numaflow server handles the pods directly instead of using a deployment/stateful set with the server just setting the replica count?

PR #2919, opened 2025-08-29
fix: detect terminating pods as soon as possible

Koalk:
When a pod is being shut down by kubernetes (evicted from node, OOMkilled, etc...) the pod will expend some time in intermediate states before garbage collectors pick up the terminated pod and remove it (populating the DeletionTimestamp field).
If we use the status.phase field we can detect pods being terminated before that so we can produce new replicas as needed a bit earlier.

### Related issues
Fixes https://github.com/numaproj/numaflow/issues/2918

### Testing

https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase

### DIFF
--- a/pkg/reconciler/vertex/controller.go
+++ b/pkg/reconciler/vertex/controller.go
@@ -602,7 +602,7 @@ func (r *vertexReconciler) findExistingPods(ctx context.Context, vertex *dfv1.Ve
 	}
 	result := make(map[string]corev1.Pod)
 	for _, v := range pods.Items {
-		if !v.DeletionTimestamp.IsZero() {
+		if r.isTerminatingPod(&v) {
 			// Ignore pods being deleted
 			continue
 		}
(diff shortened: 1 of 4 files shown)


pkg/reconciler/vertex/controller.go
```diff
@@ -615,6 +615,12 @@ func (r *vertexReconciler) findExistingPods(ctx context.Context, vertex *dfv1.Ve
 	return result, nil
 }
 
+func (r *vertexReconciler) isTerminatingPod(ctx context.Context, pod *corev1.Pod) bool {
+	// A pod is terminating if its deletion timestamp is set (i.e., non-nil)
+	// It is also terminating if its status phase is in "Failed" or "Succeeded"
+	return !pod.DeletionTimestamp.IsZero() || pod.Status.Phase == corev1.PodFailed || pod.Status.Phase == corev1.PodSucceeded
```
whynowy, line 621: This is a risky behavior. If the pod comes to failed or succeeded status due to a repeatable error, then it keeps creating new pods. Even though they won't use any cpu/memory, but some disk space is still occupied.

Koalk: Isn't that the expected behaviour or do you think it could fail too fast in quick succession and ddos itself?
I guess there could be a counter for "amounts of retries" with a sensible limit like 5 times. Another option is just using the "PodSucceeded" status since that one should be only for pods  terminating correctly, so probably not crashlooping.
Any preferences?

whynowy: @Koalk - just thought about it, I think it makes sense to include those 2 phases - I can not think of any normal reason to cause pods keeping going to terminated status (failed or succeeded).

whynowy: Please also update monovertex controller.

Koalk:
I am guessing those tests are flaky sometimes because I can't see any relation with my changes. Can someone retry them?

PR #2921, opened 2025-09-01
chore: metadata definitions for services

adarsh0728:

1. Creates a common metadata proto to be used for isb message and requests/responses for clients (source, map etc)
2. Updates the schema of KeyValue group from previous implementation in #2821 
3. Updates build logic for protos in rust by adding compilation of common protos (in future this can be extended to common items like Handshake)
4. Used `extern_path` to avoid creating duplicate types in clients/objects for metadata

### Related issues

#2811 

### Testing

Go SDK PR https://github.com/numaproj/numaflow-go/pull/208 
Testing Results in Comments

`generate-proto.sh` file has to be updated as isb/message.proto imports common/metadata


rust/numaflow-core/src/metadata.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// Key-value pairs
+    pub(crate) key_value: HashMap<String, Bytes>,
+}
+
+// Conversion implementations between protobuf and internal type
+impl From<metadata::Metadata> for Metadata {
+    fn from(metadata: metadata::Metadata) -> Self {
+        Self::from_proto(metadata)
+    }
+}
```
yhl25: Can we move the `from_proto` functionality here? Do we really need a separate `from_proto` method?

rust/numaflow-core/src/metadata.rs
```diff
@@ -0,0 +1,114 @@
...
+
+impl From<Metadata> for metadata::Metadata {
+    fn from(metadata: Metadata) -> Self {
+        // For JetStream writes: set previous_vertex to current vertex name
+        // so the next vertex knows who sent the message
+        metadata.to_proto_with_previous_vertex(get_vertex_name().to_string())
```
yhl25: same here.

rust/numaflow-core/src/metadata.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// Convert to protobuf metadata for UDF requests.
+    /// Preserves the original previous_vertex so UDFs know the actual sender.
+    pub(crate) fn to_proto(self) -> metadata::Metadata {
+        let previous_vertex = self.previous_vertex.clone();
+        self.to_proto_with_previous_vertex(previous_vertex)
+    }
+
+    /// Convert to protobuf metadata with a specific previous_vertex value.
+    /// Used internally by both conversion methods.
+    fn to_proto_with_previous_vertex(self, previous_vertex: String) -> metadata::Metadata {
```
yhl25: `to_proto` is always invoked with previous vertex right? Can we merge them and move it to from implementation?

rust/numaflow-pb/src/common/common.metadata.rs
```diff
@@ -0,0 +1,23 @@
+// This file is @generated by prost-build.
```
yhl25, line 1: rename file to `metadata.rs`?

rust/numaflow-core/src/source/user_defined.rs
```diff
@@ -171,7 +172,18 @@ impl TryFrom<read_response::Result> for Message {
             },
             headers: result.headers,
             watermark: None,
-            metadata: None,
+            metadata: Some(match result.metadata {
+                Some(source_metadata) => {
+                    let mut metadata: Metadata = source_metadata.into();
+                    metadata.previous_vertex = config::get_vertex_name().to_string();
```
vigith, line 178: is it okay to set "previous-vertex" as current vertex in source? it sounds wrong :)

yhl25: This is set in the result, which will be forwarded to the next vertex, so setting it to current vertex is correct.

PR #2926, opened 2025-09-02
chore: e2e tests for distributed throttling

yhl25:

### Related issues
Fixes #

### Testing

### DIFF
--- a/rust/numaflow-core/src/pipeline.rs
+++ b/rust/numaflow-core/src/pipeline.rs
@@ -659,36 +659,9 @@ async fn start_aligned_reduce_forwarder(
         tracker_handle,
     };
 
-    if let Some(rate_limit_config) = &config.rate_limit {
-        if should_use_redis_rate_limiter(rate_limit_config) {
-            let redis_config =
-                build_redis_rate_limiter_config(rate_limit_config, cln_token.clone()).await?;
-
-            run_reduce_forwarder::<WithRedisRateLimiter>(
-                &context,
-                reader_components.clone(),
-                reducer,
-                wal,
-                Some(redis_config.throttling_config),
-            )
-            .await?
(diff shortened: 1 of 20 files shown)


rust/numaflow-core/src/pipeline.rs
```diff
@@ -659,36 +659,9 @@ async fn start_aligned_reduce_forwarder(
         tracker_handle,
     };
 
-    if let Some(rate_limit_config) = &config.rate_limit {
```
vigith: where did this code go to?

rust/numaflow-throttling/src/lib.rs
```diff
@@ -42,7 +46,7 @@ pub struct RateLimit<W> {
     /// Max number of tokens ever filled till now. The next refill relies on the current max allowed
     /// value. For the first run this will be set to `burst` and will be slowly incremented to the
     /// `max` value.
-    max_ever_filled: Arc<AtomicUsize>,
+    max_ever_filled: Arc<Mutex<f32>>,
```
vigith, line 50: let's use `usize` and ciel to make sure it will be `usize`, do we care for `f32`. It has to be an absolute value.

rust/numaflow-core/src/pipeline.rs
yhl25: This is not required, because the rate limit is not applicable for reduce.

rust/numaflow-throttling/src/lib.rs
yhl25: `ceil` doesn't work because if the slope is 0.1 and ram up duration is 10s, the next token should be available at 10th second. If we use ceil it will be available in the 2nd second.

vigith (approved):

Issue #2924, opened 2025-09-02
Kafka source in rust hardcoded missing key value


d4munche3z:
Describe the bug
The built-in Kafka Source does not propagate the Kafka message key to the downstream Numaflow Datum's keys attribute. Messages that have a valid key in the Kafka topic arrive at the first vertex in a pipeline with an empty keys list. This prevents the use of keyed windowing (groupBy) in Reduce vertices without using a Source Transformer as a workaround to re-key the messages from the payload.
To Reproduce
Produce a Kafka message with a key: Using any standard Kafka producer, send a message to a topic, ensuring the message key is set. For example, using kafka-console-producer:
code
Bash
kafka-console-producer --bootstrap-server <kafka-broker>:9092 --topic my-keyed-topic --property "parse.key=true" --property "key.separator=:"
>my-key:{"data":"hello"}
Verify the key exists in Kafka: Using kafka-console-consumer, confirm that the message in the topic has a key.
code
Bash
kafka-console-consumer --bootstrap-server <kafka-broker>:9092 --topic my-keyed-topic --from-beginning --property print.key=true
# Expected output: my-key    {"data":"hello"}
Create a simple Numaflow pipeline: Configure a pipeline with a Kafka source reading from my-keyed-topic and a simple "inspector" Map UDF that logs the contents of the Datum it receives.
pipeline.yaml:
code
Yaml
apiVersion: numaflow.numaproj.io/v1alpha1
kind: Pipeline
metadata:
  name: key-test-pipeline
spec:
  vertices:
    - name: in
      source:
        kafka:
          brokers: ["<kafka-broker>:9092"]
          topic: "my-keyed-topic"
          consumerGroup: "key-test-group"
    - name: inspector
      udf:
        container:
          # An image for a simple inspector UDF (code provided below)
          image: my-inspector-udf:latest
  edges:
    - from: in
      to: inspector
inspector.py:
code
Python
import logging
from pynumaflow.mapper import Mapper, Messages, Datum, Message, MapServer

_LOGGER = logging.getLogger(__name__)

class Inspector(Mapper):
    def handler(self, keys: list[str], datum: Datum) -> Messages:
        _LOGGER.info(f"Received Message -> Handler Keys: {keys}, Datum Keys: {datum.keys}")
        return Messages(Message(value=datum.value, keys=datum.keys))

if __name__ == "__main__":
    logging.basicConfig(level="INFO")
    MapServer(Inspector()).start()
[...]

PR #2931, opened 2025-09-05
fix: Forward Kafka message keys

BulkBeing:
The kafka message keys were not getting forwarded when builtin kafka source is used.

### Related issues
Fixes https://github.com/numaproj/numaflow/issues/2924

### Testing
```yaml
kind: Pipeline
metadata:
  name: key-test-pipeline
spec:
  vertices:
    - name: in
      source:
        kafka:
          brokers: ["kafka:9092"]
          topic: "my-keyed-topic"
          consumerGroup: "key-test-group"
    - name: out
      scale:
        min: 1
      sink:
        log: {}

  edges:
  - from: in
    to: out
```

Sending message:
```
➜  ./kafka-console-producer.sh --bootstrap-server :9092 --topic my-keyed-topic --property "parse.key=true" --property "key.separator=:"
>my-key:{"data": "hello"}
>
```

Log sink:
```log
2025-09-05T05:23:37.788369Z  INFO numaflow_core::sink::log: Payload - {"data": "hello"} Keys - my-key EventTime - 1757049815378 Headers - X-NF-Kafka-TopicName: my-keyed-topic,  ID - out-2-0-0
```

None

### DIFF
--- a/rust/extns/numaflow-kafka/src/source.rs
+++ b/rust/extns/numaflow-kafka/src/source.rs
@@ -39,6 +39,8 @@ pub struct KafkaMessage {
     pub topic: String,
     /// The user payload.
     pub value: Bytes,
+    /// Key of the message
+    pub key: Option<String>,
     /// The partition number.
     pub partition: i32,
     /// The offset of the message.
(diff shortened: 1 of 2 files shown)


vigith (approved):

rust/extns/numaflow-kafka/src/source.rs
```diff
@@ -310,6 +312,7 @@ impl KafkaActor {
                     let message = KafkaMessage {
                         topic: message.topic().to_string(),
                         value,
+                        key: message.key().map(|k| String::from_utf8_lossy(k).to_string()),
```
hasmukhmparmar-27, line 315: Why not use ``Bytes`` type instead of ``String`` with lossy utf8? It would help user to handle errors!

PR #2935, opened 2025-09-09
fix: Rater should use ticker tick's timestamp for window

### LINKED ISSUE #2916: UI shows wrong processing rates just after re-deploying a pipeline
### Just after redeploying
Image

### Stabilizes after some time
Image

yhl25:
Closes #2916


pkg/mvtxdaemon/server/service/rater/rater.go
```diff
@@ -129,33 +130,38 @@ func NewRater(ctx context.Context, mv *v1alpha1.MonoVertex, opts ...Option) *Rat
...
 }
 
+// podTask represents a monitoring task for a pod with a common timestamp
+type podTask struct {
+	key       string
+	timestamp int64
+}
+
 // Function monitor() defines each of the worker's jobs.
 // It waits for keys in the channel, and starts a monitoring job
```
KeranYang:
```suggestion
// It waits for pod monitoring tasks in the channel, and starts them
```

pkg/mvtxdaemon/server/service/rater/pod_tracker.go
```diff
@@ -85,6 +85,9 @@ func (pt *PodTracker) Start(ctx context.Context) error {
 }
 
 func (pt *PodTracker) trackActivePods(ctx context.Context) {
+	// start updating active pods as soon as called and then after every refreshInterval
+	pt.updateActivePods()
```
KeranYang, line 89: Is this to start early so that the rater stabilize early?

pkg/daemon/server/service/rater/rater.go
```diff
@@ -264,7 +256,12 @@ func (r *Rater) getPodMetrics(vertexName, podName string) map[string]*dto.Metric
 		r.log.Warnf("[vertex name %s, pod name %s]: failed reading the metrics endpoint, the pod might have been scaled down: %v", vertexName, podName, err.Error())
 		return nil
 	}
-	defer resp.Body.Close()
+	defer func(Body io.ReadCloser) {
+		err := Body.Close()
+		if err != nil {
```
KeranYang, line 258: Thank you.

pkg/daemon/server/service/rater/rater.go
```diff
@@ -211,51 +217,37 @@ func (r *Rater) Start(ctx context.Context) error {
...
-					return result
-				}
-				return 1
-			}()))
+		case <-ticker.C:
+			// Generate a common timestamp for all pods in this tick
+			now := time.Now().Unix()
+			for range r.podTracker.GetActivePodsCount() {
+				assign(now)
+			}
```
KeranYang, line 243: If I understand correctly, this is the main fix where we assign the SAME timestamp to the entire group. Does it really help with getting accurate processing rate earlier than before? I thought the bottleneck was the time it takes for metrics being available.

yhl25: With the old approach, for a tick the metrics could end up in two different windows. Ideally all of them should belong to the same timestamp and we don't need to have 10s window, we just need to make sure we assign the same timestamp to all the pod metrics after fetching.

pkg/mvtxdaemon/server/service/rater/pod_tracker.go
yhl25: yes, to reduce the delay during boot up.

pkg/daemon/server/service/rater/rater.go
KeranYang: Is this PR to get accurate processing rate earlier than before? Have we verified that this approach helped and if so by how much?

KeranYang (approved):

PR #2937, opened 2025-09-10
test: simple attemp_acquire_n test

vigith:
Improve the testing framework for throttling. It is tricky to test throttling because time is a factor in throttling and it is not easy to manipulate time.


yhl25 (approved):
LGTM

PR #2939, opened 2025-09-12
chore: Remove Background Idle Watermark Publishing Logic

yhl25:

### Related issues
Fixes #

### Testing


yhl25:
Will create a separate PR to handle this correctly

PR #2940, opened 2025-09-12
test: Tests for verifying token refill computation

vaibhavtiwari33:
Previously added tests checked for the ability to pull for tokens at each subsequent epoch. 
This PR aims to add the ability to specify, in each test case, the number of epochs after which next set of tokens need to be pulled.

Added test scenarios include:
* Trying to acquire all tokens in the same epoch after all tokens have been fetched already.
* Trying to acquire all pending tokens in the same epoch after some tokens have been fetch already.
* Trying to acquire some tokens in the same epoch after some tokens have been fetch already. 

### Testing
Verified that the newly added unit tests are passing for both redis and in memory store.

### DIFF
--- a/rust/numaflow-throttling/src/lib.rs
+++ b/rust/numaflow-throttling/src/lib.rs
@@ -514,8 +514,9 @@ mod tests {
             pub(super) store_type: StoreType,
             // The name of the test
             pub(super) test_name: String,
-            // Tokens asked in each iteration by *each* pod
-            pub(super) asked_tokens: Vec<Option<usize>>,
+            // Tuple of tokens asked in each iteration by *each* pod
+            // and time in secs after which the tokens are asked.
+            pub(super) asked_tokens: Vec<(Option<usize>, usize)>,
             // Tokens expected to be returned by rate limiter in each iteration to *each* pod
             pub(super) expected_tokens: Vec<usize>,
         }
@@ -577,7 +578,7 @@ mod tests {
                 let mut total_expected_tokens = 0;
                 for rate_limiter in rate_limiters.iter() {
                     let tokens = rate_limiter
-                        .attempt_acquire_n(asked_tokens[i], cur_epoch)
+                        .attempt_acquire_n(asked_tokens[i].0, cur_epoch)
                         .await;
                     assert_eq!(
                         tokens, expected_tokens[i],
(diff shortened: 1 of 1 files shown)


rust/numaflow-throttling/src/lib.rs
```diff
@@ -636,6 +637,119 @@ mod tests {
             }
         }
 
+        /// A generic utility function to test rate limiter with generic state
+        /// Test runner for:
+        /// - [crate::tests::test_distributed_rate_limiter_multiple_pods_in_memory]
+        /// - [crate::tests::test_distributed_rate_limiter_multiple_pods_redis]
+        ///
+        async fn test_rate_limiter_with_state_acquire_n<S: Store + Sync>(
```
vaibhavtiwari33: Need to remove this duplicate fn

rust/numaflow-throttling/src/lib.rs
```diff
@@ -1499,6 +1541,19 @@ mod tests {
...
+            // Acquire all tokens at each epoch
+            // Immediately ask for tokens after first epoch
+            test_utils::TestCase {
+                max_tokens: 60,
+                burst_tokens: 15,
+                duration: Duration::from_secs(10),
+                pod_count: 2,
+                store_type: test_utils::StoreType::Redis,
+                test_name: "RedisStore Test params: max_tokens=60, burst_tokens=15, duration=10s, pod_count=2".to_string(),
+                asked_tokens: || -> _ { let mut a = vec![(None,1), (None,0)]; a.extend(vec![(None,1); 10]); a}(),
```
vaibhavtiwari33, line 1558: Is there a cleaner way to do this?

rust/numaflow-throttling/src/lib.rs
```diff
@@ -514,8 +514,9 @@ mod tests {
             pub(super) store_type: StoreType,
             // The name of the test
             pub(super) test_name: String,
-            // Tokens asked in each iteration by *each* pod
-            pub(super) asked_tokens: Vec<Option<usize>>,
+            // Tuple of tokens asked in each iteration by *each* pod
+            // and count of epochs after which the next set of tokens are asked.
+            pub(super) asked_tokens: Vec<(Option<usize>, usize)>,
```
vaibhavtiwari33: Second item in the tuple refers to the number of epochs after which the next set of tokens are going to be fetched, not the tokens specified in the current tuple. Please let me know if this is confusing, I can change the implementation to fetch the tokens specified in the current tuple.

Eg:
This test case: `[(None, 3), (Some(2), 1)]`, specifies that fetch all tokens at 0th epoch and fetch 2 tokens after 3 epochs.

vigith: add the above to the comment itself

rust/numaflow-throttling/src/lib.rs
vigith: i would create the a new var called `asked_tokens` before building the `TestCase` struct and then just use it.

yhl25 (approved):

rust/numaflow-throttling/src/lib.rs
vigith: nm.. this is vec! inside a vec!

PR #2947, opened 2025-09-15
fix: fallback sink validation

cosmic-chichu:
SQS was missing in fallback sink validation


pkg/apis/numaflow/v1alpha1/sink.go
```diff
@@ -150,5 +150,5 @@ func (s Sink) getFallbackUDSinkContainer(mainContainerReq getContainerReq) corev
 
 // IsAnySinkSpecified returns true if any sink is specified.
 func (a *AbstractSink) IsAnySinkSpecified() bool {
-	return a.Log != nil || a.Kafka != nil || a.Blackhole != nil || a.UDSink != nil
+	return a.Log != nil || a.Kafka != nil || a.Blackhole != nil || a.UDSink != nil || a.Serve != nil || a.Sqs != nil || a.Pulsar != nil
```
vigith: I am not really sure we can put `Serve` here. It looks good to me.

PR #2959, opened 2025-09-18
feat: Implement token deposition logic for unused tokens

### LINKED ISSUE #2913: deposit unused tokens to rate-limiter
For messages that are dropped, if required, user can ask the rate-limiter to exclude those tokens from the rate limit calculation. 

This is useful when users drop old data from a stream (without processing) and not be penalized by rate-limiter.

vaibhavtiwari33:
Closes #2913 

Change get_tokens for *WithState Ratelimiters* to utilize deposited tokens.

This PR implements using unused tokens that were deposited within the same epoch.
Implementation for modes that utilize unused tokens to compute_refill is present in #2962 

### Testing
All of the previously added unit tests are passing. New unit tests were added to test the get_tokens logic when depositing back unused tokens. 

### DIFF
--- a/rust/numaflow-throttling/src/lib.rs
+++ b/rust/numaflow-throttling/src/lib.rs
@@ -30,6 +30,11 @@ pub trait RateLimiter {
     /// available. If `n` is provided, it will try to acquire `n` tokens else it will acquire all the tokens.
     async fn acquire_n(&self, n: Option<usize>, timeout: Option<Duration>) -> usize;
 
+    /// Deposit tokens into the rate limiter.
+    /// This will be used to deposit tokens into the rate limiter.
+    ///
+    async fn deposit_unused(&self, n: usize);
+
     /// Shutdown the rate limiter and clean up resources.
     /// This will deregister the processor from the distributed store and stop any background tasks.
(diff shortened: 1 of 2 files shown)


rust/numaflow-throttling/src/lib.rs
```diff
@@ -283,6 +299,8 @@ impl RateLimiter for RateLimit<WithoutState> {
             .unwrap_or(0)
     }
 
+    async fn deposit_unused(&self, _: usize, _: u64) {}
```
vigith, line 303: fix me

rust/numaflow-throttling/src/lib.rs
```diff
@@ -362,6 +384,28 @@ impl<S: Store + Send + Sync + Clone + 'static> RateLimiter for RateLimit<WithSta
             .unwrap_or(0)
     }
```
vigith, line 386: can we directly put it back to token pool, this will make get_token way easier.

vaibhavtiwari33:
Will create a separate PR which doesn't utilize deposited_tokens variable. Will simply utilize (max_ever_refilled - token) as the tokens that were utilized.

PR #2984, opened 2025-09-26
chore: wait for serving pods to comeup before firing requests

yhl25:
* Fixes default metadata propagation when SDK's are running with older version.
* Updated the callbacks for source vertex to use empty `previous_vertex` instead of using the current vertex name.
* Serving code is updated to work for empty `previous vertex` for source vertices.
* Adds utility methods for serving e2e(wait for pods, logging etc), updated the test to wait for the pods to come up before making any requests.


rust/numaflow-core/src/mapper/map/user_defined.rs
```diff
@@ -505,7 +510,12 @@ impl From<UserDefinedMessage<'_>> for Message {
             headers: value.1.headers.clone(),
             watermark: None,
             is_late: value.1.is_late,
-            metadata: value.0.metadata.map(|m| m.into()),
+            metadata: match value.0.metadata {
```
adarsh0728, line 513: LGTM, tested!

config/apps/kafka/kafka-minimal.yaml
```diff
@@ -64,7 +64,7 @@ spec:
     spec:
       containers:
         - name: main
-          image: bitnami/kafka:3.7
+          image: bitnami/kafka
```
vigith: let's fix this via https://github.com/numaproj/numaflow/pull/2988 first?

vigith (approved):

PR #2985, opened 2025-09-26
feat: Implement immediate ramp up during re-deployments

### LINKED ISSUE #2982: Immediate Ramp up during re-deployments
# Summary

We need to keep state of each pod while deregistering and then use that information during registration to recompute the initial `max_ever_refilled`. This computation is only required if the user specifically starts the Rate Limter with `Resume` feature which takes TTL which is mandatory. If `Resume` is not set, this optional option is disabled (default behavior).

## Assumption

- We do not differentiate between different deployment strategies, we only take "time" into consideration. That means, we will ignore state, say, pod takes more than X mins to come online. In other words, everything is based on TTL.

# Use Cases

- During deployments we need to make sure the rate-limiter comes to steady state ASAP because the system was already running at full throttle
- During Pause/Resume, we need to honor **SLOW** ramp up
- During a pod restart, we should get to steady state of that pod ASAP, esp. when using longer ramp up config.

---

vaibhavtiwari33:

Previously if the pipeline restarts due to re-deployment then, this would re-trigger the ramp-up of the pipeline, i.e., it would be throttled from its min token configuration and slowly ramped-up to the max token capacity again. 

This PR adds an option to enable the ability to resume the pipeline from previously achieved TPS.
Thus, if the pipeline is restarted within `ttl` duration with the `resumedRampUp` option enabled, then it will resume from the previously achieved token capacity. 

spec:
```
    rateLimit:
      resumedRampUp: false
      ttl: "180s"
```
---
### Related issues

Closes #2982

---
### Testing
<details>

<summary>

#### Unit Tests: 

</summary>

Added individual unit tests to verify:
- Updated in-memory/redis store functionality during registration/deregistration of processor
- `attempt_acquire_n` max tokens after the processor is re-registered before/after `ttl`
- `attempt_acquire_n` max tokens after multiple processors are re-registered before/after `ttl`

</details>

##### Example of applied spec:

Screenshot 2025-10-01 at 10 38 45 AM

#####  Above spec applied with resumedRampUp. If ttl is not provided, it is defaulted to 180s (3m):

Screenshot 2025-10-01 at 10 08 54 AM

#####  Steady state of the pipeline:

Screenshot 2025-10-01 at 10 09 11 AM

#####  Pipeline ramp-up behaviour **without** resumed ramp-up:

Screenshot 2025-10-01 at 10 40 56 AM

#####  Pipeline ramp-up behaviour **with** resumed ramp-up:

WthResumePrometheusMetric

#####  Pipeline ramp-up behaviour **with** resumed ramp-up with Gap > TTL :

Screenshot 2025-10-01 at 4 48 11 PM

---

<details>

<summary>

#### ~Different refresh_interval from other tests~ (fixed, see edit)

Edit: This was fixed by adding a handler for the background task and calling abort on the handler during shutdown. 

</summary>

[This unit test](https://github.com/numaproj/numaflow/blob/4024cc027f3aff2625bd83cd08040daccbc3dfb8/rust/numaflow-throttling/src/lib.rs#L4773) added for redis store to test the following scenario:
- Start two processors in scheduled mode
- Move their `max_ever_filled` to different levels by calling `attempt_acquire_n` for different epochs for both processors (eg: `rate_limiter1.attempt_acquire_n(None, 5); rate_limiter2.attempt_acquire_n(None, 2);`) 
- shutdown both processors
- Restart only one processor
- Calling `attempt_acquire_n` on rate_limiter1 should lead get max_ever_filled of rate_limiter1

Here, refresh interval was changed from 100ms to 1000ms, otherwise sync loop of the shutdown rate_limiter1 was modifying `desired_pool_size` of restarted rate_limiter1

</details>

<details>

<summary>

#### ~Noticing additional pods not coming up in resumed pipeline~ (Discussed, this is a side-effect)

</summary>

Notice the above two SSs of metrics page of two pipelines, one with and other without resume enabled. 
The one with `resume` enabled is in steady state and gets additional pods spawned before the pipeline shutsdown, but not after it, while this behaviour is consistent in the case of pipeline with `resume` disabled.

Following metric diagram showcases this as well:

Screenshot 2025-10-01 at 4 59 06 PM

</details>


rust/numaflow-throttling/src/state/store/lua/set_steady_state.lua
```diff
@@ -0,0 +1,24 @@
+--[[
+Description:
+  Simply sets the steady state status of the processor in redis.
+
+KEYS array:
+  KEYS[1] - The base key prefix (e.g., "pipeline:vertex")
+
+ARGV array:
+  ARGV[1] - The name of the processor (e.g., "processor-alpha-1")
+  ARGV[2] - Whether the processor is in steady state (true/false)
```
vigith: shouldn't this value be `max_ever_refilled` for that processor? with just `true/false` how can we calculate the initial token size?

vaibhavtiwari33: Here I was trying to only track whether the processor had reached the max amount of tokens allowed or not. 
- If the processor had finished ramp-up and was utilizing max tokens allowed
  - then the redeployment will start back from max tokens, otherwise,
- If the processor was in the middle of ramping-up and had not reached a steady state
  - then a redeployment would lead to the processor to initialize with burst again. 

This approach allows limiting calls made to the store as opposed to tracking max_ever_filled at every refill.

Let me know if this approach is limiting and instead we want to track where in the ramp-up cycle was the processor when the redeployment was triggered.

vigith: will it make it easier, if in deregister we write the max filled, so we will know what to resume from?

vaibhavtiwari33: That's a good point, yes, I think that would be a better option. Let me make the changes to see if we can incorporate that strategy instead.

vaibhavtiwari33: I can simply use the redis TTL to let the max_ever_filled value expire for a processor after it has been added during de-registration.

vigith: Older version of Redis did not support TTL at key level in a dict. I am not sure whether 6.x/7.x supports.

vaibhavtiwari33: Seems like the support arrived in 7.4+.

vigith: is redis 7.4 supported in AWS?

vigith: We cannot use 7.4 solution because of licensing issues.

> The core Redis software for version 7.4 is not open source under the traditional open-source definition; it is dual-licensed under the `Redis Source Available` License  and the `Server Side Public` License, which are considered source-available and non-free licenses.

vaibhavtiwari33: I see

vaibhavtiwari33: I've tested this on v8.2.1, which is open source. 
Which version of Redis are we relying on currently?

vigith: I think, AWS supports <= 7.2

vaibhavtiwari33: In that case we'll have to track TTL for the max_ever_filled keys in redis ourselves. We can have a logic similar to the way heartbeats are handled.

vigith: yes, shouldn't be complex, just store time on deregister and when using it while register, check whether it is stale.

vaibhavtiwari33: yup

rust/numaflow-throttling/src/state/store/lua/register.lua
```diff
@@ -2,24 +2,33 @@
...
   ARGV[1] - The name of the processor (e.g., "processor-alpha-1")
+  ARGV[2] - The stale age in seconds
 
 Returns:
   ZCARD(heartbeats) which is the value stored at poolsize_key for this processor(heartbeats includes this processor)
-]]--
+]] --
+
+-- Define the stale age in seconds
+local STALE_AGE = tonumber(ARGV[2])
```
vaibhavtiwari33, line 20: @vigith Currently I'm utilizing the same STALE_AGE used for ~refresh interval of sync pool~ processor key TTL in stores as the stale age for determining if the processor should be resumed. Is this acceptable or should this be a separate param?

vigith: we can use the same

rust/numaflow-core/src/config/components/ratelimit.rs
```diff
@@ -16,6 +16,10 @@ pub(crate) struct RateLimitConfig {
     pub(crate) store: Option<Box<numaflow_models::models::RateLimiterStore>>,
     /// Optional modes for rate limiting.
     pub(crate) modes: Option<Box<numaflow_models::models::RateLimiterModes>>,
+    /// Resume ramp up for a processor after pause.
+    pub(crate) resume: bool,
+    /// TTL for the rate limiter state in seconds.
+    pub(crate) ttl: usize,
```
vigith, line 22: we can make resume Optional with ttl as duration, that way with one field we can achieve both.

vaibhavtiwari33: Got it, yes, we can do that. I separated it out here since we're using this [ttl](https://github.com/numaproj/numaflow/pull/2985#discussion_r2388315355) for stale pruning based on heartbeats as well. I can utilize the default value when resume is disabled, otherwise use the one specified.

vigith: that is a fair point, let's not merge it then.

rust/numaflow-throttling/src/state/store/lua/deregister.lua
vigith, line 12: please document ARGV[2]

rust/numaflow-throttling/src/state/store/lua/register.lua
```diff
@@ -30,6 +39,31 @@ local new_pool_size = heartbeat_count
...
+    -- The unpack function passes the table elements as individual arguments to ZREM
+    redis.call('ZREM', prev_max_filled_ttl_key, unpack(stale_members))
+    redis.call('HDEL', pre_max_filled_key, unpack(stale_members))
+end
+
+-- Step 3: If this processor was recently deregistered, get the prev_max_filled for it
+-- Check if there exists a prev_max_filled for the processor
+-- return that if it exists, otherwise return -1
+local stored_prev_max_filled = redis.call('HGET', pre_max_filled_key, ARGV[1])
+local prev_max_filled = -1
```
vigith: why not return 0?

vaibhavtiwari33: Since 0 is a possible legitimate value that can be held by `max_ever_filled` in a corner case, I chose -1. We can change this to 0, if that makes it cleaner

vigith: we do not have to differentiate between zero-value or real zero, right? in any case, we can set that value has the initial value.

vaibhavtiwari33: True, the effect in the corner case mentioned would be the same anyways. I'll change this to 0, since it is anyways going to be cast as usize:
```
(*state.prev_max_filled as usize).clamp(token_calc_bounds.min, token_calc_bounds.max)
```

rust/numaflow-throttling/src/state/store/lua/deregister.lua
vaibhavtiwari33: Done

rust/numaflow-throttling/src/state/store/lua/register.lua
vaibhavtiwari33: Updated

vigith (commented):
let's reduce the tests which requires `Sleep(...)`

vaibhavtiwari33:
@vigith For [this](https://github.com/numaproj/numaflow/pull/2985#pullrequestreview-3297040950), removed the 2/6 newly added tests with sleep in them. Reduced the sleep duration in the remaining tests and they're are kept for the following cases:
* 1 test for verifying expiry behaviour in `redis.store`
* 1 test for verifying expiry behaviour in `in_memory.store`
* 2 tests in `lib.rs` for verifying expiry behaviour with redis and in-memory store each.
> let's reduce the tests which requires Sleep(...)

rust/numaflow-throttling/src/state/store/in_memory_store.rs
vigith, line 21: why is this tokio mutex? are we calling `async` functions in in-memory store?

vaibhavtiwari33: Updated to use std mutex instead of tokio mutex. Although we are calling async functions in in-memory store, it is safe to use std mutex here since the locks aren't being held across `.await`

vigith: what about in Redis store.. are we using locks across await points? can we avoid that?

vaibhavtiwari33: Note: Did not convert the `compute_consensus` function to a method [here](https://github.com/numaproj/numaflow/blob/f38989f398a2eb75e6354eaca1dd220510ab32e9/rust/numaflow-throttling/src/state/store/in_memory_store.rs#L86) since that required explicitly dropping [this mutex guard](https://github.com/numaproj/numaflow/blob/f38989f398a2eb75e6354eaca1dd220510ab32e9/rust/numaflow-throttling/src/state/store/in_memory_store.rs#L187) before calling `compute_consensus` to prevent deadlock. Let me know if we want to do that.

vaibhavtiwari33: In the case of redis store, we're not using locks to access any resource and directly awaiting on the lua scripts.

vaibhavtiwari33: One place where I found usage of std mutex before (not across) a `.await` call was [in the recently added background task handler](https://github.com/numaproj/numaflow/blob/67c7c0fc4c14bd4b2739c50405869dcf034990c4/rust/numaflow-throttling/src/state.rs#L219-L224)

This, should not be an issue though since the dereferenced mutexGuard is not used once abort is called, so the compiler should drop it after that. Furthermore, we're not using that resource in the subsequent deregister call.

rust/numaflow-throttling/src/state/store.rs
```diff
@@ -10,11 +10,20 @@ pub mod redis_store;
 pub trait Store: Clone + 'static {
     /// Register the consumer with the external store. It increments the pool size by 1, and returns
     /// the initial pool size, consensus cannot be reached during registration.
-    async fn register(&self, processor_id: &str, cancel: CancellationToken)
-    -> crate::Result<usize>;
+    /// TODO: Shall the return be changed to a struct?
```
vigith: explain the new signature and remove TODO

vigith (approved):

rust/numaflow-throttling/src/state/store.rs
vaibhavtiwari33: Sry, just saw the comment. ty for addressing

PR #2987, opened 2025-09-29
doc: Document throttling modes

vaibhavtiwari33:
Add documentation for the different throttling modes with explanation and examples


vaibhavtiwari33:
~The failing check seems to be for the following two links in the quick start guide:
[403] https://docs.docker.com/desktop/kubernetes/
[403] https://docs.docker.com/get-docker/~

Not happening anymore

yhl25 (approved):

PR #2996, opened 2025-10-09
fix: autoscaling for low throughput non-source vertices

whynowy:

Today, for UDF and sinks, we use buffer availability analysis for autoscaling. This has challenges for low throughput data processing - the typical symptom is, with 80-90% of target buffer availability, 30K buffer length, and a 1-2/second throughput, it takes a long time to fill in the buffer, and trigger scaling up, especially when it starts with less replicas (e.g 1). 

```
desiredReplicas = targetBufferAvailability / (currentBufferAvailability / currentReplicas)
```

When it starts with less replicas (for example 1), the denominator becomes very close to the targetBufferAvailability, which makes the calculated desired replicas still 1. 

This would not be a problem when the currentBufferAvailability is small, or currentReplicas is large, which makes the denominator smaller, thus the autoscaling behavior would be more sensitive.

The idea for this change is, before the buffer availability is hitting some sort of threshold, we do not use buffer availability for analysis, instead using the targetProcessTime, just like the source vertex, or MonoVertex. Here the threshold we take is the targetBufferAvailability.

**Concerns**

One concern would be, when switching from using targetProcessTime to targetBufferAvailability, would there be a jump on the calculated desired replicas. Tested in the normal scenario (e.g. targetBufferAvailability=90%, targetProcessTime=20seconds), the transition is smooth. For sure this won't be always smooth for any configuration (e.g. very small targetProcessTime like 1 second, and small targetBufferAvailability like 1%), but with the configurations like replicasPerScaleUp/Down, it should still be under control.

### DIFF
--- a/pkg/reconciler/vertex/scaling/scaling.go
+++ b/pkg/reconciler/vertex/scaling/scaling.go
@@ -408,6 +408,7 @@ func (s *Scaler) desiredReplicas(_ context.Context, vertex *dfv1.Vertex, partiti
 			// we cannot go with current replicas because ideally we should scale down when pending is 0 or rate is 0.
 			continue
 		}
+		// For Soruce and non-source vertices which do not have packed buffers, use the same scaling mechanism
 		if vertex.IsASource() || pending <= bufferLength-targetAvailableBufferLength {
 			// For sources, we calculate the time of finishing processing the pending messages,
 			// and then we know how many replicas are needed to get them done in target seconds.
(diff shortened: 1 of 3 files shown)


pkg/reconciler/vertex/scaling/scaling.go
```diff
@@ -408,19 +408,20 @@ func (s *Scaler) desiredReplicas(_ context.Context, vertex *dfv1.Vertex, partiti
 			// we cannot go with current replicas because ideally we should scale down when pending is 0 or rate is 0.
 			continue
 		}
-		if vertex.IsASource() {
+		if vertex.IsASource() || pending <= bufferLength-targetAvailableBufferLength {
```
vigith, line 412: please document why we add this `pending <= ..` for posterity

PR #3007, opened 2025-10-15
docs: Update monovertex doc with information regarding supporting map udf

### LINKED ISSUE #2997: document map UDF in mono-vertex
Document how to use map UDF in mvtx

vaibhavtiwari33:
Appends to monovertex doc relevant information regarding usage of map udf in monovertex.

### Related issues
Closes #2997
Blocked by #3001

### DIFF
--- a/docs/core-concepts/monovertex.md
+++ b/docs/core-concepts/monovertex.md
@@ -49,7 +49,7 @@ few examples where you cannot use MonoVertex.
 
 ## Anatomy of MonoVertex
 
-MonoVertex supports the same Sources, Sinks, and Transformers which are used in the Pipeline spec.
+MonoVertex supports the same Sources, Map UDF, Sinks, and Transformers which are used in the Pipeline spec.
 
 ```yaml
 apiVersion: numaflow.numaproj.io/v1alpha1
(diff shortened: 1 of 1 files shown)


docs/core-concepts/monovertex.md
```diff
@@ -71,6 +68,25 @@ spec:
...
+kind: MonoVertex
+metadata:
+  name: mono-vertex-with-map
+spec:
+  source:
+    # same as the Pipeline Source
+    # ...
+  udf:
+    # same as the Pipeline Map
+    # ...
```
vigith:
```suggestion
  udf: # optional Map: same as the Pipeline Map
    # ...
```

also, just add this to the above yaml.. you do not have to create a new example spec

vaibhavtiwari33: Although it is the final state, since currently source transformer and udf aren't working together, I didn't want to add them to the same example. Let me know if we want to keep the doc based on the target state instead of current state.

vigith: that is fine.. Actually, @yhl25 is almost done with the tracker changes, he demo'ed that both are working together today morning

PR #3015, opened 2025-10-18
fix: Progress watermark for a source idling from the start

### LINKED ISSUE #3006: Support idleSource detection for empty sources
# Summary

[idleSource](https://numaflow.numaproj.io/core-concepts/watermarks/#idle-detection) should be able to progress the watermark on a source, in situations where the source has never received any data.

Currently, a watermark on the source is not set (or technically initiated to epoch 0) until the first event with event-time is received. `idleSource` is hereafter able to detect gaps in data of duration `threshold`. But, if a first event is never received, `idleSource` has no effect.

The proposal is, that idleSource can set the watermark equal to processing-time (wall-clock time) after `threshold` time has passed in the processor. (Possible `threshold+maxDelay`?).

Numaflows watermark-design is based on event time, and a coupling the process-time should generally be avoided. However, the `idleSource` is already using process-time as the max bound a watermark can progress to. Especially in settings where `stepInterval: 0s` (default setting), the watermark is effectively already incremented at instant speed to the current wall-time (or at least very near due to the finite `incrementBy`).

Add a config setting to enable this feature.

(Need to be aware of the general startup time of pipeline/pods - How to avoid that normally starting pipeline is not overrun by this feature, causing data-loss from all input streams. Should at least one stream have watermark set before this feature can trigger?)

# Use Cases

For pipelines with two or more sources joined into and accumulator or reducer vertex.
Where accumulator/reducer vertex has limited memory resources (e.g. `scale.max: 1` for ordering-guarantee and `limits.memory` set).
In the situation where the pipeline is starting up, but one of the multiple sources does not generate data, all other source will keep feeding data greedily into the accumulator/reducer, eventually causing OutOfMemory/Crashloopbackoff. The the accumulator/reducer pod restart, it will again greedily read all data from the ISB stream and crash again. Creating/recreating pipelines from current-source-offsets, thus only works, if all sources has data at the simultaneous starting point.

# Reproducible Example / Test

The following example can be used to reproduce the issue and test the new feature.

- Install numaflow 1.5.1 (helm 0.2.1) for accumulator support.
- Clone [accumulator/streamsorter](https://github.com/numaproj/numaflow-go/tree/main/examples/accumulator/streamsorter)
- Add `idleSource` and `limits` in the manifest:
```yaml
kind: Pipeline
metadata:
  name: stream-sorter
spec:
  watermark:
    idleSource:
      threshold: "5s"
      incrementBy: "3s"
      stepInterval: "2s"
  limits:
    readBatchSize: 1
  vertices:
    - name: http-one
      scale:
        min: 1
        max: 1
      source:
        http: {}
    - name: http-two
      scale:
        min: 1
        max: 1
      source:
        http: {}
    - name: accum
      udf:
        container:
          image: quay.io/numaio/numaflow-go/stream-sorter:stable
          resources:
            limits:
              memory: 64Mi
            requests:
              memory: 32Mi
        groupBy:
          window:
            accumulator:
              timeout: 10s
          keyed: true
          storage:
            persistentVolumeClaim:
              volumeSize: 1Gi
      containerTemplate:
        resources:
          limits:
            memory: 64Mi
          requests:
            memory: 32Mi
      scale:
        min: 1
        max: 1
    - name: sink
      scale:
        min: 1
        max: 1
      sink:
        log: {}
  edges:
    - from: http-one
      to: accum
    - from: http-two
      to: accum
    - from: accum
      to: sink
```
- `kubectl apply -f manifest/stream-sorter-pipeline.yaml` and let the pipeline start.
- `kubectl port-forward stream-sorter-http-one-0-xxxx 8444:8443` (only http-one)
- `yes '{"key":"value"}' | head -c 100000 > payload_100kB.json`
- `watch -n 0.1 'curl -kq -X POST --data-binary @payload_100kB.json https://localhost:8444/vertices/http-one'`
- The data never reaches sink. After ~1min the accumulator memory is reached and CrashLoopReboot occurs.

The expected output is that all data from http-one reaches sink.

(Maybe a caveat on how the new feature can cause data loss in beginning of even the earliest producing stream, here http-one.)

Image

---

vaibhavtiwari33:

Related issue describes in quite detail the problem statement this PR aims to solve. 

## Related issues
Fixes #3006

## Testing

### Test with local pipeline

Spec utilized:
```yaml
spec:
  vertices:
    - name: http-one
      scale:
        min: 1
        max: 1
      source:
        http: {}
    - name: input-two
      source:
        generator:
          rpu: 50
          duration: 1s
          keyCount: 1
    - name: sorter
      udf:
        container:
          image: quay.io/numaio/numaflow-go/stream-sorter:stable
        groupBy:
          window:
            accumulator:
              timeout: 10s
          keyed: true
          storage:
            persistentVolumeClaim:
              volumeSize: 1Gi
    - name: sink
      scale:
        min: 1
        max: 1
      sink:
        log: {}
  edges:
    - from: http-one
      to: sorter
    - from: input-two
      to: sorter
    - from: sorter
      to: sink
```

* With one source never producing any data, the pipeline is stuck with no data emitted to sink:

WithoutchangesPipeline

WithoutchangesSink

* After adding the watermark idleConfig (without InitSourceDelay), the pipeline remains stuck:

```yaml
watermark:
    maxDelay: 0s
    idleSource:
      threshold: 5s
      stepInterval: 2s
      incrementBy: 1s
```
BeforeSpec

WithChangesPipeline

WithChangesSink

* Initializing the InitSourceDelay allows progression of watermark for inactive source:

```yaml
watermark:
    maxDelay: 0s
    idleSource:
      threshold: 5s
      stepInterval: 2s
      incrementBy: 1s
      initSourceDelay: 5s
```

AfterSpec
WithSourceInitDelayPipeline

WithSourceInitDelaySink

### DIFF
--- a/docs/user-guide/reference/side-inputs.md
+++ b/docs/user-guide/reference/side-inputs.md
@@ -44,7 +44,7 @@ spec:
 
 To use the `Side Inputs` feature, a User-defined function implementing an interface defined in the Numaflow SDK
 ([Go](https://github.com/numaproj/numaflow-go/blob/main/pkg/sideinput/),
-[Python](https://github.com/numaproj/numaflow-python/blob/main/pynumaflow/sideinput/),
+[Python](https://github.com/numaproj/numaflow-python/blob/main/packages/pynumaflow/pynumaflow/sideinput/),
 [Java](https://github.com/numaproj/numaflow-java/tree/main/src/main/java/io/numaproj/numaflow/sideinput))
 is needed to retrieve the data.
 
@@ -75,7 +75,7 @@ func handle(_ context.Context) sideinputsdk.Message {
 }
 ```
 
-Similarly, this can be written in [Python](https://github.com/numaproj/numaflow-python/blob/main/examples/sideinput/simple_sideinput/example.py)
+Similarly, this can be written in [Python](https://github.com/numaproj/numaflow-python/blob/main/packages/pynumaflow/examples/sideinput/simple_sideinput/example.py)
 and [Java](https://github.com/numaproj/numaflow-java/blob/main/examples/src/main/java/io/numaproj/numaflow/examples/sideinput/simple/SimpleSideInput.java) as well.
 
 After performing the retrieval/update, the side input value is then broadcasted to all vertices that use the side input.
@@ -109,5 +109,5 @@ sideInputFileName -> "/var/numaflow/side-inputs/sideInputName"
 
 Here are some examples of watching the side input filesystem for changes in
 [Golang](https://github.com/numaproj/numaflow-go/blob/main/examples/sideinput/simple_sideinput/udf/main.go),
-[Python](https://github.com/numaproj/numaflow-python/blob/main/examples/sideinput/simple_sideinput/udf/example.py) and
+[Python](https://github.com/numaproj/numaflow-python/blob/main/packages/pynumaflow/examples/sideinput/simple_sideinput/udf/example.py) and
 [Java](https://github.com/numaproj/numaflow-java/tree/main/examples/src/main/java/io/numaproj/numaflow/examples/sideinput/udf).
--- a/docs/user-guide/sinks/fallback.md
+++ b/docs/user-guide/sinks/fallback.md
@@ -64,4 +64,4 @@ based on the use case.
 * a **fallback** response doesn't respect the sink retry strategy. It gets immediately directed to the fallback sink without getting retried.
 
 SDK methods to generate either a fallback or a failed response in a primary user-defined sink can be found here:
-[Golang](https://github.com/numaproj/numaflow-go/blob/main/pkg/sinker/types.go), [Java](https://github.com/numaproj/numaflow-java/blob/main/src/main/java/io/numaproj/numaflow/sinker/Response.java), [Python](https://github.com/numaproj/numaflow-python/blob/main/pynumaflow/sinker/_dtypes.py)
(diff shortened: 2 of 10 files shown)


rust/numaflow-core/src/watermark/idle/source.rs
```diff
@@ -50,6 +50,25 @@ impl SourceIdleDetector {
             >= self.config.threshold.as_millis() as i64
     }
 
+    fn get_source_idling_from_init_wm(&mut self) -> i64 {
+        if let Some(init_source_delay) = self.config.init_source_delay
+            && Utc::now().timestamp_millis() - self.updated_ts.timestamp_millis()
+                >= init_source_delay.as_millis() as i64
+        {
+            let now = Utc::now();
```
yhl25: avoid branching by using `let Some(x) = x else {}` ?

vaibhavtiwari33: Sure, I can do that, it required returning -1 twice with the second conditional, so it looked a little ugly to me, that's all. Let me change back

vaibhavtiwari33: Updated

rust/numaflow-core/src/watermark/idle/source.rs
```diff
@@ -50,6 +50,27 @@ impl SourceIdleDetector {
...
+
+        let Some(init_source_delay) = self.config.init_source_delay else {
+            return -1;
+        };
+
+        if now.timestamp_millis() - self.updated_ts.timestamp_millis()
+            >= init_source_delay.as_millis() as i64
+        {
+            let idle_wm = now.timestamp_millis() + init_source_delay.as_millis() as i64;
+            self.last_idle_wm_published_time = now;
```
vigith, line 66: why are we adding? shouldn't it be just "now"?

vaibhavtiwari33: That's a good point. Let me fix this

PR #3021, opened 2025-10-21
refactor: Sink Component

yhl25:
* Improves readability of sink component.
* Introduces separate modules for actor and builder code.
* Added support for exponential retry in our backoff module.
* Updated the sink code to use the exponential retry from the backoff.
* Removed the duplicate code in sink and fallback sink.

### DIFF
--- a/rust/numaflow-core/src/sinker/actor.rs
+++ b/rust/numaflow-core/src/sinker/actor.rs
@@ -131,42 +131,48 @@ where
                 });
             }
 
-            // Get next backoff delay
-            if let Some(delay) = backoff_iter.next() {
-                retry_attempt += 1;
-                warn!(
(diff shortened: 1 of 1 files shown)


rust/numaflow-core/src/sinker/actor.rs
```diff
@@ -0,0 +1,193 @@
...
+                    }
+                    OnFailureStrategy::Drop => {
+                        warn!(
+                            retry_attempts = ?retry_attempt,
+                            errors = ?error_map,
+                            "Retries exhausted, dropping messages."
+                        );
+                        dropped_messages.append(&mut messages_to_retry);
+                    }
+                    OnFailureStrategy::Retry => {
```
vigith: improve readability

PR #3038, opened 2025-10-23
chore: refactor sink response to remove serve payload

yhl25:
* Serve response is specific to `ResponseStatusFromSink::Serve` enum. It should not part of `ResponseFromSink`

### DIFF
--- a/rust/numaflow-core/src/sinker/sink.rs
+++ b/rust/numaflow-core/src/sinker/sink.rs
@@ -174,11 +174,6 @@ impl SinkWriter {
 
                 // Main processing loop
                 while let Some(batch) = chunk_stream.next().await {
-                    // empty batch, we can skip and continue
-                    if batch.is_empty() {
-                        continue;
-                    }
-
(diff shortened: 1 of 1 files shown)


rust/numaflow-core/src/sinker/sink.rs
```diff
@@ -174,7 +174,7 @@ impl SinkWriter {
 
                 // Main processing loop
                 while let Some(batch) = chunk_stream.next().await {
-                    // empty batch means end of stream
+                    // empty batch, we can skip and continue
                     if batch.is_empty() {
                         continue;
                     }
```
vaibhavtiwari33: (nit)
The doc for [`chunks_timeout`](https://docs.rs/tokio-stream/latest/tokio_stream/trait.StreamExt.html#method.chunks_timeout) states that:
> The length of the returned vector is never empty

Should be safe to remove this check here?

yhl25: yes, removed.

vigith (approved):

PR #3056, opened 2025-10-30
chore: default to use structured logging

whynowy:
Default to use json format for the logs.

`NUMAFLOW_DEBUG` is `true`:

Screenshot 2025-10-30 at 12 10 30 AM

Otherwise default to:

Screenshot 2025-10-30 at 12 11 38 AM

Still respects `RUST_LOG` if it's configured.

### DIFF
--- a/test/monovertex-e2e/monovertex_test.go
+++ b/test/monovertex-e2e/monovertex_test.go
@@ -71,9 +71,9 @@ func (s *MonoVertexSuite) TestExponentialBackoffRetryStrategy() {
 	w := s.Given().MonoVertex("@testdata/mono-vertex-exponential-retry-strategy.yaml").When().CreateMonoVertexAndWait()
 	defer w.DeleteMonoVertexAndWait()
 	w.Expect().MonoVertexPodsRunning()
-	firstRetryLog := fmt.Sprintf("retry_attempt=%d", 1)
-	secondRetryLog := fmt.Sprintf("retry_attempt=%d", 2)
-	thirdLog := fmt.Sprintf("retry_attempt=%d", 3)
+	firstRetryLog := fmt.Sprintf(`"retry_attempt":"%d"`, 1)
+	secondRetryLog := fmt.Sprintf(`"retry_attempt":"%d"`, 2)
+	thirdLog := fmt.Sprintf(`"retry_attempt":"%d"`, 3)
 	dropLog := "Retries exhausted, dropping messages."
 	w.Expect().MonoVertexPodLogContains(firstRetryLog, PodLogCheckOptionWithContainer("numa"))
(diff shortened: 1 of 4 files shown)


vigith (approved):

test/monovertex-e2e/monovertex_test.go
```diff
@@ -102,7 +102,7 @@ func (s *MonoVertexSuite) TestMonoVertexRateLimitWithRedisStore() {
 		_ = client.Close()
 	}()
 
-	w.Expect().MonoVertexPodLogContains("processed=50", PodLogCheckOptionWithContainer("numa"), PodLogCheckOptionWithCount(20))
+	w.Expect().MonoVertexPodLogContains("\"processed\":\"50", PodLogCheckOptionWithContainer("numa"), PodLogCheckOptionWithCount(20))
```
vigith, line 105: should we use JSON decoder? It might be cleaner and extensible. WDYT?

whynowy: Lots of changes, not worth of doing it...

th0ger:
Why does the log format (text/structured) depend on the log level (`NUMAFLOW_DEBUG`)?

whynowy:
`NUMAFLOW_DEBUG` is not mainly for log level, but if it's in dev/debug mode, to determine the log's human readability.  Any concern about it? @th0ger

th0ger:
Hi @whynowy.

We set `NUMAFLOW_DEBUG` on selected vertices (both dev + prod) to increase visibility with `DEBUG` logs.
I was always a bit puzzled why `DEBUG` has it's own setting, and the syntax was not generally `NUMAFLOW_LOG_LEVEL: "DEBUG"`.

If it's not "mainly for log level", then we may have misinterprepreted the intent of `NUMAFLOW_DEBUG`. I think for good reason, because the documentation says "To enable debug logs in a Vertex Pod, set environment variable `NUMAFLOW_DEBUG`". The flag name also implicitly guides you to think this is about debug log level. `NUMAFLOW_DEBUG_MODE` could have been a better name.

I just realised that it also enables [profiling](https://numaflow.numaproj.io/development/debugging/#profiling), which may not be optimal for production, do you agree?

With the new structured logging, a third behavior of `NUMAFLOW_DEBUG` is introduced. Maybe it's not best practice to have a single flag control multiple features. `NUMAFLOW_PPROF` is a good example of one-flag-one-feaure.

How do we set log level (to e.g. DEBUG or WARNING) in production?
How do we ensure that logs on prod are consistently structured for external export, regardless of the log level?
Do you recommend NOT setting `NUMAFLOW_DEBUG` on production, but only treat it as a developer productivity? Then please document that clearly.

(Side note: We want to extract `numa` logs with OTEL but are not there yet do to previous golang autoinstrumentation issues. Now with rust dataplane we should try again.)

PR #3067, opened 2025-11-06
test: Add e2e test for on-success-sink

vaibhavtiwari33:

As part of completing testing for the on-success sink feature, we want to add an end-2-end test which creates a pipeline with on-success UD sink and validates the messages being written to it. 
This PR aims to add this test and the manifest which the test uses to create the said pipeline as well as perform a basic validation based on the testing blueprint for fallback sink.

## Related issues
Closes: #3046 

## Testing
This is the test

### Blocker
Requires release of the next version of any one of the SDKs to successfully work.
Currently built go-sdk image with on-success sink changes, used in the manifest, fails to come up due to version restriction imposed in numaflow-core:
```SDK version 0.10.1 must be upgraded to at least 0.10.2, in order to work with the current numaflow version```

To bypass the blocker, we've switched to using the [redis sink implementation in Rust](https://github.com/numaproj/numaflow-rs/pull/157) with the release of v0.4.0 of the Rust SDK

### DIFF
--- a/test/e2e/testdata/simple-on-success.yaml
+++ b/test/e2e/testdata/simple-on-success.yaml
@@ -27,7 +27,6 @@ spec:
         onSuccess:
           udsink:
             container:
-              # A redis sink for e2e testing, see https://github.com/numaproj/numaflow-go/tree/main/examples/sinker/redis_sink
               image: quay.io/numaio/numaflow-rs/redis-sink:stable
               imagePullPolicy: IfNotPresent
               env:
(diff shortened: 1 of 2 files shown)


vigith (approved):
Adding logging fixed it? That sounds suspicious, can you run a couple of times to make sure it is not flaky?

vaibhavtiwari33:
Sry, that was just for figuring out what the issue was. The issue was that the redis sink implementation image that I had pushed to quay.io was private. 

Removed logging and will let the test run again.

PR #3075, opened 2025-11-14
doc: a bit more content to pipeline document

vigith:


KeranYang (approved):

docs/core-concepts/pipeline.md
```diff
@@ -14,31 +14,74 @@ The `Pipeline` is abstracted as a [Kubernetes Custom Resource](https://kubernete
...
   vertices:
-    - name: in
+    - name: input-source
       source:
-        generator:
-          rpu: 5
-          duration: 1s
-    - name: cat
+       ...
+    - name: vtx1
```
whynowy: Inconsistent name.

docs/core-concepts/pipeline.md
```diff
@@ -14,31 +14,74 @@ The `Pipeline` is abstracted as a [Kubernetes Custom Resource](https://kubernete
...
+      to: process
+    # ... more edges and conditions
+    - from: process
+      to: output-sink
 ```
 
+## Vertex Connections and Data Flow
+
+Vertices in a Pipeline are connected through `edges`, which define the data flow path and create 
+[Inter-Step Buffers](./inter-step-buffer.md) between processing stages. Each edge specifies a `from` vertex and a `to`
```
whynowy: This is not accurate. ISB is not created based on edges, but vertices.

vigith: How should i reword that such that it is easy for users to understand :)

whynowy: Something like `...flow path, with ISB providing the underneath support`?

PR #3078, opened 2025-11-19
fix: alignment and text wrapping in Errors tab

adarsh0728:
1. Adds whitespace between `bug` and `in` when sending partial responses message to the users
2. Aligned Table Columns (Fixed Ratios) [1:0.6:2.4:1] (Pod Name : Container : Message : Last Occurred)
3. Fixed Details Section Alignment [used 1:3:1 ratio] (Label : Content : Empty Space)
4. Enabled Text Wrapping (Updated the row height to be flexible (min-height: 7rem, height: auto), allowing it to expand based on content length)


adarsh0728:
> Can we fix the details page to align with the headings?

yes will do in this PR

adarsh0728:
Wrapping larger  Messages

Screenshot 2025-12-03 at 11 34 29 AM


Alignment of Details boundaries

Screenshot 2025-12-03 at 10 19 47 AM

yhl25 (approved):

PR #3080, opened 2025-11-20
feat: Numaflow MCP server

whynowy:

Screenshot 2025-11-19 at 6 14 03 PM


yhl25:
we will create a new pr to solve this.

PR #3085, opened 2025-11-25
doc: Update timeout doc for accumulator where idle detection config is required

### LINKED ISSUE #3002: docs: Improve Watermark and Backpressure behavior
Can I ask you to extend the docs for [Watermarks](https://numaflow.numaproj.io/core-concepts/watermarks) and [Back pressure](https://numaflow.numaproj.io/specifications/overview/#back-pressure), describing the behaviour for multiple sources, joins and keyed streams.

## Context 

idleness was not set for stream-sorter

### Martin

This also explains why stream-sorter container_memory_usage_bytes was seen growing and finally reaching OOMKilled/CrashLoopBackoff.
We similarly also have pipelines with dual sources and reducers where reduce_active_windows keep growing resulting in OOMKilled.

So if we have pipelines without idle detection, i.e. not accepting data loss where streams should wait for each other, how do we avoid OOM in downstream vertices? Is there a concept of back pressure where one idling (late) source causes back pressure on other sources?

### Vigith 

problem is, we cannot stop reading and create back-pressure because the pipeline is optimistic that the new data the arrives might move the watermark forward and help close the windows.. if we create back-pressure and stop reading, the pipeline will stall

### Martin

Hi vigith 🙂

I'm not sure I agree (from a conceptual perspective). If two sources join into an accumulator (or reducer) and source1 becomes stops receiving data, it should be possible to identify buffer pressure in the reducer and apply backpressure onto stream2 (only), causing the "the entire flow will stop (or slow down)", to use the wording from Back Pressure . When source1 at some point starts streaming again (or idle-detection kicks in and progresses stream2 watermark), the reducer would receive those events and progressed watermark, process its buffer and can start lifting the backpressure on stream1.

We have a documentation/understanding gap when it comes to the behaviour of Watermarks and Back pressure in pipelines with Multiple sources and Joins*.
When reading both the watermark and backpressure documentation, you get a feeling that it was authored with a single-source/no-join pipeline in mind. It is not documented whether each source "leg" of the pipeline maintains its own watermark and the join vertex combines those. Or whether there is just one overall pipeline watermark and the multiple source are updating that global state.

*) Normally multiple sources and joins are equivalent, but there are corner cases with single source being split and re-joined.

### Vigith

i totally hear you, this is something we also have been thinking.. current implementation joins the previous vertices into a single stream.. this is because the stream is owned by the (n+1)th vertex.. the fix is to make each vertex own its own stream and (n+1)th vertex reads from the previous vertices streams.. it is doable but will require some decent amount of time to make this change..

vaibhavtiwari33:
### Related issues
Closes #3002

### DIFF
--- a/docs/user-guide/user-defined-functions/reduce/windowing/session.md
+++ b/docs/user-guide/user-defined-functions/reduce/windowing/session.md
@@ -33,10 +33,10 @@ The `timeout` is the duration of inactivity (no data flowing in for the particul
 considered to be closed.
 
 Note: The determination of whether a key is inactive, or has timed out, is based on the watermark progressing.
-In order to close the accumulator window, we compare this timeout against the watermark.
+In order to close the session window, we compare this timeout against the watermark.
 If the watermark progression for the vertex has stalled for some reason, eg: due to one of the sources idling in a multi-source setup,
 the timeout may not be triggered without configuring [idle watermark detection](https://numaflow.numaproj.io/core-concepts/watermarks/#idle-detection).
-Currently, in such cases, the accumulator window may not close as it continues to hold on to the state for the key while ingesting
+Currently, in such cases, the session window may not close as it continues to hold on to the state for the key while ingesting
 more data, hoping to progress watermark with the next datum. This might lead to OOM situations.
 
 ## Example


vigith (approved):

vigith:
```
[./docs/user-guide/use-cases/monitoring-and-observability.md]:
[403] https://blog.numaproj.io/numaflow-letting-golden-signals-work-for-you-1bce18e472da | Rejected status code (this depends on your "accept" configuration): Forbidden
```

make 403 accept?

docs/user-guide/user-defined-functions/reduce/windowing/session.md
```diff
@@ -32,6 +32,13 @@ and a unit suffix, such as "300ms", "1.5h" or "2h45m". Valid time units are "ns"
 The `timeout` is the duration of inactivity (no data flowing in for the particular key) after which the session is
 considered to be closed.
 
+Note: The determination of whether a key is inactive, or has timed out, is based on the watermark progressing.
+In order to close the accumulator window, we compare this timeout against the watermark.
+If the watermark progression for the vertex has stalled for some reason, eg: due to one of the sources idling in a multi-source setup,
+the timeout may not be triggered without configuring [idle watermark detection](https://numaflow.numaproj.io/core-concepts/watermarks/#idle-detection).
+Currently, in such cases, the accumulator window may not close as it continues to hold on to the state for the key while ingesting
+more data, hoping to progress watermark with the next datum. This might lead to OOM situations.
```
vigith:
```suggestion
In order to close the session window, we compare this timeout against the watermark.
If the watermark progression for the vertex has stalled for some reason, eg: due to one of the sources idling in a multi-source setup,
the timeout may not be triggered without configuring [idle watermark detection](https://numaflow.numaproj.io/core-concepts/watermarks/#idle-detection).
Currently, in such cases, the session window may not close as it continues to hold on to the state for the key while ingesting
more data, hoping to progress watermark with the next datum. This might lead to OOM situations.
```

Issue #3037, opened 2025-10-22
Enable Opentelemetry protocol for metrics


vigith:
### Discussed in https://github.com/numaproj/numaflow/discussions/3035

<div type='discussions-op-text'>

<sup>Originally posted by **cdangelo-kline** October 22, 2025</sup>
Dear all,
Currently, Numaflow supports integration with Prometheus using an Operator for scraping metrics.
To make the system more vendor-neutral, Numaflow could consider supporting the OpenTelemetry protocol. This would allow it to expose metrics via both Prometheus scraping and OTLP.
As a reference, this could be implemented similarly to [Argo Workflows' OTLP integration](https://argo-workflows.readthedocs.io/en/latest/metrics/#opentelemetry-protocol) and  [Argo Workflow Prometheus Scraping](https://argo-workflows.readthedocs.io/en/latest/metrics/#prometheus-scraping).
Does this approach make sense?

</div>

juniemariam:
Hi! I’m new here and would love to work on this issue if it’s available.

vigith:
I have assigned this to you

juniemariam:
Hi everyone,
I’ve been exploring how Numaflow exposes metrics through Prometheus and wanted to check if the following direction makes sense before starting a PR.

Summary of what I’m thinking:
1. Keep Prometheus scraping as-is (controller on :9090, daemons on their existing ports).
2. Add OTLP exporters (enabled when OTEL_EXPORTER_OTLP_ENDPOINT is set).
3. Start with the controller-manager, then extend to daemon servers, UI backend, and Rust serving components in phases.
4. Add an otlp: block in the controller/server ConfigMaps for endpoint, protocol (grpc/http), TLS, and temporality settings.
5. Update Helm/Kustomize overlays to propagate these settings as environment variables.
6. Keep Prometheus the default; OTLP is opt-in.
7. Document how to enable the dual-exporter setup (similar to Argo Workflows).

Questions:
Does this plan match the direction you were thinking? And are there any existing design or telemetry guidelines I should follow before starting?
Note: I plan to start with controller-only OTLP support first and expand later if that makes sense.

Thanks!

whynowy:
@juniemariam - thanks for looking into it.

Agreed that we should keep Prometheus scraping as-is.

I would suggest starting from the daemon servers for the exploration. Unfortunately, controller-runtime (which controller-manager depends on) does not support otel at this moment, we need to how that goes.

juniemariam:
[...]

PR #3086, opened 2025-11-25
feat: enable OTLP metrics exporter for daemon services

### LINKED ISSUE #3087: Enable OTLP metrics export for Daemon services
# Summary
This is a child issue under #3037.

Numaflow currently exposes metrics only through Prometheus scraping.  
To make the metrics pipeline more vendor-neutral, the daemon-server and MonoVertex daemon should support exporting metrics via the OpenTelemetry Protocol (OTLP), in addition to Prometheus.

# Use Cases
- Preserve existing Prometheus behavior without modification
- Enable OTLP export only when `OTEL_EXPORTER_OTLP_ENDPOINT` is configured

This issue is scoped only to:
- daemon-server
- mvtx-daemon-server

PR addressing this: #3086

---

juniemariam:
This PR adds optional support for exporting Numaflow metrics through the OpenTelemetry Protocol (OTLP), alongside the existing Prometheus scraping. The goal is to make the metrics pipeline more vendor-neutral, similar to Argo Workflows’ approach, where metrics can be scraped by Prometheus or pushed via OTLP. Prometheus behavior remains unchanged. OTLP export is enabled only when an OTEL endpoint environment variable is set.

### Related issues
Fixes #3087
Part of #3037
Discussed in #3035

### Testing
- Started an OpenTelemetry Collector locally (debug exporter).
- Ran both daemon-server and mvtx-daemon-server with:
```
export OTEL_EXPORTER_OTLP_ENDPOINT=127.0.0.1:4317
export OTEL_EXPORTER_OTLP_INSECURE=true
```
Verified:

- OTLP exporter initialized successfully.
image

image

- Metrics appeared in the OTEL collector output.
- Prometheus /metrics endpoint still returned all metrics correctly.
- Confirmed that when the OTEL env vars are not set, behavior falls back to Prometheus-only.

- No changes to existing Prometheus scraping.
- OTLP path is optional and does not impact users unless configured.
- This PR intentionally touches only the daemon and MonoVertex daemon (controller-manager is not updated due to controller-runtime limitations).
- Added only required OTEL dependencies.


vigith:
please run "make codegen".. also this does not fix the parent issue, can you open a sub issue under the parent issue?

pkg/shared/telemetry/otel.go
```diff
@@ -0,0 +1,101 @@
...
+	"go.uber.org/zap"
+
+	"github.com/numaproj/numaflow/pkg/shared/logging"
+)
+
+// InitOTLPExporter initializes OpenTelemetry Protocol (OTLP) exporter if
+// OTEL_EXPORTER_OTLP_ENDPOINT environment variable is set. This allows metrics
+// to be exported to an OTLP collector while keeping Prometheus scraping unchanged.
+func InitOTLPExporter(ctx context.Context, componentName, componentInstance string, gatherer prometheus.Gatherer) (func(context.Context) error, error) {
+	endpoint := os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT")
```
whynowy: Use `pkg/shared/util.LookupEnvStringOr`

pkg/shared/telemetry/otel.go
```diff
@@ -0,0 +1,101 @@
...
+	otlpExporter, err := otlpmetricgrpc.New(ctx, opts...)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create OTLP metric exporter: %w", err)
+	}
+
+	// Create resource with service information
+	res, err := resource.New(ctx,
+		resource.WithAttributes(
+			attribute.String("service.name", componentName),
+			attribute.String("service.namespace", "numaflow"),
```
whynowy: Use `pkg/apis.Project` to replace `"numaflow"`.

pkg/daemon/server/daemon_server.go
```diff
@@ -152,6 +154,16 @@ func (ds *daemonServer) Run(ctx context.Context) error {
 
 	go ds.exposeMetrics(ctx)
 
+	// Initialize OTLP exporter if OTEL_EXPORTER_OTLP_ENDPOINT is set
+	shutdown, err := telemetry.InitOTLPExporter(ctx, v1alpha1.ComponentDaemon, ds.pipeline.Name, prometheus.DefaultGatherer)
+	if err != nil {
+		log.Warnw("Failed to initialize OTLP exporter, continuing without OTLP", zap.Error(err))
+	} else {
+		defer func() {
+			_ = shutdown(context.Background())
```
vigith, line 163:
```suggestion
			_ = shutdown(ctx)
```

any reason why we can't do this?

juniemariam: I used `context.Background()` here because the server’s `ctx` is cancelled during shutdown which prevents the OTLP exporter from flushing its final metrics. 

This follows OTEL Go examples, which also use context.Background() for graceful shutdown https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric#section-readme

That said, I’m happy to adjust based on what we prefer in Numaflow
- should we wrap it in a timed context (eg ctx with a timeout) to avoid hanging on shutdown
- should we skip calling `Shutdown` entirely and just rely on process exit?

Please let me know if we need a change here.

PR #3098, opened 2025-12-10
chore: clean up managedFields from k8s objects from UI display

whynowy:

Screenshot 2025-12-09 at 10 38 16 PM


whynowy:
> how does it look like after the cleanup?

All the lines from 17 to 72 will go away.

Screenshot 2025-12-12 at 8 38 07 AM

PR #3105, opened 2025-12-12
feat: Spec changes for short-circuiting mvtx

### LINKED ISSUE #3100: Controller/CRD changes for supporting short-circut in MonoVertex

vaibhavtiwari33:
This PR aims to isolate CRD changes and corresponding glue code for reading them to config for the short-circuiting feature for mvtx.

Currently proposed rules for short-circuiting:

```yaml
spec:
   # conditional routing to edges.
  bypass:
    fallback:
      tags:
        operator: or
        values:
          - tag-a
    onSuccess:
      tags:
        operator: or
          values:
            - tag-a
    sink:
      tags:
        operator: or
        values:
          - tag-a
            ...
```

### Related issues
Fixes #3100

### Testing

### DIFF
--- a/pkg/apis/numaflow/v1alpha1/mono_vertex_types.go
+++ b/pkg/apis/numaflow/v1alpha1/mono_vertex_types.go
@@ -518,15 +518,6 @@ type MonoVertexSpec struct {
 	// If specified, the bypass will be triggered if the conditions are met at any of the components.
 	// The first level of the bypass spec specifies the destination to which the message will be forwarded to,
 	// and the next level specifies the conditions to trigger the said bypass.
-	// For example, if the bypass spec is:
-	// bypass:
-	//   fallback:
-	//     tags:
-	//       operator: or
-	//       values:
-	//       - "error"
-	// Then a bypass to fallback sink will be triggered if the message is encountered at any
-	// component (source, UDF) that has a tag "error".
 	// +optional
 	Bypass *MonoVertexBypassCondition `json:"bypass,omitempty" protobuf:"bytes,15,opt,name=bypass"`
 }
--- a/api/json-schema/schema.json
+++ b/api/json-schema/schema.json
@@ -21177,7 +21177,7 @@
         },
         "bypass": {
           "$ref": "#/definitions/io.numaproj.numaflow.v1alpha1.MonoVertexBypassCondition",
-          "description": "Bypass defines the bypass destination and conditions to trigger bypass for the mono vertex components. If specified, the bypass will be triggered if the conditions are met at any of the components. The first level of the bypass spec specifies the destination to which the message will be forwarded to, and the next level specifies the conditions to trigger the said bypass. For example, if the bypass spec is: bypass:\n  fallback:\n    tags:\n      operator: or\n      values:\n      - \"error\"\nThen a bypass to fallback sink will be triggered if the message is encountered at any component (source, UDF) that has a tag \"error\"."
+          "description": "Bypass defines the bypass destination and conditions to trigger bypass for the mono vertex components. If specified, the bypass will be triggered if the conditions are met at any of the components. The first level of the bypass spec specifies the destination to which the message will be forwarded to, and the next level specifies the conditions to trigger the said bypass."
         },
         "containerTemplate": {
           "$ref": "#/definitions/io.numaproj.numaflow.v1alpha1.ContainerTemplate",
--- a/api/openapi-spec/swagger.json
+++ b/api/openapi-spec/swagger.json
@@ -21172,7 +21172,7 @@
           "type": "boolean"
         },
         "bypass": {
-          "description": "Bypass defines the bypass destination and conditions to trigger bypass for the mono vertex components. If specified, the bypass will be triggered if the conditions are met at any of the components. The first level of the bypass spec specifies the destination to which the message will be forwarded to, and the next level specifies the conditions to trigger the said bypass. For example, if the bypass spec is: bypass:\n  fallback:\n    tags:\n      operator: or\n      values:\n      - \"error\"\nThen a bypass to fallback sink will be triggered if the message is encountered at any component (source, UDF) that has a tag \"error\".",
+          "description": "Bypass defines the bypass destination and conditions to trigger bypass for the mono vertex components. If specified, the bypass will be triggered if the conditions are met at any of the components. The first level of the bypass spec specifies the destination to which the message will be forwarded to, and the next level specifies the conditions to trigger the said bypass.",
           "$ref": "#/definitions/io.numaproj.numaflow.v1alpha1.MonoVertexBypassCondition"
         },
         "containerTemplate": {
--- a/docs/APIs.md
+++ b/docs/APIs.md
@@ -6309,10 +6309,7 @@ for the mono vertex components. If specified, the bypass will be
 triggered if the conditions are met at any of the components. The first
 level of the bypass spec specifies the destination to which the message
 will be forwarded to, and the next level specifies the conditions to
-trigger the said bypass. For example, if the bypass spec is: bypass:
-fallback: tags: operator: or values: - “error” Then a bypass to fallback
-sink will be triggered if the message is encountered at any component
-(source, UDF) that has a tag “error”.
+trigger the said bypass.
 </p>
 
 </td>
@@ -6957,10 +6954,7 @@ for the mono vertex components. If specified, the bypass will be
 triggered if the conditions are met at any of the components. The first
 level of the bypass spec specifies the destination to which the message
 will be forwarded to, and the next level specifies the conditions to
-trigger the said bypass. For example, if the bypass spec is: bypass:
-fallback: tags: operator: or values: - “error” Then a bypass to fallback
-sink will be triggered if the message is encountered at any component
-(source, UDF) that has a tag “error”.
+trigger the said bypass.
 </p>
 
 </td>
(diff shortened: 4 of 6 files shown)


pkg/apis/numaflow/v1alpha1/udf.go
```diff
@@ -29,6 +29,10 @@ type UDF struct {
 	Container *Container `json:"container" protobuf:"bytes,1,opt,name=container"`
 	// +optional
 	GroupBy *GroupBy `json:"groupBy" protobuf:"bytes,3,opt,name=groupBy"`
+	// +optional
+	// MonoVertexSinkConditions is used to define conditions based on which the data from
+	// the UDF is forwarded to the different sinks. Only used for mono vertex.
+	MonoVertexSinkConditions *MonoVertexSinkerCondition `json:"sinkConditionals" protobuf:"bytes,4,opt,name=sinkConditionals"`
```
whynowy: Does it make sense to add it directly in `udf` or `transformer` definition? Would it be better to add a dedicated field in mvtx spec to do so?

I understand both `udf` and `transformer` may return tags, can they use the same config? One extra question - since `udf` is supported in mvtx, shall we deprecate `transformer`, as least not encourage ppl to use it in mvtx? @vigith

vaibhavtiwari33: Something similar to edges in a pipeline? 
Since we'll only have one transformer and one udf in mvtx, it seemed more intuitive to add it as part of their spec to specify conditions for forwarding messages to sink.

whynowy: > Something similar to edges in a pipeline? Since we'll only have one transformer and one udf in mvtx, it seemed more intuitive to add it as part of their spec to specify conditions for forwarding messages to sink.

The `udf` and `transformer` are also used in pipeline, how to interpret that in that case?

vigith: +1 to @whynowy's question.

I thought you were planning to add something like

```yaml
    <type>: # can be transformer, udf, or even source
      to: 
        fallback:
          operator: not
          tags: 
           - error-tag
        onsuccess:
          operator: and
          tags:
           - success-tag
        sink:
          operator: or
          tags:
           - dynamo-db
```

vigith: I also agree that we could add a new field.

vaibhavtiwari33: Sure, will add a separate field, so that we don't confuse users implementing udf/transformer in pipeline with a field that is not usable there.

rust/numaflow-core/src/config/monovertex.rs
```diff
@@ -60,11 +62,13 @@ impl Default for MonovertexConfig {
                 read_ahead: false,
                 source_type: SourceType::Generator(GeneratorConfig::default()),
             },
+            source_to_sink_condition: None,
```
whynowy: Logic of using these conditions not implemented?

pkg/apis/numaflow/v1alpha1/mono_vertex_forwarding.go
```diff
@@ -0,0 +1,33 @@
...
+limitations under the License.
+*/
+
+package v1alpha1
+
+type MonoVertexForwarding struct {
+	// +optional
+	SourceForwarding *MonoVertexSinkerCondition `json:"fromSource" protobuf:"bytes,1,opt,name=source"`
+	// +optional
+	MapForwarding *MonoVertexSinkerCondition `json:"fromMap" protobuf:"bytes,2,opt,name=map"`
```
whynowy: Use consistent names for json and protobuf unless there's particular reasons.

pkg/apis/numaflow/v1alpha1/mono_vertex_forwarding.go
```diff
@@ -0,0 +1,33 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+type MonoVertexForwarding struct {
+	// +optional
+	SourceForwarding *MonoVertexSinkerCondition `json:"fromSource" protobuf:"bytes,1,opt,name=source"`
```
whynowy: Why name them as `fromSource` and `fromMap`?

rust/numaflow-core/src/config/monovertex.rs
vaibhavtiwari33: I wanted to get feedback regarding the CRD changes first and finalize the schema before moving forward with dataplane implementation.

pkg/apis/numaflow/v1alpha1/mono_vertex_forwarding.go
vaibhavtiwari33: This should be `source` and `map`, let me change them.

pkg/apis/numaflow/v1alpha1/mono_vertex_forwarding.go
vaibhavtiwari33: I think this got missed, it should be `source` and `map` respectively

vaibhavtiwari33: updated

pkg/apis/numaflow/v1alpha1/mono_vertex_forwarding.go
vaibhavtiwari33: updated

pkg/apis/numaflow/v1alpha1/mono_vertex_forwarding.go
```diff
@@ -0,0 +1,33 @@
...
+*/
+
+package v1alpha1
+
+type MonoVertexForwarding struct {
+	// +optional
+	SourceForwarding *MonoVertexSinkerCondition `json:"source" protobuf:"bytes,1,opt,name=source"`
+	// +optional
+	UdfForwarding *MonoVertexSinkerCondition `json:"udf" protobuf:"bytes,2,opt,name=udf"`
+}
```
vigith: Can you please document all the exported fields? It will help in openapi spec  and all other place whether there is code generation.

vaibhavtiwari33: Sure, yes, my bad, basic mistake, let me add documentation for this.

vaibhavtiwari33: done

pkg/apis/numaflow/v1alpha1/mono_vertex_types.go
```diff
@@ -514,6 +514,21 @@ type MonoVertexSpec struct {
 	// +kubebuilder:default={"desiredPhase": Running}
 	// +optional
 	Lifecycle MonoVertexLifecycle `json:"lifecycle,omitempty" protobuf:"bytes,14,opt,name=lifecycle"`
+	// Bypass defines the bypass destination and conditions to trigger bypass for the mono vertex components.
+	// If specified, the bypass will be triggered if the conditions are met at any of the components.
+	// The first level of the bypass spec specifies the destination to which the message will be forwarded to,
```
whynowy, line 519: Do not need to give concrete example here, it won't be readable in generated docs.

rust/numaflow-core/src/monovertex/stream_splitter.rs
vigith, line 1: call it splitter. You can change it when you implement.

pkg/apis/numaflow/v1alpha1/mono_vertex_types.go
```diff
@@ -514,6 +514,21 @@ type MonoVertexSpec struct {
...
+	// bypass:
+	//   fallback:
+	//     tags:
+	//       operator: or
+	//       values:
+	//       - "error"
+	// Then a bypass to fallback sink will be triggered if the message is encountered at any
+	// component (source, UDF) that has a tag "error".
+	// +optional
+	Bypass *MonoVertexBypassCondition `json:"bypass,omitempty" protobuf:"bytes,15,opt,name=bypass"`
```
vigith, line 522: I feel `bypasses` is a better terminology since this is not about "a bypass" rule rather "a set of bypass" rules. 

@yhl25 @whynowy @vaibhavtiwari33 WDYT?

vaibhavtiwari33: Without context it can also be inferred as a verb, as in, `this bypasses that`.
So `bypasses` kind of also evokes the notion that these rules are set to bypass components listed in the bypass spec.

pkg/apis/numaflow/v1alpha1/mono_vertex_types.go
vaibhavtiwari33: Got it.

pkg/apis/numaflow/v1alpha1/mono_vertex_types.go
whynowy: `bypass` should be okay, it's not a `[]`, and the word would be easier to use than `bypasses`.

pkg/apis/numaflow/v1alpha1/mono_vertex_types.go
vaibhavtiwari33: done

whynowy (approved):

PR #3110, opened 2025-12-15
feat (UI): support `onSuccess` sink in MonoVertex graph

adarsh0728:
Supports onSuccess Sink in MonoVertex Graph

- The design/style remains same in cases where only one of the two sinks(fallback and onSuccess) is present or none is present
- The design/stlye changes when both fallback and onSuccess sinks are present
   - From primary sink - we have two directions upArrow --> onSuccess Sink and downArrow --> fallback Sink
- Wrapper classes are created so that above design could be accommodated cleanly
- Currently the image used for onSuccess sink is same as primary Sink, this can be changed in a follow up PR if required

### All w/o Fallback

image

### All w/o onSuccess

image

### All with onSuccess and Fallback

image

### Simple Pipeline (no change)
image

### DIFF
--- a/ui/src/components/pages/Pipeline/partials/Graph/partials/CustomNode/index.tsx
+++ b/ui/src/components/pages/Pipeline/partials/Graph/partials/CustomNode/index.tsx
@@ -133,16 +133,7 @@ const CustomNode: FC<NodeProps> = ({
     return style;
   };
 
-  const nodeStyle = useMemo(() => {
-    return {
-      border: `${isSelected(highlightValues[data?.name])} ${getBorderColor(
-        data?.type
-      )}`,
-      display: "flex",
-      flexDirection: "column",
-      ...commonStyle,
-    };
-  }, [highlightValues, data]);
+
 
   const genStyle = (text: string) => {
     return {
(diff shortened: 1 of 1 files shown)


vigith:
It looks awesome. Can you open an issue to fix the pipeline?

ui/src/components/pages/Pipeline/partials/Graph/partials/CustomNode/index.tsx
```diff
@@ -449,9 +585,12 @@ const CustomNode: FC<NodeProps> = ({
           arrow
           placement={"bottom-end"}
         >
-          <Box className={"node-rate"}>
+          <Box className={nodeRateWrapperClass} style={nodeRateStyle}>
             {formatRate(data?.vertexMetrics?.ratePerMin)}
           </Box>
+          {/* <Box className={"node-rate"} style={nodeRateStyle}>
```
BulkBeing: remove ?

adarsh0728:
> Awesome! Can you make sure the length of the arrow between the transformer and sink is same as between the source and the transformer? Also, please give some more space between the onsuccess and the primary sink.

i think you meant, length of arrow from primary sink to onsuccess(or fallback) sink should be same as from source to sink?

vigith:
Why is OnSuccess sink same color as Primary (unlike Fallback)? Should we change?

adarsh0728:
> Why is OnSuccess sink same color as Primary (unlike Fallback)? Should we change?

Yes, it will be done in follow up (mentioned in description). Let me create issue as well.

PR #3128, opened 2026-01-08
fix: avoid panic in sink code

yhl25:
* handle actor receiver failures to avoid messages being acked instead of getting nacked.

### DIFF
--- a/rust/extns/numaflow-kafka/src/lib.rs
+++ b/rust/extns/numaflow-kafka/src/lib.rs
@@ -1,5 +1,5 @@
 use rdkafka::config::ClientConfig;
-use tracing::{warn};
+use tracing::warn;
 
 pub mod sink;
 pub mod source;
--- a/rust/numaflow-core/src/mapper/map/user_defined.rs
+++ b/rust/numaflow-core/src/mapper/map/user_defined.rs
@@ -520,9 +520,11 @@ impl UserDefinedStreamMap {
         // only insert if we are able to send the message to the server
         if let Err(e) = self.read_tx.send(message.into()).await {
             error!(?e, "Failed to send message to server");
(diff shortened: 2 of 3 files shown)


vaibhavtiwari33 (approved):

rust/numaflow-core/src/pipeline/isb/jetstream/js_reader.rs
vigith, line 1: was the problem that we were ignoring errors by `let _ = ...` syntax?

vaibhavtiwari33: Other changes seem to be for proper error handling, but during the morning discussion it was identified that the sink writer had panicked instead of handling and propagating the error:

```rust
-        rx.await.unwrap()
----------------------------
+        rx.await
+           .map_err(|e| Error::ActorPatternRecv(e.to_string()))?
```

vigith: why will `panic` cause data loss?

vigith: our code is riddled with `expect` which is closer to `panic`

vaibhavtiwari33: With [this change](https://github.com/numaproj/numaflow/pull/3011), the messages are acked when dropped, unless [explicitly marked for failure](https://github.com/numaproj/numaflow/blob/29037686d4278215ec10d5676499872207687c69/rust/numaflow-core/src/message.rs#L79-L89). Hence, panicking didn't allow sink writer to nack the concerned message and it was effectively acked when dropped after panic

PR #3132, opened 2026-01-10
chore: refactor map component

yhl25:
* Remove actor pattern overhead in mapper.
* Separate package for map operations - unary, stream and batch.

Performance testing:

7% increase in the overall throughput.
Screenshot 2026-01-12 at 9 01 26 AM

15% decrease in the avg processing time.
Screenshot 2026-01-12 at 9 01 36 AM

### DIFF
--- a/rust/numaflow-core/src/mapper/map/batch.rs
+++ b/rust/numaflow-core/src/mapper/map/batch.rs
@@ -57,33 +57,38 @@ impl UserDefinedBatchMap {
         Ok(mapper)
     }
 
+    /// Broadcasts a batch map gRPC error to all pending senders and records error metrics.
+    fn broadcast_error(sender_map: &ResponseSenderMap, error: tonic::Status) {
+        let senders =
+            std::mem::take(&mut *sender_map.lock().expect("failed to acquire poisoned lock"));
+
+        for (_, (_, sender)) in senders {
+            let _ = sender.send(Err(Error::Grpc(Box::new(error.clone()))));
+            pipeline_metrics()
+                .forwarder
+                .udf_error_total
+                .get_or_create(pipeline_metric_labels(VERTEX_TYPE_MAP_UDF))
+                .inc();
+        }
+    }
+
     /// receive responses from the server and gets the corresponding oneshot response sender from the map
     /// and sends the response.
     async fn receive_batch_responses(
         sender_map: ResponseSenderMap,
         mut resp_stream: Streaming<MapResponse>,
     ) {
-        while let Some(resp) = match resp_stream.message().await {
-            Ok(message) => message,
-            Err(e) => {
-                let senders = {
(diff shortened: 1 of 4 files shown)


rust/numaflow-core/src/mapper/map/batch.rs
```diff
@@ -0,0 +1,310 @@
...
+        sender_map: ResponseSenderMap,
+        mut resp_stream: Streaming<MapResponse>,
+    ) {
+        while let Some(resp) = match resp_stream.message().await {
+            Ok(message) => message,
+            Err(e) => {
+                let senders = {
+                    let mut senders = sender_map.lock().expect("failed to acquire poisoned lock");
+                    senders.drain().collect::<Vec<_>>()
+                };
```
BulkBeing: You can replace this part with:
```rust
 let senders = std::mem::take(&mut *sender_map.lock().expect("failed to acquire poisoned lock"));
```
`senders` will be a hashmap instead of `Vec`, but the rest of the code won't change. Also avoids iterating over and creating a new `Vec`

rust/numaflow-core/src/mapper/map/batch.rs
```diff
@@ -0,0 +1,310 @@
...
+                        .expect("failed to send error response");
+                    pipeline_metrics()
+                        .forwarder
+                        .udf_error_total
+                        .get_or_create(pipeline_metric_labels(VERTEX_TYPE_MAP_UDF))
+                        .inc();
+                }
+                None
+            }
+        } {
```
BulkBeing: There's an extra scope created here (with `{`). Is this needed ? Makes it difficult to read as it feels like part of above `while let` loop.

rust/numaflow-core/src/mapper/map/batch.rs
```diff
@@ -0,0 +1,310 @@
...
+        let sender_entry = sender_map
+            .lock()
+            .expect("failed to acquire poisoned lock")
+            .remove(&msg_id);
+
+        if let Some((msg_info, sender)) = sender_entry {
+            let mut response_messages = vec![];
+            for (i, result) in resp.results.into_iter().enumerate() {
+                response_messages.push(UserDefinedMessage(result, &msg_info, i as i32).into());
+            }
```
BulkBeing:
```suggestion
            let response_messages: Vec<_> = resp
                .results
                .into_iter()
                .enumerate()
                .map(|(i, result)| UserDefinedMessage(result, &msg_info, i as i32).into())
                .collect();
```

BulkBeing: or `Vec::with_capacity`

BulkBeing (approved):

rust/numaflow-core/src/mapper/map/batch.rs
BulkBeing: Same changes can be done for other map implementations as well

PR #3138, opened 2026-01-19
chore: add dummy rust daemon server 

KeranYang:

Beginning of rust re-write of daemon server.

Next step: proto autogen for daemon server side APIs.

### Related issues

N/A.

### Testing

Started a MonoVertex and see the log on the daemon server.

### DIFF
--- a/docs/operations/metrics/metrics.md
+++ b/docs/operations/metrics/metrics.md
@@ -241,3 +241,53 @@ spec:
       - key: numaflow.numaproj.io/isbsvc-name
         operator: Exists
 ```
+
+## Metrics Exposition Format
+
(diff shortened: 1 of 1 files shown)


rust/Cargo.toml
```diff
@@ -7,6 +7,7 @@ members = [
     "numaflow",
     "numaflow-monitor",
     "numaflow-sideinput",
+    "numaflow-mvtx-daemon",
```
vigith:
```suggestion
    "numaflow-daemon",
```

let's not create a daemon crate per resource type, rather just have one crate that contains both daemons.

KeranYang: Done. Thanks!

Issue #3136, opened 2026-01-16
Metric name change after upgrading from v1.5 -> v1.7


MarkusOvervad:
Hi Numaflow.

We recently upgraded from v1.5 -> v1.7 and when that happened one of our key metrics, that we use to observe our pipelines changed its name. 

The issue was discovered because we scrape metrics from Numaflow via Prometheus and Metricbeat at the same time. 
Looking at the metrics via curl. We see the following output:

```text
# HELP forwarder_read_bytes Total number of bytes read.
# TYPE forwarder_read_bytes counter
forwarder_read_bytes_total{vertex="vertex_name",pipeline="pipeline_name",vertex_type="ReduceUDF",replica="0",partition_name="partion_name"} 24329610471
```

The changed that happened was forwarder_read_bytes_total became forwarder_read_bytes instead. The same goes for forwarder_ack_error_total -> forwarder_ack_error.

Looking at the Prometheus metric documentation: [Prometheus Docs](https://prometheus.io/docs/instrumenting/exposition_formats/#comments-help-text-and-type-information) 
We would expect the metric_name coming after both # HELP and # TYPE to have the same metric_name as in the sample. Unless the # TYPE would be e.g. histogram.

When viewing the metrics in Grafana the name of the metric is still forwarder_read_bytes_total, which suggests Prometheus uses the sample name. When viewing the metrics in Elasticsearch the metric name changed, which suggests Metricbeat uses the metric_name from # TYPE and not from the samples.

This should not be an issue according to the documentation from Prometheus. We expect this to be corrected, so the metric naming is consistent.

We believe the changes to effect in v.1.6, since a lot of changes were made to the codebase in regard to generating the metrics.

Not sure of how many metrics from Numaflow this applies to.

- Numaflow: [ v1.7.0 ]
- Metricbeat: [v8.17.3 ]

Best Regards,
Markus

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

>For quick help and support, join our [slack](https://numaproj.slack.com/join/shared_invite/zt-19svuv47m-YKHhsQ~~KK9mBv1E7pNzfg#/shared-invite/email) channel.

vaibhavtiwari33:
The difference unfortunately comes down to using golang vs rust client for prometheus. 

### Numaflow v1.7

[...]

PR #3139, opened 2026-01-19
chore: Metrics changes for Mvtx UDF, on-success sink and drop metrics

vaibhavtiwari33:

The PR aims to fix the following issues:
* Add metrics for whenever upstream components mark messages to be dropped in sink.
* Add metrics for monovertex udf/mapper
* Add metrics for on-success sink.

### Testing

### DIFF
--- a/rust/numaflow-core/src/sinker/sink.rs
+++ b/rust/numaflow-core/src/sinker/sink.rs
@@ -653,28 +653,25 @@ impl SinkWriter {
 
 /// Sends count of messages marked for explicit drop by the user
 /// Currently pub(crate) to allow usage by the bypass_router.
-/// TODO: Should this be a separate metric? Currently reason is different for pipeline
 pub(crate) fn send_drop_metrics(is_mono_vertex: bool, dropped_messages_count: usize) {
-    if dropped_messages_count > 0 {
-        if is_mono_vertex {
-            monovertex_metrics()
-                .sink
-                .dropped_total
-                .get_or_create(mvtx_forward_metric_labels())
-                .inc_by(dropped_messages_count as u64);
-        } else {
-            // The reason here is different from the one used when
-            // messages are dropped after 3 failed retries
-            pipeline_metrics()
-                .forwarder
-                .drop_total
-                .get_or_create(&pipeline_drop_metric_labels(
-                    VERTEX_TYPE_SINK,
-                    get_vertex_name(),
-                    "Dropped Upstream",
-                ))
-                .inc_by(dropped_messages_count as u64);
-        }
+    if is_mono_vertex {
+        monovertex_metrics()
+            .sink
+            .dropped_total
+            .get_or_create(mvtx_forward_metric_labels())
+            .inc_by(dropped_messages_count as u64);
+    } else {
+        // The reason here is different from the one used when
+        // messages are dropped after 3 failed retries
+        pipeline_metrics()
+            .forwarder
+            .drop_total
+            .get_or_create(&pipeline_drop_metric_labels(
+                VERTEX_TYPE_SINK,
+                get_vertex_name(),
+                "Dropped Upstream",
+            ))
+            .inc_by(dropped_messages_count as u64);
     }
 }
 


vigith:
> Patch coverage is 67.37288% with 77 lines in your changes missing coverage. Please review.

nothing we can do?

vaibhavtiwari33:
> > Patch coverage is 67.37288% with 77 lines in your changes missing coverage. Please review.
> 
> nothing we can do?

My bad, looks like the condition for generating mvtx specific metrics might be causing this. Let me add tests for the same.

docs/operations/metrics/metrics.md
```diff
@@ -22,6 +22,8 @@ These metrics can be used to determine throughput of your pipeline.
...
 | `forwarder_fbsink_write_total`             | Counter     | `pipeline=<pipeline-name>` <br> `vertex=<vertex-name>` <br> `vertex_type=<vertex-type>` <br> `replica=<replica-index>` <br> `partition_name=<partition-name>` | Provides the total number of messages written to a fallback sink                                                |
 | `forwarder_fbsink_write_bytes_total`       | Counter     | `pipeline=<pipeline-name>` <br> `vertex=<vertex-name>` <br> `vertex_type=<vertex-type>` <br> `replica=<replica-index>` <br> `partition_name=<partition-name>` | Provides the total number of bytes written to a fallback sink                                                   |
+| `forwarder_onssink_write_total`            | Counter     | `pipeline=<pipeline-name>` <br> `vertex=<vertex-name>` <br> `vertex_type=<vertex-type>` <br> `replica=<replica-index>` <br> `partition_name=<partition-name>` | Provides the total number of messages written to a on-success sink                                              |
```
adarsh0728: should we change the name? `forwarder_onsuccess_sink_write_total` or something informative so that it is easier for users to understand.

docs/operations/metrics/metrics.md
```diff
@@ -11,7 +11,7 @@ These metrics in combination can be used to determine the overall health of your
 These metrics can be used to determine throughput of your pipeline.
```
adarsh0728, line 11: one general suggestion - can we have following headings instead of Traffic, Latency, Errors etc

- Pipeline Metrics
- MonoVertex Metrics
- ISB Metrics
- Controller Metrics
- SDK Info Metrics

WDYT?  Can be taken up in follow up PR as well.

vaibhavtiwari33: That would be easier to navigate, we can have sub-sections within each as required.

docs/operations/metrics/metrics.md
vaibhavtiwari33: Sure, yes. I went with this format since the on success sink is named `ons-sink`
For metrics purposes, I think it should be fine to have long names.

vaibhavtiwari33:
> no need to test the metrics per se, but we should add negative tests that invoke drop behavior?

Added a test that verifies drop behavior in [this PR](https://github.com/numaproj/numaflow/pull/3134)

vaibhavtiwari33:
> > Patch coverage is 67.37288% with 77 lines in your changes missing coverage. Please review.
> 
> nothing we can do?

Tried adding test for this, but seems like setting env variable would require unsafe block

vigith:
let's also address the comments in https://github.com/numaproj/numaflow/issues/3136#issuecomment-3776627056 and https://github.com/numaproj/numaflow/issues/3136#issuecomment-3776659273 in this PR

vigith:
> > > Patch coverage is 67.37288% with 77 lines in your changes missing coverage. Please review.
> > 
> > 
> > nothing we can do?
> 
> Tried adding test for this, but seems like setting env variable would require unsafe block

the trick is to move `env` one level up and make the method accept env as a `HashMap`. This is what we do in config, etc.

docs/operations/metrics/metrics.md
vaibhavtiwari33: done

rust/numaflow-core/src/sinker/sink.rs
```diff
@@ -603,6 +651,33 @@ impl SinkWriter {
     }
 }
 
+/// Sends count of messages marked for explicit drop by the user
+/// Currently pub(crate) to allow usage by the bypass_router.
+/// TODO: Should this be a separate metric? Currently reason is different for pipeline
+pub(crate) fn send_drop_metrics(is_mono_vertex: bool, dropped_messages_count: usize) {
+    if dropped_messages_count > 0 {
```
vigith: NIT: I don't think you need this check.. `incr_by(0)` is valid and code will look cleaner.

rust/numaflow-core/Cargo.toml
```diff
@@ -74,5 +74,6 @@ aws-smithy-mocks = "0.1.1"
 k8s-openapi = { version = "0.24.0", features = ["v1_32"] }
 numaflow-kafka = { workspace = true, features = ["kafka-tests-utils"] }
 numaflow-pulsar = { workspace = true, features = ["pulsar-tests-utils"] }
+serial_test = "3.3.1"
```
vigith, line 77: the reason you need this is to make sure tests which require metrics update should run serially since metrics is a global state, correct?

vigith:
I did not review `metrics.md` :)

rust/numaflow-core/Cargo.toml
vaibhavtiwari33: Yes, that's correct

rust/numaflow-core/src/sinker/sink.rs
vaibhavtiwari33: That's a good point actually, let me address this.

vaibhavtiwari33:
Following is the updated metrics.md that relate to updating numaflow version in the exposition format section:


Screenshot 2026-01-21 at 7 09 28 PM

docs/operations/metrics/metrics.md
vaibhavtiwari33: Created a separate issue for this

rust/numaflow-core/src/sinker/sink.rs
vaibhavtiwari33: Updated

Issue #3140, opened 2026-01-20
Daemon server re-write in rust


KeranYang:
- [ ] gRPC+HTTP connections.
- [ ] MonoVertex API implementations.
- [ ] Pipeline API implementations.

# Summary

What change needs making?

# Use Cases

When would you use this?

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #3143, opened 2026-01-21
chore: support gRPC for rust mvtx daemon server

KeranYang:

Add a Rust gRPC server for the MonoVertex daemon, with mock data.

Next step is to support HTTP.

### Related issues

Relates to #3140 

### Testing

Tested that both numa controller and numa server can receive the mock data sent from the MonoVertex daemon server.

### DIFF
--- a/rust/numaflow-core/src/reduce/reducer/aligned/reducer.rs
+++ b/rust/numaflow-core/src/reduce/reducer/aligned/reducer.rs
@@ -1469,7 +1469,13 @@ mod tests {
             let proto_message = numaflow_pb::objects::isb::Message::decode(&data[..]).unwrap();
 
             // Extract and store the key
-            let key = proto_message.header.unwrap().keys[0].clone();
+            let key = proto_message
+                .header
+                .unwrap()
+                .keys
+                .first()
+                .expect("Expected at least one key")
+                .clone();
             received_keys.insert(key);
 
             result_count += 1;
--- a/rust/numaflow-core/src/reduce/reducer/aligned/user_defined.rs
+++ b/rust/numaflow-core/src/reduce/reducer/aligned/user_defined.rs
@@ -353,7 +353,7 @@ mod tests {
         let window = Window::new(window_start, window_end);
 
         // Create messages
-        let messages = vec![
+        let messages = [
             Message {
                 typ: Default::default(),
                 keys: Arc::from(vec!["key1".into()]),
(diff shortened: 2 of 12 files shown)


whynowy:
Is there a way to start both gRPC and HTTP together?

KeranYang:
> Is there a way to start both gRPC and HTTP together?

Yeah, I need to figure it out since our controller uses gRPC and UI can optionally use HTTP. Will investigate options and let you know. @whynowy

hack/generate-proto.sh
```diff
@@ -41,11 +41,6 @@ go-to-protobuf \
         --output-dir="${GOPATH}/src/" \
         --proto-import ./vendor
 
-# Following 2 proto files are needed
-mkdir -p ${GOPATH}/src/google/api
-curl -Ls https://raw.githubusercontent.com/googleapis/googleapis/master/google/api/annotations.proto -o ${GOPATH}/src/google/api/annotations.proto
-curl -Ls https://raw.githubusercontent.com/googleapis/googleapis/master/google/api/http.proto -o ${GOPATH}/src/google/api/http.proto
-
```
KeranYang, line 48: pinned the google api protos so that they can be shared between golang and rust.

vigith: why not just `curl` it to a common location?

rust/numaflow-daemon/src/lib.rs
vigith, line 1: are these dead code?

hack/generate-proto.sh
KeranYang: I prefer `pin` it because `curl` generates extra network call and technically(unlikely though) could generate backward compatibility issue. To me .proto files shouldn't be something that's dynamically changing.

rust/numaflow-daemon/src/lib.rs
KeranYang: Yes, as long as [this line](https://github.com/numaproj/numaflow/blob/09c1db512fd878ab68a632522d9772349b1c0206/pkg/apis/numaflow/v1alpha1/mono_vertex_types.go#L234) is commented out, they are dead code.

hack/generate-proto.sh
KeranYang: Also tried `curling` to a common location. Don't particularly like it because I needed to make changes at several places(scripts and Makefile).

PR #3168, opened 2026-01-29
chore: write a macro for emitting critical errors 

vigith:
Make emitting critical errors easier by exposing a macro

### DIFF
--- a/rust/numaflow-core/src/transformer.rs
+++ b/rust/numaflow-core/src/transformer.rs
@@ -15,9 +15,8 @@ use crate::config::{get_vertex_name, is_mono_vertex};
 use crate::error::Error;
 use crate::message::Message;
 use crate::metrics::{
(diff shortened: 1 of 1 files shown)


rust/numaflow-core/src/lib.rs
```diff
@@ -11,6 +11,44 @@ use tokio_util::sync::CancellationToken;
...
+///
+/// # Example
+/// ```ignore
+/// use crate::critical_error;
+/// use crate::config::pipeline::VERTEX_TYPE_SINK;
+///
+/// critical_error!(VERTEX_TYPE_SINK, "eot_received_from_sink");
+/// ```
+#[macro_export]
+macro_rules! critical_error {
```
vigith, line 32: we need to refactor metrics mod a bit and make then move this macro to there

PR #3172, opened 2026-01-30
chore: support for keys in http source

yhl25:


vigith (commented):
Please add to https://numaflow.numaproj.io/user-guide/sources/http/

PR #3178, opened 2026-02-01
feat: ISB Writer Trait integration (propagated via NumaflowTypeConfig)

### LINKED ISSUE #3176: define ISB Writer Trait and add it as Associated Type in NumaflowTypeConfig
`isb/writer.rs` should be referencing the `ISBWriter` rather than `JetstreamWriter`.

vigith:
Closes #3176

### DIFF
--- a/rust/numaflow-core/src/pipeline/isb/jetstream/js_writer.rs
+++ b/rust/numaflow-core/src/pipeline/isb/jetstream/js_writer.rs
@@ -31,8 +31,8 @@ type MetricLabels = Arc<Vec<(String, String)>>;
 pub(crate) enum WriteError {
     /// Buffer is full, cannot write (retryable)
     BufferFull,
-    /// Publish operation failed (retryable)
-    PublishFailed(String),
+    /// Write operation failed (retryable)
+    WriteFailed(String),
 }
 
 /// Buffer information for a JetStream stream.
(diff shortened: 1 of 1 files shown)


rust/numaflow-core/src/pipeline/isb/jetstream/js_writer.rs
```diff
@@ -377,6 +377,72 @@ impl JetStreamWriter {
...
+
+    async fn async_write(
+        &self,
+        message: Message,
+    ) -> std::result::Result<Self::PendingWrite, crate::pipeline::isb::WriteError> {
+        // Delegate to the existing async_write method
+        self.async_write(message).await.map_err(|e| match e {
+            WriteError::BufferFull => crate::pipeline::isb::WriteError::BufferFull,
+            WriteError::PublishFailed(msg) => crate::pipeline::isb::WriteError::WriteFailed(msg),
+        })
```
yhl25, line 392: Do we need this additional enum matching?

rust/numaflow-core/src/pipeline/isb/writer.rs
```diff
@@ -400,25 +409,26 @@ impl ISBWriter {
...
                         ?e, stream = ?write_result.stream,
                         "Failed to resolve the future, trying blocking write",
                     );
                     Self::publish_paf_error_metrics(write_result.stream.name, &e);
+                    // Fallback to blocking write - wrap offset in ResolveResult
                     writer
                         .blocking_write(message.clone(), cln_token.clone())
                         .await
+                        .map(ResolveResult::new)
                 }
```
yhl25, line 426: Can we make both `blocking_write` and `resolve` return same return type? that way we don't have to leak the resolve result to the orchestrator?

vigith: this needs a rewrite of the orchestrator, from a trait point this is cleaner because there is separation of concern because of how the orchestrator is designed.

PR #3181, opened 2026-02-02
fix: Wait for response stream to close before finishing the receiver method for map

vaibhavtiwari33:

### Related issues
Fixes #

### Testing

### DIFF
--- a/rust/numaflow-core/src/mapper/map/unary.rs
+++ b/rust/numaflow-core/src/mapper/map/unary.rs
@@ -80,8 +80,16 @@ impl UserDefinedUnaryMap {
                 Ok(None) => break,
                 Err(e) => {
                     error!(?e, "Error reading message from unary map gRPC stream");
-                    Self::broadcast_error(&sender_map, e);
-                    continue;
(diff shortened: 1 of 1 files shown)


rust/numaflow-core/src/mapper/map/unary.rs
```diff
@@ -80,7 +80,15 @@ impl UserDefinedUnaryMap {
...
-                    Self::broadcast_error(&sender_map, e);
+                    Self::broadcast_error(&sender_map, e.clone());
+                    loop {
+                        match resp_stream.message().await {
+                            Ok(None) => break,
+                            Ok(Some(_)) | Err(_) => {
+                                Self::broadcast_error(&sender_map, e.clone());
+                            }
+                        }
+                    }
```
vigith, line 91: why not use `while let Some(..)` syntax? this doesn't make sense

vaibhavtiwari33: made that change in the original PR, was testing out with this one. Closing this PR for now

Issue #2645, opened 2025-05-17
Logs in OTEL format in "numa" container


vigith:
# Summary

Enable [OTEL Subscriber](https://docs.rs/tracing-opentelemetry/latest/tracing_opentelemetry/) if user want logs to be compatible with [OTEL](https://opentelemetry.io/ ) format. 

We can only support for "numa" (Rust) container because the logs created using the [tracing](https://github.com/tokio-rs/tracing/) module.

The numaflow controller (Golang) uses structure dlogging since it uses [zap](https://github.com/uber-go/zap)

## SDKs
Perhaps we should use structure logging or do not log except for "warn"/"error" to `stderr`. We should not be importing "otel" libraries into the SDK as it can create dependency issues.

# Use Cases

Users want to collect the logs using OTEL Collector.

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

th0ger:
How does this task relate to https://github.com/numaproj/numaflow/pull/3056 ?
Is the structures logging in https://github.com/numaproj/numaflow/pull/3056 OTEL compatible?

vigith:
"structured logging" in those issues just mean JSON format. We still need to figure out what exactly should be done to make it compatible with OTEL.

madsbv:
I am planning to try to enable `tracing_opentelemetry` in the dataplane code for internal use, and would like to work towards submitting a PR for this issue. I'm aware of #3121 and #3185 as well and would be happy to work towards a more comprehensive implementation. Are there any further discussions or decisions from the numaflow team I should be aware of?

My plan would be to start by adding `tracing_opentelemetry`  as a subscriber without changing any existing logging, then start working on tracing a la #3185.

vigith:
@madsbv i can assign https://github.com/numaproj/numaflow/issues/3121 to you if you are interested. If you are, please comment on that issue.

PR #3185, opened 2026-02-03
doc: Tracing design proposal

adarsh0728:
Proposal: OpenTelemetry Tracing Design for Numaflow


docs/specifications/tracing.md
```diff
@@ -0,0 +1,238 @@
...
+     - Covers overhead (ISB read/write, serialization).
+   - Prepares UDF Request. Starts Span `S3` (Parent: `S2`) -> **"UDF RPC Call"**.
+     - Covers gRPC roundtrip to UDF container.
+   - Injects `T1-S3` into UDF Request Metadata.
+
+4. **UDF (SDK)**:
+   - Receives Request.
+   - Extracts `T1-S3`.
+   - Starts Span `S4` (Parent: `S3`). runs user code... Ends `S4`.
+   - **Crucial**: SDK sends Span `S4` data to Collector.
```
whynowy: Expecting the sdk to handle data sending?

adarsh0728: Yes, thought of both core and sdk to send data
```
Trace X:
  Core Span: Vertex Processing
    SDK will start Span for: UDF Execution
      User starts Span (optional): some calculation
```
      
For sdk/user spans to work, user has to initiate tracing(we can provide a helper fn via sdk to set up global tracer for udf process with same OTEL endpoint env variable)

docs/specifications/tracing.md
```diff
@@ -0,0 +1,257 @@
...
+## 1. Objective
+The goal is to enable distributed tracing in Numaflow to track message propagation from Source to Sink. This will provide visibility into:
+- The path a message takes through the pipeline.
+- Latency introduced by each component (Source, ISB, UDFs, Sink).
+- Bottlenecks in the system.
+
+## 2. Key Architectural Decisions
+
+### 2.1 Format: W3C Trace Context
+We will use the **W3C Trace Context** standard (`traceparent`, `tracestate`) for context propagation.
```
vigith: hyperlink to the spec?

docs/specifications/tracing.md
```diff
@@ -0,0 +1,257 @@
...
+- **Rationale**: It is the industry standard, natively supported by OpenTelemetry, compact (1-2 headers), and vendor-neutral.
+
+### 2.2 Carrier: System Metadata (`sys_metadata`)
+Trace context will be stored in the `sys_metadata` field of the Numaflow `Message` protobuf.
+
+- **Location**: `message.metadata.sys_metadata`
+- **Key**: `"tracing"`
+- **Value**: A `KeyValueGroup` containing the W3C headers.
+
+**Message Structure:**
```
vigith: we do not need to add anything to our proto spec, right? since this is already in place

docs/specifications/tracing.md
```diff
@@ -0,0 +1,257 @@
...
+  2. **Links**: We add the contexts of all *other* messages as **Span Links**. In distributed tracing visualization tools, these appear as "Related Traces" or "Links", allowing users to jump from any input message trace to the aggregated reduce trace.
+- **Limit**: To prevent span bloat in massive windows, limit the number of links (e.g., max 50). If more, sample representative links (e.g., first 10, last 10).
+
+### 3.6 Handling FlatMap (Fan-Out)
+When a Map vertex produces multiple output messages from a single input (1 -> N):
+- **Forking**: The trace context "forks" automatically.
+- **Implementation**: The Core injects the **same** current span context (parent) into `sys_metadata` for **every** output message.
+- **Result**: Downstream vertices will create N separate spans, all pointing back to the same parent span from the Map vertex.
+
+## 4. Implementation Plan: Numaflow SDKs (e.g., Go/Rust/Python)
```
vigith: let's not worry about SDKs, we just need server spans. Adding dependencies on SDK might be an overkill.

docs/specifications/tracing.md
adarsh0728: yes

th0ger:
I applaude the objective of this proposal 👍🏼Great synergy with https://github.com/numaproj/numaflow/issues/2645.

vigith:
@adarsh0728 let's work on this PR itself since there are lot of eyes on this 😄

docs/specifications/tracing.md
```diff
@@ -0,0 +1,257 @@
...
+2. **Start Span**: Start a new span (`name="UserFunction"`, `kind=Server`).
+3. **Execute**: Pass the `context` (containing the span) to the user's function.
+   - This allows users to create their own child spans using standard OTel APIs: `otel.Tracer("my-udf").Start(ctx, "sub-task")`.
+4. **End Span**: Close the span when the function returns.
+
+## 5. End-to-End Workflow
+
+1. **Source**:
+   - Generates data.
+   - Starts Trace `T1`, Span `S1`.
```
BulkBeing: Do we create the trace id in SDK after receiving messages from UDF ? Only then we can propagate `traceparent` coming from external systems like Kafka right?

docs/specifications/tracing.md
```diff
@@ -0,0 +1,257 @@
...
+  2. **Extract**: Look for `sys_metadata["tracing"]`. Use `MetadataExtractor` to get the parent context.
+  3. **Start Span**: Create a new span (Span Kind to be decided) linked to the parent context.
+     * **Available Span Kinds**:
+       - `Server`: Covers the server-side handling of an RPC or other remote request.
+       - `Client`: Covers the client-side of an RPC or other remote request.
+       - `Producer`: Describes a child of an asynchronous request (e.g., sending to a queue).
+       - `Consumer`: Describes a child of an asynchronous request (e.g., receiving from a queue).
+       - `Internal`: Default. Represents an internal operation within an application.
+  4. **Process**:
+     - **Before UDF**: Create a child span (e.g., `kind=Client`) for the UDF call. Inject this context into a *copy* of the metadata passed to the UDF.
```
BulkBeing: Similarly, should we set `kind=Producer` and `kind=Consumer` when writing to ISB and reading from ISB ?

docs/specifications/tracing.md
```diff
@@ -0,0 +1,257 @@
...
+     - **After UDF**: Receive results.
+  5. **Write**:
```
BulkBeing: I think handling traces in Sink should be a separate section.
We could add more details for retries, dropped etc. Eg:
```rust
  // Record a Span Event with details
  span.add_event("message.dropped", vec![
      KeyValue::new("drop.reason", "fallback_strategy"),
      KeyValue::new("drop.after_retries", 5),
  ]);
```

Tracing backend will see something like:
```json
{
  "trace_id": "4bf92f...",
  "span_id": "aabb...",
  "name": "numaflow.my-pipeline.sink-vertex.process",
  "events": [
    {
      "name": "message.dropped",
      "timestamp": "2026-04-01T12:00:00Z",
      "attributes": {
        "drop.reason": "fallback_strategy",
        "drop.after_retries": 5
      }
    }
  ]
}
```

docs/specifications/tracing.md
```diff
@@ -0,0 +1,257 @@
...
+
+2. **ISB (JetStream)**:
+   - Persists `M1`.
+
+3. **Map Vertex (Core)**:
+   - Reads `M1`.
+   - Extracts context `T1-S1`.
+   - Starts Span `S2` (Parent: `S1`) -> **"Vertex Processing"**.
+     - Covers overhead (ISB read/write, serialization).
+   - Prepares UDF Request. Starts Span `S3` (Parent: `S2`) -> **"UDF RPC Call"**.
```
BulkBeing: Let's use predictable span names? eg
```
numaflow.{pipeline}.{vertex}.process
numaflow.{pipeline}.{vertex}.udf
numaflow.{pipeline}.{vertex}.sink.retry
```

docs/specifications/tracing.md
```diff
@@ -0,0 +1,257 @@
...
+#### Option: Trace Links
+This approach preserves the relationship between all input messages and the aggregated result without creating a messy parent-child graph.
+- **Ingest Phase**: As messages arrive in the window, we extract their trace contexts.
+- **Aggregate Phase**: When the window closes and we create the span for the Reduce UDF execution:
+  1. **Primary Parent**: We select **one** context (e.g., from the first message) to be the direct `Parent` of the new span. This maintains the main timeline.
+  2. **Links**: We add the contexts of all *other* messages as **Span Links**. In distributed tracing visualization tools, these appear as "Related Traces" or "Links", allowing users to jump from any input message trace to the aggregated reduce trace.
+- **Limit**: To prevent span bloat in massive windows, limit the number of links (e.g., max 50). If more, sample representative links (e.g., first 10, last 10).
+
+### 3.6 Handling FlatMap (Fan-Out)
+When a Map vertex produces multiple output messages from a single input (1 -> N):
```
BulkBeing: How does this work with batchmap (`N` inputs to `N` outputs in one grpc invocation) ?

yhl25:
I think we should User Metadata for Trace Propagation When UDF is Instrumented                                                                                                    
                                                                                                                                                                                              
 If a UDF is instrumented, it should inject its active span context into outgoing user metadata (e.g. `user_metadata["tracing"]`). Downstream UDFs should then prefer this over `sys_metadata["tracing"]`.                                                                             
                                                                                                                                                                                              
 Order would be:
  1. `user_metadata["tracing"]` — set by upstream UDF (preferred)                                                                                                                         
  2. `sys_metadata["tracing"]` — set by platform (fallback)
  3.  Fresh root context                                                                                                                                                                       
  
 `sys_metadata["tracing"]` carries the platform's span context, not the UDF's. Without this, downstream UDF spans                    become siblings of upstream UDF spans instead of children.

vigith:
`source.read` as PRODUCER but map/sink UDFs as CONSUMER is inconsistent. isn't this counter-intuit? (OTel messaging semantics typically use PRODUCER for the component sending to a queue and CONSUMER for the one receiving. source.read reads from an external system) what if we use  CONSUMER or CLIENT which is more accurate?

also, sink.write as CLIENT which is ok, but numaflow.platform.map is INTERNAL while its UDF child is CONSUMER breaks the parent-child kind pattern, won't it?

yhl25:
Implementation is done https://github.com/numaproj/numaflow/pull/3440.

PR #3190, opened 2026-02-04
refactor: `write` and `resolve` should return the same type and Orchestrator to orchestrate retries

### LINKED ISSUE #3182: isb writer orchestrator should deal with retries not trait methods
this needs a rewrite of the orchestrator, from a trait point this is cleaner because there is separation of concern because of how the orchestrator is designed.

_Originally posted by @vigith in https://github.com/numaproj/numaflow/pull/3178#discussion_r2756278838_

if retries are handled in orchestrator, then `write` and `resolve` after `async_write` can return the same type and orchestrator can deal with error handling

vigith:
Closes #3182

refactor orchestrator to remove retry complexity away from trait methods. earlier we had a method called `blocking_write` (renamed now to `write`) which retries for ever until `ctx` was cancelled, this means every trait implementer have to do this; this retry logic is now moved to Orchestrator.

### DIFF
--- a/rust/numaflow-core/src/pipeline/isb/writer.rs
+++ b/rust/numaflow-core/src/pipeline/isb/writer.rs
@@ -454,9 +454,7 @@ impl<C: NumaflowTypeConfig> ISBWriterOrchestrator<C> {
                 Ok(result) => return Ok(result),
                 Err(WriteError::BufferFull) => {
                     // Buffer is full, wait and retry
-                    debug!(
-                        "Buffer full during write retry, waiting before next attempt"
-                    );
+                    debug!("Buffer full during write retry, waiting before next attempt");
                 }
                 Err(WriteError::WriteFailed(ref e)) => {
                     error!(?e, "Write failed during retry, will retry");


rust/numaflow-core/src/pipeline/isb/writer.rs
```diff
@@ -443,18 +441,47 @@ impl<C: NumaflowTypeConfig> ISBWriterOrchestrator<C> {
...
+        message: Message,
+        cln_token: CancellationToken,
+    ) -> std::result::Result<WriteResult, WriteError> {
+        loop {
+            match writer.write(message.clone()).await {
+                Ok(result) => return Ok(result),
+                Err(WriteError::BufferFull) => {
+                    // Buffer is full, wait and retry
+                    debug!("Buffer full during write retry, waiting before next attempt");
+                }
```
yhl25, line 458: Use `retry` crate instead of writing custom retry logic?

rust/numaflow-core/src/pipeline/isb.rs
```diff
@@ -165,21 +168,19 @@ pub(crate) trait ISBWriter: Send + Sync + Clone {
...
-    /// Writes a message and blocks until confirmed, returning the offset.
+    /// Writes a message and waits for confirmation, returning the result.
+    ///
+    /// This is a single-attempt write that returns immediately after the write
+    /// completes or fails. The orchestrator is responsible for retry logic.
     ///
-    /// This is the simple blocking write method with infinite retries until success
-    /// or cancellation. Use this when you don't need the high-performance async pattern.
+    /// Returns `Err(WriteError::BufferFull)` if the buffer is full.
+    /// Returns `Err(WriteError::WriteFailed)` if the write operation fails.
```
yhl25, line 179: Should we introduce `Retryable` and `NonRetryable` error kinds?

yhl25 (approved):
Followup PR:
* Throttled logging for buffer full errors.
* Use `retry` crate for retry logic.
* Introduce `Retryable` and `NonRetryable` errors.

PR #3200, opened 2026-02-05
feat(ui): show onSuccess and fallback sink container icons in Pipeline graph

### LINKED ISSUE #3118: Show `onSuccess` sink container icon in Pipeline UI
Currently we do not show fallback sink and onSuccess container in Pipeline UI.

related: https://github.com/numaproj/numaflow/issues/2623

sapkota-aayush:
Fixes #3118

Sink vertices in the Pipeline graph now show separate container icons when sink.onSuccess or sink.fallback are set in the spec:
- Sink – main sink container
- OnSuccess – shown when sink.onSuccess is configured
- Fallback – shown when sink.fallback is configured

Behavior matches how MonoVertex nodes show multiple containers. Changes are limited to:
- CustomNode/index.tsx – render Sink / OnSuccess / Fallback icons for sink vertices
- CustomNode/style.css – styles for the new container icons

Screenshot (sink vertex with main sink, onSuccess, and Fallback Sink Container icons):
Numaflow pr image

Question for reviewers: For the new Sink / OnSuccess / Fallback container icons, should clicking each icon open different sidebar content (e.g. logs/config for that container), or should all continue to open the same vertex details? Current behavior is unchanged (same vertex details for all).

### DIFF
--- a/ui/src/utils/fetcherHooks/pipelineViewFetch.ts
+++ b/ui/src/utils/fetcherHooks/pipelineViewFetch.ts
@@ -113,7 +113,7 @@ export const usePipelineViewFetch = (
     };
 
     fetchPipeline();
-  }, [requestKey, addError]);
+  }, [requestKey, addError, namespaceId, pipelineId]);
 
   // Call to get buffers
   useEffect(() => {
--- a/ui/src/components/pages/Pipeline/partials/Graph/partials/CustomNode/index.tsx
+++ b/ui/src/components/pages/Pipeline/partials/Graph/partials/CustomNode/index.tsx
@@ -293,6 +293,12 @@ const CustomNode: FC<NodeProps> = ({
     ? "mono-vertex-img-wrapper-small"
     : "mono-vertex-img-wrapper";
   const imgClass = hasBothSinks ? "mono-vertex-img-small" : "mono-vertex-img";
+  const pipelineSinkWrapperClass = hasBothSinks
+    ? "pipeline-sink-container-wrapper-small"
+    : "pipeline-sink-container-wrapper";
+  const pipelineSinkImgClass = hasBothSinks
+    ? "pipeline-sink-container-img-small"
+    : "pipeline-sink-container-img";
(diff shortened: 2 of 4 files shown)


adarsh0728:
Can we add similarly within source - with source and source transformer as a horizontal graph

sapkota-aayush:
@adarsh0728 Thanks for the review! 
Yes, I can add a similar horizontal layout for source vertices showing the source container and source transformer container icons side-by-side, similar to what I did for sink/onSuccess/fallback.
Would you prefer I add this to this PR, or should I create a follow-up PR to keep this one focused on sinks?

vigith:
@sapkota-aayush you can do a follow up, please comment on https://github.com/numaproj/numaflow/issues/3206 and I will assign to you

ui/src/utils/fetcherHooks/pipelineViewFetch.ts
```diff
@@ -113,7 +113,7 @@ export const usePipelineViewFetch = (
     };
 
     fetchPipeline();
-  }, [requestKey, addError]);
+  }, [requestKey, addError, namespaceId, pipelineId]);
```
adarsh0728, line 116: reason for this change?

sapkota-aayush: > reason for this change?

I added `namespaceId` and `pipelineId` to the dependency array because `BASE_API` uses these values. This ensures the fetch runs again when switching pipelines or namespaces. This was likely required by React's dependency rules, though I may be mistaken.

sapkota-aayush:
> if possible, lets have uparrow and downarrow emanate from primary sink , currently there is a gap.
> 
> follow #3110
> 
> Image

Thanks for the feedback! I'll update the arrow positioning so the up/down arrows emanate directly from the primary sink icon (removing the gap) in a follow-up commit or in PR #3206. I'll reference #3110 for the expected behavior.

PR #3201, opened 2026-02-05
feat: use initContainerStatuses in health checks

### LINKED ISSUE #3160: InitContainerStatus not used for health checking
**Describe the bug**

Numaflow can either create SideCar `Containers` or `InitContainers`, depending on the version of Kubernetes plus the presence of the `NUMAFLOW_K8S_SERVER_VERSION` environment variable. 

However, much of the code was written assuming `Containers` and not `InitContainers`. 

For example, [this](https://github.com/numaproj/numaflow/blob/v1.7.1/pkg/reconciler/util.go#L55) and [this](https://github.com/numaproj/numaflow/blob/v1.7.1/pkg/reconciler/util.go#L93).

**To Reproduce**

I noticed this when I created a MonoVertex on a cluster with an invalid image path and it had a `PodsHealthy` condition. 

**Expected behavior**

`PodsHealthy` to equal false

**Screenshots**

- Kubernetes: v1.32.9-eks-3025e55
- Numaflow: 1.7.0 (when I deployed numaflow 1.5.2, it did not use InitContainers so I didn't observe the issue, although I think the issue would exist for K8S > 1.34)

**Additional context**

---

[monovertex-bad-image.yaml](https://github.com/user-attachments/files/24871472/monovertex-bad-image.yaml)
[pod-bad-image.yaml](https://github.com/user-attachments/files/24871471/pod-bad-image.yaml)

dpadhiar:

Health checks currently do not check both container and initContainer statuses when determining a pod's health. This can be an issue due to Numaflow creating sidecars in either. We remedy this by checking initContainerStatuses of the pod in all places that we are already checking the containerStatues.

### Related issues
Fixes #3160 

### Testing

Unit testing.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2727,6 +2727,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "numaflow-testing"
+version = "0.1.0"
+dependencies = [
+ "bytes",
+ "chrono",
+ "parking_lot",
+ "thiserror 2.0.17",
+ "tokio",
+]
+
+[[package]]
 name = "numaflow-throttling"
 version = "0.1.0"
 dependencies = [
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -10,6 +10,7 @@ members = [
     "numaflow-daemon",
     "numaflow-shared",
     "numaflow-throttling",
+    "numaflow-testing",
     # extensions
     "extns/numaflow-pulsar",
     "extns/numaflow-sqs",
--- a/rust/numaflow-core/src/transformer.rs
+++ b/rust/numaflow-core/src/transformer.rs
@@ -1,9 +1,9 @@
 use bytes::Bytes;
+use futures::stream::{self, StreamExt};
 use numaflow_monitor::runtime;
 use numaflow_pb::clients::sourcetransformer::source_transform_client::SourceTransformClient;
-use std::sync::Arc;
 use std::time::Duration;
-use tokio::sync::{Semaphore, mpsc, oneshot};
+use tokio::sync::{mpsc, oneshot};
 use tokio_util::sync::CancellationToken;
 use tonic::transport::Channel;
 use tonic::{Code, Status};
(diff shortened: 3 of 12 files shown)


server/apis/v1/health.go
```diff
@@ -207,6 +207,17 @@ func isVertexHealthy(h *handler, ns string, pipeline string, vertex *dfv1.Vertex
...
 			}
 		}
+		// Iterate over all the initContainers in the pod
+		for _, containerStatus := range pod.Status.InitContainerStatuses {
+			// check that the container terminated successfully
+			if containerStatus.State.Terminated != nil && containerStatus.State.Terminated.Reason == "Error" {
+				return false, &resourceHealthResponse{
+					Message: fmt.Sprintf("Init container %q in pod %q is not healthy", containerStatus.Name, pod.Name),
+					Code:    "V3",
+				}, nil
```
juliev0, line 217: Nice job finding a way to look for health in a way that doesn't require the container to be "Running" like it does for regular Containers.

juliev0:
@dpadhiar I'm slightly nervous if this was only tested with unit testing, though. It will be good to get more confidence through observing that the Pod Condition is set true when we expect it to be and false when we expect it not to be. Also, that the Numaflow Server is showing the right thing.

dpadhiar:
> @dpadhiar I'm slightly nervous if this was only tested with unit testing, though. It will be good to get more confidence through observing that the Pod Condition is set true when we expect it to be and false when we expect it not to be. Also, that the Numaflow Server is showing the right thing.

Got it, I will do some manual testing and we could wait to merge this in until Monday.

server/apis/v1/health.go
```diff
@@ -207,6 +207,17 @@ func isVertexHealthy(h *handler, ns string, pipeline string, vertex *dfv1.Vertex
 				}, nil
 			}
 		}
+		// Iterate over all the initContainers in the pod
+		for _, containerStatus := range pod.Status.InitContainerStatuses {
+			// check that the container terminated successfully
+			if containerStatus.State.Terminated != nil && containerStatus.State.Terminated.Reason == "Error" {
```
whynowy, line 213: Keep an eye on it to see if it covers everything.

PR #3203, opened 2026-02-06
chore(perf): Avoid message cloning in Sink

BulkBeing:
Avoids cloning the messages in Sink wherever possible.

### Testing
Tested manually by sending 100,000 messages and validating all messages on the Sink side.

### DIFF
--- a/rust/numaflow-core/src/sinker/actor.rs
+++ b/rust/numaflow-core/src/sinker/actor.rs
@@ -93,32 +93,30 @@ where
                 .map(|resp| (resp.id, resp.status))
                 .collect::<HashMap<_, _>>();
 
-            // Classify messages based on responses
-            let mut i = 0;
-            while let Some(msg) = messages_to_retry.get(i) {
+            // Classify messages based on responses, preserving original ordering.
+            // Take ownership of the current batch so we can iterate by move and push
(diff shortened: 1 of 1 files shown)


rust/numaflow-core/src/sinker/actor.rs
```diff
@@ -94,34 +94,35 @@ where
...
-                match result_map.remove(&msg.id.to_string()) {
+            let mut i = 0;
+            while let Some(msg) = messages_to_retry.get(i) {
+                let msg_id = msg.id.to_string();
+                match result_map.remove(&msg_id) {
                     Some(ResponseStatusFromSink::Success) => {
-                        false // remove from retry list
+                        // Remove from retry list, don't need the message anymore
+                        // Don't increment i here, since we swapped the element at i with the last element
+                        messages_to_retry.swap_remove(i);
```
vigith: won't we lose ordering because of replace with the last element during unhappy path?

BulkBeing: Yes.. Should we preserve the order ?

vigith: Yes, because during retries, we do not want to switch semantics to unordered.

Issue #3198, opened 2026-02-05
inmemory ISB for testing


vigith:
# Summary

We need to build an in-memory ISB for testing, currently our unit tests are not covering failure cases since it is difficult to mimic those in distributed standalone ISB.

# Use Cases

- Prove how efficient our traits are for ISB.
- This will **only be** used for unit testing

PR #3213, opened 2026-02-08
feat(test): a simple in-memory buffer for simulating ISB

vigith:
Closes https://github.com/numaproj/numaflow/issues/3198

### DIFF
--- a/rust/numaflow-testing/src/simplebuffer/error_injector.rs
+++ b/rust/numaflow-testing/src/simplebuffer/error_injector.rs
@@ -21,12 +21,16 @@ pub struct ErrorInjector {
     fail_next_nacks: AtomicUsize,
     /// Fail the next N wip acks.
     fail_next_wip_acks: AtomicUsize,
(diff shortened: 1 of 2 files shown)


yhl25 (commented):
Can we make write method return a future and resolve it using resolve method? Also let's make sure we can inject errors for during future resolution?

rust/numaflow-testing/src/simplebuffer/error_injector.rs
```diff
@@ -0,0 +1,317 @@
...
+
+use std::sync::atomic::{AtomicBool, AtomicU64, AtomicUsize, Ordering};
+use std::time::Duration;
+
+/// Controller for injecting errors during testing.
+///
+/// This allows tests to force specific error conditions like buffer full,
+/// ack failures, nack failures, etc.
+#[derive(Debug)]
+pub struct ErrorInjector {
```
vaibhavtiwari33, line 11: this is really nice

PR #3215, opened 2026-02-08
test: Utilities for unit testing framework

vaibhavtiwari33:
Create helper utilities for testing individual (source, sink, mapper, transformer) components. 

### Testing

### DIFF
--- a/ui/yarn.lock
+++ b/ui/yarn.lock
@@ -7790,9 +7790,9 @@ jsonfile@^6.0.1:
     graceful-fs "^4.1.6"
 
 jsonpath@^1.1.1:
-  version "1.2.0"
-  resolved "https://registry.yarnpkg.com/jsonpath/-/jsonpath-1.2.0.tgz#744fbb8c9a5ba4e3b815432c0fe1756516be7f50"
-  integrity sha512-EVm29wT2coM0QfZd8TREEeMTOxZcyV3oCQ61AM0DrMkVaVCKXtPEm0oJccEbz5P9Oi+JwRkkIt0Bkn63gqCHjg==
+  version "1.2.1"
+  resolved "https://registry.yarnpkg.com/jsonpath/-/jsonpath-1.2.1.tgz#2b74a4bcc78948e43e33ac971138ce0c68bce701"
+  integrity sha512-Jl6Jhk0jG+kP3yk59SSeGq7LFPR4JQz1DU0K+kXTysUhMostbhU3qh5mjTuf0PqFcXpAT7kvmMt9WxV10NyIgQ==
   dependencies:
     esprima "1.2.5"
     static-eval "2.1.1"


rust/numaflow-core/src/mapper/map.rs
vigith, line 1: 👍 

see whether we can remove `global-state-tests` altogether now that we have multi-threaded server. you can do in a follow up PR.

rust/numaflow-core/src/shared/test_utils/server.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    fn do_shutdown(&mut self) {
+        if let Some(tx) = self.shutdown_tx.take() {
+            // It's okay if the receiver is already gone (server exited early).
+            let _ = tx.send(());
+        }
+        if let Some(handle) = self.thread_handle.take() {
+            // FIXME: waiting to join the server thread can block in case of single threaded tokio test runtime
+            // From claude:
+            // * `handle.join()` blocks the tokio runtime thread, waiting for the source server thread to exit
```
yhl25: Do we need `From claude` ?

rust/numaflow-core/src/shared/test_utils/server.rs
```diff
@@ -0,0 +1,128 @@
...
+/// Dropping the handle sends a shutdown signal and joins the thread.
+pub(crate) struct TestServerHandle {
+    /// The UDS socket path where the server is listening.
+    sock_file: PathBuf,
+    /// Channel to trigger server shutdown.
+    shutdown_tx: Option<oneshot::Sender<()>>,
+    /// thread handle for the server runtime.
+    thread_handle: Option<std::thread::JoinHandle<()>>,
+    /// Temp directory kept alive so the socket file persists.
+    _tmp_dir: TempDir,
```
yhl25: Do we need to cleanup the sock file?

rust/numaflow-core/src/shared/test_utils/server.rs
```diff
@@ -0,0 +1,128 @@
...
+{
+    let tmp_dir = TempDir::new().expect("failed to create temp dir");
+    let sock_file = tmp_dir.path().join(format!("{name}.sock"));
+    let server_info_file = tmp_dir.path().join(format!("{name}-server-info"));
+
+    let (shutdown_tx, shutdown_rx) = oneshot::channel();
+
+    let sock = sock_file.clone();
+
+    let thread_handle = std::thread::Builder::new()
```
vigith, line 117: please document why we need this

rust/numaflow-core/src/shared/test_utils/server.rs
vigith: won't it get auto deleted since this it is using TempDir?

vaibhavtiwari33: That is a good question. As you mentioned, from [TempDir doc](https://docs.rs/tempdir/latest/tempdir/struct.TempDir.html) we have:

```
The TempDir type creates a directory on the file system that is deleted once it goes out of scope.
```

But without persisting the TempDir handle, I was actually facing test flakiness due to the client sometimes failing to connect with the server. It turns out there was a race condition between the client connecting to the server using the sock file and the `TempDir` getting deleted (deleting the sock file with it) after it is dropped when the `start_server` method exits.

yhl25: Why not move the scope of the `TempDir` to the server thread? the moment server goes out of scope the dir will get deleted.

vaibhavtiwari33: That's a good point actually, let me do that instead

rust/numaflow-core/src/shared/test_utils/server.rs
vaibhavtiwari33: removed

rust/numaflow-core/src/shared/test_utils/server.rs
vaibhavtiwari33: Moved

rust/numaflow-core/src/shared/test_utils/server.rs
vaibhavtiwari33: Done

PR #3217, opened 2026-02-10
refactor: introduce ISBFactory for creating ISB reader/writer

vigith:
Even though we have made ISB generic via traits (reader and writer), unfortunately we have Jetstream Context being propagated all the way from pipeline. We need to introduce an ISBFactory that can be used to create reader and writer.

### DIFF
--- a/rust/numaflow-core/src/pipeline.rs
+++ b/rust/numaflow-core/src/pipeline.rs
@@ -1,3 +1,4 @@
+use std::marker::PhantomData;
 use std::time::Duration;
 
 use async_nats::jetstream::Context;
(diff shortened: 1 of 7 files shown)


rust/numaflow-core/src/pipeline/isb/jetstream/factory.rs
```diff
@@ -0,0 +1,95 @@
...
+    #[ignore = "requires NATS server"]
+    async fn test_jetstream_factory_creation() {
+        let client = async_nats::connect("localhost:4222").await.unwrap();
+        let context = async_nats::jetstream::new(client);
+        let factory = JetStreamFactory::new(context.clone());
+
+        // Verify the context is accessible
+        assert!(std::ptr::eq(factory.context(), &factory.context));
+    }
+}
```
yhl25, line 94: Do we need these?

yhl25 (approved):

PR #3226, opened 2026-02-11
feat: check initContainerStatuses are in Waiting or Terminated state during health check 

### LINKED ISSUE #3160: InitContainerStatus not used for health checking

dpadhiar:

Follow-up PR for #3201 

### Related issues
Fixes #3160 

### Testing

Unit testing and locally.

### DIFF
--- a/server/apis/v1/health.go
+++ b/server/apis/v1/health.go
@@ -211,16 +211,16 @@ func isVertexHealthy(h *handler, ns string, pipeline string, vertex *dfv1.Vertex
 		// Iterate over all the initContainers in the pod
 		for _, containerStatus := range pod.Status.InitContainerStatuses {
 			// if waiting, check that there is no backoff error
-			if containerStatus.State.Waiting != nil && slices.Contains([]string{"CrashLoopBackOff", "ImagePullBackOff"}, containerStatus.State.Waiting.Reason) {
+			if containerStatus.State.Waiting != nil && slices.Contains(dfv1.UnhealthyWaitingStatus, containerStatus.State.Waiting.Reason) {
 				return false, &resourceHealthResponse{
-					Message: fmt.Sprintf("Init container %q in pod %q is not healthy", containerStatus.Name, pod.Name),
+					Message: fmt.Sprintf("Init container %q in pod %q is in waiting state and not healthy", containerStatus.Name, pod.Name),
 					Code:    "V3",
 				}, nil
 			}
 			// if terminated, check that it did not error
 			if containerStatus.State.Terminated != nil && containerStatus.State.Terminated.Reason == "Error" {
 				return false, &resourceHealthResponse{
-					Message: fmt.Sprintf("Init container %q in pod %q is not healthy", containerStatus.Name, pod.Name),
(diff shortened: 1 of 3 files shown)


server/apis/v1/health.go
```diff
@@ -209,7 +210,14 @@ func isVertexHealthy(h *handler, ns string, pipeline string, vertex *dfv1.Vertex
 		}
 		// Iterate over all the initContainers in the pod
 		for _, containerStatus := range pod.Status.InitContainerStatuses {
-			// check that the container terminated successfully
+			// if waiting, check that there is no backoff error
+			if containerStatus.State.Waiting != nil && slices.Contains([]string{"CrashLoopBackOff", "ImagePullBackOff"}, containerStatus.State.Waiting.Reason) {
+				return false, &resourceHealthResponse{
+					Message: fmt.Sprintf("Init container %q in pod %q is not healthy", containerStatus.Name, pod.Name),
```
juliev0: can we distinguish the message here from the one below so we know which one it is?

server/apis/v1/health.go
```diff
@@ -209,7 +210,14 @@ func isVertexHealthy(h *handler, ns string, pipeline string, vertex *dfv1.Vertex
 		}
 		// Iterate over all the initContainers in the pod
 		for _, containerStatus := range pod.Status.InitContainerStatuses {
-			// check that the container terminated successfully
+			// if waiting, check that there is no backoff error
+			if containerStatus.State.Waiting != nil && slices.Contains([]string{"CrashLoopBackOff", "ImagePullBackOff"}, containerStatus.State.Waiting.Reason) {
```
juliev0: I guess ideally we would use [this](https://github.com/numaproj/numaflow/blob/main/pkg/reconciler/util.go#L32) value from both code locations. I wonder if there's a file in numaflow for shared constants/vars?

dpadhiar: There isn't a shared place right now for this variable. We could create a shared/constants.go file but this will be the only value in it for now.

PR #3231, opened 2026-02-12
test: use SimpleBuffer for testing ISBReaderOrchestrator 

vigith:
Use SimpleBuffer to do more edge case testing for ISBReaderOrchestrator

### DIFF
--- a/rust/numaflow-core/src/pipeline/isb/reader.rs
+++ b/rust/numaflow-core/src/pipeline/isb/reader.rs
@@ -1591,9 +1591,6 @@ mod simplebuffer_tests {
             .await
         });
 
-        // Give it time to start retrying
-        sleep(Duration::from_millis(50)).await;
-
         // Cancel should stop the retry loop
         cancel.cancel();
 
@@ -1622,9 +1619,6 @@ mod simplebuffer_tests {
 
         // Read message
         let msg = rx.next().await.expect("Should receive message");
-
-        // Wait a bit for WIP to be attempted (and fail)
-        sleep(Duration::from_millis(50)).await;
 
         // Ack by dropping msg (is_failed defaults to false)
         drop(msg);
@@ -1761,9 +1755,6 @@ mod simplebuffer_tests {
         // Ack first message to free a permit by taking it out and dropping
         let first_msg = inflight.remove(0);
         drop(first_msg);
-
-        // Small delay to allow the wip_loop to complete ack and release permit
-        sleep(Duration::from_millis(100)).await;
 
         // Now 4th message should come through
         let fourth = tokio::time::timeout(Duration::from_millis(500), rx.next())


rust/numaflow-core/src/pipeline/isb/simplebuffer.rs
```diff
@@ -0,0 +1,491 @@
...
+            ),
+            Offset::String(str_offset) => {
+                // For testing, we use sequence as the first part of the string
+                let seq: i64 = std::str::from_utf8(&str_offset.offset)
+                    .ok()
+                    .and_then(|s| s.parse().ok())
+                    .unwrap_or(0);
+                numaflow_testing::simplebuffer::Offset::new(seq, str_offset.partition_idx)
+            }
+        };
```
vaibhavtiwari33: nit: Should we implement `From` for `impl From<&Offset> for numaflow_testing::simplebuffer::Offset`? 
I see that we already have `impl From<&numaflow_testing::simplebuffer::Offset> for Offset`

Mentioning since `nack` and `mark_wip` also use the same logic

rust/numaflow-core/src/pipeline/isb/simplebuffer.rs
```diff
@@ -0,0 +1,491 @@
...
+    }
+}
+
+/// Test type configuration that uses SimpleBuffer for ISB operations.
+///
+/// This allows testing ISBReaderOrchestrator and ISBWriterOrchestrator
+/// without requiring external infrastructure like NATS.
+#[derive(Clone)]
+#[allow(dead_code)] // Will be used when ISBReaderOrchestrator/ISBWriterOrchestrator integration tests are added
+pub(crate) struct WithSimpleBuffer;
```
vaibhavtiwari33: nit: I see it being constructed in the reader

vigith: I have a local commit for this, will push it in.

rust/numaflow-core/src/pipeline/isb/simplebuffer.rs
vigith: good catch

rust/numaflow-core/src/pipeline/isb/reader.rs
```diff
@@ -1322,3 +1322,497 @@ mod tests {
...
+            fourth.is_err(),
+            "4th message should block due to backpressure"
+        );
+
+        // Ack first message to free a permit by taking it out and dropping
+        let first_msg = inflight.remove(0);
+        drop(first_msg);
+
+        // Small delay to allow the wip_loop to complete ack and release permit
+        sleep(Duration::from_millis(100)).await;
```
yhl25: This if flaky!

rust/numaflow-core/src/pipeline/isb/reader.rs
```diff
@@ -1322,3 +1322,497 @@ mod tests {
...
+                &reader,
+                &offset,
+                &cancel_clone,
+            )
+            .await
+        });
+
+        // Give it time to start retrying
+        sleep(Duration::from_millis(50)).await;
+
```
yhl25: flaky!

yhl25 (approved):

PR #3238, opened 2026-02-15
test: ISBWriterOrchestrator tests using SimpleBuffer

vigith:


rust/numaflow-testing/src/simplebuffer/error_injector.rs
```diff
@@ -113,7 +124,19 @@ impl ErrorInjector {
...
+        // First check if we need to skip this write
+        let skip = self.skip_writes_before_fail.load(Ordering::Relaxed);
+        if skip > 0 {
+            self.skip_writes_before_fail
+                .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |c| {
+                    if c > 0 { Some(c - 1) } else { None }
+                })
+                .ok();
+            return false;
+        }
```
vaibhavtiwari33: nit: check the fetched value directly?

```suggestion
        let skip = self.skip_writes_before_fail
            .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |c| {
                if c > 0 { Some(c - 1) } else { None }
            }).is_ok();
        if skip {
            return false;
        }
```

rust/numaflow-core/src/pipeline/isb/writer.rs
```diff
@@ -997,3 +997,803 @@ mod tests {
...
+
+        // Message should be acked (we exit the loop on cancel)
+        let ack = tokio::time::timeout(Duration::from_secs(2), ack_rx)
+            .await
+            .expect("Should receive ack")
+            .unwrap();
+        assert_eq!(ack, ReadAck::Ack);
+
+        handle.await.unwrap().unwrap();
+    }
```
vaibhavtiwari33, line 1315: nit: Check to assert that no message was added?

```suggestion
        handle.await.unwrap().unwrap();
        assert_eq!(adapter.pending_count(), 0);
    }
```

rust/numaflow-core/src/pipeline/isb/writer.rs
```diff
@@ -997,3 +997,803 @@ mod tests {
...
+
+        // Should succeed via fallback write
+        let ack = tokio::time::timeout(Duration::from_secs(2), ack_rx)
+            .await
+            .expect("Should receive ack")
+            .unwrap();
+        assert_eq!(ack, ReadAck::Ack);
+
+        handle.await.unwrap().unwrap();
+    }
```
vaibhavtiwari33, line 1354: Assertion to check the message was written to buffer
```suggestion
        handle.await.unwrap().unwrap();
        assert_eq!(adapter.pending_count(), 1);
    }
```

rust/numaflow-core/src/pipeline/isb/writer.rs
```diff
@@ -997,3 +997,803 @@ mod tests {
...
+            .await
+            .expect("Should receive ack response")
+            .unwrap();
+        assert_eq!(ack, ReadAck::Ack);
+
+        cancel.cancel();
+        drop(tx);
+
+        handle.await.unwrap().unwrap();
+    }
```
vaibhavtiwari33, line 1401: same here
```suggestion
        handle.await.unwrap().unwrap();
        assert_eq!(adapter.pending_count(), 1);
    }
```

rust/numaflow-core/src/pipeline/isb/writer.rs
```diff
@@ -997,3 +997,803 @@ mod tests {
...
+
+        let (msg, ack_rx) = create_test_message(1, "hello", None);
+        tx.send(msg).await.unwrap();
+
+        // Message should succeed via duplicate detection in fallback write
+        let ack = tokio::time::timeout(Duration::from_secs(2), ack_rx)
+            .await
+            .expect("Should receive ack response")
+            .unwrap();
+        assert_eq!(ack, ReadAck::Ack);
```
vaibhavtiwari33: nit: Instead of unwrapping the timeout, should we assert on the timeout result?

```suggestion
        let ack = tokio::time::timeout(Duration::from_secs(2), ack_rx)
            .await
            .expect("Timeout future should not error out");
        assert!(ack.is_ok(), "Should receive ack before timeout");
        assert_eq!(ack.expect("Should receive an ack"), ReadAck::Ack);
```

rust/numaflow-core/src/pipeline/isb/writer.rs
```diff
@@ -997,3 +997,803 @@ mod tests {
...
+
+        // Only 1 message should be in buffer (duplicate was detected)
+        assert_eq!(adapter.pending_count(), 1);
+    }
+
+    // ==================== PAF Concurrency Tests ====================
+
+    /// Test: PAF concurrency limits concurrent resolutions
+    #[tokio::test]
+    async fn test_paf_concurrency_semaphore() {
```
vaibhavtiwari33: How does this explicitly test paf concurrency resolution?
Should we rather do the negative test, i.e., set paf concurrency to 0 and ensure that no messages are written?

rust/numaflow-core/src/pipeline/isb/writer.rs
```diff
@@ -997,3 +997,803 @@ mod tests {
...
+            writers.insert(*stream_name, adapter.writer());
+
+            let writer_config = BufferWriterConfig {
+                streams: vec![stream],
+                buffer_full_strategy: buffer_full_strategy.clone(),
+                ..Default::default()
+            };
+
+            vertex_configs.push(ToVertexConfig {
+                name: Box::leak(format!("vertex-{}", idx).into_boxed_str()),
```
vaibhavtiwari33, line 1622: nit: use vertex_name from above?

rust/numaflow-core/src/pipeline/isb/writer.rs
```diff
@@ -997,3 +997,803 @@ mod tests {
...
+        // The key assertion: even though the first message failed,
+        // the semaphore was released (permit dropped when spawn task completed).
+        // We verify this by checking that the orchestrator completed successfully
+        // (wait_for_paf_resolvers was able to acquire all permits).
+    }
+
+    /// Test: Multiple messages with paf_concurrency=1 are processed sequentially,
+    /// and the semaphore properly gates concurrent PAF resolutions.
+    #[tokio::test]
+    async fn test_paf_concurrency_one_processes_sequentially() {
```
vaibhavtiwari33: nit: keep single stream tests together? move this one above multi stream tests

rust/numaflow-core/src/pipeline/isb/writer.rs
```diff
@@ -997,3 +997,803 @@ mod tests {
...
+        let adapter = SimpleBufferAdapter::new(SimpleBuffer::new(100, 0, "test-buffer"));
+        // Add resolve latency to make timing observable
+        adapter.error_injector().set_resolve_latency(10);
```
vaibhavtiwari33: nit: bump this up to 20 to account for overhead in CI?

vigith: why do you think so? we are looking for atleast 10ms latency, so CI overhead shouldn't matter.

vaibhavtiwari33: this test ensures that the messages are processed sequentially by asserting on the total time taken for processing. 
just thinking out loud, assuming CI overhead takes 10ms and the messages get processed within 20ms, the test would still pass. 
Assuming the overhead remains the same across both scenarios (resolve latency 10 vs 20), there would be more chance of the test failing in the second scenario if the messages are getting processed quicker (for some reason)

rust/numaflow-core/src/pipeline/isb/writer.rs
vigith: will remove this test

rust/numaflow-core/src/pipeline/isb/writer.rs
```diff
@@ -997,3 +997,804 @@ mod tests {
...
+        // Stream 1 should have the message (write succeeded)
+        assert_eq!(adapter1.pending_count(), 1);
+        // Stream 2 should also have the message from initial write (before resolve failed)
+        assert_eq!(adapter2.pending_count(), 1);
+    }
+
+    /// Test: Semaphore permit is properly released even when PAF resolution fails,
+    /// allowing subsequent messages to be processed.
+    #[tokio::test]
+    async fn test_semaphore_released_on_paf_failure() {
```
vaibhavtiwari33, line 1717: sry, I didn't get what is the difference between this test and the test above?

`test_multi_vertex_partial_failure_naks_message`

rust/numaflow-core/src/pipeline/isb/writer.rs
vigith: ain't we doing a `>`?

vaibhavtiwari33: True, my statement still stands though, but claiming ~5-10ms CI overhead is also a bit of a stretch, so I think your change should be fine.

rust/numaflow-core/src/pipeline/isb/writer.rs
vigith: i updated `test_semaphore_released_on_paf_failure` to send a second message to prove that it is not blocked. if you think it is not required, we can remove that test.

rust/numaflow-core/src/pipeline/isb/writer.rs
vigith: it shouldn't matter even if CI overhead is > 100ms too, right? since we are checking the lower bound of time

rust/numaflow-core/src/pipeline/isb/writer.rs
```diff
@@ -997,3 +997,774 @@ mod tests {
...
+        assert_eq!(ack2, ReadAck::Ack);
+
+        drop(tx);
+        handle.await.unwrap().unwrap();
+
+        // Both messages should have been written to both streams
+        // (msg1's retry detected duplicate, so still only 1 entry per message)
+        assert_eq!(adapter1.pending_count(), 2);
+        assert_eq!(adapter2.pending_count(), 2);
+    }
```
vaibhavtiwari33, line 1774: Add a test for conditional forwarding verification.

```suggestion
    fn create_multi_vertex_orchestrator(
    adapters: Vec<(&'static str, &SimpleBufferAdapter, Option<Box<ForwardConditions>>)>,
    buffer_full_strategy: BufferFullStrategy,
    paf_concurrency: usize,
) -> (ISBWriterOrchestrator<WithSimpleBuffer>, CancellationToken) {
    let cancel = CancellationToken::new();

    let mut writers = HashMap::new();
    let mut vertex_configs = vec![];

    for (idx, (stream_name, adapter, condition)) in adapters.iter().enumerate() {
        let vertex_name: &'static str = Box::leak(format!("vertex-{}", idx).into_boxed_str());
        let stream = Stream::new(stream_name, vertex_name, 0);
        writers.insert(*stream_name, adapter.writer());

        let writer_config = BufferWriterConfig {
            streams: vec![stream],
            buffer_full_strategy: buffer_full_strategy.clone(),
            ..Default::default()
        };

        vertex_configs.push(ToVertexConfig {
            name: Box::leak(format!("vertex-{}", idx).into_boxed_str()),
            partitions: 1,
            writer_config,
            conditions: condition.clone(),
            to_vertex_type: VertexType::Sink,
        });
    }

    let components = ISBWriterOrchestratorComponents {
        config: vertex_configs,
        writers,
        paf_concurrency,
        watermark_handle: None,
        vertex_type: VertexType::Source,
    };

    let orchestrator = ISBWriterOrchestrator::<WithSimpleBuffer>::new(components);
    (orchestrator, cancel)
}

    /// Test: When writing to multiple downstream vertices and one vertex's PAF resolution fails
    /// (and retry also fails due to cancellation), the message should be NAK'd.
    ///
    /// This tests the `resolved_offsets.len() != n` path in resolve_and_finalize.
    #[tokio::test]
    async fn test_multi_vertex_partial_failure_naks_message() {
        // Create two separate buffers with independent error injection
        let adapter1 = SimpleBufferAdapter::new(SimpleBuffer::new(100, 0, "stream-1"));
        let adapter2 = SimpleBufferAdapter::new(SimpleBuffer::new(100, 1, "stream-2"));

        // Stream 2 configuration:
        // 1. Initial write succeeds (skip 1 write before failing)
        // 2. Resolve fails (fail_resolves(1))
        // 3. Retry writes fail (fail all writes after skipping the first one)
        // 4. Cancellation will cause write_with_retry to exit with error
        adapter2.error_injector().fail_resolves(1);
        adapter2
            .error_injector()
            .skip_writes_then_fail(1, usize::MAX);

        let (orchestrator, cancel) = create_multi_vertex_orchestrator(
            vec![("stream-1", &adapter1, None), ("stream-2", &adapter2, None)],
            BufferFullStrategy::RetryUntilSuccess,
            10,
        );

        let (tx, rx) = mpsc::channel(10);
        let messages_stream = ReceiverStream::new(rx);

        let handle = orchestrator
            .streaming_write(messages_stream, cancel.clone())
            .await
            .unwrap();

        let (msg, ack_rx) = create_test_message(1, "hello", None);
        tx.send(msg).await.unwrap();

        // Cancel after a short delay to allow the initial write and resolve to happen,
        // then cause the retry loop to exit with error
        let cancel_clone = cancel.clone();
        tokio::spawn(async move {
            // Wait for the resolve to fail and retry to start
            sleep(Duration::from_millis(50)).await;
            cancel_clone.cancel();
        });

        // The message should be NAK'd because stream-2's retry fails due to cancellation
        let ack = tokio::time::timeout(Duration::from_secs(2), ack_rx)
            .await
            .expect("Should receive ack response")
            .unwrap();
        assert_eq!(ack, ReadAck::Nak);

        drop(tx);

        handle.await.unwrap().unwrap();

        // Stream 1 should have the message (write succeeded)
        assert_eq!(adapter1.pending_count(), 1);
        // Stream 2 should also have the message from initial write (before resolve failed)
        assert_eq!(adapter2.pending_count(), 1);
    }

    /// Test: Semaphore permit is properly released even when PAF resolution fails,
    /// allowing subsequent messages to be processed.
    ///
    /// This test differs from `test_multi_vertex_partial_failure_naks_message` in that:
    /// - It uses `paf_concurrency = 1` (vs 10) to make semaphore behavior critical
    /// - It sends a **second message after the first fails** to explicitly verify
    ///   the semaphore permit was released
    ///
    /// With `paf_concurrency = 1`, if the permit wasn't released after the first
    /// message's failure, the second message would never be processed (deadlock).
    /// The first test only verifies NAK behavior; this test verifies resource cleanup.
    #[tokio::test]
    async fn test_semaphore_released_on_paf_failure() {
        let adapter1 = SimpleBufferAdapter::new(SimpleBuffer::new(100, 0, "stream-1"));
        let adapter2 = SimpleBufferAdapter::new(SimpleBuffer::new(100, 1, "stream-2"));

        // First message: stream-2 will fail resolve only (no write failures)
        // The resolve failure will trigger write_with_retry, which will succeed
        // (detecting the duplicate from the initial async_write).
        // The key is that the first message's PAF resolution "fails" initially.
        adapter2.error_injector().fail_resolves(1);

        // Use paf_concurrency = 1 to ensure sequential processing
        // This makes semaphore release critical - if not released, second message blocks forever
        let (orchestrator, cancel) = create_multi_vertex_orchestrator(
            vec![("stream-1", &adapter1, None), ("stream-2", &adapter2, None)],
            BufferFullStrategy::RetryUntilSuccess,
            1,
        );

        let (tx, rx) = mpsc::channel(10);
        let messages_stream = ReceiverStream::new(rx);

        let handle = orchestrator
            .streaming_write(messages_stream, cancel.clone())
            .await
            .unwrap();

        // Send first message (resolve will fail on stream-2, but retry write succeeds as duplicate)
        let (msg1, ack_rx1) = create_test_message(1, "first", None);
        tx.send(msg1).await.unwrap();

        // First message should be ACK'd (resolve failed but retry succeeded via duplicate detection)
        let ack1 = tokio::time::timeout(Duration::from_secs(2), ack_rx1)
            .await
            .expect("Should receive ack response for msg1")
            .unwrap();
        assert_eq!(ack1, ReadAck::Ack);

        // Send second message - this is the key assertion!
        // If the semaphore permit wasn't released after msg1's PAF resolution completed,
        // this message would never be processed (blocked waiting for permit)
        let (msg2, ack_rx2) = create_test_message(2, "second", None);
        tx.send(msg2).await.unwrap();

        // Second message should be ACK'd (proves semaphore was released)
        let ack2 = tokio::time::timeout(Duration::from_secs(2), ack_rx2)
            .await
            .expect("Should receive ack response for msg2 - semaphore should have been released")
            .unwrap();
        assert_eq!(ack2, ReadAck::Ack);

        drop(tx);
        handle.await.unwrap().unwrap();

        // Both messages should have been written to both streams
        // (msg1's retry detected duplicate, so still only 1 entry per message)
        assert_eq!(adapter1.pending_count(), 2);
        assert_eq!(adapter2.pending_count(), 2);
    }

    /// Test: Conditional forwarding based on tag values.
    ///
    /// This test verifies that messages are forwarded to the correct streams based on tag values.
    #[tokio::test]
    async fn test_conditional_forwarding() {
        let adapter1 = SimpleBufferAdapter::new(SimpleBuffer::new(100, 0, "stream-1"));
        let adapter2 = SimpleBufferAdapter::new(SimpleBuffer::new(100, 1, "stream-2"));

        let condition1: Option<Box<ForwardConditions>> = Some(Box::new(ForwardConditions {
            tags: Box::new(TagConditions {
                operator: Some(String::from("or")),
                values: vec![String::from("tag1")],
            }),
        }));

        let condition2: Option<Box<ForwardConditions>> = Some(Box::new(ForwardConditions {
            tags: Box::new(TagConditions {
                operator: Some(String::from("or")),
                values: vec![String::from("tag2")],
            }),
        }));

        // Use paf_concurrency = 1 to ensure sequential processing
        // This makes semaphore release critical - if not released, second message blocks forever
        let (orchestrator, cancel) = create_multi_vertex_orchestrator(
            vec![("stream-1", &adapter1, condition1), ("stream-2", &adapter2, condition2)],
            BufferFullStrategy::RetryUntilSuccess,
            1,
        );

        let (tx, rx) = mpsc::channel(10);
        let messages_stream = ReceiverStream::new(rx);

        let handle = orchestrator
            .streaming_write(messages_stream, cancel.clone())
            .await
            .unwrap();

        let (msg1, ack_rx1) = create_test_message(1, "first", Some(vec!["tag1"]));
        tx.send(msg1).await.unwrap();

        let (msg2, ack_rx2) = create_test_message(2, "second", Some(vec!["tag2"]));
        tx.send(msg2).await.unwrap();

        let ack1 = tokio::time::timeout(Duration::from_secs(2), ack_rx1)
            .await
            .expect("Should receive ack response for msg1")
            .unwrap();
        assert_eq!(ack1, ReadAck::Ack);

        let ack2 = tokio::time::timeout(Duration::from_secs(2), ack_rx2)
            .await
            .expect("Should receive ack response for msg2")
            .unwrap();
        assert_eq!(ack2, ReadAck::Ack);

        drop(tx);
        handle.await.unwrap().unwrap();

        // Only one message should have been written to each stream
        assert_eq!(adapter1.pending_count(), 1);
        assert_eq!(adapter2.pending_count(), 1);
    }
```

PR #3241, opened 2026-02-16
fix: minor fixes for reduce 

yhl25:
* Fix late message handling - Improved late message detection logic to handle edge cases when watermark is -1 (pipeline just started), and moved watermark updates to occur only after message acceptance
* Add drop metrics for unaligned reducer
* Fix WAL message serialization - Preserve the is_late flag when serializing/deserializing messages to/from WAL.
* Fix rater partition counting - Sum metric counts for the same partition name instead of overwriting, as there may be multiple metric series with the same partition but different label values
* Add source partition label - Introduce a new source_partition metric label to get metrics at partition level.
* Improve logging - Added watermark summary logs and improved log messages for better debugging of watermark progression and late message handling

### DIFF
--- a/pkg/daemon/server/service/rater/rater_test.go
+++ b/pkg/daemon/server/service/rater/rater_test.go
@@ -27,10 +27,12 @@ import (
 	"testing"
 	"time"
 
+	dto "github.com/prometheus/client_model/go"
 	"github.com/stretchr/testify/assert"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
+	"github.com/numaproj/numaflow/pkg/metrics"
 	sharedqueue "github.com/numaproj/numaflow/pkg/shared/queue"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/daemon/server/service/rater/rater.go
```diff
@@ -359,35 +359,37 @@ func (r *Rater) getPodPendingCounts(vertexName, podName string, metricsData map[
...
-		metricsList := value.GetMetric()
-		partitionReadCount := make(map[string]float64)
-		for _, ele := range metricsList {
-			var partitionName string
-			for _, label := range ele.Label {
-				if label.GetName() == metrics.LabelPartitionName {
-					partitionName = label.GetValue()
-					break
-				}
+	value, ok := metricsData[readTotalMetricName]
```
KeranYang, line 362: Thanks for making this more readable. Just to confirm, there is no logical change here.

pkg/mvtxdaemon/server/service/rater/rater.go
```diff
@@ -268,19 +268,22 @@ func (r *Rater) getPodMetrics(podName string) map[string]*dto.MetricFamily {
...
+
+	metricsList := value.GetMetric()
+	// Sum all metric values as there may be multiple series with different labels
+	// (e.g., different source_partition values)
+	// We use Untyped here as the counter metric family shows up as untyped from the rust client
+	// https://github.com/prometheus/client_rust/issues/194
+	var totalCount float64
+	for _, ele := range metricsList {
+		totalCount += ele.Untyped.GetValue()
+	}
```
KeranYang, line 285: nit: is it possible to unit test it?

pkg/daemon/server/service/rater/rater.go
yhl25: No logical change, the logical change was made in my previous PR.

vigith (approved):

PR #3252, opened 2026-02-20
fix: move BoxedFuture to stack pinning for async_write

vigith:
Change PendingWrite from `Pin<Box<dyn Future>>` to an `enum` with concrete variants (`JetStream`, `SimpleBuffer`).

This avoids our layer of heap allocation in `async_write`, which is a hot code path. The enum implements `Future` directly and maintains object-safety for future dynamic dispatch. 

Joke is, `async_nats` library's `PublishAckFuture` internally boxes via `IntoFuture::into_future()`, so there is one allocation but, we eliminate the additional boxing we were doing on top of that.


vigith:
This is getting a bit complex in the name of efficiency.

PR #3260, opened 2026-02-24
chore: parse daemon responses properly

KeranYang:

This PR ensures rust daemon server can serve golang daemon client. We make it work by parsing the protobuf response to json following [ProtoJSON](https://protobuf.dev/programming-guides/json/) format.

### Testing

Tested by setting `server.daemon.client.protocol` to both `http` and `grpc` and verifying dummy responses are properly rendered on UI.

Screenshot 2026-02-23 at 5 30 56 PM

Screenshot 2026-02-23 at 5 30 45 PM

Screenshot 2026-02-23 at 5 31 21 PM

### DIFF
--- a/rust/numaflow-daemon/Cargo.toml
+++ b/rust/numaflow-daemon/Cargo.toml
@@ -16,7 +16,6 @@ tower.workspace = true
 axum.workspace = true
 axum-server.workspace = true
 rustls.workspace = true
-rustls-pki-types = "1.11"
 bytes.workspace = true
 http.workspace = true
 serde.workspace = true
@@ -25,6 +24,7 @@ chrono.workspace = true
 prost-types.workspace = true
 hyper = "1.6.0"
 http-body-util = "0.1"
+rustls-pki-types = "1.11"
 
 [lints]
 workspace = true


rust/numaflow-daemon/src/tls.rs
```diff
@@ -1,14 +1,22 @@
...
 
+use std::sync::Arc;
+
 use crate::error::{Error, Result};
 use axum_server::tls_rustls::RustlsConfig;
 use rcgen::{
     CertificateParams, DistinguishedName, DnType, ExtendedKeyUsagePurpose, KeyPair, KeyUsagePurpose,
 };
+use rustls::server::ServerConfig;
+use rustls_pemfile::{certs, private_key};
```
vigith: https://github.com/rustls/pemfile i think it is archived. please confirm?

rust/Cargo.toml
vigith, line 1: revert

rust/numaflow-daemon/src/service.rs
```diff
@@ -4,10 +4,18 @@ use numaflow_pb::servers::mvtxdaemon::{
...
 use std::collections::HashMap;
+use std::env;
 use std::result::Result;
 use tonic::{Request, Response, Status};
 
+/// Env var set by daemon deployment; matches pkg/apis/numaflow/v1alpha1/const.go EnvMonoVertexName.
+/// This is a temporary solution to get the MonoVertex name.
+/// On golang mvtx daemon end, we load the entire NUMAFLOW_MONO_VERTEX_OBJECT to get the name.
+/// TODO - switch to the full CR when we need more informations.
+const ENV_MONO_VERTEX_NAME: &str = "NUMAFLOW_MONO_VERTEX_NAME";
```
vigith, line 17: Fix this in the follow up PR. Perhaps you can move loading `NUMAFLOW_MONO_VERTEX_OBJECT` from `numaflow-core` to `numaflow-shared` and use that wiring here?

KeranYang: will do in the follow up PR.

rust/Cargo.toml
KeranYang: Sorry I don't think we need to revert, I added 3 new dependencies that are all needed for this change. I did notice the sequence was wrong so fixed in the new commit.

vigith (approved):

Issue #3177, opened 2026-02-01
Watermark handling should be generic over K/V Store


vigith:
We need a pluggable Watermark handling Store. Our design of Watermark progression is based on a Key-Value Store with Watch semantics

## Writer

ISB Writer internally publishes the Watermark and relies on Jetstream semantics. We need a trait for Publishing the Watermark and it should be injected into ISB Writer.

Due to this, `write` does not return an `Offset` and it is consumed internally.

PR #3261, opened 2026-02-24
feat: in-memory kv store with impl KVStore for testing

vigith:
In-memory Key Value store track watermark propagation. The store is capable of handling other K/V semantics too.

Related to #3177

### DIFF
--- a/rust/numaflow-testing/src/simplekvstore/store.rs
+++ b/rust/numaflow-testing/src/simplekvstore/store.rs
@@ -321,8 +321,6 @@ struct SimpleKVWatchStream {
     error_injector: Arc<KVErrorInjector>,
     /// Whether the stream is closed.
     closed: bool,
-    /// Pending receive future.
-    pending_recv: Option<Pin<Box<dyn std::future::Future<Output = RecvResult> + Send + 'static>>>,
 }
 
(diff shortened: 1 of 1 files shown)


rust/numaflow-testing/src/simplekvstore/store.rs
```diff
@@ -0,0 +1,649 @@
...
+            history_index: 0,
+            receiver,
+            close_receiver,
+            close_after,
+            items_emitted: 0,
+            error_injector,
+            closed: false,
+            pending_recv: None,
+        }
+    }
```
yhl25:
```suggestion
}

/// Result type for receive operation.
type RecvResult = Result<KVHistoryEntry, broadcast::error::RecvError>;

impl SimpleKVWatchStream {
    fn new(
        history: Vec<KVHistoryEntry>,
        receiver: broadcast::Receiver<KVHistoryEntry>,
        close_receiver: broadcast::Receiver<()>,
        close_after: usize,
        error_injector: Arc<KVErrorInjector>,
    ) -> Self {
        Self {
            history,
            history_index: 0,
            receiver,
            close_receiver,
            close_after,
            items_emitted: 0,
            error_injector,
            closed: false,
        }
    }
```

Issue #3271, opened 2026-02-27
Pipeline Definition changes to support Ordered Processing


vigith:
# Summary

We need to indicate that the processing needs to be ordered. User should be able to define it either at the vertex level or at pipeline scope (all vertices will be ordered).

This means, we will add 
```yaml
spec:
  ordered:
    enabled: true
``` 
at top level, or at each vertex like below

```
  vertices:
    - name: example
      ordered:
        enabled: false
```

The vertex will override the top-level spec. The reduce will have no impact due to this because it is already partitioned.

```yaml
apiVersion: numaflow.numaproj.io/v1alpha1
kind: Pipeline
metadata:
  name: simple-pipeline
spec:
  ordered:
    enabled: true
    ..
  vertices:
    - name: in
      source:
        # A self data generating source
        generator:
          rpu: 5
          duration: 1s
          keyCount: 5
          value: 5
    - name: cat
      partitions: 2
      ordered:
        enabled: false
      udf:
        container:
          image: quay.io/numaio/numaflow-go/map-cat:stable # A UDF which simply cats the message
          imagePullPolicy: Always
    - name: out
      sink:
        # A simple log printing sink
        log: {}
  edges:
    - from: in
      to: cat
    - from: cat
      to: out
```

PR #3278, opened 2026-02-27
feat(controller): enable ordered processing spec

vigith:
Closes: #3271


pkg/reconciler/validator/pipeline_validate.go
```diff
@@ -767,3 +771,17 @@ func validateSQSSink(sqs dfv1.SqsSink) error {
 
 	return nil
 }
+
+// validateOrderedProcessing validates the ordered processing configuration.
+// The ordered configuration is simple (just a boolean), so minimal validation is needed.
+// Note: Reduce vertices ignore the ordered setting as they are already partitioned.
+func validateOrderedProcessing(pl dfv1.Pipeline) error {
```
vigith, line 778: in the followup PR, we need to throw an warning or throw an error if autoscaling is set.

PR #3281, opened 2026-03-01
feat(controller): fix scaling for ordered processing and validations

### LINKED ISSUE #3272: Controller to use fixed min/max scale with Ordered Processing
# Summary

Controller should set min and max replicas to same as the number of partitions set. We should disable the autoscaling when ordered processing is configured. This is because each partition can be only processed by one pod.

### Validations
Validations are only applicable to Map and Sink vertices. Source is always ordered and Reduce does not required ordering.
* If partitions is not set, then we should throw an error
* Validation should fail if scale is set by user.

vigith:
Closes #3272

### DIFF
--- a/pkg/reconciler/validator/pipeline_validate.go
+++ b/pkg/reconciler/validator/pipeline_validate.go
@@ -774,7 +774,6 @@ func validateSQSSink(sqs dfv1.SqsSink) error {
 
 // validateOrderedProcessing validates the ordered processing configuration.
 // For map and sink vertices with ordered processing enabled:
-// - Partitions must be set
 // - Scale config (min/max) must not be set by user
 // Note: Source vertices are always ordered, and reduce vertices are already partitioned,
 // so ordered processing only applies to map and sink vertices.
@@ -797,13 +796,7 @@ func validateOrderedProcessing(pl dfv1.Pipeline) error {
 		}
 
 		// For map and sink vertices with ordered processing enabled:
-
-		// 1. Partitions must be set
-		if v.Partitions == nil || *v.Partitions <= 0 {
-			return fmt.Errorf("vertex %q: partitions must be set and greater than 0 when ordered processing is enabled", v.Name)
-		}
-
-		// 2. Scale config (min/max) must not be set by user
+		// Scale config (min/max) must not be set by user
 		// Users should not configure autoscaling for ordered processing vertices
 		if v.Scale.Min != nil || v.Scale.Max != nil {
 			return fmt.Errorf("vertex %q: scale (min/max) should not be set when ordered processing is enabled. Replicas will be fixed to partition count", v.Name)
--- a/pkg/reconciler/validator/pipeline_validate_test.go
+++ b/pkg/reconciler/validator/pipeline_validate_test.go
@@ -1495,7 +1495,7 @@ func TestValidateOrderedProcessing(t *testing.T) {
 			wantErr: false,
 		},
 		{
-			name: "map vertex with ordered but no partitions - should fail",
+			name: "map vertex with ordered but no partitions - should succeed (defaults to 1)",
 			pl: dfv1.Pipeline{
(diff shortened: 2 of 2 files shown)


pkg/reconciler/validator/pipeline_validate.go
```diff
@@ -773,15 +773,42 @@ func validateSQSSink(sqs dfv1.SqsSink) error {
...
 
-	// Vertex-level validation: ensure vertex-level ordered config is valid.
-	// Since it's just a boolean field, no additional validation is needed.
-	// Reduce vertices will ignore this setting (handled in GetEffectiveOrderedConfig).
+		// For map and sink vertices with ordered processing enabled:
+
+		// 1. Partitions must be set
+		if v.Partitions == nil || *v.Partitions <= 0 {
+			return fmt.Errorf("vertex %q: partitions must be set and greater than 0 when ordered processing is enabled", v.Name)
+		}
```
vigith: we cannot scale to 0.

pkg/reconciler/pipeline/controller.go
```diff
@@ -661,8 +661,19 @@ func buildVertices(pl *dfv1.Pipeline) map[string]dfv1.Vertex {
...
+		} else if v.GetEffectiveOrderedConfig(pl.Spec.Ordered) && (v.IsMapUDF() || v.IsASink()) {
+			// For map and sink vertices with ordered processing enabled, set replicas to partition count
+			// This ensures each partition is processed by exactly one pod for ordered processing
+			// Note: Source vertices are always ordered, reduce vertices are already handled above
+			partitions := pl.NumOfPartitions(v.Name)
+			replicas = int32(partitions)
+			// Disable autoscaling for ordered processing vertices by setting min = max = partitions
+			vCopy.Scale.Disabled = true
+			vCopy.Scale.Min = &replicas
+			vCopy.Scale.Max = &replicas
```
yhl25, line 676: Should we override the value or throw a validation error so that user can set it?

vigith: we throw validation error https://github.com/numaproj/numaflow/pull/3281/changes#diff-e474f76c584e90fc6e5032ecf11887abcacb844443dec8e431933aa53e2c99abR808-R810

pkg/reconciler/validator/pipeline_validate.go
```diff
@@ -773,15 +773,42 @@ func validateSQSSink(sqs dfv1.SqsSink) error {
...
+			continue
+		}
 
-	// Vertex-level validation: ensure vertex-level ordered config is valid.
-	// Since it's just a boolean field, no additional validation is needed.
-	// Reduce vertices will ignore this setting (handled in GetEffectiveOrderedConfig).
+		// For map and sink vertices with ordered processing enabled:
+
+		// 1. Partitions must be set
+		if v.Partitions == nil || *v.Partitions <= 0 {
```
yhl25: Should we default to 1?

pkg/reconciler/validator/pipeline_validate.go
```diff
@@ -773,15 +773,42 @@ func validateSQSSink(sqs dfv1.SqsSink) error {
...
+			return fmt.Errorf("vertex %q: partitions must be set and greater than 0 when ordered processing is enabled", v.Name)
+		}
+
+		// 2. Scale config (min/max) must not be set by user
+		// Users should not configure autoscaling for ordered processing vertices
+		if v.Scale.Min != nil || v.Scale.Max != nil {
+			return fmt.Errorf("vertex %q: scale (min/max) should not be set when ordered processing is enabled. Replicas will be fixed to partition count", v.Name)
```
yhl25: Do we fail even if users has set min and max the same as the partition count?

vigith: Yes, if `!nil`, it will fail

yhl25: But that is a valid config right?

PR #3282, opened 2026-03-01
test: Tests for sink forwarder using simple buffer backed ISBs

vaibhavtiwari33:
Tests for sink forwarder using simple buffer backed ISBs

### Testing
This is the test

### DIFF
--- a/rust/numaflow-core/src/pipeline/forwarder/sink_forwarder.rs
+++ b/rust/numaflow-core/src/pipeline/forwarder/sink_forwarder.rs
@@ -519,7 +519,8 @@ mod simple_buffer_tests {
 
         // Shutdown
         cln_token.cancel();
-        let forwarder_result = tokio::time::timeout(Duration::from_secs(TIMEOUT), forwarder_handle).await;
+        let forwarder_result =
+            tokio::time::timeout(Duration::from_secs(TIMEOUT), forwarder_handle).await;
         assert!(
             forwarder_result.is_ok(),
             "Forwarder task should complete gracefully"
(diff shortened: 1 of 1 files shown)


rust/numaflow-core/src/pipeline/forwarder/sink_forwarder.rs
```diff
@@ -342,6 +342,527 @@ where
...
+                .await
+                .unwrap();
+
+        // Create and start the SinkForwarder
+        let forwarder = SinkForwarder::<WithSimpleBuffer>::new(isb_reader, sink_writer).await;
+
+        let forwarder_cln = cln_token.clone();
+        let forwarder_handle = tokio::spawn(async move { forwarder.start(forwarder_cln).await });
+
+        tokio::time::sleep(Duration::from_millis(500)).await;
```
vigith: this can be flaky, right?

rust/numaflow-core/src/pipeline/forwarder/sink_forwarder.rs
```diff
@@ -342,6 +342,527 @@ where
...
+        } = SinkTestHandle::create_sink(
+            TestSinkType::UserDefined(sink_svc),
+            None,
+            None,
+            batch_size,
+        )
+        .await;
+
+        // ISB Reader Orchestrator
+        let input_stream = Stream::new("sink-input-buffer", "test-in", 0);
```
vigith, line 466: why is `-buffer` in there but not in https://github.com/numaproj/numaflow/pull/3282/changes#diff-6b6458796b57b7564d42246f83ddb31f9411fc3275a7e652ae7f06ba008276e2R437

rust/numaflow-core/src/pipeline/forwarder/sink_forwarder.rs
vaibhavtiwari33: removed

rust/numaflow-core/src/pipeline/forwarder/sink_forwarder.rs
vaibhavtiwari33: The name of the simple buffer adaptor could be different from the stream itself IMO.

vigith (approved):

PR #3283, opened 2026-03-02
test: tests for source forwarder using simple buffer ISB

vaibhavtiwari33:
tests for source forwarder using simple buffer ISB

### Testing
this is the test

### DIFF
--- a/docs/user-guide/reference/ordered-processing.md
+++ b/docs/user-guide/reference/ordered-processing.md
@@ -1,0 +1,153 @@
+# Ordered Processing
+
+> **Available from v1.8**
+
+By default, Numaflow improves throughput by distributing work across any available processing unit, which results in 
+out-of-order processing. However, some workflows require messages to be processed in a deterministic order — for 
+example, a create-update-delete sequence where you cannot update a record before it has been created.
+
+Ordered processing in Numaflow provides partitioned FIFO semantics: within a partition, the N-th message is processed 
+only after the (N-1)-th message completes.
+
+## How It Works
+
+Ordered processing works differently depending on the vertex type:
+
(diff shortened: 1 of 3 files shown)


rust/numaflow-core/src/pipeline/forwarder/source_forwarder.rs
```diff
@@ -252,6 +252,449 @@ where
...
+            watermark_handle: None,
+            vertex_type: VertexType::Source,
+        };
+
+        ISBWriterOrchestrator::new(writer_components)
+    }
+
+    // Test source forwarder with a single output stream using SimpleBuffer.
+    // Reads from a UD source and writes to a SimpleBuffer-backed ISB for verification.
+    #[tokio::test(flavor = "multi_thread")]
```
vigith, line 389: why do we have to make it multi_thread? default is single thread? what will happen if we do not make it multi threaded?

vaibhavtiwari33: It should be fine in this case tbh. I added this since the UDSource was to be spawned in a separate os thread, but it will not affect the outcome here if we remove multi-thread

vigith: let's figure out why `multi_thread` is really required only for failure path.

vaibhavtiwari33: Updated the comment above explaining why we need it.

vigith (approved):
Should we add tests for edge cases and negative cases
  - Source that returns zero messages (empty batch)
  - Cancellation token triggered mid-forwarding (graceful shutdown)

PR #3287, opened 2026-03-05
refactor: Merge Heartbeat and OT Buckets for Watermark Progression

yhl25:
## Problem Statement
The existing watermark system used two separate KV buckets for tracking watermark progression:

**Offset-Timeline (OT) Bucket**: Stored watermark-offset mappings to track message progression through the pipeline.
**Heartbeat (HB) Bucket**: A dedicated bucket solely for tracking processor liveness, processors would periodically write to this bucket to signal they're alive.

This dual-bucket architecture introduced several challenges:

* Synchronization Complexity: Having two sources of truth for processor state created race conditions and synchronization issues. A processor could appear "alive" in the heartbeat bucket while its OT entries were stale, or vice versa. The system had to reconcile state from both buckets, adding complexity to processor lifecycle management.

* No Single Source of Truth: When determining if a processor is alive and what its current watermark is, the system had to consult both buckets. This made reasoning about the system harder and introduced edge cases where the two buckets could be out of sync.

* Infrastructure Overhead: Each watermark-enabled edge required creating, managing, and monitoring an extra KV bucket. This doubled the number of buckets in the system for watermark tracking.

Additionally, for source watermarks, there was no mechanism to know how many processors (partitions) should be reporting before the watermark could be considered valid. During startup or rebalancing, fetching watermarks before all processors were up would result in incorrect watermark values.

## Solution

**Unified Bucket with KV Creation Timestamps**: Instead of maintaining a separate heartbeat bucket, we now use the KV store's native entry creation timestamp (`KVEntry.created`) as the liveness signal. When a processor publishes a `WMB`  to the OT bucket, the KV store records when that entry was created. The `ProcessorManager` uses this timestamp to determine if a processor is still alive, if the entry is too old, the processor is marked inactive or deleted.

**Single Source of Truth**: All processor state (watermark, offset, and liveness) now comes from a single OT bucket. The ProcessorManager watches this bucket for updates and uses the entry timestamps to track processor health. No reconciliation between buckets is needed.

**Processor Count in WMB for Source Watermarks**: Sources now include partition information in the WMB itself:

* Each source partition publishes its watermark with `processor_count` indicating the total number of partitions.
* The `SourceWatermarkFetcher` reads the maximum `processor_count` from all active processors' head WMBs.
* Watermarks are only computed when the number of active processors matches the expected count.
* This ensures watermark stability during startup, we wait for all processors to be up before fetching a valid watermark.

**Periodic Publishing for Liveness** : The ISBWatermarkPublisher now publishes WMBs periodically (based on a configurable delay) even when the watermark hasn't changed. This ensures the KV entry timestamps stay fresh, allowing downstream vertices to detect processor liveness without a separate heartbeat mechanism.

## Key Changes

* Removed HB bucket creation from the controller, only OT buckets are created now.
* `ProcessorManager` determines liveness using KVEntry.created timestamp instead of watching a separate HB bucket.
* `WMB` struct now includes an optional `processor_count` field for source watermarks.
* `SourceWatermarkFetcher` waits for all expected processors (based on `processor_count` in WMBs) before computing a valid watermark.
* Source implementations (Kafka, Pulsar, JetStream, NATS, SQS, etc.) now report `active_partitions` and `total_partitions`, enabling proper watermark coordination.
* `SourceIdleDetector` handles both heartbeat publishing (via periodic WMB updates) and idle detection.

## Testing
* Unit tests are updated to progress watermark using single bucket.
* Performance testing is done and we don't see any regression, we see slight improvements in all the parameters.

Screenshot 2026-03-12 at 10 32 04 PM
Screenshot 2026-03-12 at 10 31 55 PM
Screenshot 2026-03-12 at 10 32 14 PM

### DIFF
--- a/rust/numaflow-core/src/watermark/source.rs
+++ b/rust/numaflow-core/src/watermark/source.rs
@@ -54,8 +54,6 @@ struct SourceWatermarkState {
     /// Cache of partitions that have been active during the current interval.
     /// Used for ISB idle watermark publishing.
     active_input_partitions: HashMap<u16, bool>,
-    /// Total partitions in the source.
-    total_partitions: Option<u32>,
 }
 
 impl SourceWatermarkState {
@@ -72,13 +70,7 @@ impl SourceWatermarkState {
             isb_idle_manager,
             source_idle_manager,
             active_input_partitions: HashMap::new(),
-            total_partitions: None,
         }
-    }
-
-    /// Sets the initialized partitions for source idle watermark publishing.
-    fn set_total_partitions(&mut self, partitions: u32) {
-        self.total_partitions = Some(partitions);
     }
 
     /// Handles generating and publishing source watermark with computation
@@ -406,7 +398,6 @@ impl SourceWatermarkHandle {
     ) {
         let mut state = self.state.lock().await;
         if let Some(p) = total_partitions {
-            state.set_total_partitions(p);
             state.publisher.set_processor_count(p);
         }
 
--- a/rust/numaflow-core/src/watermark/source/source_wm_publisher.rs
+++ b/rust/numaflow-core/src/watermark/source/source_wm_publisher.rs
@@ -66,7 +66,7 @@ impl SourceWatermarkPublisher {
     }
 
     /// Helper to create OT KV stores from bucket configs using the JetStream context.
-    /// Heartbeat is now embedded in WMB, so no separate hb_stores are needed.
+    /// Heartbeat is now embedded in WMB, so no separate store needed to track heartbeats.
     async fn create_ot_stores(
         js_context: &async_nats::jetstream::Context,
         bucket_configs: &[BucketConfig],


rust/numaflow-core/src/watermark/wmb.rs
```diff
@@ -5,13 +5,19 @@ use prost::Message;
...
 #[allow(clippy::upper_case_acronyms)]
 pub(crate) struct WMB {
     pub(crate) idle: bool,
     pub(crate) offset: i64,
     pub(crate) watermark: i64,
     pub(crate) partition: u16,
+    /// Optional expected processor count for source watermarks.
+    /// When set, the fetcher will wait until this many processors are active before
+    /// computing a valid watermark.
+    pub(crate) processor_count: Option<u32>,
```
vaibhavtiwari33, line 20: Why do we have the `processor_count` in the WMB? It seems to be only getting used in source_wm_fetcher, so maybe we can directly initialize the processor_count in the source_wm_fetcher?
Raising this concern since processor_count is only relevant for source.

rust/numaflow-core/src/watermark/source.rs
```diff
@@ -49,8 +48,14 @@ struct SourceWatermarkState {
...
     isb_idle_manager: ISBIdleDetector,
-    source_idle_manager: Option<SourceIdleDetector>,
+    /// Source idle detector is always present (with default 100ms heartbeat interval).
+    /// It handles both heartbeat publishing and idle detection (if user configured).
+    source_idle_manager: SourceIdleDetector,
+    /// Cache of partitions that have been active during the current interval.
+    /// Used for ISB idle watermark publishing.
     active_input_partitions: HashMap<u16, bool>,
+    /// Total partitions in the source.
+    total_partitions: Option<u32>,
```
vaibhavtiwari33: nit: `total_partitions` information from the `SourceWatermarkState` doesn't seem to be getting used anywhere. I see the `processor_count` in the `publisher` being used to populate the WMB instead.

rust/numaflow-core/src/watermark/source/source_wm_fetcher.rs
```diff
@@ -29,29 +29,66 @@ impl SourceWatermarkFetcher {
...
+                }
 
-            if head_wm < min_wm {
-                min_wm = head_wm;
+                // Get the processor_count from the head WMB and track the maximum
+                // This is the source of truth for expected processor count
+                if let Some(head_wmb) = timeline.get_head_wmb()
+                    && let Some(count) = head_wmb.processor_count
+                {
+                    max_expected_count = Some(max_expected_count.map_or(count, |c| c.max(count)));
```
vaibhavtiwari33, line 72: I think we already discussed it earlier but can this not create stalling behavior when pipeline restarts and the source has changed partitions to a lower value?

rust/numaflow-core/src/watermark/source/source_wm_publisher.rs
```diff
@@ -33,26 +43,35 @@ impl SourceWatermarkPublisher {
...
-    /// Returns (ot_stores, hb_stores) tuple.
-    async fn create_kv_stores(
+    /// Sets the total processor count (number of source partitions).
+    /// This count is included in every published WMB.
+    pub(crate) fn set_processor_count(&mut self, count: u32) {
+        self.processor_count = Some(count);
+    }
+
+    /// Helper to create OT KV stores from bucket configs using the JetStream context.
+    /// Heartbeat is now embedded in WMB, so no separate hb_stores are needed.
```
vaibhavtiwari33: nit: Since we're removing `hb_stores`, the reference might be confusing for new eyes

```suggestion
    /// Heartbeat is now embedded in WMB, so no separate store needed to track heartbeats.
```

rust/numaflow-core/src/watermark/source/source_wm_publisher.rs
```diff
@@ -18,11 +21,18 @@ use numaflow_shared::kv::jetstream::JetstreamKVStore;
...
     to_vertex_configs: Vec<BucketConfig>,
     publishers: HashMap<String, ISBWatermarkPublisher>,
+    /// Total number of source partitions (processors) for this source.
+    /// This is included in every published WMB to help the fetcher determine
+    /// when all processors have reported.
+    processor_count: Option<u32>,
+    /// Pre-created OT stores for source watermark publishers
+    source_ot_stores: HashMap<&'static str, Arc<dyn KVStore>>,
+    /// Pre-created OT stores for ISB watermark publishers
+    isb_ot_stores: HashMap<&'static str, Arc<dyn KVStore>>,
```
vaibhavtiwari33, line 35: This makes the code cleaner. Ty! 🚀

rust/numaflow-core/src/watermark/source/source_wm_fetcher.rs
vaibhavtiwari33: I guess it would be fine since the idle watermark that we publish should carry the new total partition information

rust/numaflow-core/src/watermark/wmb.rs
yhl25: `processor_count` needs to be propagated as part of wmb because we don't have separate wmbs for source.

PR #3290, opened 2026-03-07
fix(test): update tokio runtime flavor used in the spawned OS thread for component servers

vaibhavtiwari33:
* Clean up code comments around why the deadlock is happening in single threaded test environment.
* Change the flavor of tokio runtime in the spawned os thread for the server components to current-thread instead of multi-thread. 
* Add a test which uses the in-built http source to prevent a ud source server from spawning, and which tests that a panic in a UDSink server doesn't cause the single threaded test to hang because of the map udf. 
* Remove explicit shutdowns of the different component servers at the end of `start_forwarder_test`

### DIFF
--- a/rust/extns/numaflow-http/src/lib.rs
+++ b/rust/extns/numaflow-http/src/lib.rs
@@ -288,7 +288,18 @@ impl HttpSourceActor {
     }
 
     async fn pending(&self) -> Option<usize> {
-        Some(self.server_rx.len())
+        if !self.server_rx.is_empty() {
+            info!(
+                "Pending messages in HTTP Internal Channel: {}",
+                self.server_rx.len()
+            );
+        }
+
+        // we have to always return None because, if we return Some, the autoscaler will scale up thinking
+        // there are pending messages and will start autoscaling.
+        // Pending should only be supported for sources which are not bounded like kafka, jetstream
+        // pulsar, sqs, etc.
+        None
     }
 
     async fn read(&mut self, count: usize) -> Option<Result<Vec<HttpMessage>>> {
@@ -1122,7 +1133,10 @@ mod tests {
             .unwrap();
 
         let pending_count = pending_rx.await.unwrap();
-        assert_eq!(pending_count, Some(0)); // Should be 0 since we read all messages
+        assert_eq!(
+            pending_count, None,
+            "HTTP source pending should always be None"
+        );
 
         // Note: We already tested ack above when we acked the messages to complete the HTTP requests
 
@@ -1229,7 +1243,7 @@ mod tests {
 
         // Test pending count
         let pending = handle.pending().await;
-        assert_eq!(pending, Some(5), "Should have 5 pending messages");
+        assert_eq!(pending, None, "Should have 5 pending messages");
 
         // Test read method
         let messages = handle.read(3).await.unwrap().unwrap();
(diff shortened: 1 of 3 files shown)


rust/numaflow-core/src/monovertex/forwarder.rs
```diff
@@ -1442,9 +1500,9 @@ mod tests {
 
         let SinkTestHandle {
             sink_writer,
-            ud_sink_server_handle,
-            fb_ud_sink_server_handle,
-            ons_ud_sink_server_handle,
+            ud_sink_server_handle: _ud_sink_server_handle,
+            fb_ud_sink_server_handle: _fb_ud_sink_server_handle,
+            ons_ud_sink_server_handle: _ons_ud_sink_server_handle,
```
vigith, line 1505: why do we need `_` prefix for these? use the short notation?

rust/numaflow-core/src/source/test_utils.rs
```diff
@@ -134,11 +135,79 @@ impl SourceTestHandle {
...
             source_transformer_test_handle: transformer_test_handle,
         }
     }
+
+    /// Create a built-in HTTP source component.
+    ///
+    /// Binds to a random port, starts the embedded HTTP server, and spawns a background
+    /// task that sends `num_messages` POST requests to it. No separate gRPC server is
+    /// started — the HTTP server is embedded in the source itself.
+    pub(crate) async fn create_http_source(
```
vigith, line 149: why did we move away from ud-source?

vaibhavtiwari33: Currently we're seeing the deadlock behavior for the following combo:
```
udsource (on OS thread)
panicking udf (on OS thread)
built-in sink
```

Using this helper method I added a test which allowed me to ensure that the same deadlock behavior is not happening for the following combo:
```
built-in source (http source above)
udf (on OS thread)
panicking udsink (on OS thread)
```

i.e., the deadlock is avoided in the latter case due to structured concurrency in mapper.

rust/numaflow-core/src/monovertex/forwarder.rs
vaibhavtiwari33: I've removed the explicit shutdown calls to the component servers and relying completely on drop for these handles to trigger shutdown. Added the `_` to circumvent clippy to allow holding these handles without utilizing them

rust/numaflow-core/src/source/test_utils.rs
vigith: with `tokio::multi_thread`, we were fine, right? why do we need to do this if we are planning to fix the structured concurrency problem later on?

vigith: @yhl25 WDYT?

vaibhavtiwari33: The test that I added `test_map_sink_panic` works with the `current_thread` scheduler (`#[tokio::test]`).
We essentially don't need to do this, but I just wanted to ensure that structured concurrency will solve our problem by showcasing it for mapper.

vaibhavtiwari33: Also, in regards to your original comment:
> why did we move away from ud-source?

We still have a helper for creating ud-source (`pub(crate) async fn create_ud_source<S, T>`), just that this helper method (`create_http_source`) has been added alongside it.

PR #3300, opened 2026-03-10
test: E2E test for ordered processing

### LINKED ISSUE #3298: E2E Tests for Ordered Processing
```yaml
kind: Pipeline
metadata:
  name: ordered-pipeline
spec:
  limits:
    readBatchSize: 1
  ordered:
    enabled: true
  vertices:
    - name: in-1
      source:
        http: {}
    - name: in-2
      source:
        http: {}
    - name: cat
      partitions: 3
      udf:
        container:
          image: quay.io/numaio/numaflow-rs/map-cat:stable
          imagePullPolicy: IfNotPresent
    - name: out
      partitions: 3
      udsink:
        container:
          # A redis sink for e2e testing, see https://github.com/numaproj/numaflow-go/tree/main/examples/sinker/redis_sink
          image: quay.io/numaio/numaflow-rs/redis-sink:stable
          imagePullPolicy: IfNotPresent
          env:
            - name: ORDERED_OUTPUT
              # this makes sure that we use LIST instead of HSET
              value: "true"
            - name: SINK_KEY
              # The key is set in the format of "pipeline_name-vertex_name"
              value: "ordered-processing-sink-output"
  edges:
    - from: in-1
      to: cat
    - from: in-2
      to: cat
    - from: cat
      to: out
```

The way we will test is that, we will pump in ordered data and will make sure it is ordered by the key. E.g., we will have 3 payloads `["create", "update", "delete"]` with keys `["A", "M", "Z"]`, we will make sure there is ordering within each key.

vaibhavtiwari33:
E2E test for ordered processing

### Related issues
Fixes #3298

### Testing
The new E2E test passes

### DIFF
--- a/test/e2e/functional_test.go
+++ b/test/e2e/functional_test.go
@@ -410,6 +410,44 @@ func (s *FunctionalSuite) TestOrderedProcessing() {
 	}
 }
 
+func (s *FunctionalSuite) TestOrderedProcessingMultiSource() {
+	w := s.Given().Pipeline("@testdata/ordered-processing-multi-source.yaml").
+		When().
+		CreatePipelineAndWait()
+	defer w.DeletePipelineAndWait()
+	pipelineName := "ordered-multi-source"
+
+	w.Expect().VertexPodsRunning()
+
+	// 10 numeric keys, 100 messages per key = 1000 total messages.
+	// Messages are sent round-robin across two HTTP sources (in-1, in-2).
+	numKeys := 10
+	msgsPerKey := 100
+	sources := []string{"in-1", "in-2"}
+
+	for seq := 0; seq < msgsPerKey; seq++ {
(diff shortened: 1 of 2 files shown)


test/e2e/functional_test.go
```diff
@@ -380,6 +378,38 @@ func (s *FunctionalSuite) TestPipelineUserMetadataPropagation() {
...
+	pipelineName := "ordered-processing"
+
+	// wait for all the pods to come up
+	w.Expect().VertexPodsRunning()
+
+	// Send messages with different keys, each key getting values in a specific order.
+	// The x-numaflow-keys header sets the message key for partition routing.
+	keys := []string{"A", "M", "Z"}
+	values := []string{"create", "update", "delete"}
+
```
yhl25, line 395: Can you send atleast 1000 messages and use two sources?

PR #3315, opened 2026-03-16
opex: emit SQS metrics

ayildirim21:

### Related issues
Fixes #

### Testing

Currently only adding producer side metrics as that is what is required for golden signals:

- publish latency (histogram)
- publish success (counter)
- publish failure (counter

### DIFF
--- a/rust/numaflow-core/src/metrics/mod.rs
+++ b/rust/numaflow-core/src/metrics/mod.rs
@@ -33,6 +33,9 @@ use crate::sinker::sink::SinkWriter;
 use crate::source::Source;
 use crate::watermark::WatermarkHandle;
 
+pub(crate) mod sqs;
+pub(crate) use sqs::sqs_metrics;
+
 // SDK information
 const SDK_INFO: &str = "sdk_info";
 const COMPONENT: &str = "component";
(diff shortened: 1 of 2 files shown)


rust/extns/numaflow-sqs/src/sink.rs
```diff
@@ -98,6 +98,7 @@ impl TryFrom<BatchEntryInput> for SendMessageBatchRequestEntry {
 pub struct SqsSink {
     client: Client,
     queue_url: &'static str,
+    pub queue_name: &'static str,
```
yhl25, line 102: Do we need the queue name to be pub?

ayildirim21: Yes since we need it for the labels in numaflow-core

vigith: please document the exported fields.

rust/numaflow-core/src/metrics/mod.rs
vigith, line 1: Can we move SQS metrics under `metrics` mod? This file will grow unbounded otherwise.

PR #3317, opened 2026-03-17
feat: introduce concurrency env for udf invocation

yhl25:
### Summary                                                                                                                                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                                                            
  - Adds a first-class Concurrency field to `MonoVertexLimits`, `PipelineLimits`, and `VertexLimits`, replacing the undocumented `NUMAFLOW_UDF_CONCURRENCY` and `MAX_ACK_PENDING` env vars.                                                                                                                                                                                                                                                           
  - Splits the two knobs cleanly: readBatchSize controls only the size of one read; concurrency caps how many messages may be in-flight (read but not yet acked).
  - Promotes `READ_AHEAD` to a controller-injected env var with sensible defaults — false for source vertices and MonoVertex (cheap re-reads, source ordering preserved), true for Map/Sink/Reduce (keeps ISBs full). Operators can override on the container template.                                                                                                                                                                       
  - Wires both knobs through the Rust dataplane (monovertex + pipeline forwarders, ISB reader, source streaming loop).                                                                                                                                                                                                                                                                                                                      
  - Ordered processing now auto-forces concurrency = 1 on Map/Sink vertices in the controller, so users no longer need to remember to set readBatchSize: 1 to get FIFO semantics.                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                                                                                                                                            


### Behavior                                                                                                                                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                                                                                            
  With read-ahead enabled, the maximum in-flight count per vertex is concurrency + readBatchSize (the data plane keeps reading until in-flight hits concurrency, then may pre-fetch one more batch). With read-ahead disabled, it's min(`concurrency`, `readBatchSize`).                                                                                                                                                                        
   
  To force strictly sequential processing, set concurrency: 1 (and disable `READ_AHEAD` for non-source vertices, or rely on the source-vertex default).


rust/numaflow-core/src/source.rs
```diff
@@ -463,14 +467,20 @@ impl<C: crate::typ::NumaflowTypeConfig> Source<C> {
...
-            // is enabled you can have upto (max_ack_pending / read_batch_size) ack tasks. We
-            // divide by read_batch_size because we do batch acking in source.
+            // The semaphore caps the number of in-flight ack tasks. With read-ahead disabled, we
+            // allow exactly one batch in flight (sequential processing). With read-ahead enabled,
+            // we allow up to `concurrency` in-flight messages, divided by `read_batch_size` because
+            // we do batch acking. We always allow at least one task so a `concurrency` smaller than
+            // `read_batch_size` still makes progress.
             let max_ack_tasks = match &self.read_ahead {
-                true => MAX_ACK_PENDING / self.read_batch_size,
+                true => std::cmp::max(1, self.concurrency / self.read_batch_size.max(1)),
```
vaibhavtiwari33, line 503: This drop is significant. We're effectively dropping max ack pending from 10000 -> 500. It should be clearly noted in release notes.

rust/numaflow-core/src/config/pipeline.rs
```diff
@@ -563,7 +571,10 @@ impl PipelineConfig {
                 name: Box::leak(edge.from.clone().into_boxed_str()),
                 reader_config: BufferReaderConfig {
                     streams,
-                    max_ack_pending,
+                    // Cap inflight messages on this ISB reader to the vertex's `concurrency`. The
+                    // reader holds a semaphore of this size so that, even with read-ahead, we never
+                    // have more than `concurrency + read_batch_size` messages in flight.
+                    max_ack_pending: concurrency,
```
vaibhavtiwari33, line 577: We're dropping the default max ack pending from 25k -> 500 for ISB reader. This should also be noted as part of release notes since the change is quite significant.

pkg/apis/numaflow/v1alpha1/vertex_types.go
```diff
@@ -509,6 +522,16 @@ func (v VertexLimits) GetReadBatchSize() uint64 {
...
+// GetConcurrency returns the maximum number of in-flight (read-but-not-acked) messages allowed
+// at any time. It defaults to the read batch size when unset, which preserves the historical
+// behavior where concurrency was implicitly bounded by the batch size.
+func (v VertexLimits) GetConcurrency() uint64 {
+	if v.Concurrency != nil {
+		return *v.Concurrency
+	}
+	return v.GetReadBatchSize()
+}
+
```
vaibhavtiwari33: nit: Is this getting used anywhere?

pkg/reconciler/pipeline/controller.go
```diff
@@ -659,6 +659,18 @@ func buildVertices(pl *dfv1.Pipeline) map[string]dfv1.Vertex {
...
+		// by other means (single-stream-per-replica routing for sources, partition-by-key for
+		// reduce), so we leave their concurrency alone. We override even if the user set
+		// concurrency explicitly: ordered processing without concurrency=1 would silently violate
+		// the FIFO contract.
+		if vCopy.IsOrdered() && (vCopy.IsMapUDF() || vCopy.IsASink()) {
+			if vCopy.Limits == nil {
+				vCopy.Limits = &dfv1.VertexLimits{}
+			}
+			vCopy.Limits.Concurrency = ptr.To[uint64](1)
+		}
```
vaibhavtiwari33, line 677: Should we throw an error when concurrency > 1 for ordered processing instead of overwriting?

docs/user-guide/reference/mvtx-tuning.md
```diff
@@ -19,4 +20,37 @@ spec:
...
+  limits:
+    readBatchSize: 100
+    concurrency: 500
+  containerTemplate:
+    env:
+      - name: READ_AHEAD
+        value: "true"
+```
+
+With read-ahead enabled the upper bound on in-flight messages becomes **`concurrency + readBatchSize`** (the data plane keeps reading until the in-flight count hits `concurrency`, then may pre-fetch one more batch).
```
vaibhavtiwari33: I don't think this is correct. On source with read ahead enabled we cap at min(concurrency, readBatchSize) in this implementation

pkg/apis/numaflow/v1alpha1/vertex_types.go
vaibhavtiwari33, line 534: nit: is this getting used anywhere?

vaibhavtiwari33 (approved):

docs/user-guide/reference/mvtx-tuning.md
vaibhavtiwari33: Updated

pkg/apis/numaflow/v1alpha1/vertex_types.go
yhl25: Not used, but still want to keep it because of the default value.

vigith (approved):

PR #3342, opened 2026-03-31
refactor: Nack Messages by default to avoid acking failed messages

yhl25:
Previously, there was a potential for messages to be accidentally acked if a panic or error occurred mid-processing. This could happen because the default behavior was to ack messages when dropped, so any message that didn't get explicitly marked as failed could be silently acked and lost.

This PR introduces a `MessageHandle` type that wraps messages and tracks acknowledgment state throughout the pipeline. The default behavior is now to nack, and a message is only acked after it has been fully processed and all downstream messages have been successfully written. Explicit `mark_success()` and `mark_failed()` calls replace the previous manual isFailed flag manipulation, making the ack/nack logic clearer and safer.

This ensures that in panic or error scenarios, messages are correctly nacked and retried rather than being silently lost.


rust/numaflow-core/src/message.rs
```diff
@@ -53,41 +90,117 @@ pub(crate) struct Message {
...
+pub(crate) struct MessageHandle {
+    pub(crate) message: Message,
+    pub(crate) ack_handle: Arc<AckHandle>,
+}
+
+impl Clone for MessageHandle {
+    fn clone(&self) -> Self {
+        self.ack_handle
+            .ref_count
+            .fetch_add(1, std::sync::atomic::Ordering::SeqCst);
```
vaibhavtiwari33: Why do we need sequential consistency for this? Cloning would get expensive. Especially since this is for each message.

rust/numaflow-core/src/message.rs
```diff
@@ -53,41 +90,117 @@ pub(crate) struct Message {
...
     }
 }
 
 impl Drop for AckHandle {
     fn drop(&mut self) {
         if let Some(ack_handle) = self.ack_handle.take() {
-            if self.is_failed.load(std::sync::atomic::Ordering::Relaxed) {
-                ack_handle.send(ReadAck::Nak).expect("Failed to send nak");
+            // NAK if ref_count is not 0 (meaning not all references were marked as success)
+            if self.ref_count.load(std::sync::atomic::Ordering::SeqCst) != 0 {
```
vaibhavtiwari33: Similarly, here, IMO we don't need sequential consistency here. Wrapping AckHandle in Arc guarantees that at the time of drop, there wouldn't exist another thread holding this AckHandle for it to increase the value of the ref_count.

rust/numaflow-core/src/message.rs
```diff
@@ -53,41 +90,117 @@ pub(crate) struct Message {
...
+        }
+    }
+
+    /// Mark the message as successfully processed.
+    /// This decrements the reference count. When all references are marked as success
+    /// (ref_count reaches 0), the message will be ACK'd when the AckHandle is dropped.
+    pub(crate) fn mark_success(&self) {
+        self.ack_handle
+            .ref_count
+            .fetch_sub(1, std::sync::atomic::Ordering::SeqCst);
```
vaibhavtiwari33: Ok, I see why we're probably using `SeqCst`, to try to avoid scenarios where `ref_count` becomes negative. But I think, at the end of the day, we still can get away with `Relaxed` ordering since `AckHandle` is wrapped in `Arc` which already uses `Release`/`Acquire`, so the final count during `Drop` will be correct. 

Also, similar to how there can be bugs because of missing out on marking a message as success, we can also create bugs by marking a message as success too many times (ref_count < 0). Should we limit the ref_count to only be decreased till 0?

rust/numaflow-core/src/source.rs
```diff
@@ -528,14 +527,14 @@ impl<C: crate::typ::NumaflowTypeConfig> Source<C> {
                     );
 
                     let (resp_ack_tx, resp_ack_rx) = oneshot::channel();
-                    message.ack_handle = Some(Arc::new(AckHandle::new(resp_ack_tx)));
+                    let ack_handle = Arc::new(AckHandle::new(resp_ack_tx));
 
                     // insert the offset and the ack one shot in the tracker.
```
vaibhavtiwari33:
```suggestion
                    // insert the message (with offset) into the tracker
```

rust/numaflow-core/src/message.rs
```diff
@@ -53,41 +90,117 @@ pub(crate) struct Message {
...
     fn drop(&mut self) {
         if let Some(ack_handle) = self.ack_handle.take() {
-            if self.is_failed.load(std::sync::atomic::Ordering::Relaxed) {
-                ack_handle.send(ReadAck::Nak).expect("Failed to send nak");
+            // NAK if ref_count is not 0 (meaning not all references were marked as success)
+            if self.ref_count.load(std::sync::atomic::Ordering::SeqCst) != 0 {
+                let _ = ack_handle.send(ReadAck::Nak);
             } else {
-                ack_handle.send(ReadAck::Ack).expect("Failed to send ack");
+                let _ = ack_handle.send(ReadAck::Ack);
```
vaibhavtiwari33: I agree that we shouldn't panic if sending the ack/nack fails, but since the tokio tasks spawned for listening to these ack/nacks aren't structured, we should have a warn log here to notify that the receiver task exit before receiving ack/nack.

rust/numaflow-core/src/pipeline/isb/writer.rs
```diff
@@ -39,6 +40,85 @@ struct PendingWriteResult {
...
 }
 
+/// ISBWriteTask encapsulates all the context needed to execute a write operation per message.
+/// Similar to MapUnaryTask, it spawns as a tokio task and calls mark_success() at the end.
+struct ISBWriteTask<C: NumaflowTypeConfig> {
+    orchestrator: ISBWriterOrchestrator<C>,
+    /// Permit to achieve structured concurrency by ensuring we do not exceed the concurrency limit
+    /// and all the tasks are cleaned up when the component is shutting down.
+    permit: OwnedSemaphorePermit,
+    read_message: MessageHandle,
```
vaibhavtiwari33: nit: 
`msg_handle: MessageHandle`

rust/numaflow-core/src/mapper/map/unary.rs
```diff
@@ -45,7 +45,7 @@ pub(in crate::mapper) struct MapUnaryTask {
     /// Permit to achieve structured concurrency by ensuring we do not exceed the concurrency limit
     /// and all the tasks are cleaned up when the component is shutting down.
     pub permit: OwnedSemaphorePermit,
-    pub message: Message,
+    pub read_message: MessageHandle,
```
vaibhavtiwari33: nit:
`msg_handle: MessageHandle`

rust/numaflow-core/src/mapper/map/stream.rs
```diff
@@ -106,53 +106,52 @@ impl MapStreamTask {
                             .await
                             .expect("failed to update tracker");
 
-                        let bypassed =
+                        // Downstream messages are independent - they use a no-op AckHandle.
+                        // The original read_message is ACK'd via mark_success! below.
+                        let read_msg: MessageHandle = mapped_message.into();
```
vaibhavtiwari33: Why a no-op AckHandle? Shouldn't the final ack/nack for the message only be triggered when downstream (isb writer/sink) finishes processing?

rust/numaflow-core/src/mapper/map/batch.rs
```diff
@@ -95,30 +104,42 @@ impl MapBatchTask {
                         )
                         .await?;
 
+                    // Downstream messages are independent - they use a no-op AckHandle.
+                    // The original read_batch is ACK'd via mark_success_batch! below.
                     for mapped_message in mapped_messages {
-                        let bypassed = if let Some(ref bypass_router) = self.bypass_router {
-                            bypass_router
-                                .try_bypass(mapped_message.clone())
+                        let read_msg: MessageHandle = mapped_message.into();
```
vaibhavtiwari33: Same here, why independent ackHandles?

rust/numaflow-core/src/watermark/source.rs
```diff
@@ -42,6 +42,26 @@ pub(crate) mod source_wm_fetcher;
...
+        let partition_id = match &msg.offset {
+            Offset::Int(o) => o.partition_idx,
+            Offset::String(o) => o.partition_idx,
+        };
+        Self {
+            partition_id,
+            event_time_ms: msg.event_time.timestamp_millis(),
+        }
+    }
+}
```
vaibhavtiwari33, line 63: I don't think we need to introduce this tbh. We're abstracting away the partition idx extraction logic which is only being used once.

Let's remove this.

rust/numaflow-core/src/message.rs
```diff
@@ -53,41 +120,146 @@ pub(crate) struct Message {
...
+        self.ack_handle
+            .ref_count
+            .fetch_sub(1, std::sync::atomic::Ordering::Relaxed);
+    }
+
+    /// Mark the message as failed (consumes the handle), recording the reason it will be nacked.
+    /// ref_count is not decremented, so the message will be NAK'd when the AckHandle is dropped.
+    /// The error is logged at NAK time.
+    pub(crate) fn mark_failed(self, reason: impl fmt::Display) {
+        *self.ack_handle.failure_reason.lock().unwrap() = Some(reason.to_string());
```
vaibhavtiwari33, line 219: Failure reasons would be overwritten everytime we call mark_failed for the same ack handle. 

Also, can we remove unwrap here?

rust/numaflow-core/src/message.rs
```diff
@@ -53,41 +120,146 @@ pub(crate) struct Message {
...
-    pub(crate) ack_handle: Option<oneshot::Sender<ReadAck>>,
-    pub(crate) is_failed: AtomicBool,
+struct AckHandle {
+    sender: Option<oneshot::Sender<ReadAck>>,
+    /// Reference count to track active references. Starts at 1 when created.
+    /// Incremented when cloned (via MessageHandle::clone or with_message), decremented when
+    /// mark_success is called. On drop: NAK if ref_count != 0, ACK if ref_count == 0.
+    ref_count: AtomicUsize,
+    /// Set by mark_failed to record why the message is being nacked.
+    failure_reason: Mutex<Option<String>>,
```
vaibhavtiwari33, line 138: I'm not sure about the `Mutex` here.
Our final goal is to only capture the error message, it should not require locking overhead. 

I see that we're overwriting the error message every time anyways. Let's use `OnceLock` to only capture the first `mark_failure` message. This will avoid `Mutex` usage.

rust/numaflow-core/src/source.rs
```diff
@@ -519,23 +518,21 @@ impl<C: crate::typ::NumaflowTypeConfig> Source<C> {
 
                 let mut ack_handles = vec![];
```
vaibhavtiwari33, line 519: Can we change this to msg_handles instead, since that is what we're tracking in this vector. Will avoid confusions with actual `ack_handle`

rust/numaflow-core/src/monovertex/bypass_router.rs
```diff
@@ -278,28 +288,27 @@ impl BypassRouterReceiver {
                     let mut primary_messages: Vec<Message> = vec![];
                     let mut fallback_messages: Vec<Message> = vec![];
                     let mut on_success_messages: Vec<Message> = vec![];
-                    let mut ack_handles = vec![];
+                    let mut read_messages: Vec<MessageHandle> = vec![];
```
vaibhavtiwari33: nit: rename to `msg_handles`

rust/numaflow-core/src/mapper/map/batch.rs
```diff
@@ -42,8 +43,8 @@ pub(in crate::mapper) struct BatchSenderMapState {
 /// MapBatchTask encapsulates all the context needed to execute a batch map operation.
 pub(in crate::mapper) struct MapBatchTask {
     pub mapper: UserDefinedBatchMap,
-    pub batch: Vec<Message>,
-    pub output_tx: mpsc::Sender<Message>,
+    pub read_batch: Vec<MessageHandle>,
```
vaibhavtiwari33: nit: rename `read_batch`

rust/numaflow-core/src/watermark/source.rs
yhl25: Keeping `SourceWatermarkEntry` so the watermark code doesn't need to know about `MessageHandle`. Without it we'd pass `&[(u16, i64)]` which is unclear. `From` is the standard way to do conversions in Rust.

rust/numaflow-core/src/message.rs
yhl25: fixed.

vaibhavtiwari33 (approved):

rust/numaflow-core/src/watermark/source.rs
vaibhavtiwari33: I think without it, we were directly passing the `Message` around earlier, but fair that we want to limit the scope of info available for watermark code.

PR #3344, opened 2026-04-01
tests: Watermark tests

vaibhavtiwari33:
Additional tests for watermark modules. Primarily added to wm_fetcher and wm_publisher modules. 

### Scenarios covered

* Implemented existing nats-tests with PublishScenario, PublishStep and ExpectedWMB pattern
* **wm_fetcher module**
  * A Deleted processor whose head_offset < input offset should be removed from the processor map after fetch_watermark
  * A Deleted processor whose head_offset >= input offset should be retained.
  * InActive processors are still included in fetch_watermark's watermark computation
  * After fetching watermarks for different partitions, the returned watermark should be the minimum across ALL partitions' last_processed_wm, not just the current one.
  * With two edges, fetch_head_idle_wmb should return the minimum idle WMB across both edges, but only if ALL active processors in ALL edges are idle.
  * With two edges, if one edge has a non-idle processor, fetch_head_idle_wmb should return None.
  * Deleted and InActive processors should be excluded from fetch_head_idle_wmb
  * If a Deleted processor were the only non-idle one, it should be excluded, allowing the idle result to come through
  * When all active processors have head watermark = -1 (default/uninitialized), fetch_head_watermark should return -1
* **wm_publisher module**:
  * Basic publish scenarios:
    * Publish a single watermark to one edge, one partition.
    * Publish increasing watermarks, verify only the latest persists.
    * Publish to two partitions on the same edge. Each partition tracks state independently, but they share the same KV key (processor_name), so the last write wins in the store.
    * Publish to two separate edges, verify each edge's store has the correct WMB.
    * Verify the idle flag is correctly propagated to the WMB.
  * Monotonicity and regression scenarios:
    * Higher offset with lower watermark: the publisher should keep the higher watermark (regression correction). The published offset advances but watermark does not regress.
    * Lower offset with higher watermark: offset should NOT regress, but watermark updates.
    * Lower offset with lower watermark: nothing changes (both are below tracked state).
    * Same offset, higher watermark: watermark should update, offset stays.
    * Watermark of -1 (initial sentinel) should not cause regression detection.
  * Processor count scenarios
    * Verify processor_count is propagated to the WMB.
    * Verify processor_count=None results in None in the WMB.
  * Source publisher (is_source=true) behavior
    * Source publishers suppress regression warnings but still keep the higher watermark.
  * Multi-edge multi-partition stress scenario
    * Publish across 2 edges x 2 partitions, verify each edge has its correct final WMB.
    * Transition from idle to active: publish idle, then non-idle on the same partition.
  * KV store failure resilience: When the KV store fails, the publisher should not panic and should continue functioning on subsequent calls
  * Delay-based WMB publish throttling scenarios
    * With a long delay configured, publishes within the window are suppressed.
    * After the delay has elapsed, the publish should go through.
    * With no delay (None), every publish goes through.
    * Delay accumulates state: publishes within the window update tracked state, and when the delay elapses, the best values are published.
* **manager module**
  * Processor transitions from Active to InActive when it stops publishing WMBs.
  * Processor transitions from InActive to Deleted when it has been gone for a long time.
  * Processor re-activation from InActive to Active when a new WMB arrives.
  * Deleted processors are NOT re-activated by new WMBs.
  * KVWatchOp::Delete marks the specific processor as Deleted.
  * KVWatchOp::Purge marks ALL processors as Deleted.
* **timeline module**
  * Watermark regression: inserting a WMB with a watermark smaller than the head should be silently ignored. The timeline remains unchanged.
  * Same watermark, smaller offset: the new WMB should be skipped. The head remains the entry with the larger offset.
  * Greater watermark, equal offset: the head WMB's watermark is mutated in-place without adding a new entry to the deque.


rust/numaflow-core/src/watermark/processor/manager.rs
```diff
@@ -742,4 +802,309 @@ mod tests {
...
+                .expect("processor should exist");
+            processor.set_status(Status::Deleted);
+        }
+
+        // Publish a new WMB
+        store.put("processor1", wmb_bytes(200, 2, 0)).await.unwrap();
+
+        // Give the OT watcher time to process the WMB
+        tokio::time::sleep(Duration::from_millis(500)).await;
+
```
vigith, line 1002: let's try not sleep in test.

vigith:
> Please move the existing tests to new pattern.

what is the new pattern? please create and issue and link it?

rust/numaflow-core/src/watermark/isb/wm_publisher.rs
```diff
@@ -509,3 +510,1174 @@ mod tests {
...
+
+    /// Describes a complete publish-then-verify scenario.
+    struct PublisherScenario {
+        name: &'static str,
+        processor_name: &'static str,
+        bucket_configs: Vec<BucketConfig>,
+        is_source: bool,
+        steps: Vec<PublishStep>,
+        expected: Vec<ExpectedWMB>,
+    }
```
vaibhavtiwari33, line 556: @vigith 
Not essentially a new pattern, but just that we're defining here the scenario, steps and the expected value. 

I'll try to replicate the existing nats tests to see if any of them are transferable to this pattern

vaibhavtiwari33: I can do it as a separate PR.

PR #3350, opened 2026-04-03
fix: replace single sleep before fetching pending count with a timeout

vaibhavtiwari33:
From this: https://github.com/numaproj/numaflow/pull/3349

### DIFF
--- a/rust/extns/numaflow-nats/src/jetstream.rs
+++ b/rust/extns/numaflow-nats/src/jetstream.rs
@@ -705,7 +705,7 @@ mod tests {
         .await;
 
         assert!(
-            result.is_err(),
+            result.is_ok(),
             "{description}: expected Some({expected}) but got {:?} after {timeout:?}",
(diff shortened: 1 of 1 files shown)


rust/extns/numaflow-nats/src/jetstream.rs
```diff
@@ -684,6 +684,33 @@ mod tests {
...
+                tokio::time::sleep(interval).await;
+            }
+        })
+        .await;
+
+        assert!(
+            result.is_err(),
+            "{description}: expected Some({expected}) but got {:?} after {timeout:?}",
+            source.pending_messages().await.unwrap()
+        );
```
vigith, line 711: are you expecting this to be an error? if yes, fix the message.

vaibhavtiwari33: Fixed the assertion. It should've been the other way round. ty!

PR #3358, opened 2026-04-10
fix: start lastScaledAt as creationTime by default

### LINKED ISSUE #3357: fix: vertex stuck not reaching Running due to lastScaledAt null validation error after SSA migration
## Bug description

After PR #2570 migrated vertex status updates to Server-Side Apply (SSA), vertices can get permanently stuck in a reconcile error loop and never reach `Running` phase, producing no pods.

## Error

The controller logs the following error on every reconcile loop:

```
Vertex.numaflow.numaproj.io "<vertex-name>" is invalid: status.lastScaledAt: Invalid value: "null": lastScaledAt in body must be of type string: "null"
```

## Root cause

With SSA, the entire `VertexStatus` struct is serialized and sent as a JSON patch. A zero-value `metav1.Time` (i.e. `LastScaledAt` that has never been set) serializes to JSON `null`. The full CRD schema defines `lastScaledAt` as `type: string, format: date-time` without `nullable: true`, so the Kubernetes API server rejects the patch.

This creates a deadlock:
1. Vertex created → `LastScaledAt` is zero → SSA writes `null` → API rejects
2. Vertex never reaches `Running`
3. Autoscaler skips it: "Vertex not in Running phase, skip scaling"
4. `LastScaledAt` never gets set → back to step 1

The trigger condition is any vertex where `currentReplicas == desiredReplicas` on first reconcile (e.g. a source vertex with `min: 1, max: 1`, or any vertex during pipeline pause with `min: 0, max: 0`), because the scale branch that sets `LastScaledAt` is never taken.

## Affected versions

All versions since #2570 was merged (v1.7.x+). The minimal CRD install is not affected since `status` uses `x-kubernetes-preserve-unknown-fields: true`.

## Fix

Add `// +nullable` to `LastScaledAt` in `VertexStatus` and `MonoVertexStatus` so the generated CRD schema includes `nullable: true`, allowing the API server to accept `null` for this field.

Koalk:

When a vertex or monovertex is first reconciled and replicas have not changed
(i.e. `currentReplicas == desiredReplicas`), the `LastScaledAt` field in the
status is never set and stays at its Go zero value (`metav1.Time{}`).

Since #2570 migrated status updates to Server-Side Apply (SSA), the full status
struct is serialised and sent as a JSON patch. A zero `metav1.Time` serialises
as JSON `null`. The full CRD schema defines `lastScaledAt` as
`type: string, format: date-time` without `nullable: true`, so the Kubernetes
API server rejects the patch with:

```
Vertex.numaflow.numaproj.io "<name>" is invalid: status.lastScaledAt:
  Invalid value: "null": lastScaledAt in body must be of type string: "null"
```

This creates a deadlock on every reconcile loop:

1. Vertex created → `LastScaledAt` is zero → SSA writes `null` → API rejects
2. Vertex never reaches `Running` phase
3. Autoscaler skips it: _"Vertex not in Running phase, skip scaling"_
4. `LastScaledAt` never gets set → back to step 1

The trigger condition is any vertex where `currentReplicas == desiredReplicas`
on the first reconcile — for example a source vertex with `scale.disabled: true`,
or any vertex during a pipeline pause where `min: 0, max: 0` means
`currentReplicas` and `desiredReplicas` are both 0.

**Fix:** after the existing replica-change branch, initialise `LastScaledAt` to
the vertex's own `CreationTimestamp` if it is still zero. This ensures SSA never
writes `null` for this field regardless of scale configuration, while also being
semantically accurate (the vertex was "last scaled" when it was created).

### Related issues

Fixes #3357

### Testing

Reproduced locally against a Minikube cluster running Numaflow v1.7.3:

1. Deployed a pipeline with a source vertex configured with `scale.disabled: true`
   (i.e. `currentReplicas == desiredReplicas` from the start)
2. Confirmed the controller logged the `lastScaledAt: null` reconciler error in a
   tight loop and the vertex never reached `Running` phase
3. Applied the fix and redeployed — the error disappeared, the vertex reached
   `Running`, and `kubectl get vertex <name> -o jsonpath='{.status.lastScaledAt}'`
   returned the vertex's creation timestamp

- The root cause was introduced by #2570 (SSA migration). The minimal CRD
  install was not affected because its `status` uses
  `x-kubernetes-preserve-unknown-fields: true` and skips typed validation.
- Only `controller.go` for vertex and monovertex are changed — no CRD schema
  changes, no API changes.
- The same pattern (initialise to `CreationTimestamp` when zero) could apply to
  any other `metav1.Time` status fields added in future if they share the same
  SSA serialisation path.

### DIFF
--- a/pkg/reconciler/monovertex/controller.go
+++ b/pkg/reconciler/monovertex/controller.go
@@ -307,6 +307,9 @@ func (mr *monoVertexReconciler) orchestratePods(ctx context.Context, monoVtx *df
 		monoVtx.Status.Replicas = uint32(desiredReplicas)
 		monoVtx.Status.LastScaledAt = metav1.Time{Time: time.Now()}
 	}
+	if monoVtx.Status.LastScaledAt.IsZero() {
+		monoVtx.Status.LastScaledAt = monoVtx.CreationTimestamp
+	}
 	if monoVtx.Status.Selector == "" {
 		selector, _ := labels.Parse(dfv1.KeyComponent + "=" + dfv1.ComponentMonoVertex + "," + dfv1.KeyMonoVertexName + "=" + monoVtx.Name)
 		monoVtx.Status.Selector = selector.String()
--- a/pkg/reconciler/monovertex/controller_test.go
+++ b/pkg/reconciler/monovertex/controller_test.go
@@ -304,6 +304,34 @@ func Test_orchestratePods(t *testing.T) {
 			assert.True(t, strings.HasPrefix(n, testObj.Name+"-mv-0") || strings.HasPrefix(n, testObj.Name+"-mv-1"))
 		}
 	})
+
+	t.Run("test lastScaledAt initialised to CreationTimestamp when zero", func(t *testing.T) {
+		cl := fake.NewClientBuilder().Build()
+		r := fakeReconciler(t, cl)
+		testObj := testMonoVtx.DeepCopy()
+		creationTime := time.Now()
+		testObj.CreationTimestamp = metav1.Time{Time: creationTime}
+		err := r.orchestratePods(context.TODO(), testObj)
+		assert.NoError(t, err)
+		assert.Equal(t, creationTime.Unix(), testObj.Status.LastScaledAt.Unix())
+	})
+
+	t.Run("test lastScaledAt not overwritten when already set", func(t *testing.T) {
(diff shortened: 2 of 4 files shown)


Koalk:
I am not sure the test failure is related to this PR in specific

vigith (approved):

vaibhavtiwari33 (approved):

PR #3360, opened 2026-04-13
chore: remove phantom data markers

vaibhavtiwari33:

From the [nomicon explanation](https://doc.rust-lang.org/nomicon/phantom-data.html) it seems like the `PhantomData` is required to be declared in a struct's field, to help with static analysis, when the data it holds isn't declared in any other field. 

Since, the `ISBWriterOrchestrator<C>` already holds the generic type `C` that the PhantomData holds, i figure it should be safe to remove the phantom data marker.


yhl25 (approved):

vigith (approved):
It was introduced because we did not have a field that was associated with `NumaflowTypeConfig` Generics, so we needed a place-holder.

PR #3395, opened 2026-04-25
fix: introduce path field in messageID to track unique messages across fanout components

### LINKED ISSUE #3391: Source-transformer fan-out breaks UDF map stage leading to potential data loss
**Describe the bug**
In a MonoVertex with source → transformer → map → sink, if the source transformer fans out (emits N > 1 messages per input), the UDF map stage downstream mis-routes or drops responses because all N fanned-out children are sent to the mapper with the same MapRequest.id. The sink ends up receiving fewer messages than expected
  
**Reproduction**                                                                                                                                                                                                   
                                  

```
kind: MonoVertex
metadata:
  name: simple-mono-vertex
spec:
  source:
    udsource:
      container:
        image: quay.io/numaio/numaflow-go/source-simple-source:stable
        imagePullPolicy: IfNotPresent
    transformer:
      container:
        image: quay.io/numaio/numaflow-go/mapt-fan-out:fan_out
        imagePullPolicy: IfNotPresent
  udf:
    container:
      image: quay.io/numaio/numaflow-go/map-cat:stable
      imagePullPolicy: IfNotPresent
  sink:
    udsink:
      container:
        image: quay.io/numaio/numaflow-go/sink-log:stable
        imagePullPolicy: IfNotPresent
```                                                                                                                                                                                                                                                                                                                                                                      
  - Source: emits 100 messages
  - Source transformer: emits 3 messages per input (300 total)                                                                                                                                           
  - Map UDF:  cat                                             
  - Sink: log                                                                                                                                                                               
                                                                                                                                                                                                                 
 **Expected**: sink receives 300 messages; source acks all 100 offsets.                                                                                                                                               
  
**Results**:                                                                                                                                                                                                       
  - Sink received 100 messages (expected 300).
  - All 100 unique source offsets are represented exactly once — no offset got 2 or 3 children through.                                                                                                            
  - Which child won is arbitrary: 86× -1, 57× -3, 63× -2 


```
     User Defined Sink: 63-2
     User Defined Sink: 86-1
     User Defined Sink: 57-3
     User Defined Sink: 56-1
     User Defined Sink: 61-3
     User Defined Sink: 60-1
     User Defined Sink: 87-1
     User Defined Sink: 58-3
     User Defined Sink: 62-3
     User Defined Sink: 59-1
```


 **Affected versions**: current main . Applies to all three map modes: unary, stream, batch.

**Additional context**
```
  ERROR failed to map message  e: ActorPatternRecv("channel closed")                                                                                                                                             
        offset: b"MA==" (= "0")                                                                                                                                                                                    
        target: numaflow_core::mapper::map::unary   
```                                                                                                                                                               
                                                                                                                                                                                                                   
the first oneshot sender's tx was dropped when the second fan-out sibling's insert overwrote it. The error fires per source offset, twice (offsets 0 / MA==, 1 / MQ==, 2 / Mg==, ... each  
  appear 2× in the errors — the 2 losers out of 3 children). 

**despite 200 map errors, the pipeline did not halt or nack**


**Transform**
```
const fanOutSize = 3

// FanOut emits multiple output messages for each incoming message.
type FanOut struct{}
func (f *FanOut) Transform(_ context.Context, keys []string, d sourcetransformer.Datum) sourcetransformer.Messages {
	results := sourcetransformer.MessagesBuilder()
	eventTime := d.EventTime()
	payload := string(d.Value())

	for i := 1; i <= fanOutSize; i++ {
		output := fmt.Sprintf("%s-%d", payload, i)
		log.Printf("FanOut: emitting message %d/%d: %s", i, fanOutSize, output)
		results = results.Append(
			sourcetransformer.NewMessage([]byte(output), eventTime).WithKeys(keys),
		)
	}

	return results
}
```

vaibhavtiwari33:
### Issue 1
The issue reported in the linked bug report is caused due to using message offset as a key for `ResponseSenderMap` which holds the callback sender to send the results returned by the map udf server. 

This logic broke down when map is followed by a source transformer that implements flatmap, since, the offset for each message coming out of the transformer, that originated from a single message, will be the same. 

### Fix 1
Easy fix for it is to use the MessageID, which contains index info, instead of message offset. The flatmap transformer will increment the index in message ID for each message which originated from the same offset:

```rust
for (i, result) in resp.results.into_iter().enumerate() {
    let message = UserDefinedTransformerMessage(result, &msg_info, i as i32).into();
    response_messages.push(message);
}
```

### Issue 2
But this unearths another issue. Each map mode also essentially uses the same logic for flatmap. The index in messageID for each message is overwritten by map after the responses are returned from the udf. For example, in unary:

```rust
for (i, result) in results.into_iter().enumerate() {
    mapped_messages.push(UserDefinedMessage(result, &parent_info, i as i32).into());
}
```

since, we're overwritting the index part in the mapper, the message IDs for each message originating from the same offset are essentially being 'commonized', so to speak, i.e., losing their uniqueness. 

eg:

MessageID format: {vtx}-{offset}-{index}

For a given offset "X", and vertex "V", we have the following flow of a message ID through the mvtx:

```mermaid
flowchart LR
      Source["Source(`V-X-0`)"] --> Transformer
      Transformer -->|"`V-X-0`"| M1["MapTask(`V-X-0`)"]
      Transformer -->|"`V-X-1`"| M2["MapTask(`V-X-0`)"]
      Transformer -->|"`V-X-1`"| M3["MapTask(`V-X-0`)"]
      M1 --> Sink["Sink ('error due to non unique messages')"]
      M2 --> Sink
      M3 --> Sink
```

### Fix 2

Introduce path field in MessageID struct/proto to track the different indices at each component. 

## Related issues
Fixes #3391

## Testing
Added unit tests in monovertex forwarder to cover the following scenarios:
* Flatmap source transformer followed by a cat map udf
* Flatmap source transformer followed by a flatmap map udf


vaibhavtiwari33:
It was discussed that adding another field in messageID for the purpose of uniqueness across monovertex components is over-engineering. We'll store the sequence of indices from each component using offset field instead.

PR #3397, opened 2026-04-28
feat: add spans for tracing

### LINKED ISSUE #3379: OTel tracing core spans — platform.process, source.read, map, sink.write
Parent: #3377
Depends on: #3378

## What

Add the core platform spans for MonoVertex message processing: `numaflow.platform.process` (root), `numaflow.monovertex.source.read`, `numaflow.monovertex.map`, `numaflow.monovertex.sink.batch_wait`, and `numaflow.monovertex.sink.write`. Uses the dual-key `sys_metadata` pattern for UDF context propagation.

### 1. `rust/numaflow-core/src/message.rs` — AckHandle
- Add `pipeline_span: Mutex<Option<tracing::Span>>` field
- `set_pipeline_span()` method (interior mutability)
- Update all `AckHandle::new()` call sites (6+ files)

### 2. `rust/numaflow-core/src/source.rs` — platform.process + source.read
- Per-message: extract upstream context (W3C/B3 headers)
- Create `numaflow.platform.process` span (root, stored in AckHandle for accurate duration)
- Create `numaflow.monovertex.source.read` span (child, short-lived marker)
- Inject `platform.process` context into `sys_metadata["tracing"]`

### 3. `rust/numaflow-core/src/mapper/map/unary.rs` (+ batch.rs, stream.rs)
- Extract `platform.process` from `sys_metadata["tracing"]` → parent
- Create `numaflow.monovertex.map` span (sibling of source.read)
- Before UDF: inject map context into `sys_metadata["tracing_udf"]`
- After UDF: remove `tracing_udf` from results, leave `tracing` intact

### 4. `rust/numaflow-core/src/sinker/sink.rs`
- Per-message `numaflow.monovertex.sink.write` spans using OTel API (accurate batch duration)
- `numaflow.monovertex.sink.batch_wait` span (time spent waiting to be batched)
- Inject `sink.write` context into `sys_metadata["tracing_udf"]` for sink UDF

### Span Hierarchy
```
numaflow.platform.process                    (root, 1.5s — full lifecycle)
├── numaflow.monovertex.source.read          (0ms marker)
├── numaflow.monovertex.map                  (288ms — covers UDF gRPC call)
│   └── udf.map.process                      (10ms — UDF, created by SDK)
├── numaflow.monovertex.sink.batch_wait      (~1s — batching timeout)
└── numaflow.monovertex.sink.write           (273ms — covers batch UDF call)
    └── udf.sink.process                     (5ms — UDF, created by SDK)
```

## Testing
- `cargo check` / `cargo test` pass
- Deploy MonoVertex with Jaeger → verify 10+ spans per trace with correct hierarchy
- All spans have `numaflow.topology=monovertex` attribute

adarsh0728:

- Per-message spans for source, transformer, map, sink, fallback, and on-success stages
- No span/context work unless OTLP tracing is configured.
- Defaulting enabled tracing to parent-based 10% sampling
- The root span is named `numaflow.vertex.process` instead of `numaflow.platform.process` so the naming works for both MonoVertex and future Pipeline tracing.
- `numaflow.monovertex.source.dispatch` measures source-stage work after read, including tracker insert, transform, watermark, and downstream dispatch. Per-message spans can only be created after the batch read returns
- `numaflow.monovertex.source.transform` as a child span for source transformer UDF time
- `numaflow.monovertex.map` spans for map UDF processing.
- `numaflow.monovertex.sink.write`, `numaflow.monovertex.sink.fallback`, and `numaflow.monovertex.sink.on_success` spans for sink stages

```
numaflow.vertex.process
├── numaflow.monovertex.source.dispatch
│   └── numaflow.monovertex.source.transform
├── numaflow.monovertex.map
├── numaflow.monovertex.sink.write
└── numaflow.monovertex.sink.on_success
```

```
numaflow.vertex.process
├── numaflow.monovertex.source.dispatch
│   └── numaflow.monovertex.source.transform
├── numaflow.monovertex.map
├── numaflow.monovertex.sink.write
├── numaflow.monovertex.sink.fallback
```

### Follow Up Work

- Sink batching span 
- Pipeline spans (ISB etc)

### Configuration

Tracing is enabled only when an OTLP endpoint is configured:
```
env:
  - name: OTEL_EXPORTER_OTLP_ENDPOINT
    value: http://<collector>:4317
  - name: OTEL_SERVICE_NAME
    value: numaflow
```
Optional sampler override:

```
env:
  - name: OTEL_TRACES_SAMPLER
    value: parentbased_traceidratio
  - name: OTEL_TRACES_SAMPLER_ARG
    value: "0.1"
```

### Related issues
Fixes #3379 

### Testing

Screenshot 2026-04-28 at 12 28 17 PM
Screenshot 2026-04-28 at 12 28 32 PM

#### Verified that no major change(with this PR) in performance

comparison of [d383f63](https://github.com/numaproj/numaflow/pull/3397/commits/d383f63edd38457ad7d3ead24373249a1ff9142e) with [f520417](https://github.com/numaproj/numaflow/pull/3397/commits/f52041739f14f18b7a30cdab015e58039e4b6b99)

image

- Huge difference because
  - no sampling 
  - global::tracer("numaflow-core") was called per span --> this is now cached
  - Changed SDK span attributes to be set only when span.is_recording(), avoiding Vec"KeyValue" allocation for noop/dropped spans.
  - Added early returns in source/map/sink/transform tracing helpers so the no-exporter path skips:
    - parent context extraction
    - message id string allocation for span attributes
    - span creation
    - metadata injection
    - span guard storage/drop work

comparison of [d383f63](https://github.com/numaproj/numaflow/pull/3397/commits/d383f63edd38457ad7d3ead24373249a1ff9142e) with main

image

In MonoVertex, one `numaflow.vertex.process` span covers the full source → map → sink lifecycle for a message. In Pipeline mode, each vertex can later create its own `numaflow.vertex.process` span for the work owned by that vertex, such as source processing, map processing, or sink processing. The numaflow.topology, numaflow.pipeline.name, and numaflow.vertex.name attributes distinguish where the span came from.

This keeps the root span name topology-neutral while allowing operation names like monovertex.process today and pipeline.process in a future pipeline tracing PR.


rust/numaflow-core/src/mapper/map/batch.rs
```diff
@@ -150,6 +173,40 @@ impl MapBatchTask {
     }
 }
 
+fn inject_batch_map_spans_if_enabled(
+    msg_handles: &mut [MessageHandle],
+    tracing_enabled: bool,
+) -> Option<otel::ContextSpanGuard> {
+    if !tracing_enabled {
+        return None;
```
yhl25: Can we have the `From<>` implementations and move the otel related types to `otel.rs`?

yhl25 (commented):
Please make sure the span types are correct. The source span should have `consumer` span type.

adarsh0728:
> Please make sure the span types are correct. The source span should have `consumer` span type.

this is the current state

| Span name | Kind | Logic |
|---|---:|---|
| `numaflow.vertex.process` | `Internal` | Runtime lifecycle/root span for processing one message inside Numaflow. |
| `numaflow.monovertex.source.dispatch` | `Consumer` | consuming a message from the source side into the processing graph. |
| `numaflow.monovertex.source.transform` | `Client` | calls the source transformer UDF, so it represents an outbound call from Numaflow to user code. |
| `numaflow.monovertex.map` | `Client` | calls the map UDF, so it is modeled as a client call to user code. |
| `numaflow.monovertex.sink.write` | `Client` | calls the primary sink, an outbound write/call from Numaflow to sink code. |
| `numaflow.monovertex.sink.fallback` | `Client` | calls the fallback sink after primary failure, also an outbound sink call. |
| `numaflow.monovertex.sink.on_success` | `Client` | calls the on-success sink after primary success, also an outbound sink call. |

adarsh0728:
@yhl25 even if we keep sampling at 0 by default, spans are dropped only inside the SDK span creation path when the SDK calls `sampler.should_sample()`. By that point, we already do work such as context extraction, span builder setup, message id formatting, and metadata propagation checks.

So sampling=0 reduces recording/export cost, but not the platform span work.

yhl25:
> @yhl25 even if we keep sampling at 0 by default, spans are dropped only inside the SDK span creation path when the SDK calls `sampler.should_sample()`. By that point, we already do work such as context extraction, span builder setup, message id formatting, and metadata propagation checks.
> 
> So sampling=0 reduces recording/export cost, but not the platform span work.

Do you see any performance drop? because span creation itself is cheap

adarsh0728:
> Do you see any performance drop? because span creation itself is cheap

yes there is drop

vigith (approved):

PR #3400, opened 2026-04-29
fix: use min event_time across in-flight msgs before publishing idle src wm

### LINKED ISSUE #3399: Slow PAF resolutions at stalled source allows idle watermark publishing
## **Describe the bug**

* When the numaflow source vertex's read loop stalls, either, because in-flight PAFs from the JetStream writer are slow to resolve and back-pressure propagates back through the writer's permit pool; or for an external reason while the PAF resolutions are slow, the source partition's idle threshold may get crossed (In the validation pipeline we saw PAF resolutions taking > 5s)
* The source idle watermark publisher then begins incrementing the source-OT head wm by `increment_by` every `step_interval`, even though there are still in-flight messages whose event_times are below the head wm being published. 
* The same idle path also stamps those inflated values directly onto every downstream ISB-OT bucket via `publish_isb_idle_watermark`. 
* Downstream readers then read those WMBs, advance their `current_watermark` past valid in-flight messages.
* After PAFs finally resolve for these in-flight messages, before publishing them to ISB, we get the watermark from the source-OT (at the present moment, which has advanced) and stamp these messages with this advanced watermark.
* The reducer downstream, drops these messages with `"Old message popped up, Watermark has progressed past event time"`, as it compares the event times of these messages against the current watermark. 

### Flow connecting main components involved

```mermaid
flowchart LR
    subgraph "source pod"
        TICK[background task to<br/>move idle wm]
        S[Source.read]
        W[ISB writer]
        PSI[publish_source_isb_watermark<br/>writes WMB to ISB-OT]
        SP[publish_source_watermark<br/>writes to source-OT]
    end

    subgraph "JetStream"
        JSDATA[(ISB stream)]
        SOT[(source-OT KV)]
        IOT[(ISB-OT KV)]
    end

    subgraph "reduce pod"
        R[Reader]
        RD[Reducer]
    end

    S --> SP
    SP --> SOT
    S --> W
    W -->|PUB| JSDATA
    W -->|PAF resolves| PSI
    PSI -.->|fetch source-OT head| SOT
    PSI --> IOT
    R -.->|fetch ISB-OT WMB| IOT
    R -->|read message| JSDATA
    R --> RD
    TICK --> SP
    TICK --> PSI
```

### Logs showcasing the issue

We see the following log on reduce vertex:
```json
{                                                                                                                                                                                                                                     
    "current_watermark": 1451607012000,                                                           
    "is_late": false,                                                                                                                                                                                                                   
    "level": "ERROR",                                                                                                                                                                                                                   
    "message": "Old message popped up, Watermark has progressed past event time",                                                                                                                                                       
    "message_event_time": 1451607009432,                                                                                                                                                                                                
    "message_watermark": 1451607012000,                                                                                                                                                                                               
    "offset": {                                                                                                                                                                                                                         
      "Int": {                                                                                                                                                               
        "offset": 25377,
        "partition_idx": 4
      }                                                                                                                                                                                                                                 
    },
    "target": "numaflow_core::reduce::reducer::aligned::reducer",                                                                                                                                                                       
    "timestamp": "2026-04-28T10:37:39.351011Z"                                                                                                                               
  }
```

A log was added on source ISB wm publishing, to check whether the wm fetched from source-OT bucket has advanced past the message's event time, while the message wasn't marked as late at source.

([rust/numaflow-core/src/watermark/source.rs](https://github.com/numaproj/numaflow/blob/27e11afbe8e05c115fe14d1e4ff24575d352dc5f/rust/numaflow-core/src/watermark/source.rs#L141))
```rust
if message.event_time.timestamp_millis() < watermark.timestamp_millis() && !message.is_late
{
    warn!(
        offset = offset.offset,
        input_partition,
        msg_event_time = message.event_time.timestamp_millis(),
        src_watermark = watermark.timestamp_millis(),
        delta_ms = watermark.timestamp_millis() - message.event_time.timestamp_millis(),
        "Source stamping ISB-OT WMB with watermark ahead of message event_time"
    );
}
```

We see this to be the case in logs:

```json
{                                                                                                             
    "delta_ms": 2568,
    "input_partition": 0,                                                                                                                                                                                                               
    "level": "WARN",                                                                                                                                                                                                                    
    "message": "Source stamping ISB-OT WMB with watermark ahead of message event_time",                                                                                                                                                 
    "msg_event_time": 1451607009432,                                                                                                                                                                                                    
    "offset": 25377,                                                                                                                                          
    "src_watermark": 1451607012000,
    "target": "numaflow_core::watermark::source",
    "timestamp": "2026-04-28T10:37:38.595372Z"   
  }
```

We then tracked, which code path is publishing this advanced watermark: [rust/numaflow-core/src/watermark/isb/wm_publisher.rs](https://github.com/numaproj/numaflow/blob/56c4bd23f5d88dbc0a22debc6bf60b568c12cd61/rust/numaflow-core/src/watermark/isb/wm_publisher.rs#L213-L223)

and see the following log:
```json
{                                                                                                                                                                                                                                     
    "code_path": "tick->publish_source_idle_watermark->publish_source_watermark->",                                                                           
    "idle": true,                                                                  
    "init_offset": 1777372655579342,
    "init_watermark": 1451607012000,
    "level": "INFO",                
    "message": "publish_watermark_with_processor_count",                                                                                                                                                                                
    "partition": 0,                                     
    "publish_offset": 1777372655579342,                                                                                                                                                                                                 
    "publish_watermark": 1451607012000,                                                                                                                             
    "target": "numaflow_core::watermark::isb::wm_publisher",
    "timestamp": "2026-04-28T10:37:35.582184Z",                                                                                                                                                                                         
    "vertex": "source" 
  }   
```

Thus, idle source watermark publishing kicked in after idle threshold was breached. Incrementing the watermark beyond the event times of the current messages that are in-flight. 

### The stall, as a sequence diagram

```mermaid
sequenceDiagram
    autonumber
    participant Src as Source main loop
    participant W as ISB writer
    participant Sema as PAF permit pool
    participant Idle as SourceIdleDetector
    participant Tick as bg idle tick (2 s)
    participant SOT as source-OT
    participant ISB as ISB-OT
    participant Rdr as Reduce reader
    participant Red as Reducer

    Src->>W: writes flow normally, head_wm=1451606955000
    W->>Sema: take permit, await PAF
    Note over Sema: ⚠ PAFs slow to resolve<br/>(NATS hiccup)<br/>permits drained

    Src->>W: messages_tx.send(...) blocks<br/>(channel full, no permits)
    Note over Src: read loop stalls — no batches<br/>generate_and_publish_source_watermark<br/>not called → reset() not called

    Note over Idle: updated_ts last reset at 27.103<br/>after 5 s, threshold crossed → IDLE

    loop every step_interval (2 s)
        Tick->>Idle: partitions_needing_publish?
        Idle-->>Tick: [partition 0]
        Tick->>Idle: is_partition_idle(0)?
        Idle-->>Tick: true
        Tick->>SOT: head + increment_by = head + 3 s
        Tick->>ISB: stamp WMB(idle=true, wm=head)<br/>on every downstream partition
        Note over SOT,ISB: head_wm climbs:<br/>1451606955000 → 1451607009000 → 1451607012000
    end

    Note over Sema: PAFs eventually resolve
    Sema-->>W: permits freed
    W->>SOT: fetch_source_watermark(msg_offset)
    SOT-->>W: returns inflated head 1451607012000
    W->>ISB: stamp WMB(idle=false, wm=1451607012000)
    Note over W,ISB: poisoned WMB stamped a second time

    Src->>W: read loop unblocks<br/>data batch with event_times<br/>BELOW the inflated head

    Rdr->>ISB: fetch_watermark for each msg
    ISB-->>Rdr: returns 1451607012000
    Rdr->>Red: forward msg<br/>(msg.watermark = 1451607012000)

    Red->>Red: current_watermark = max(prev, 1451607012000)
    loop For each subsequent message
        Red->>Red: msg.event_time < 1451607012000?
        Note over Red: YES → DROP<br/>"Old message popped up"
    end
```

### **To Reproduce**
This is not guaranteed to reproduce the behavior, as it requires a couple of well timed conditions. 
1. set `max_delay` as 0
2. set idle threshold to a low value
3. start a high throughput pipeline and kill some isb pods to introduce delay in PAF resolutions
4. stop producing messages to source/wait for PAF concurrency to fill up because of slow PAF resolutions. 

### **Expected behavior**
Valid inflight messages should be successfully processed downstream and not discarded. 

- Kubernetes:
- Numaflow: main-e058e42

**Additional context**
Thanks to @BulkBeing 's validation framework this bug was surfaced.

---


<!-- Issue Author: Don't delete this message to encourage other users to support your issue! -->

vaibhavtiwari33:

As described in the linked issue, the background task responsible for publishing idle watermarks can increment the head WMB in source-OT, and subsequently in ISB-OT, even when there are messages in-flight, just that they're taking too long to be written to the ISB stream. 
The issue arises since the source idle manager only has the view of ```When was the last time any messages was published to a particular source partition?```, it doesn't care what is happening with the messages that have already been read and are still inflight. 

The ISB reader handles this correctly by finding out the [lowest watermark among in-flight messages](https://github.com/numaproj/numaflow/blob/27e11afbe8e05c115fe14d1e4ff24575d352dc5f/rust/numaflow-core/src/watermark/isb.rs#L193-L198) and incorporating that info as well when calculating the watermark to publish. 
```rust
// if window manager is not configured, we should use the lowest watermark among all the
// inflight messages.
self.get_lowest_watermark()
    .await
    .unwrap_or(Watermark::from_timestamp_millis(-1).unwrap())
```
 
This in turn calls the tracker to infer the lowest watermark:
```rust
/// Gets the lowest watermark among all the inflight requests using the tracker
async fn get_lowest_watermark(&self) -> Option<Watermark> {
    if let Ok(lowest_watermark) = self.tracker.lowest_watermark().await {
        Some(Watermark::from_timestamp_millis(lowest_watermark.timestamp_millis()).unwrap())
    } else {
        None
    }
}
```

At source, we cannot infer the lowest watermark using the tracker because at the time of message insertion into the tracker, we haven't assigned watermark to the messages yet, as they're fresh off of the [read call](https://github.com/numaproj/numaflow/blob/27e11afbe8e05c115fe14d1e4ff24575d352dc5f/rust/numaflow-core/src/source.rs#L502-L531).

Thus, we'll need to also track `event_time` of the messages in the tracker and use it for watermark publishing. But, since the event_times can be out of order at the source, we need to find the min `event_time` across all the in-flight messages in the tracker (and not just the ones with min offset like we do to get min watermark). 

### Related issues
Fixes #3399

### Testing
This PR was tested using the validation framework. 

### DIFF
--- a/rust/numaflow-core/src/tracker.rs
+++ b/rust/numaflow-core/src/tracker.rs
@@ -29,6 +29,8 @@ struct TrackerEntry {
     serving_callback_info: Option<ServingCallbackInfo>,
     /// Watermark for the message
     watermark: Option<DateTime<Utc>>,
+    /// EventTime for the message
+    event_time: DateTime<Utc>,
 }
 
 /// TrackerState holds the mutable state of the tracker.
@@ -211,6 +213,7 @@ impl Tracker {
             TrackerEntry {
                 serving_callback_info: callback_info,
                 watermark: message.watermark,
+                event_time: message.event_time,
             },
         );
         Ok(())
@@ -335,6 +338,19 @@ impl Tracker {
             })
             .min();
         Ok(watermark.unwrap_or(DateTime::from_timestamp_millis(-1).unwrap()))
+    }
+
+    /// Returns the lowest event_time among all the tracked offsets.
+    pub(crate) async fn lowest_event_time(&self) -> Result<DateTime<Utc>> {
+        let state = self.state.read().await;
+        // Get the lowest event_time across all partitions
+        let event_time = state
+            .entries
+            .values()
+            .flat_map(|partition_entries| partition_entries.values())
+            .map(|entry| entry.event_time)
+            .min();
+        Ok(event_time.unwrap_or(DateTime::from_timestamp_millis(-1).unwrap()))
     }
 
     /// Sets the idle status of the tracker. Setting idle offset to None means the partition is not
--- a/rust/numaflow-core/src/pipeline/forwarder.rs
+++ b/rust/numaflow-core/src/pipeline/forwarder.rs
@@ -37,8 +37,6 @@
 //! [Actor Pattern]: https://ryhl.io/blog/actors-with-tokio/
 
 use crate::config::pipeline::PipelineConfig;
-use crate::config::pipeline::watermark::WatermarkConfig;
-use crate::watermark::source::SourceWatermarkHandle;
 use crate::{config, error, pipeline};
 use tokio_util::sync::CancellationToken;
 use tracing::info;
@@ -67,27 +65,11 @@ pub(crate) async fn start_forwarder(
         config::pipeline::VertexConfig::Source(source) => {
             info!("Starting source forwarder");
 
-            // create watermark handle, if watermark is enabled
-            // Uses idle_config.step_interval (default 100ms) as the unified heartbeat interval
-            let source_watermark_handle = match &config.watermark_config {
-                Some(WatermarkConfig::Source(source_config)) => Some(
-                    SourceWatermarkHandle::new(
-                        js_context.clone(),
-                        &config.to_vertex_config,
-                        source_config,
-                        cln_token.clone(),
-                    )
-                    .await?,
-                ),
-                _ => None,
-            };
-
             source_forwarder::start_source_forwarder(
                 cln_token,
                 js_context,
                 config.clone(),
                 source.clone(),
-                source_watermark_handle,
             )
             .await?;
         }
--- a/rust/numaflow-core/src/pipeline/forwarder/source_forwarder.rs
+++ b/rust/numaflow-core/src/pipeline/forwarder/source_forwarder.rs
@@ -7,6 +7,7 @@ use crate::metrics::{
 };
 use crate::pipeline::PipelineContext;
 
+use crate::config::pipeline::watermark::WatermarkConfig;
 use crate::pipeline::isb::writer::{ISBWriterOrchestrator, ISBWriterOrchestratorComponents};
 use crate::shared::create_components;
 use crate::shared::metrics::start_metrics_server;
@@ -73,7 +74,6 @@ pub(crate) async fn start_source_forwarder(
     js_context: Context,
     config: PipelineConfig,
     source_config: SourceVtxConfig,
-    source_watermark_handle: Option<SourceWatermarkHandle>,
 ) -> error::Result<()> {
     let serving_callback_handler = if let Some(cb_cfg) = &config.callback_config {
         Some(
@@ -90,6 +90,22 @@ pub(crate) async fn start_source_forwarder(
     };
 
     let tracker = Tracker::new(serving_callback_handler, cln_token.clone());
+
+    // create watermark handle, if watermark is enabled
+    // Uses idle_config.step_interval (default 100ms) as the unified heartbeat interval
+    let source_watermark_handle = match &config.watermark_config {
+        Some(WatermarkConfig::Source(source_config)) => Some(
+            SourceWatermarkHandle::new(
+                js_context.clone(),
+                &config.to_vertex_config,
+                source_config,
+                cln_token.clone(),
+                tracker.clone(),
+            )
+            .await?,
+        ),
+        _ => None,
+    };
 
     // Create the ISB factory from the JetStream context
     use crate::pipeline::isb::ISBFactory;
@@ -1117,9 +1133,6 @@ mod tests {
                 panic!("Expected source vertex config");
             };
 
-        // For this test, we don't have watermark config, so watermark handle is None
-        let source_watermark_handle = None;
-
         let cancellation_token = CancellationToken::new();
         let forwarder_task = tokio::spawn({
             let cancellation_token = cancellation_token.clone();
@@ -1130,7 +1143,6 @@ mod tests {
                     context,
                     pipeline_config,
                     source_vtx_config,
-                    source_watermark_handle,
                 )
                 .await
                 .unwrap();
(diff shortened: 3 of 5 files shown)


rust/numaflow-core/src/tracker.rs
```diff
@@ -211,6 +213,7 @@ impl Tracker {
             TrackerEntry {
                 serving_callback_info: callback_info,
                 watermark: message.watermark,
+                event_time: message.event_time,
```
yhl25, line 216: Do we need to store the event time? because for source, watermark will be set to event time and the `lowest_watermark` should be same as the `lowest_event_time`?

vaibhavtiwari33: At the time when the messages are inserted into the tracker, after reading them off of the source, we may/may not have the watermark information in the message. Depends on the source. But we'll definitely have event_time information. 

Since, event_time of the messages can be out of order, thus we need to compare against all of the inflight messages.

vaibhavtiwari33: The way I'm saying this is based off of looking at some of our in-built source implementations, for example for http source we have:

```rust
impl From<HttpMessage> for Message {
    fn from(value: HttpMessage) -> Self {
        Message {
            typ: Default::default(),
            keys: Arc::from(value.keys),
            tags: None,
            value: value.body,
            offset: Offset::String(StringOffset::new(value.id.clone(), *get_vertex_replica())),
            event_time: value.event_time,
            watermark: None,                  // <----------------------
            id: MessageID {
                vertex_name: get_vertex_name().to_string().into(),
                offset: value.id.into(),
                index: 0,
            },
            headers: Arc::new(value.headers),
            // Set default metadata so that metadata is always present.
            metadata: Some(Arc::new(Metadata::default())),
            is_late: false,
        }
    }
}
```

yhl25 (approved):

PR #3404, opened 2026-05-04
fix: nack messages during token cancellation in ISBWrite task

### LINKED ISSUE #3403: Spurious ACKs on token cancellation in ISB writer
### **Describe the bug**

When the cancellation token (`cln_token`) is cancelled while `ISBWriteTask` is in the retry loop inside `write_to_stream`, the function returns `None`. That `None` is treated by callers as an *intentional skip* (the same sentinel used for `BufferFullStrategy::DiscardLatest` and `should_forward == false`). As a result, the message ends up `mark_success!`'d even though it was never written to the ISB stream. The source's `oneshot::Sender<ReadAck>` fires `ReadAck::Ack`, the source commits its offset, and on restart the message is **not** redelivered.

### **To Reproduce**
Following steps are not guaranteed to reproduce the issue every time. 

1. Start a high throughput pipeline with map UDF
2. Randomly issue sigkills to map vertex pods

### **Expected behavior**
Each message sent by the source should be received at the sink

### **Screenshots**
Following are the splunk query screenshots for logs from a run in validation dashboard. In the validation run, first run succeeded while the next run failed. 

#### First run: 5:50AM - 6:19AM PDT

Query: 
```spl
index=numaperfharness source="iks2/ip-paas-ppd-use2-k8s/us-east-2*/*/oss-analytics-numaflowperfharness-use2-e2e/map-*/numa/stdout" 
"Processed message*" 
| spath processed | timechart span=30s sum(processed) by host
```

Image

#### Second Run: 6:24AM - 6:57AM PDT

Query:
```spl
index=numaperfharness source="iks2/ip-paas-ppd-use2-k8s/us-east-2*/*/oss-analytics-numaflowperfharness-use2-e2e/map-*/numa/stdout" 
"Processed message*" 
| spath processed 
| timechart span=30s sum(processed) by host
```

Image

Notice the long tail at the end (6:51AM-6:57AM) when none of the pods were processing any messages before getting killed by the validation platform.
To ensure that there really were no messages to process in any of pods, we check pending at each pod:

Query: 
```spl
index=numaperfharness 
source="iks2/ip-paas-ppd-use2-k8s/us-east-2*/*/oss-analytics-numaflowperfharness-use2-e2e/map-*/numa/stdout" 
source!="iks2/ip-paas-ppd-use2-k8s/us-east-2*/*/oss-analytics-numaflowperfharness-use2-e2e/map-*source*/numa/stdout" 
| spath level | search level=INFO 
| spath message | search message="Pending messages*" 
| rex field=_raw "Pending messages Some\((?<pending>\d+)\)"
| timechart span=1m avg(pending) as pending_messages by host useother=f limit=0
```

Image

So, during that tail period in the second run, there were no pending messages either. But since the validation pipeline was waiting for additional messages, this means that the messages must be lost somewhere in the second run, i.e., they got positively acknowledged when they weren't supposed to. 

I checked all the different types of errors that got raised during each pipeline run:

Query:
```spl
index=numaperfharness source="iks2/ip-paas-ppd-use2-k8s/us-east-2*/*/oss-analytics-numaflowperfharness-use2-e2e/map-*/numa/stdout" 
| spath level | search level!=INFO 
| spath message | search message!="stream is full (throttled logging)"  
| timechart count by message useother=f limit=0
```

Image

All the different types of errors that occurred in the first run, also occurred in the second run, except one:

```json
{"timestamp":"2026-05-04T16:40:48.909701Z","level":"ERROR","message":"Shutdown signal received, exiting write loop","target":"numaflow_core::pipeline::isb::writer"}
```

At first I ignored this error, since this is only happening when we've cancelled the token and shutting everything down. But then I checked other failed runs as well. This error was common in all of these runs. 

This meant that it could be possible that during this exit flow in the ISB writer, we're somehow acknowledging the messages to be written to the stream, without actually writing them to the stream. Following the actual flow from this error leads us to the answer, as stated at the top. 

- Numaflow: main-ce2ac17

**Additional context**
Important bug highlighted with the help of @BulkBeing's validation platform

---

vaibhavtiwari33:
Bug is described in detail in the referenced issue.

### Related issues
Fixes #3403

### Testing
Added unit test to verify that the behaviour doesn't happen anymore. Running the PR through the validation pipeline

### DIFF
--- a/rust/numaflow-core/src/pipeline/isb/writer.rs
+++ b/rust/numaflow-core/src/pipeline/isb/writer.rs
@@ -3,9 +3,9 @@ use std::sync::Arc;
 use tokio::sync::OwnedSemaphorePermit;
 use tokio::sync::Semaphore;
 use tokio::task::JoinHandle;
-use tokio::time::{Duration, Instant, sleep};
-use tokio_stream::StreamExt;
+use tokio::time::{sleep, Duration, Instant};
 use tokio_stream::wrappers::ReceiverStream;
+use tokio_stream::StreamExt;
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, error, warn};
 
@@ -15,15 +15,15 @@ use crate::error::Error;
 use crate::mark_success;
 use crate::message::{Message, MessageHandle, Offset};
 use crate::metrics::{
-    PIPELINE_PARTITION_NAME_LABEL, pipeline_drop_metric_labels, pipeline_metric_labels,
-    pipeline_metrics,
+    pipeline_drop_metric_labels, pipeline_metric_labels, pipeline_metrics,
+    PIPELINE_PARTITION_NAME_LABEL,
 };
 use crate::pipeline::isb::error::ISBError;
 use crate::pipeline::isb::{ISBWriter, PendingWrite, WriteError, WriteResult};
 use crate::shared::forward;
 use crate::typ::NumaflowTypeConfig;
 use crate::watermark::WatermarkHandle;
-use crate::{Result, mark_failed};
+use crate::{mark_failed, Result};
 
 const DEFAULT_RETRY_INTERVAL_MILLIS: u64 = 10;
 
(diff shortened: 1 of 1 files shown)


yhl25 (approved):

rust/numaflow-core/src/pipeline/isb/writer.rs
```diff
@@ -1234,7 +1261,7 @@ mod simple_buffer_tests {
 
         // Message should still be acked (dropped messages are considered processed)
         let ack = ack_rx.await.unwrap();
-        assert_eq!(ack, ReadAck::Ack);
+        assert_eq!(ack, ReadAck::Nak);
```
vigith: Isn't this a regression?

vigith (changes requested):

yhl25 (approved):

vigith (approved):

PR #3406, opened 2026-05-06
chore: consider graceful timeout for batch map operations

yhl25:


rust/numaflow-core/src/mapper/map.rs
```diff
@@ -234,14 +235,15 @@ impl MapHandle {
...
                     error!(?error, "error received while performing map operation");
-
-                    // Store only the first Grpc error (root cause). Ignore Mapper errors (like "mapper closed") which are consequences.
-                    if matches!(error, Error::Grpc(_)) && self.final_result.is_ok() {
-                        // cancel the token to let the upstream know that we are shutting down so
-                        // that they stop sending new messages.
-                        ctx.cln_token.cancel();
-                        self.final_result = Err(error);
+                    if self.final_result.is_ok() {
+                        upstream_cln_token.cancel();
```
vigith, line 241: why are cancelling only if final_result is is_ok?

vaibhavtiwari33: As I understand, intent is to only cancel when we receive the first error.

vaibhavtiwari33 (approved):

vaibhavtiwari33 (changes requested):
~Every mapper error will restart the pod, at high throughput I'm seeing this to be a bit frequent. Let's gate the shutdown back to gRPC error~

rust/numaflow-core/src/mapper/map.rs
```diff
@@ -234,14 +235,15 @@ impl MapHandle {
...
                 Some(error) = ctx.error_rx.recv() => {
                     error!(?error, "error received while performing map operation");
-
-                    // Store only the first Grpc error (root cause). Ignore Mapper errors (like "mapper closed") which are consequences.
-                    if matches!(error, Error::Grpc(_)) && self.final_result.is_ok() {
-                        // cancel the token to let the upstream know that we are shutting down so
-                        // that they stop sending new messages.
-                        ctx.cln_token.cancel();
-                        self.final_result = Err(error);
+                    if self.final_result.is_ok() {
```
vigith, line 240:
```suggestion
                    // we only cancel when we get the first error
                    if self.final_result.is_ok() {
```

vigith (approved):

vigith:
> Every mapper error will restart the pod, at high throughput I'm seeing this to be quite frequent. Let's gate the shutdown back to gRPC error

why would that happen with this change?

rust/numaflow-core/src/mapper/map.rs
```diff
@@ -234,14 +235,16 @@ impl MapHandle {
...
                     error!(?error, "error received while performing map operation");
-
-                    // Store only the first Grpc error (root cause). Ignore Mapper errors (like "mapper closed") which are consequences.
-                    if matches!(error, Error::Grpc(_)) && self.final_result.is_ok() {
-                        // cancel the token to let the upstream know that we are shutting down so
-                        // that they stop sending new messages.
-                        ctx.cln_token.cancel();
-                        self.final_result = Err(error);
+                    // we only cancel when we get the first error
+                    if self.final_result.is_ok() {
```
vaibhavtiwari33, line 240: > > Every mapper error will restart the pod, at high throughput I'm seeing this to be quite frequent. Let's gate the shutdown back to gRPC error
>
> why would that happen with this change?

earlier we had a check to only cancel the token on Grpc errors, but mapper errors can also be sent to this channel, taking unary map for eg:
- https://github.com/numaproj/numaflow/blob/e5ed35efeb3f69f783df490a19b5b151ddaf58e8/rust/numaflow-core/src/mapper/map/unary.rs#L235-L241
- https://github.com/numaproj/numaflow/blob/e5ed35efeb3f69f783df490a19b5b151ddaf58e8/rust/numaflow-core/src/mapper/map/unary.rs#L264-L272

Any errors returned from the unary method above are then sent on the `error_tx` channel:
- https://github.com/numaproj/numaflow/blob/e5ed35efeb3f69f783df490a19b5b151ddaf58e8/rust/numaflow-core/src/mapper/map/unary.rs#L73-L86

vaibhavtiwari33:
Let's remove the `hard_shutdown_token` passed to batch map's context and move it back to the `cln_token`, any error encountered in batch map will only cancel the `hard_shutdown_token`, it won't signal iSB reader that it needs to stop sending messages

rust/numaflow-core/src/mapper/map.rs
vigith: shouldn't we be handling these errors? if they can be ignored, then let's not propagate. we need to get to the root cause on why these errors are generated.

vaibhavtiwari33: I think my observation was off, the constant restarts that I saw was in batch-map pods, these changes only concern themselves with unary and stream map.

vaibhavtiwari33: I'll resolve the conversation.

vaibhavtiwari33:
I've updated the `process_batch_messages` to cancel upstream token on error but monitor hard_shutdown_token for closing up shop.

vaibhavtiwari33 (approved):

Issue #2760, opened 2025-06-29
MonoVertex Card always shows health status as "healthy"


adarsh0728:
Currently health status is hardcoded as healthy for MonoVertex in UI.

https://github.com/numaproj/numaflow/blob/d84980e200c90c8dababba535752f0cec85fd408/ui/src/components/pages/Namespace/partials/MonoVertexCard/index.tsx#L474

Also mentioned in [PR](https://github.com/numaproj/numaflow/pull/2547) as a TODO. 
Creating the issue for tracking.

![Image](https://github.com/user-attachments/assets/a8624b1a-371b-4158-b3bf-e56c4aae6dae)

PR #3408, opened 2026-05-06
fix: implement `getMonoVertexStatus` to return correct health status

### LINKED ISSUE #2763: Backend Changes if required

suryapratap-01:

`getMonoVertexStatus` in `server/apis/v1/handler.go` was a stub that always returned `"healthy"` regardless of actual MonoVertex state. This caused the cluster summary API to always report all MonoVertices as active/healthy — even when paused, failed, or degraded — making the health dashboard unreliable for MonoVertices.

This PR implements real status logic based on `Status.Phase`, `Status.IsHealthy()`, and `Spec.Lifecycle.GetDesiredPhase()`, consistent with how `getPipelineStatus` and `getIsbServiceStatus` already work.

#### Status logic

| `Status.Phase` | `Spec.Lifecycle.DesiredPhase` | Conditions | Status returned | Reason |
|---|---|---|---|---|
| `Unknown` | any | any | `unknown` | Controller has not yet reconciled — phase is not set |
| `Failed` | any | any | `critical` | Controller marked deployment as failed |
| `Paused` | any | any | `inactive` | MonoVertex is fully paused, 0 replicas running |
| `Running` | `Paused` | any | `inactive` | Pause requested but controller has not yet updated `Status.Phase` — catches the in-progress pause window |
| `Running` | `Running` | any `False` | `warning` | Running but one or more conditions degraded (`Deployed`, `DaemonHealthy`, or `PodsHealthy`) |
| `Running` | `Running` | all `True` | `healthy` | Fully running and all conditions healthy |

The `Spec.Lifecycle.GetDesiredPhase()` check is critical — without it, there is a reconciliation lag window where the user has requested a pause but `Status.Phase` still shows `Running`, causing the MonoVertex to be incorrectly reported as `healthy` or `warning` during the transition.

The cluster summary buckets `inactive` and `unknown` both into the `Inactive` counter, since neither represents a MonoVertex that is actively processing data.

### Related issues

Fixes #2763
Part of #2760
Completes the backend TODO noted in #2547

### Testing

**Unit tests** — 6 cases covering all branches:

```bash
go test -v -race -short -timeout 60s ./server/apis/v1/ -run TestGetMonoVertexStatus
```

```
--- PASS: TestGetMonoVertexStatus/running_and_healthy
--- PASS: TestGetMonoVertexStatus/running_but_daemon_unhealthy
--- PASS: TestGetMonoVertexStatus/paused
--- PASS: TestGetMonoVertexStatus/failed
--- PASS: TestGetMonoVertexStatus/unknown_phase
--- PASS: TestGetMonoVertexStatus/running_but_pausing_in_progress
```

**E2E test:**

```bash
go test -v -timeout 10m -count 1 --tags test -p 1 ./test/api-e2e -run TestAPISuite/TestAPIsForIsbAndPipelineAndMonoVertex
```

```
--- PASS: TestAPISuite/TestAPIsForIsbAndPipelineAndMonoVertex (10.20s)
```

**Full suite:** `make test` — all 35 packages PASS. `make lint` — 0 issues.

**Manual verification on kind cluster** — 3 MonoVertices in different states:

```json
{
  "monoVertexSummary": {
    "active": { "Healthy": 1, "Warning": 1, "Critical": 0 },
    "inactive": 1
  }
}
```

Before this fix: always returned `{"active": {"Healthy": 3}, "inactive": 0}` regardless of actual state.

- The `Spec.Lifecycle.GetDesiredPhase() == Paused` check in the `Running` case is intentional — it handles the reconciliation lag between when the user requests a pause and when the controller updates `Status.Phase`. Without this check, the MonoVertex appears healthy during the pause transition.
- The E2E test `TestAPIsForIsbAndPipelineAndMonoVertex` had a latent race condition — it created a MonoVertex and immediately checked the cluster summary without waiting for it to become healthy. The stub masked this race. This PR adds an `assert.Eventually` poll to fix it properly.
- No frontend changes needed — the UI already handles all status values (`healthy`, `warning`, `critical`, `inactive`, `unknown`) via `GetConsolidatedHealthStatus()`. The frontend was built for this but the backend never returned anything other than `"healthy"`.
- Fix follows the same pattern as `getPipelineStatus` (line 1483) and `getIsbServiceStatus` (line 1499) in `handler.go`.

### DIFF
--- a/server/apis/v1/handler.go
+++ b/server/apis/v1/handler.go
@@ -323,7 +323,8 @@ func (h *handler) GetClusterSummary(c *gin.Context) {
 			h.respondWithError(c, fmt.Sprintf("Failed to fetch cluster summary, failed to get the status of the mono vertex %s, %s", monoVertex.Name, err.Error()))
 			return
 		}
-		if status == dfv1.MonoVertexStatusInactive {
+		// MonoVertices with inactive or unknown status are not actively processing data
+		if status == dfv1.MonoVertexStatusInactive || status == dfv1.MonoVertexStatusUnknown {
 			summary.monoVertexSummary.Inactive++
 		} else {
 			summary.monoVertexSummary.Active.increment(status)
@@ -1513,12 +1514,18 @@ func getIsbServiceStatus(isbsvc *dfv1.InterStepBufferService) (string, error) {
 func getMonoVertexStatus(mvt *dfv1.MonoVertex) (string, error) {
 	switch mvt.Status.Phase {
 	case dfv1.MonoVertexPhaseUnknown:
-		return dfv1.MonoVertexStatusInactive, nil
+		// Phase not yet set by the controller — reconciliation is pending
+		return dfv1.MonoVertexStatusUnknown, nil
 	case dfv1.MonoVertexPhaseFailed:
 		return dfv1.MonoVertexStatusCritical, nil
 	case dfv1.MonoVertexPhasePaused:
 		return dfv1.MonoVertexStatusInactive, nil
 	case dfv1.MonoVertexPhaseRunning:
+		// Check desired phase to catch the pausing-in-progress case:
+		// the user has requested a pause but the controller has not yet updated Status.Phase
+		if mvt.Spec.Lifecycle.GetDesiredPhase() == dfv1.MonoVertexPhasePaused {
+			return dfv1.MonoVertexStatusInactive, nil
+		}
 		if !mvt.Status.IsHealthy() {
 			return dfv1.MonoVertexStatusWarning, nil
 		}
--- a/server/apis/v1/handler.go
+++ b/server/apis/v1/handler.go
@@ -1531,7 +1531,8 @@ func getMonoVertexStatus(mvt *dfv1.MonoVertex) (string, error) {
 		}
 		return dfv1.MonoVertexStatusHealthy, nil
 	default:
-		return dfv1.MonoVertexStatusHealthy, nil
+		// Unhandled phase (e.g. Pausing, Deleting) — status is not deterministic
+		return dfv1.MonoVertexStatusUnknown, nil
 	}
 }
 
--- a/server/apis/v1/handler.go
+++ b/server/apis/v1/handler.go
@@ -323,12 +323,10 @@ func (h *handler) GetClusterSummary(c *gin.Context) {
 			h.respondWithError(c, fmt.Sprintf("Failed to fetch cluster summary, failed to get the status of the mono vertex %s, %s", monoVertex.Name, err.Error()))
 			return
 		}
-		// if the mono vertex is healthy, increment the active count, otherwise increment the inactive count
-		// TODO - add more status types for mono vertex and update the logic here
-		if status == dfv1.MonoVertexStatusHealthy {
-			summary.monoVertexSummary.Active.increment(status)
-		} else {
+		if status == dfv1.MonoVertexStatusInactive {
 			summary.monoVertexSummary.Inactive++
+		} else {
+			summary.monoVertexSummary.Active.increment(status)
 		}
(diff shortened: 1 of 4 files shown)


server/apis/v1/handler.go
```diff
@@ -323,12 +323,10 @@ func (h *handler) GetClusterSummary(c *gin.Context) {
 			h.respondWithError(c, fmt.Sprintf("Failed to fetch cluster summary, failed to get the status of the mono vertex %s, %s", monoVertex.Name, err.Error()))
 			return
 		}
-		// if the mono vertex is healthy, increment the active count, otherwise increment the inactive count
-		// TODO - add more status types for mono vertex and update the logic here
-		if status == dfv1.MonoVertexStatusHealthy {
-			summary.monoVertexSummary.Active.increment(status)
-		} else {
+		if status == dfv1.MonoVertexStatusInactive {
```
adarsh0728: lets return `MonoVertexStatusUnknown` when phase is unknown - and update this to increment inactive counter even on `MonoVertexStatusUnknown`

adarsh0728: also add a comment

server/apis/v1/handler.go
```diff
@@ -1513,8 +1511,21 @@ func getIsbServiceStatus(isbsvc *dfv1.InterStepBufferService) (string, error) {
 }
 
 func getMonoVertexStatus(mvt *dfv1.MonoVertex) (string, error) {
-	// TODO - add more logic to determine the status of a mono vertex
-	return dfv1.MonoVertexStatusHealthy, nil
+	switch mvt.Status.Phase {
+	case dfv1.MonoVertexPhaseUnknown:
+		return dfv1.MonoVertexStatusInactive, nil
```
adarsh0728: make it `MonoVertexStatusUnknown`

server/apis/v1/handler.go
```diff
@@ -1513,8 +1511,21 @@ func getIsbServiceStatus(isbsvc *dfv1.InterStepBufferService) (string, error) {
...
+	case dfv1.MonoVertexPhaseFailed:
+		return dfv1.MonoVertexStatusCritical, nil
+	case dfv1.MonoVertexPhasePaused:
```
adarsh0728, line 1521: lets add this check as well for cases where reconciliation could be pending and phase is not yet updated for pause.`mvt.Spec.Lifecycle.GetDesiredPhase() == dfv1.MonoVertexPhasePaused`

adarsh0728 (approved):

server/apis/v1/handler.go
```diff
@@ -1513,8 +1512,27 @@ func getIsbServiceStatus(isbsvc *dfv1.InterStepBufferService) (string, error) {
...
+		// the user has requested a pause but the controller has not yet updated Status.Phase
+		if mvt.Spec.Lifecycle.GetDesiredPhase() == dfv1.MonoVertexPhasePaused {
+			return dfv1.MonoVertexStatusInactive, nil
+		}
+		if !mvt.Status.IsHealthy() {
+			return dfv1.MonoVertexStatusWarning, nil
+		}
+		return dfv1.MonoVertexStatusHealthy, nil
+	default:
+		return dfv1.MonoVertexStatusHealthy, nil
```
vigith: shouldn't it return `MonoVertexStatusUnknown`?

vigith (changes requested):

server/apis/v1/handler.go
suryapratap-01: change implemented

vigith (approved):
can you do a follow up PR to fix `getPipelineStatus` which has weaker logic than what you have now implemented for MonoVertex

yhl25 (approved):

PR #3410, opened 2026-05-07
feat: add platform-wide `NUMAFLOW_LOG_LEVEL` support

adarsh0728:

- Users need a way to raise log levels for daemon and controller pods to reduce log noise.
- Adds `NUMAFLOW_LOG_LEVEL` support for Go components using zap log levels.
- Documents Go and Rust log-level controls, including examples for daemon, controller, webhook, server, ISB jobs, vertex, MonoVertex.
- Moves noisy idle MonoVertex rater metric logs from info to debug.

### DIFF
--- a/docs/user-guide/reference/configuration/log-levels.md
+++ b/docs/user-guide/reference/configuration/log-levels.md
@@ -1,36 +1,36 @@
 # Log Levels
 
-Numaflow pods use two separate logging systems depending on whether they run Go or Rust code. This page explains how to control the log level for each.
+Numaflow-owned pods use `NUMAFLOW_LOG_LEVEL` as the standard log-level control. Data-plane pods also support `RUST_LOG` for advanced filtering.
 
 ## Quick reference
 
-| Pod / container | Runtime | Log-level env var | Default level |
-|---|---|---|---|
-| Pipeline daemon | Go | `NUMAFLOW_LOG_LEVEL` | `info` |
-| MonoVertex daemon | Go | `NUMAFLOW_LOG_LEVEL` | `info` |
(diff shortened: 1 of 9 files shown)


docs/user-guide/reference/configuration/log-levels.md
```diff
@@ -0,0 +1,186 @@
...
+
+```yaml
+apiVersion: numaflow.numaproj.io/v1alpha1
+kind: MonoVertex
+spec:
+  daemonTemplate:
+    containerTemplate:
+      env:
+        - name: NUMAFLOW_LOG_LEVEL
+          value: warn
```
yhl25, line 59: I think we should just use `DAEMON_LOG_LEVEL` and not `NUMAFLOW_LOG_LEVEL`

yhl25: Should we explicitly mention `CONTROLLER_LOG_LEVEL`(daemon, controller etc) and `DATAPLANE_LOG_LEVEL` (should work like `RUST_LOG`)

BulkBeing: I think a single variable would be better, like NUMAFLOW_LOG_LEVEL.
We won't need to use DAEMON_LOG and NUMAFLOW_LOG at the same time for a single container anyway right?

Even if there is a situation like this, what do you think about treating them like components/modules and use Rust's log/tracing syntax:
```
NUMAFLOW_LOG_LEVEL=debug,builtin_udf=warn
```

yhl25: makes sense, lets go with one single log level env.

PR #3421, opened 2026-05-11
fix: prevent int32 overflow in MonoVertex autoscaler desiredReplicas

### LINKED ISSUE #3415: Integer overflow in MonoVertex autoscaler desiredReplicas causes incorrect replica count
**Issue**                                                                                                                                                                                                                                                  
  
The MonoVertex autoscaler logs a large negative desired replica count (e.g. `-1981856851`) due to an unchecked `float64 → int32` cast in `desiredReplicas`           
  (`scaling.go:314`). The two upstream rate guards only reject `rate < 0` and `rate == 0`, allowing a near-zero but valid rate (e.g. `0.001`) to pass through. Combined
   with a large message backlog, the intermediate `float64` result exceeds `math.MaxInt32` (~2.1B) and wraps to a large negative on cast.                              
                                                                                                                                                                       
  The pipeline itself is not broken — the negative value is clamped to `minReplicas` downstream — but the computed replica count is wrong and the log is misleading.   
    
**Logs**

The pipeline in which issue was observed had a huge lag, processing rate was around 6000 tps and only 1 pod was running because of the above mentioned bug

<img width="1289" height="504" alt="Image" src="https://github.com/user-attachments/assets/c7d13723-d632-4509-80c6-8271fea64551" />

```
{"level":"info","ts":"2026-05-08T11:16:43.194355004Z","logger":"numaflow.mvtx-autoscaler","caller":"scaling/scaling.go:240","msg":"Calculated desired replica number of MonoVertex \"**\" is: -1981856851.","worker":"11","monoVtxKey":"**"}
{"level":"info","ts":"2026-05-08T11:16:43.194546158Z","logger":"numaflow.mvtx-autoscaler","caller":"scaling/scaling.go:249","msg":"Calculated desired replica number 1 of MonoVertex \"**\" is smaller than min, using min 1.","worker":"11","monoVtxKey":"**
```
                                                                                                                                                                
  **To Reproduce**                                                                                                                                                     
  1. Deploy a MonoVertex with a large message backlog (e.g. 100,000+ pending messages).                                                                                
  2. Allow the processing rate to drop to a very small but nonzero value (e.g. pipeline restarts, rate briefly near `0.001 msg/sec`).                                
  3. Observe the autoscaler log: `Calculated desired replica number of MonoVertex "..." is: -1981856851.`                                                              
  4. Observe the follow-up log clamping to min: `smaller than min, using min 1.`                                                                                       
                                                                                                                                                                       
  **Expected behavior**                                                                                                                                                
  The desired replica count should never be negative. With a large backlog and low rate, it should be clamped to `math.MaxInt32` before the `int32` cast (the caller already clamps to `maxReplicas` after).                                                                                                                              
                                                                                                                                                                       
  **Root cause**
  `scaling.go:314`:                                                                                                                                                    
  ```go                                                 
  desired = int32(math.Round(                                                                                                                                          
      ((float64(pending) / processingRate) / float64(targetSeconds)) * float64(readyReplicas)
  ))
```                                                                                                                                                                   
  With pending=100000, processingRate=0.001, targetSeconds=20, readyReplicas=1:                                                                                        
  → intermediate float64 = 5,000,000,000 → int32 overflow → -1981856851        
                                                                                                                                                                       
  **Environment**                                                                                                                                                          
  - Kubernetes: any                                                                                                                                                    
  - Numaflow: affected on current main                                                                                                                                 
                                                        
  **Additional context**                                                                                                                                                   
  
The monovertex scaling test file (scaling_test.go) is currently empty — no unit tests exist for desiredReplicas. The fix involves clamping the float64 to math.MaxInt32 before casting, and adding test coverage for the overflow case.

### PR DESCRIPTION
**Pull request by suryapratap-01**

The MonoVertex autoscaler computed `desiredReplicas` in `pkg/reconciler/monovertex/scaling/scaling.go` via an unchecked `float64 → int32` cast. With a large backlog and a small-but-nonzero processing rate (e.g. `pending=100,000`, `rate=0.001 msg/s`, `targetProcessingSeconds=20`, `readyReplicas=1`), the intermediate `float64` (`5,000,000,000`) exceeds `math.MaxInt32` (`~2.147B`) and wraps to a large negative `int32` (`-1981856851`). The downstream `desired < minReplicas` branch then clamps to `minReplicas`, so the MonoVertex stays stuck at `min` (usually 1) under heavy lag with a misleading negative log line. The original `int32(pending)` cap also wraps when `pending > math.MaxInt32`, so the same failure mode is reachable purely from a huge backlog (e.g. ~2.35B pending) at any rate.

This PR makes the cast overflow-safe:

- Clamp the `float64` result to `math.MaxInt32` before casting to `int32`.
- Treat any non-positive cast result as `1` (changed from `== 0` to `<= 0`) as defense-in-depth.
- Guard the `pending`-based cap so that `pending > math.MaxInt32` produces a `pendingCap = math.MaxInt32` instead of wrapping when narrowed to `int32`.

The caller still clamps to `spec.scale.max`, so the maximum effective desired remains policy-controlled.

## Related issues

Fixes #3415

## Testing

Unit tests (new + existing) in `pkg/reconciler/monovertex/scaling/scaling_test.go`:

```bash
go test -race -short -timeout 60s ./pkg/reconciler/monovertex/scaling/... -run TestDesiredReplicas -v
```

All 9 subtests pass, including 3 new regression cases:

- `overflow_fromIssue3415` — `pending=100_000`, `rate=0.001`, `targetSec=20`, `readyReplicas=1` → expected `100_000` (would be a large negative without the fix).
- `extremeOverflow` — `pending=1_000_000`, `rate=0.0001`, `targetSec=1`, `readyReplicas=10` → expected `1_000_000`.
- `pendingExceedsMaxInt32` — `pending = int64(math.MaxInt32) + 1000`, `rate=1e9`, `targetSec=20` → expected `1` (pending-cap path does not wrap to negative).

Lint:

```bash
golangci-lint run --timeout 5m ./pkg/reconciler/monovertex/scaling/...
# 0 issues
```

## Special notes for reviewers

- Pure-Go change confined to `desiredReplicas` in `pkg/reconciler/monovertex/scaling/scaling.go`. No proto / CRD / Rust / API changes.
- No behavior change for healthy inputs — the new path is only entered when the float result would have wrapped (`raw > math.MaxInt32`) or when `pending > math.MaxInt32`. All pre-existing tests and call-site contracts are preserved.
- Optional follow-up (out of scope): factor the cast into a small `clampToInt32(float64) int32` helper that also rejects `NaN` / `±Inf`. Current upstream rate guards (`rate < 0`, `rate == 0`) make `NaN` / `±Inf` unreachable today, but a helper would harden future call sites.
- Rollout risk: low. Worst case under the fix is `desired == math.MaxInt32`, which the caller immediately clamps to `spec.scale.max`.

### DIFF
--- a/pkg/reconciler/monovertex/scaling/scaling.go
+++ b/pkg/reconciler/monovertex/scaling/scaling.go
@@ -308,17 +308,30 @@ func (s *Scaler) desiredReplicas(_ context.Context, monoVtx *dfv1.MonoVertex, pr
 		return int32(monoVtx.Status.Replicas)
 	}
 
-	var desired int32
 	// We calculate the time of finishing processing the pending messages,
 	// and then we know how many replicas are needed to get them done in target seconds.
-	desired = int32(math.Round(((float64(pending) / processingRate) / float64(monoVtx.Spec.Scale.GetTargetProcessingSeconds())) * float64(monoVtx.Status.ReadyReplicas)))
+	// Clamp the float64 result before casting to int32 to prevent wraparound when the
+	// intermediate value exceeds math.MaxInt32 (can happen with large pending + near-zero rate).
+	raw := math.Round(((float64(pending) / processingRate) / float64(monoVtx.Spec.Scale.GetTargetProcessingSeconds())) * float64(monoVtx.Status.ReadyReplicas))
+	if raw > math.MaxInt32 {
+		raw = math.MaxInt32
+	}
+	desired := int32(raw)
 
 	// we only scale down to zero when the pending and rate are both zero.
-	if desired == 0 {
+	if desired <= 0 {
 		desired = 1
 	}
-	if desired > int32(pending) && pending > 0 { // For some corner cases, we don't want to scale up to more than pending.
-		desired = int32(pending)
+	// For some corner cases, we don't want to scale up to more than pending.
+	// Guard the int32 cast on pending to prevent wraparound when pending > math.MaxInt32.
+	if pending > 0 {
+		pendingCap := int32(math.MaxInt32)
+		if pending <= math.MaxInt32 {
+			pendingCap = int32(pending)
+		}
+		if desired > pendingCap {
+			desired = pendingCap
+		}
 	}
 	return desired
 }
--- a/pkg/reconciler/monovertex/scaling/scaling.go
+++ b/pkg/reconciler/monovertex/scaling/scaling.go
@@ -312,18 +312,19 @@ func (s *Scaler) desiredReplicas(_ context.Context, monoVtx *dfv1.MonoVertex, pr
 	// and then we know how many replicas are needed to get them done in target seconds.
 	// Clamp the float64 result before casting to int32 to prevent wraparound when the
 	// intermediate value exceeds math.MaxInt32 (can happen with large pending + near-zero rate).
-	raw := math.Round(((float64(pending) / processingRate) / float64(monoVtx.Spec.Scale.GetTargetProcessingSeconds())) * float64(monoVtx.Status.ReadyReplicas))
-	if raw > math.MaxInt32 {
-		raw = math.MaxInt32
+	desiredRaw := math.Round(((float64(pending) / processingRate) / float64(monoVtx.Spec.Scale.GetTargetProcessingSeconds())) * float64(monoVtx.Status.ReadyReplicas))
+	if desiredRaw > math.MaxInt32 {
+		desiredRaw = math.MaxInt32
 	}
-	desired := int32(raw)
+	desired := int32(desiredRaw)
 
 	// we only scale down to zero when the pending and rate are both zero.
-	if desired <= 0 {
+	if desired == 0 {
 		desired = 1
 	}
 	// For some corner cases, we don't want to scale up to more than pending.
-	// Guard the int32 cast on pending to prevent wraparound when pending > math.MaxInt32.
+	// pending is int64 (matches the daemon's wrapperspb.Int64Value type), but desired
+	// replicas must fit in int32 per the Kubernetes replica spec, so we guard the cast.
 	if pending > 0 {
 		pendingCap := int32(math.MaxInt32)
 		if pending <= math.MaxInt32 {
--- a/pkg/reconciler/monovertex/scaling/scaling_test.go
+++ b/pkg/reconciler/monovertex/scaling/scaling_test.go
@@ -15,3 +15,130 @@ limitations under the License.
 */
 
 package scaling
+
+import (
+	"context"
+	"math"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	dfv1 "github.com/numaproj/numaflow/pkg/apis/numaflow/v1alpha1"
+)
+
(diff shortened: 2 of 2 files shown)


pkg/reconciler/monovertex/scaling/scaling.go
```diff
@@ -308,17 +308,30 @@ func (s *Scaler) desiredReplicas(_ context.Context, monoVtx *dfv1.MonoVertex, pr
...
+	// intermediate value exceeds math.MaxInt32 (can happen with large pending + near-zero rate).
+	raw := math.Round(((float64(pending) / processingRate) / float64(monoVtx.Spec.Scale.GetTargetProcessingSeconds())) * float64(monoVtx.Status.ReadyReplicas))
+	if raw > math.MaxInt32 {
+		raw = math.MaxInt32
+	}
+	desired := int32(raw)
 
 	// we only scale down to zero when the pending and rate are both zero.
-	if desired == 0 {
+	if desired <= 0 {
```
yhl25:
```suggestion
	if desired == 0 {
```

Since we are handling the overflow.

pkg/reconciler/monovertex/scaling/scaling.go
```diff
@@ -308,17 +308,30 @@ func (s *Scaler) desiredReplicas(_ context.Context, monoVtx *dfv1.MonoVertex, pr
...
 	}
-	if desired > int32(pending) && pending > 0 { // For some corner cases, we don't want to scale up to more than pending.
-		desired = int32(pending)
+	// For some corner cases, we don't want to scale up to more than pending.
+	// Guard the int32 cast on pending to prevent wraparound when pending > math.MaxInt32.
+	if pending > 0 {
+		pendingCap := int32(math.MaxInt32)
+		if pending <= math.MaxInt32 {
+			pendingCap = int32(pending)
+		}
```
yhl25, line 332: Can we please verify what type is used in daemon and use the same here? that way we don't have to do a overflow check here.

pkg/reconciler/monovertex/scaling/scaling.go
suryapratap-01: Done

pkg/reconciler/monovertex/scaling/scaling.go
suryapratap-01: Verified the daemon types:

1. MonoVertexMetrics.Pendings is map[string]*wrapperspb.Int64Value (pkg/apis/proto/mvtxdaemon/mvtxdaemon.pb.go)

2. CalculatePending returns int64 (pkg/mvtxdaemon/server/service/rater/helper.go:66)


The pending parameter in desiredReplicas is already int64, so types match. The overflow guard isn't there because of a type mismatch with the daemon — it's because desiredReplicas returns int32 (Kubernetes replica counts are int32: monoVtx.Status.Replicas, GetMaxReplicas() etc.). Any int64 → int32 cast on values that could exceed MaxInt32 will wrap, so we need to clamp before the cast. Updated the comment to make this explicit.

adarsh0728 (approved):

yhl25 (approved):

PR #3445, opened 2026-05-28
fix: remove concurrency default set by the kubebuilder 

vaibhavtiwari33:
remove concurrency default set by the kubebuilder to allow dataplane to set the concurrency value

### Related issues
Fixes recently observed flakiness in reduce-two-e2e test. 

### Testing
The test passes in CI consistently


vigith (approved):
What is the concurrency now if nothing is set? Who sets it?

vaibhavtiwari33:
It fallsback to the batch_size in the pipeline/monovertex config when the concurrency field is not set in the spec:

https://github.com/numaproj/numaflow/blob/8ad757118a1a5d7e8c8ec7e2d2b0e9bc79a7d9de/rust/numaflow-core/src/config/pipeline.rs#L347-L355

Since kubebuilder was always setting a default to 500 in the spec, that value was taking precedence.

PR #3449, opened 2026-06-02
feat: add MCP friendly diagnostic APIs

adarsh0728:
### Summary

This PR adds MCP/agent-friendly diagnostic APIs for Numaflow pipelines and mono-vertices. The goal is to expose small, typed, read-only responses that LLMs, automation tools can query directly.

### Why

The existing server APIs are useful for the UI, but many of them return raw daemon/Kubernetes shapes, large resource objects, making automated debugging harder because agents first need to discover topology, correlate health/metrics/errors/events across several APIs, and normalize response formats themselves.

### What Changed

Added single-purpose typed endpoints for diagnostics:

- Vertex metrics and pending counts.
- Pipeline buffer info and buffer pending counts.
- Pipeline and mono-vertex data health.
- Pipeline watermark lag.
- Pipeline, vertex, and mono-vertex status summaries.
- Vertex and mono-vertex logs.
- Vertex and mono-vertex runtime errors.
- Debug snapshots for both pipeline and monovertex
- Events and Discovery APIs

### Backward Compatibility

Existing UI/raw endpoints are preserved. New fields are additive.

### DIFF
--- a/pkg/apis/numaflow/v1alpha1/pipeline_types.go
+++ b/pkg/apis/numaflow/v1alpha1/pipeline_types.go
@@ -101,6 +101,13 @@ type Pipeline struct {
 	Status PipelineStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
 }
 
+// GetVertexFullName returns the generated Vertex CR name for a logical vertex
+// name, which is "<pipeline>-<vertex>". This is the name used for the Vertex
+// object and as the involvedObject name on its Kubernetes events.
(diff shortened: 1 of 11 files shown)


adarsh0728:
@suryapratap-01 let's think from MCP and revamped UI perspective, and take this forward.

suryapratap-01:
I have already taken your API-first code into consideration and built MCP on top of it. I’ll do another pass to see how we can make it better for the UI as well.

PR #3450, opened 2026-06-02
feat: support micro batching alternative at source

### LINKED ISSUE #3452: Support microbatching alternative at source for mvtx
# Summary

Support an optional (opt-in) `streaming` field that allows enabling per message level concurrency at source in monovertex. 

---

vaibhavtiwari33:
* Currently source reads 1 batch of messages at a time and waits for ack/nack of the whole batch of messages it read and sent downstream. 
* This is fine for pipeline which can read multiple batches (read ahead)
* In case of monovertex, it essentially blocks reading of additional messages until the current batch gets fully acked/nacked (which can be delayed in cases of large tail latency).

This PR aims to add an opt-in feature for monovertex to enable per message ack loop (instead of per batch acking)

* Added a field in monovertex spec to enable streaming (off by default)
* When streaming is enabled, acks for messages are tracked individually instead of in a batch
* Max in-flight messages are still capped using `concurrency`
* Semaphores (==`concurrency`) are distributed per message instead of per batch
* Most of the new logic lives in `streaming_source` in `source.rs` where we're doing the same operations as before, but on a per message level. 

Implemented spec:

```yaml
kind: MonoVertex
metadata:
  name: my-mvtx
spec:
  streaming: true
...
```

### Related issues
Closes https://github.com/numaproj/numaflow/issues/3452

### Testing
* Unit tests and an e2e test was added
* Tested with running a normal ud source with and without streaming enabled

#### Tracing Test:

**Streaming disabled; transformer returns results**
Screenshot 2026-06-03 at 5 01 03 PM

**Streaming enabled: transformer returns results**
Screenshot 2026-06-03 at 5 02 21 PM

**Streaming disabled; transformer omits results**
Screenshot 2026-06-03 at 5 08 31 PM

**Streaming enabled; transformer omits results**
Screenshot 2026-06-03 at 5 09 31 PM

This is a monovertex only feature currently since pipeline already supports reading multiple batches.


rust/numaflow-core/src/source.rs
```diff
@@ -491,218 +499,489 @@ impl<C: crate::typ::NumaflowTypeConfig> Source<C> {
...
+                // permit is acquired and ack task is spawned.
+                //
+                // transform_batch is called with a single-element Vec (per-message transform).
+                // This trades batch transformer throughput for per-message forward latency,
+                // which is acceptable in streaming mode.
+                let mut transformed_handles = match self.transformer.as_mut() {
+                    None => vec![msg_handle],
+                    Some(transformer) => {
+                        match transformer
+                            .transform_batch(vec![msg_handle], cln_token.clone(), None)
```
vigith: did you test tracing with this?

vaibhavtiwari33: I haven't tested end to end tracing with this change but all similar tracing changes have been made as the normal path, so should be good. I'll test this branch as well.

One more change that needs to be made in a separate PR is message level metrics (replacing batch level metrics for source)

vaibhavtiwari33: You're right, I missed creating spans for transformer in this branch. Fixing this.

vaibhavtiwari33: Updated, ty for catching this!

vigith (approved):
added a WARN if streaming and readAhead are set simultaneously

rust/numaflow-core/src/source.rs
```diff
@@ -502,6 +502,12 @@ impl<C: crate::typ::NumaflowTypeConfig> Source<C> {
             streaming = self.streaming,
             "Started streaming source"
         );
+        if self.streaming && self.read_ahead {
+            warn!(
+                "streaming=true supersedes read_ahead=true; read_ahead is ignored in streaming mode"
+            );
+        }
+
```
vaibhavtiwari33, line 505: Since currently streaming is a mvtx only feature, I'll move this inside the following block:
```rust
if self.streaming {
    self.streaming_source(
        pipeline_labels,
        mvtx_labels,
        bypass_router,
        messages_tx,
        cln_token,
    )
    .await
}
```

yhl25 (commented):
Can you run performance and validation with this change and attach the results?

vaibhavtiwari33:
I don't think there is a way for me to set "streaming: true" in the performance test. So ideally the numbers should've been closer than this, let me run the test again. 
Screenshot 2026-06-10 at 9 53 19 AM

vaibhavtiwari33:
Ran the benchmark again, the numbers came up in the same relative ballpark. This is expected, looks good to me:

Screenshot 2026-06-10 at 11 35 35 PM

vaibhavtiwari33:
Successfully passed validation for 12ish hrs
Screenshot 2026-06-11 at 10 45 26 AM

Issue #3368, opened 2026-04-15
Numa container should not restart on UDF failures/crashes


yhl25:
## Parent Issue
Part of #3367 — Remove monitor container and move functionality to numa container

## Overview

The numa container must remain running and healthy even when the UDF container crashes or restarts. Currently, numa container restarts can disrupt error reporting and state management when a UDF fails.

## Requirements

- Numa container should detect UDF container crashes without restarting itself
- Numa container should continue serving its endpoints (including the error endpoint) during and after a UDF crash
- Appropriate logging/signaling when a UDF failure is detected

## Acceptance Criteria

- [ ] Numa container stays running when the UDF container crashes
- [ ] Numa container correctly detects and logs UDF failures
- [ ] No disruption to numa container's own health checks during UDF restarts

PR #3459, opened 2026-06-08
feat: add udf shared helpers and resilient error persistence

adarsh0728:
### Summary

Introduces shared helpers, UDF-specific error classification, and safer runtime error persistence. The actual source/transformer/map/sink reconnect behaviour **will be wired in follow-up PRs.**

### Related Issues
https://github.com/numaproj/numaflow/issues/3368
### What Changed

- Added Error::UdfRedrive for UDF stream failures that should be redriven after reconnect instead of treated as fatal.
- Added shared `create_*_client` helpers for source, sink, source transformer, and mapper UDF clients.
  - Create a fresh UDS-backed gRPC channel.
  - Apply the configured gRPC max message size.
  - Wait for the typed client to become ready.
  - Re-read SDK server info from the running sidecar.
- Updated UDF startup paths to use the same shared client creation helpers that future reconnect paths will use.
- Updated runtime error persistence to support repeated and concurrent UDF errors.
   - Removed single-error persistence behaviour.
   - Added collision-safe filenames: <unix_nanos>-<sequence>-numa.{tmp,json}.

### Behaviour Change

- This PR **does not** yet change source/transformer/map/sink runtime behaviour. UDF failures are not fully reconnected in place until follow-up PRs wire these helpers into the UDF clients.
- The **behaviour that does change in this PR is runtime error persistence**

### DIFF
--- a/rust/numaflow-core/src/shared/grpc.rs
+++ b/rust/numaflow-core/src/shared/grpc.rs
@@ -18,7 +18,7 @@ use tonic::transport::{Channel, Endpoint};
 use tower::service_fn;
 use tracing::{info, warn};
 
-use numaflow_shared::server_info::sdk_server_info;
+use numaflow_shared::server_info::{ServerInfo, sdk_server_info};
 
 use crate::error;
 use crate::error::Error;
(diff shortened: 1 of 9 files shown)


rust/numaflow-core/src/shared/grpc.rs
```diff
@@ -179,6 +194,84 @@ pub(crate) async fn create_multi_rpc_channel(endpoints: Vec<String>) -> error::R
...
+    retry_interval: Duration,
+) -> error::Result<SinkClient<Channel>> {
+    let channel = create_rpc_channel_with_interval(socket_path, retry_interval).await?;
+    let mut client = SinkClient::new(channel)
+        .max_encoding_message_size(grpc_max_message_size)
+        .max_decoding_message_size(grpc_max_message_size);
+    wait_until_sink_ready(&cln_token, &mut client).await?;
+    let _server_info = sdk_server_info(server_info_path, cln_token.clone()).await?;
+    Ok(client)
+}
```
yhl25: why a separate method for `reconnect`? can we just use the existing connect method or call it `create_sink_client`? similarly for other clients as well.

adarsh0728: will use `create_sink_client` and similar naming for other components

adarsh0728: added different clients (will be used at startup and reconnect) , may be can add a common helper like `create_udf_client`  to avoid duplication of logic inside those helpers.. but current helpers look simple and readabale.

yhl25 (approved):

PR #3471, opened 2026-06-12
fix: pod logs streaming blocked by cleanResponseMiddleware buffering

### LINKED ISSUE #3470: Pod logs panel stuck on "Loading logs..." — follow=true streaming broken by cleanResponseMiddleware
**Describe the bug**

The pod logs panel in the Numaflow UI always shows "Loading logs..." and never displays any output. The same issue affects curl requests with follow=true.

  Root cause: Every api/v1 route is wrapped by cleanResponseMiddleware (added in #3098 to strip managedFields from JSON responses). This middleware replaces gin.ResponseWriter with a custom in-memory buffer
  (customResponseWriter) and only forwards the captured body to the client after the handler returns.

 The PodLogs handler (handler.go) opens a Kubernetes log stream with `Follow: true` and loops indefinitely, writing and flushing one log line at a time. Because the middleware has swapped the writer for a buffer:

  1. All log bytes accumulate in the in-memory buffer and are never sent to the network.
  2. c.Writer.Flush() (called per line) falls through to the embedded gin.ResponseWriter.Flush(), which is a no-op since nothing was written to the real writer.
  3. Since the follow stream never ends, c.Next() never returns, so the middleware's post-handler forwarding code is never reached.

  The client receives an open HTTP connection with zero body bytes indefinitely.

**To Reproduce**

1. Open the Numaflow UI and navigate to any running pod's logs panel.
  2. Observe the panel stays stuck on "Loading logs...".
  3. Alternatively:
  curl -k "https://<numaflow-server>/api/v1/namespaces/<ns>/pods/<pod>/logs?container=<container>&follow=true"
 
**Expected behavior**
Log lines stream to the client in real time.

**Screenshots**

- Numaflow: [1.8.0]

**Additional context**
- The UI hardcodes follow=true, so the logs panel is always affected.
  - Non-streaming endpoints (all other api/v1 routes) are unaffected — they return before the middleware flushes, so JSON cleaning works correctly.
---

suryapratap-01:

The UI logs panel gets stuck on "Loading logs..." and `follow=true` pod log requests return zero bytes.

`cleanResponseMiddleware` wraps the response writer in a buffer so it can strip `managedFields` from JSON, but only flushes after the handler returns. For streaming endpoints (pod logs with `follow=true`) the handler never returns while the client is connected, so no bytes reach the UI.

This PR makes the middleware detect streaming dynamically by overriding `Flush()` on the wrapped writer. When a handler calls `c.Writer.Flush()`, the writer drains its buffer, flips into pass-through mode, and forwards subsequent writes directly. JSON cleanup is skipped for streaming responses. 

### Related issues

Fixes #3470

### Testing

image

image

### Notes for reviewers

- Non-streaming endpoints are unchanged.
- A handler that streams without ever calling `Flush()` would still be buffered. All current streaming handlers flush.
- Thanks @adarsh0728 for the `Flush()`-detection suggestion.

### DIFF
--- a/server/routes/clean_response_middleware.go
+++ b/server/routes/clean_response_middleware.go
@@ -25,21 +25,13 @@ import (
 )
 
 var (
-	// "managedFields" is too annoying
+	// Fields stripped from buffered JSON responses.
 	cleanupFields = []string{"managedFields"}
 )
 
-// customResponseWriter wraps gin.ResponseWriter to buffer the response body for
-// post-processing. A handler that calls Flush() is treated as streaming: the
-// writer drains any buffered bytes, flips into pass-through mode, and forwards
-// every subsequent Write directly to the underlying ResponseWriter. Handlers
-// that never call Flush() remain fully buffered so the middleware can rewrite
-// the response (e.g. stripping managedFields).
-//
-// Note: a handler that streams without ever calling Flush() would still be
-// buffered. All current streaming endpoints (pod logs follow, c.Stream-based
-// handlers) call Flush(), so this is sufficient in practice. Hijacked
-// connections (e.g. websockets) bypass Write entirely and are unaffected.
+// customResponseWriter buffers non-streaming responses so the middleware can
+// rewrite them. Calling Flush marks the response as streaming: buffered bytes
+// are drained and subsequent writes pass through directly.
 type customResponseWriter struct {
 	gin.ResponseWriter
 	body      *bytes.Buffer
@@ -55,14 +47,12 @@ func (w *customResponseWriter) Write(b []byte) (int, error) {
 
 func (w *customResponseWriter) WriteString(s string) (int, error) {
 	if w.streaming {
-		return w.ResponseWriter.Write([]byte(s))
+		return w.ResponseWriter.WriteString(s)
 	}
 	return w.body.WriteString(s)
 }
 
-// Flush is the streaming signal. On the first call we drain any bytes that
-// were buffered before the handler decided to stream, flip into pass-through
-// mode, and then forward to the underlying writer's Flush.
+// Flush switches to streaming mode and forwards the flush to the underlying writer.
 func (w *customResponseWriter) Flush() {
 	if !w.streaming {
 		w.streaming = true
@@ -111,7 +101,7 @@ func cleanResponseMiddleware() gin.HandlerFunc {
 	}
 }
 
-// removeFieldRecursively removes all occurrences of the given field from a (possibly nested) JSON structure represented as map[string]interface{} or []interface{}.
+// removeFieldsRecursively removes fields from nested JSON objects and arrays.
 func removeFieldsRecursively(data interface{}, fields ...string) {
 	switch val := data.(type) {
 	case map[string]interface{}:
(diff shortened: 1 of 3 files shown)


server/routes/clean_response_middleware.go
```diff
@@ -46,6 +46,12 @@ func (w customResponseWriter) WriteString(s string) (int, error) {
 // cleanResponseMiddleware is a Gin middleware to clean up data in the response body.
 func cleanResponseMiddleware() gin.HandlerFunc {
 	return func(c *gin.Context) {
+		// Skip buffering for streaming endpoints (e.g. pod logs with follow=true),
+		// otherwise the response would never be written to the client.
+		if strings.HasSuffix(c.FullPath(), "/pods/:pod/logs") {
```
adarsh0728: we should implement a generic solution for future streaming endpoints eg:, new logs API

server/routes/clean_response_middleware.go
```diff
@@ -46,6 +46,12 @@ func (w customResponseWriter) WriteString(s string) (int, error) {
...
+			c.Next()
+			return
+		}
```
vaibhavtiwari33: ~There could be additional apis facing the same issue~

edit: adarsh pointed out the same thing, ignore

server/routes/clean_response_middleware.go
suryapratap-01: wilco

server/routes/clean_response_middleware.go
suryapratap-01: wilco

server/routes/clean_response_middleware.go
suryapratap-01: Done. Reworked this into a generic streaming-route bypass.

[cleanResponseMiddleware](cci:1://file:///Users/ssingh130/Documents/Projects/numaflow/server/routes/clean_response_middleware.go:47:0-87:1) now accepts configurable skippers, and I added a route-pattern skipper based on Gin `c.FullPath()`. Pod logs is configured through that mechanism instead of being hardcoded, so future streaming endpoints can be added declaratively.

Tests now cover configured streaming bypass, unconfigured route buffering, and grouped pod logs route wiring.

Validated with `go test ./server/routes/` and targeted `golangci-lint ./server/routes/...`.

suryapratap-01: Updated PR description

server/routes/clean_response_middleware.go
```diff
@@ -77,6 +88,18 @@ func cleanResponseMiddleware() gin.HandlerFunc {
 	}
 }
 
+// skipCleanResponseForRoutes matches Gin route patterns returned by c.FullPath().
+func skipCleanResponseForRoutes(routes ...string) cleanResponseSkipper {
```
adarsh0728: with this approach also, we would need to maintain a skip list. Right? Consider overriding Flush() on `customResponseWriter` to detect streaming

suryapratap-01: done

adarsh0728 (approved):

PR #3466, opened 2026-06-09
feat(UI): jetstream summary & raft details

suryapratap-01:
### What this PR does / why we need it

Adds ISB-level JetStream debug information in the UI, focused on JetStream Summary and RAFT Meta Group details.

### Related issues

#3235

### Testing

Tested with focused UI and backend checks:

```bash
npm test -- --runTestsByPath src/components/common/SlidingSidebar/partials/ISBUpdate/index.test.tsx --watchAll=false --coverage=false --ci
go test ./server/apis/v1 ./server/cmd/server
```

Also manually verified the ISB sidebar shows JetStream Summary, RAFT Meta Group Information, refresh behavior, and clickable ISB service details.

### Special notes for reviewers

This PR intentionally keeps scope small: Streams, Consumers and KV Stores per vertex are deferred to follow-up work.

https://github.com/user-attachments/assets/25084427-a443-48e3-b638-a2a07153628c

Issue #3235, opened 2026-02-13
UI enhancements for debugging


yhl25:
# Summary

We should be able to display the below information from the UI for improving the debugging experience to the User.

- Watermark per partition/pod
- KV store info
- Consumer and stream info

suryapratap-01:
I can take this up

PR #3472, opened 2026-06-12
feat: isb stream consumer kv debug info

suryapratap-01:

Surfaces ISB (JetStream) debug info in the UI instead of exec-ing into nats pods:

- 3 read-only endpoints: `.../isb/{streams,consumers,kv-stores}` with optional vertex/edge/partition filters (built on the monitor from #3466).
- New **ISB** tab in vertex and edge sidebars; edges also get per-partition **Watermarks**.
- Rollout-safe: shows "ISB information is not available yet" against older servers; renders partial data when some endpoints fail.

### Related issues

Part of #3235 — covers stream/consumer/KV info. Watermark per pod will come in the next PR.

### Testing

- Go handler tests (scoping, filters, validation, partial/total failures); UI suite 65/65 green, new tests for empty/partial/error states.
- Manual: kind cluster + 3-replica JetStream ISB; responses cross-checked against raw NATS `/jsz` — exact match. Verified filters and old-server fallback.

- Read-only endpoints; no CRD or codegen changes; no breaking API changes.
- UI degrades gracefully against older daemons/servers (relevant for in-place upgrades).
- **Open question on placement**: the stream/consumer/KV info currently lives in a dedicated **ISB** tab. It could instead be folded into the existing **Buffers** tab (single place for all buffer-related details) — let me know if that's the direction you'd prefer and I'll restructure.


server/apis/v1/jetstream_monitor.go
```diff
@@ -36,25 +36,26 @@ import (
...
+type jetStreamMonitorSnapshot struct {
+	pod    corev1.Pod
+	jsInfo *natsserver.JSInfo
+}
+
 // GetISBServiceJetStream returns JetStream monitor summary and RAFT meta group information for an ISB service.
 func (h *handler) GetISBServiceJetStream(c *gin.Context) {
 	ns, isbsvcName := c.Param("namespace"), c.Param("isb-service")
-	_, pods, ok := h.getJetStreamISBMonitorPods(c, ns, isbsvcName)
+	snapshots, errors, ok := h.collectJetStreamMonitorSnapshots(c, ns, isbsvcName)
```
adarsh0728, line 47: is this API becoming too overloaded?If yes, can we split this (both from UI and agent perspective) into two endpoints. One for summary and other for RAFT

server/apis/v1/pipeline_isb_debug.go
```diff
@@ -0,0 +1,549 @@
+package v1
```
adarsh0728, line 1: check if DTO builders can be moved to a separate file

server/apis/v1/pipeline_isb_debug.go
```diff
@@ -0,0 +1,549 @@
...
+
+func buildPipelineISBConsumerDTOs(snapshots []jetStreamMonitorSnapshot, expected map[string]pipelineISBResourceMetadata) []PipelineISBConsumerDTO {
+	consumers := map[string]PipelineISBConsumerDTO{}
+	for _, snapshot := range snapshots {
+		forEachJetStreamDetail(snapshot.jsInfo, func(stream natsserver.StreamDetail) {
+			metadata, ok := expected[stream.Name]
+			if !ok {
+				return
+			}
+			for _, consumer := range append(stream.Consumer, stream.DirectConsumer...) {
```
adarsh0728: use two loops (with a helper fn) instead of append

ui/src/components/common/SlidingSidebar/partials/EdgeDetails/index.tsx
```diff
@@ -1,20 +1,51 @@
...
+const ISB_TAB_INDEX = 1;
+
 export interface EdgeDetailsProps {
+  namespaceId?: string;
+  pipelineId?: string;
   edgeId: string;
-  watermarks: (string|number)[];
+  from?: string;
+  to?: string;
+  watermarks: (string | number)[];
```
adarsh0728: watermarks can be undefined..make it optional `? `

ui/src/components/common/SlidingSidebar/partials/PipelineISBDebugInfo/index.tsx
```diff
@@ -0,0 +1,329 @@
...
+
+const formatOptional = (value?: string | number) => {
+  if (value === undefined || value === "") {
+    return "-";
+  }
+  return typeof value === "number" ? formatNumber(value) : value;
+};
+
+const formatSeconds = (value?: number) => {
+  if (!value) {
```
adarsh0728: what if value is 0?

ui/src/components/common/SlidingSidebar/partials/PipelineISBDebugInfo/index.tsx
```diff
@@ -0,0 +1,329 @@
...
+  const streamRows = streams?.streams || [];
+  const consumerRows = consumers?.consumers || [];
+  const kvStoreRows = kvStores?.kvStores || [];
+  const monitorErrors = collectErrors(streams, consumers, kvStores);
+  const showSharedNotice = edgeScoped || hasSharedTargetVertexRows(streams, consumers);
+  const hasAnyInformation =
+    !!streamRows.length ||
+    !!consumerRows.length ||
+    !!kvStoreRows.length ||
+    !!monitorErrors.length;
```
adarsh0728: if there are errors and no info, then we would end up showing PARTIAL_ISB_UNAVAILABLE_MESSAGE. Separate it out so that we show ISB_UNAVAILABLE_MESSAGE in no rows + errors case

ui/src/utils/fetchWrappers/pipelineISBDebugFetch.ts
```diff
@@ -0,0 +1,153 @@
...
+    if (!shouldFetch) {
+      return;
+    }
+    if (streamsLoading || consumersLoading || kvStoresLoading) {
+      setLoading(true);
+      return;
+    }
+    const fetchError = streamsError || consumersError || kvStoresError;
+    const apiError = streamsData?.errMsg || consumersData?.errMsg || kvStoresData?.errMsg;
+    const availableData = {
```
adarsh0728, line 148: this can show stale data - we are just checking `streamsData?.data`

adarsh0728: we do not know the data belongs to the current request...setData and setError to undefined before making new requests. You can refer how we are doing it using `useEffect`

server/apis/v1/jetstream_monitor.go
suryapratap-01: ```GetISBServiceJetStream``` returns both summary and RAFT data, but both are derived from the same``` /jsz``` pod snapshots and the UI consumes them together. Splitting endpoints would duplicate monitor calls.

server/apis/v1/pipeline_isb_debug.go
suryapratap-01: Done. Moved the pipeline ISB expected-resource and DTO builder helpers into `pipeline_isb_debug_builders.go` so `pipeline_isb_debug.go` stays focused on handlers, and request validation.

server/apis/v1/pipeline_isb_debug.go
suryapratap-01: Done. Added a helper that visits regular consumers and direct consumers with separate loops, so we avoid `append(stream.Consumer, stream.DirectConsumer...)`. Also added coverage for the `DirectConsumer` path.

ui/src/components/common/SlidingSidebar/partials/EdgeDetails/index.tsx
suryapratap-01: Made `watermarks` optional in `EdgeDetailsProps` and added a regression test that renders edge details without passing watermarks.

ui/src/components/common/SlidingSidebar/partials/PipelineISBDebugInfo/index.tsx
suryapratap-01: Fixed `formatSeconds` to only treat `undefined` as unavailable, so `0` now renders as `0.00s`. Added test coverage for this case.

ui/src/components/common/SlidingSidebar/partials/PipelineISBDebugInfo/index.tsx
suryapratap-01: Done. Split “has displayable rows” from monitor errors. If there are no stream/consumer/KV rows, the UI now shows the full unavailable message, while still rendering monitor error details when available.

ui/src/utils/fetchWrappers/pipelineISBDebugFetch.ts
suryapratap-01: `refresh()` now clears `data` and `error` immediately, and request identity changes also clear them before accepting the next response. Added tests for both parameter changes and manual refresh.

adarsh0728 (approved):
stale data case still not fully addressed `usePipelineISBDebugFetch`: `availableData` still uses `streamsData?.data` / `consumersData?.data` / `kvStoresData?.data` w/o checking the endpoint error. `useFetch` can retain old data after a failed request, gate each entry with `!streamsError`, `!consumersError`, `!kvStoresError`, or clear old data in `useFetch`.

adarsh0728:
discussed offline, 

1. remove unnecessary fields,
2. move it under buffer, and expand only when user wants to see it
3. remove from edges for now (joins etc)

suryapratap-01:
https://github.com/user-attachments/assets/c4a4cdc6-7a15-4529-8ab5-b529c5456010





This is what I have come up with. Can I get a feedback @yhl25 @adarsh0728

yhl25:
> Screen.Recording.2026-06-25.at.2.24.12.PM.mov 
> This is what I have come up with. Can I get a feedback @yhl25 @adarsh0728

Showing ackPending in both the buffers view and consumer view can be confusing, also the numbers are not matching.

suryapratap-01:
Screenshot 2026-07-01 at 4 47 10 PM

ui/src/components/pages/Pipeline/partials/Graph/partials/CustomEdge/index.tsx
```diff
@@ -341,9 +341,18 @@ const CustomEdge: FC<EdgeProps<Edge<Record<string, any>>>> = ({
...
-                      <span>AckPending: {data?.ackPendingLabel || 0}</span>
+                    <div
+                      style={{
+                        display: "flex",
+                        flexDirection: "column",
+                        alignItems: "center",
+                        gap: "2px",
+                      }}
+                    >
+                      <span>Queued messages: {data?.pendingLabel || 0}</span>
```
adarsh0728: why labels/names are changed here?

adarsh0728:
remove `.` from all new tooltips added

ui/src/components/common/SlidingSidebar/partials/VertexDetails/partials/Buffers/index.tsx
```diff
@@ -30,79 +149,260 @@ export function Buffers({
...
+      key: "bufferLength",
+      label: "Buffer Length",
+      width: "12rem",
+      align: "center",
+      testId: "bufferLength",
+      tooltip: "Maximum configured buffer capacity in messages.",
+      render: (buffer) => buffer?.bufferLength,
+    },
+    {
+      key: "bufferUsageLimit",
```
adarsh0728: this is not required - user provided value (already in spec)

ui/src/components/common/SlidingSidebar/partials/VertexDetails/partials/Buffers/index.tsx
```diff
@@ -30,79 +149,260 @@ export function Buffers({
...
+      render: (buffer) => buffer?.ackPendingCount,
+    },
+    {
+      key: "totalMessages",
+      label: "Total Messages",
+      width: "14rem",
+      align: "center",
+      testId: "totalMessages",
+      tooltip:
+        "Total messages currently counted for this buffer, with metrics drilldown when available.",
```
adarsh0728: remove this tooltip, and why we changed it to Total Messages instead of Total Pending Messages

adarsh0728:
`ISB Information is not available yet.` fallback message is seen every time we collapse the accordion.

PR #3475, opened 2026-06-17
feat: keep numa alive across source, transformer, and sink UDF failures

adarsh0728:
### Summary

This PR wires UDF reconnect behavior for source, source transformer, and sink paths so `numa` does not exit on UDF failures.

### Testing

Kubernetes smoke tests covering panic/application errors, stream closures, hard container kills.

1. Rust SDK - validated transformer(panic, stream close mid flight, hard kill), sink(panic, stream close mid flight, hard kill), source (fail_mode read,ack,nack) --> fail_mode=nack with panic sink confirmed source offsets were nacked, UDSource nack() was called, the source process aborted, numa reconnected/redrove and stayed alive
**Issue Observed**: https://github.com/numaproj/numaflow-rs/pull/180, and panic mssgs didn't include container name
2. Go SDK - validated transformer(panic, stream close mid flight, hard kill), sink(panic, stream close mid flight, hard kill)
3. JAVA SDK - validated transformer(panic, stream close mid flight, hard kill), sink(panic, stream close mid flight, hard kill)
4. Python SDK- validated transformer(panic, stream close mid flight, hard kill), sink(panic, stream close mid flight, hard kill)

**Behaviour confirmed:** numa did not exit, udf restarts were tolerated, in flight mssgs were redriven, critical_error_total metric incremented(with low cardinality) , runtime errors persisted 

### Related Issues
part of https://github.com/numaproj/numaflow/issues/3368

### What Changed

Instead of allowing **tonic::Status / gRPC stream failures** to bubble up as terminal `Error::Grpc`, these paths now classify UDF side gRPC failures as `Error::UdfRedrive`. The forwarder then reconnects the UDF client and redrives/retries in flight messages, preserving at least once behaviour.

- In source, failures are handled at read(), ack() and nack()
- In transformer, If resp_stream.message().await returns an error:, record the error -> drain pending senders with Error::UdfRedrive -> break response loop so caller reconnects --> redrive in flight messages
- In sink, during sink_once(), failure sending requests to sink becomes Error::UdfRedrive , sink response stream close becomes Error::UdfRedrive , any gRPC error from sink response stream becomes Error::UdfRedrive --> then in sink(), Error::UdfRedrive  causes sink batch to be retried after reconnect()

### Follow Up

Add map related changes
Add container reference in UDFRedrive just in case sdk doesn't include container name
Handle EOT and other critical errors - no numa restart so that numaplane analysis template can be made more robust


vigith (commented):
@adarsh0728 could you please bump up the codecov a bit?

rust/numaflow-core/src/sinker/sink.rs
```diff
@@ -227,6 +230,10 @@ impl SinkWriter {
                             mark_success_batch!(read_batch);
                         }
                         Err(e) => {
+                            if matches!(e, Error::UdfRedrive(_)) {
+                                error!(?e, "redrivable sink error");
+                                continue;
+                            }
```
vigith, line 237: Should we call `mark_failed_batch!(read_batch, &e)` before the `continue`, (or retry the handshake inside `reconnect()` so `UdfRedrive` never escapes `sink()`)? The current silent auto-NAK makes incidents invisible in logs. Right?

rust/numaflow-core/src/source.rs
```diff
@@ -932,8 +932,12 @@ impl<C: crate::typ::NumaflowTypeConfig> Source<C> {
...
                             cln_token.clone(),
                         )
-                        .await
-                        {
+                        .await;
+                        if let Err(e) = result {
+                            if matches!(e, Error::UdfRedrive(_)) {
+                                error!(?e, "Redrivable error while invoking ack");
+                                return;
+                            }
```
vigith, line 943: shouldn't we `nack` here or will it be retried?

adarsh0728: we reach here after `invoke_ack_single` has chosen disposition --> now  if disposition is :

1. `ack`,   and `ack_with_retry` returns Error::UdfRedrive , that means source ack was not confirmed by numa. If source did not commit, so it should be redelivered after reconnect  OR if ack reached the source but the response failed, which is also acceptable.
2. `nack`, `nack_with_retry` currently swallows the retry result and returns Ok(()), so Error::UdfRedrive from source nack() generally will not reach this match. 

So this branch mainly means the source ack path hit a recoverable UDF redrive error and the ack retry loop stopped before `numa` could confirm the ack. `nack` is not required in that case

https://github.com/numaproj/numaflow/blob/83c554d5930d22582d901120a5d8fcae0c822c24/rust/numaflow-core/src/source.rs#L1106

vigith (approved):

rust/numaflow-core/src/transformer/user_defined.rs
```diff
@@ -206,12 +274,39 @@ impl UserDefinedTransformer {
...
+
+        if let Err(e) = self.reconnect().await {
+            let _ = respond_to.send(Err(e));
+            return;
+        }
+
+        self.senders
+            .lock()
+            .await
+            .insert(key.clone(), (msg_info, respond_to));
```
yhl25: we are doing send operation twice inside the transformer, what happens if it fails after the reconnect + send?

yhl25 (commented):
I think there is scope for simplifying the reconnection handling inside transformer.

rust/numaflow-core/src/transformer/user_defined.rs
adarsh0728: We try sending once. If the stream is closed, reconnect once. If sending still fails after reconnect, we return UdfRedrive so the message is retried 

```
                let transformed_messages = loop {
                    match Transformer::transform(
                        transform_handle.clone(),
                        read_msg.clone(),
                        hard_shutdown_token.clone(),
                    )
                    .await
                    {
                        Ok(messages) => break messages,
                        Err(Error::UdfRedrive(e)) => {
                            error!(?e, ?offset, "transformer stream redrive requested");
                        }
                        Err(e) => return Err(e),
                    }
                };
```

rust/numaflow-core/src/transformer/user_defined.rs
```diff
@@ -167,36 +199,91 @@ impl UserDefinedTransformer {
...
+
+    async fn queue_request(
+        &mut self,
+        key: &str,
+        msg_info: ParentMessageInfo,
+        respond_to: oneshot::Sender<Result<Vec<Message>>>,
+        request: SourceTransformRequest,
+    ) -> Option<(ParentMessageInfo, oneshot::Sender<Result<Vec<Message>>>)> {
+        self.senders
+            .lock()
```
vigith: Why are we using tokio mutex?

adarsh0728: will check this

adarsh0728: std::sync::Mutex is sufficient , we do not hold guard across await points @yhl25

adarsh0728:
Merging this PR, will take up map related changes as a follow up

PR #3481, opened 2026-06-23
feat: surface OOM (and other) reasons as free-form text

vaibhavtiwari33:

During Progressive Rollout when a Pipeline or MonoVertex fails, we capture its Status in the rollout’s Status so we can see later any information to why it failed. It would be very helpful to have information related to OOM, etc if that was the reason that it failed. (Also, maybe image pull failure would also be useful.)

At the Pod level, this is in the form of free-form text. If it’s not too long and we want to keep the logic simple, we could consider just taking all containers that are failed and appending the message string.

### Testing
Added unit tests for these. 

#### Testing for status change: 
Watching for status change in mvtx: [mvtxStatusCapture](https://gist.github.com/vaibhavtiwari33/72af64b8f2e29b26004e3d2b9d642edf)
Watching for status change in pod: [podStatusCapture](https://gist.github.com/vaibhavtiwari33/9a44b4a94e9b6b58a95599f646fcd51c)

### DIFF
--- a/pkg/reconciler/util.go
+++ b/pkg/reconciler/util.go
@@ -130,15 +130,7 @@ func formatContainerFailure(c corev1.ContainerStatus, reason, msg string) string
 // formatRecentRestart renders a container that recently OOM/Error-restarted but is no longer in a
 // failed state. The OOMKilled reason is inferred from exit code 137 when not explicitly set.
 func formatRecentRestart(c corev1.ContainerStatus, x *corev1.ContainerStateTerminated) string {
-	reason := x.Reason
-	if reason == "" {
-		if x.ExitCode == 137 {
-			reason = "OOMKilled"
-		} else {
-			reason = "Error"
-		}
-	}
-	return fmt.Sprintf("container %q restarted recently: %s (exit code %d)", c.Name, reason, x.ExitCode)
+	return fmt.Sprintf("container %q restarted recently: %s %s (exit code %d)", c.Name, x.Reason, x.Message, x.ExitCode)
 }
 
 func NumOfReadyPods(pods corev1.PodList) int {
--- a/pkg/reconciler/util.go
+++ b/pkg/reconciler/util.go
@@ -127,22 +127,15 @@ func formatContainerFailure(c corev1.ContainerStatus, reason, msg string) string
 	return detail
 }
 
-// formatRecentRestart renders a container that recently OOM/Error-restarted. The detail is normally
-// read from the last termination, but the current terminated state is preferred when it carries a
-// reason, and OOMKilled is given top preference wherever it appears: the runtime tags OOM
-// best-effort, so the current and last terminations can disagree (one "OOMKilled", one "Error").
+// formatRecentRestart renders a container that recently OOM/Error-restarted.
 // Reason, message and exit code are reported verbatim (no OOM is inferred); the message is included
 // only when present, since it is empty for a typical OOM kill.
 func formatRecentRestart(c corev1.ContainerStatus, x *corev1.ContainerStateTerminated) string {
-	t := x
-	if cur := c.State.Terminated; cur != nil && cur.Reason != "" && x.Reason != "OOMKilled" {
-		t = cur
+	detail := fmt.Sprintf("container %q restarted recently: %s", c.Name, x.Reason)
+	if x.Message != "" {
+		detail += " " + x.Message
 	}
-	detail := fmt.Sprintf("container %q restarted recently: %s", c.Name, t.Reason)
-	if t.Message != "" {
-		detail += " " + t.Message
-	}
-	return detail + fmt.Sprintf(" (exit code %d)", t.ExitCode)
+	return detail + fmt.Sprintf(" (exit code %d)", x.ExitCode)
 }
 
 func NumOfReadyPods(pods corev1.PodList) int {
--- a/pkg/reconciler/util.go
+++ b/pkg/reconciler/util.go
@@ -19,6 +19,7 @@ package reconciler
 import (
 	"fmt"
 	"slices"
+	"strings"
 	"time"
 
 	appv1 "k8s.io/api/apps/v1"
@@ -34,8 +35,8 @@ func CheckPodsStatus(pods *corev1.PodList) (healthy bool, reason string, message
 		return true, "NoPodsFound", "No Pods found", false
 	} else {
 		for _, pod := range pods.Items {
-			if podHealthy, msg, transient := isPodHealthy(&pod); !podHealthy {
-				return false, "Pod" + msg, fmt.Sprintf("Pod %s is unhealthy", pod.Name), transient
+			if podHealthy, podReason, details, transient := isPodHealthy(&pod); !podHealthy {
+				return false, "Pod" + podReason, fmt.Sprintf("Pod %s: %s", pod.Name, details), transient
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


pkg/reconciler/util.go
```diff
@@ -46,50 +47,103 @@ func CheckPodsStatus(pods *corev1.PodList) (healthy bool, reason string, message
...
+	}
+	return detail
+}
+
+// formatRecentRestart renders a container that recently OOM/Error-restarted but is no longer in a
+// failed state. The OOMKilled reason is inferred from exit code 137 when not explicitly set.
+func formatRecentRestart(c corev1.ContainerStatus, x *corev1.ContainerStateTerminated) string {
+	reason := x.Reason
+	if reason == "" {
+		if x.ExitCode == 137 {
```
juliev0: hmm, unfortunately it seems like 137 doesn't always mean OOM killed:

Exit code 137 means the process exited because it received signal 9 (SIGKILL). An OOM kill is one common reason a process gets SIGKILL, but it's not the only one.

juliev0: Does c.lastTerminationState not have any Message in it?

juliev0: I think it does:
`c.LastTerminationState.Terminated.Message` I think?

juliev0: Can we just use that?

pkg/reconciler/util.go
```diff
@@ -46,50 +47,103 @@ func CheckPodsStatus(pods *corev1.PodList) (healthy bool, reason string, message
...
+	if len(details) > 0 {
+		return false, firstReason, strings.Join(details, "; "), false
+	}
+	// Otherwise, a recent OOM/error restart on an otherwise-running container (within the last 2 mins).
 	if !lastRestartTime.IsZero() && lastRestartTime.Add(2*time.Minute).After(time.Now()) {
-		return false, "RecentRestart", true
+		var restartDetails []string
+		for _, c := range containers {
+			if x := c.LastTerminationState.Terminated; x != nil &&
+				(x.ExitCode == 137 || (x.ExitCode == 143 && x.Reason == "Error")) {
```
juliev0: I wonder if we need to check for these specific error codes? any reason you see why we shouldn't just do it for all error codes?

vaibhavtiwari33: I've kind of basically re-used what checks we already had. Let me see if we can use everything.

juliev0: I now see that the line Derek added above where it says "// TODO: revisit later if needed" is also only checking for these specific error codes.

Okay, in that case, I'm thinking that the fact that he's doing that check makes it so that it's impossible for `!lastRestartTime.IsZero()` to be true in the case of a different error code, and therefore we don't need `(x.ExitCode == 137 || (x.ExitCode == 143 && x.Reason == "Error")` down here - wdyt?

juliev0: If you're agreement on that, I'm okay if we don't address the [issue](https://github.com/numaproj/numaflow/issues/3482) you wrote now (up to you)

vaibhavtiwari33: > doing that check makes it so that it's impossible for !lastRestartTime.IsZero() to be true in the case of a different error code

This is correct. But that just tells us that there existed at least one container with the error code gated conditions. Since, I'm rebuilding the actual error message list separately, I need to again check the specified condition. 

But this does give me an idea that I should probably simplify this and build the `restartDetails` list within the same first error code gated conditional block in the default case. The difference would just be active vs lazy initialization of `restartDetails` but I feel that is fine.

pkg/reconciler/util.go
vaibhavtiwari33: Updated

vaibhavtiwari33: So, seems like the message can be empty in case of OOMKilled ([ref](https://github.com/kubernetes/kubernetes/issues/69676)) so I've reverted to using reason.

pkg/reconciler/util.go
juliev0: > But that just tells us that there existed at least one container with the error code gated conditions. Since, I'm rebuilding the actual error message list separately, I need to again check the specified condition.

Good call - I missed that!

juliev0: > But this does give me an idea that I should probably simplify this and build the restartDetails list within the same first error code gated conditional block in the default case. 

Good idea.

yhl25 (approved):

pkg/reconciler/util.go
juliev0: Discussed with @vaibhavtiwari33 that we can just format the string with `c.Message`, `c.ExitCode`, and the original `c.Reason` in the output

pkg/reconciler/util.go
```diff
@@ -46,50 +47,95 @@ func CheckPodsStatus(pods *corev1.PodList) (healthy bool, reason string, message
...
+		return true, "", "", false
 	}
 
 	// check both container and initContainer statuses
-	healthy, reason, isTransientUnhealthy = checkContainerStatuses(pod.Status.ContainerStatuses)
+	healthy, reason, message, isTransientUnhealthy = checkContainerStatuses(pod.Status.ContainerStatuses)
 	if !healthy {
-		return healthy, reason, isTransientUnhealthy
+		return healthy, reason, message, isTransientUnhealthy
 	}
```
juliev0, line 62: I guess if `containerStatuses` is unhealthy, we won't check `initContainerStatuses`. 
But maybe the only containerStatus in latest numaflow is the "numa" container? I guess the fact that we are checking `if x.ExitCode == 137 || (x.ExitCode == 143 && x.Reason == "Error") {` in `checkContainerStatuses(pod.Status.ContainerStatuses)` is saving us from prematurely returning here when numa container shows this:
```
    lastState:
      terminated:
        containerID: containerd://1257b3c0267b364c344893ec0e99c6fd6dee3e8c1dc3f30b8171b19db77c96ad
        exitCode: 1
        finishedAt: "2026-07-01T18:59:14Z"
        reason: Error
        startedAt: "2026-07-01T18:54:28Z"
```

So, this seems okay I think...

juliev0 (approved):
looks overall good to me. 

one question: you mentioned that when you tested image pull backoff scenario, the monovertex status looks like this:
```
  - lastTransitionTime: "2026-07-01T21:30:42Z"
    message: 'Pod simple-pipeline-cat-0-hwzzg: container "udf" ImagePullBackOff: Back-off
      pulling image "quay.io/numaio/numaflow-rs/map-cat-oom:stable": ErrImagePull:
      failed to pull and unpack image "quay.io/numaio/numaflow-rs/map-cat-oom:stable":
      failed to resolve reference "quay.io/numaio/numaflow-rs/map-cat-oom:stable":
      unexpected status from HEAD request to https://quay.io/v2/numaio/numaflow-rs/map-cat-oom/manifests/stable:
      401 UNAUTHORIZED'
    reason: PodImagePullBackOff
    status: "False"
    type: PodsHealthy
```

I'm curious: does it reliably stay in this state if you do a watch?

vaibhavtiwari33:
@juliev0 
[Here](https://gist.github.com/vaibhavtiwari33/6fe872871bf0609e159492c161bda77b) is the gist which captured 2 continuous statuses of udf vertex when there was a image pull back error. 

We can see that we got the same message in both statuses. Let me know if I should capture the statuses longer

PR #3484, opened 2026-06-29
fix(isbsvc): use quorum-aware health threshold for JetStream

Koalk:

ISBSvc health reporting was comparing ready pod count against the full replica count, so a 3-replica JetStream cluster would be marked unhealthy whenever a single pod was temporarily unavailable (e.g. during a GKE rolling node upgrade). NATS JetStream only needs ⌊replicas/2⌋+1 pods for write quorum, so this caused spurious `ReconcileFixedResourcesFailed` conditions on all dependent pipelines. This PR makes the health threshold quorum-aware: a cluster is healthy as long as enough pods are up to maintain quorum. Single-replica ISBSvc instances retain all-or-nothing semantics since their quorum evaluates to 1.

### Related issues

Fixes #

### Testing

- Unit tests added for `CheckStatefulSetStatus` covering 3-replica quorum boundary (2/3 healthy, 1/3 unhealthy) and 1-replica all-or-nothing semantics.
- Integration tests added for `CheckChildrenResourceStatus` on the `jetStreamInstaller` verifying the full reconciler path marks the ISBSvc healthy at quorum and unhealthy below quorum.

`CheckStatefulSetStatus` now takes a `minReadyReplicas int32` parameter. The existing call sites that needed all-or-nothing semantics (passing the full replica count) retain that behaviour unchanged. Only the JetStream ISBSvc installer passes a quorum value. If other StatefulSet-backed components warrant quorum-aware checks in future, the same parameter is available.

### DIFF
--- a/pkg/reconciler/isbsvc/installer/jetstream.go
+++ b/pkg/reconciler/isbsvc/installer/jetstream.go
@@ -557,6 +557,10 @@ func (r *jetStreamInstaller) getPVCs(ctx context.Context) ([]corev1.PersistentVo
 	return pvcl.Items, nil
 }
 
+// CheckChildrenResourceStatus evaluates the JetStream StatefulSet and pod health.
+// Returns (needsRequeue, error): needsRequeue is true when the failure is transient
+// (a pod had a recent OOM or error restart) and the controller must explicitly requeue,
+// since no Kubernetes event fires when the 2-minute transient window expires.
 func (r *jetStreamInstaller) CheckChildrenResourceStatus(ctx context.Context) (bool, error) {
 	var isbStatefulSet appv1.StatefulSet
 	if err := r.client.Get(ctx, client.ObjectKey{
--- a/pkg/reconciler/isbsvc/installer/jetstream.go
+++ b/pkg/reconciler/isbsvc/installer/jetstream.go
@@ -575,11 +575,29 @@ func (r *jetStreamInstaller) CheckChildrenResourceStatus(ctx context.Context) er
 	// the ISBSvc unhealthy when the cluster can still serve writes (e.g. during a rolling node upgrade).
 	// Single-replica ISBSvc (replicas==1) keeps all-or-nothing semantics: quorum evaluates to 1.
 	quorum := int32(r.isbSvc.Spec.JetStream.GetReplicas()/2 + 1)
-	if status, reason, msg := reconciler.CheckStatefulSetStatus(&isbStatefulSet, quorum); status {
-		r.isbSvc.Status.MarkChildrenResourceHealthy(reason, msg)
-	} else {
+	if status, reason, msg := reconciler.CheckStatefulSetStatus(&isbStatefulSet, quorum); !status {
 		r.isbSvc.Status.MarkChildrenResourceUnHealthy(reason, msg)
+		return nil
 	}
+	// Quorum is met, but individual pods may be in a persistent failure state
+	// (CrashLoopBackOff, ImagePullBackOff, OOMKill) that will never self-resolve.
+	// Terminating pods from a node eviction are skipped by CheckPodsStatus, so this
+	// check distinguishes transient unavailability from permanent pod failures.
+	// Limitation: a pod stuck in Pending (e.g. unschedulable) is not detected here;
+	// a follow-up time-based escalation would be needed to catch that case.
+	podList := &corev1.PodList{}
+	if err := r.client.List(ctx, podList, &client.ListOptions{
+		Namespace:     r.isbSvc.Namespace,
+		LabelSelector: labels.SelectorFromSet(r.labels),
+	}); err != nil {
+		r.isbSvc.Status.MarkChildrenResourceUnHealthy("ListPodsFailed", err.Error())
+		return err
+	}
+	if healthy, reason, msg, _ := reconciler.CheckPodsStatus(podList); !healthy {
+		r.isbSvc.Status.MarkChildrenResourceUnHealthy(reason, msg)
+		return nil
+	}
+	r.isbSvc.Status.MarkChildrenResourceHealthy("Healthy", "JetStream cluster is healthy")
 	return nil
 }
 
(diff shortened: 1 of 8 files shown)


pkg/reconciler/util.go
```diff
@@ -173,17 +173,18 @@ func getDeploymentCondition(status appv1.DeploymentStatus, condType appv1.Deploy
...
 	if sts.Status.ObservedGeneration == 0 || sts.Generation > sts.Status.ObservedGeneration {
 		return false, "Progressing", "Waiting for statefulset spec update to be observed..."
 	}
 	if sts.Status.UpdateRevision != sts.Status.CurrentRevision {
 		return false, "Progressing", fmt.Sprintf("waiting for statefulset rolling update to complete %d pods at revision %s...",
 			sts.Status.UpdatedReplicas, sts.Status.UpdateRevision)
 	}
-	if sts.Spec.Replicas != nil && sts.Status.ReadyReplicas < *sts.Spec.Replicas {
-		return false, "Unavailable", fmt.Sprintf("Waiting for %d pods to be ready...\n", *sts.Spec.Replicas-sts.Status.ReadyReplicas)
+	if sts.Spec.Replicas != nil && sts.Status.ReadyReplicas < minReadyReplicas {
```
vigith: `sts.Spec.Replicas != nil` is not use any more, should we need this guard?

vigith (commented):
it looks good otherwise

whynowy:
Thanks for raising this PR. This is a good improvement for the case of node rotation, but meanwhile, for the scenario of a node not being able to start for whatever reason, it will lose the awareness. It would be better to have the ability to detect that scenario alone with this change.

Koalk:
Updated with two additional changes based on the review:

**1. Removed the `sts.Spec.Replicas != nil` guard** (as @vigith noted — it was there to prevent a nil dereference of `*sts.Spec.Replicas`, but since we no longer dereference that field, the guard was dead logic).

**2. Added pod-level health check after the quorum check passes** — if `ReadyReplicas >= quorum` but any pod is in `CrashLoopBackOff` or `ImagePullBackOff`, the ISBSvc is still marked unhealthy. This addresses @whynowy's concern about losing visibility into persistent failures.

The key distinction this relies on: pods being evicted during a node rotation have `DeletionTimestamp` set (they are "Terminating") and are skipped by the existing `CheckPodsStatus` logic. Pods in a persistent failure state (`CrashLoopBackOff` etc.) are `Waiting` with a non-nil error reason and are flagged.

---

**Known limitation of this approach:** a pod stuck in `Pending` indefinitely (e.g. unschedulable due to resource pressure or a taint/toleration mismatch) is *not* detected here. `Pending` pods have no container statuses yet, so `CheckPodsStatus` treats them as healthy. The quorum count would still be met if 2 of 3 pods are running, and the stuck pod would silently go unnoticed.

To catch that case properly, a follow-up would be needed — likely a time-based escalation: if `ReadyReplicas >= quorum` but `ReadyReplicas < desired` for longer than a configurable threshold (e.g. 5 minutes), escalate to unhealthy. This would require either a new timestamp field in the ISBSvc status or an annotation to track when the cluster first entered the degraded-but-quorum state.

vigith (approved):

pkg/reconciler/isbsvc/installer/jetstream.go
```diff
@@ -571,12 +571,33 @@ func (r *jetStreamInstaller) CheckChildrenResourceStatus(ctx context.Context) er
...
+	// a follow-up time-based escalation would be needed to catch that case.
+	podList := &corev1.PodList{}
+	if err := r.client.List(ctx, podList, &client.ListOptions{
+		Namespace:     r.isbSvc.Namespace,
+		LabelSelector: labels.SelectorFromSet(r.labels),
+	}); err != nil {
+		r.isbSvc.Status.MarkChildrenResourceUnHealthy("ListPodsFailed", err.Error())
+		return err
+	}
+	if healthy, reason, msg, _ := reconciler.CheckPodsStatus(podList); !healthy {
```
vigith: shouldn't we have to handle `transientUnhealthy` the last return val for `CheckPodsStatus`?

Koalk: that makes sense, I'll amend on this PR

vigith: thank you @Koalk; LGTM

Koalk:
Do you want me to open an Issue with the problem this is trying to solve and stuff or are you happy to merge it as is?

vigith:
@whynowy any concerns. I am good with this PR.

pkg/reconciler/isbsvc/installer/jetstream.go
```diff
@@ -566,18 +566,41 @@ func (r *jetStreamInstaller) CheckChildrenResourceStatus(ctx context.Context) er
...
+	if status, reason, msg := reconciler.CheckStatefulSetStatus(&isbStatefulSet, quorum); !status {
 		r.isbSvc.Status.MarkChildrenResourceUnHealthy(reason, msg)
+		return false, nil
+	}
+	// Quorum is met, but individual pods may be in a persistent failure state
+	// (CrashLoopBackOff, ImagePullBackOff, OOMKill) that will never self-resolve.
+	// Terminating pods from a node eviction are skipped by CheckPodsStatus, so this
+	// check distinguishes transient unavailability from permanent pod failures.
+	// Limitation: a pod stuck in Pending (e.g. unschedulable) is not detected here;
+	// a follow-up time-based escalation would be needed to catch that case.
```
whynowy, line 591: +1

pkg/reconciler/isbsvc/installer/jetstream.go
```diff
@@ -557,7 +557,7 @@ func (r *jetStreamInstaller) getPVCs(ctx context.Context) ([]corev1.PersistentVo
 	return pvcl.Items, nil
 }
 
-func (r *jetStreamInstaller) CheckChildrenResourceStatus(ctx context.Context) error {
+func (r *jetStreamInstaller) CheckChildrenResourceStatus(ctx context.Context) (bool, error) {
```
whynowy, line 564: Doc to tell what the `bool` return means would be helpful.

whynowy (approved):

yhl25 (approved):

Issue #3371, opened 2026-04-15
e2e tests to cover UDF failure scenarios


yhl25:
## Parent Issue
Part of #3367 — Remove monitor container and move functionality to numa container

## Overview

Add end-to-end tests that exercise UDF crash/failure paths and verify that errors are correctly captured by the numa container, propagated through the daemon server, and visible via the UI server.

## Test Scenarios

- UDF container crashes mid-processing — verify numa container stays up and error is recorded
- UDF container restarts multiple times — verify errors accumulate and are queryable
- Daemon server returns correct critical errors after a UDF failure
- UI server displays the errors surfaced by the daemon server

## Acceptance Criteria

- [ ] e2e test for UDF crash with numa container remaining healthy
- [ ] e2e test verifying errors are retrievable via daemon server API
- [ ] e2e test verifying errors are visible in the UI
- [ ] Tests are integrated into the CI pipeline

Issue #3367, opened 2026-04-15
Remove monitor container and move functionality to numa container


yhl25:
## Overview

Remove the dedicated monitor container from Numaflow and consolidate its functionality into the numa container. This simplifies the architecture by reducing the number of sidecar containers while maintaining all monitoring and error-reporting capabilities.

## Motivation

The monitor container currently runs as a sidecar alongside UDF containers to handle error reporting and monitoring. Moving this responsibility to the numa container reduces operational overhead, simplifies the pod spec, and consolidates control-plane concerns in one place.

## Tasks

- [ ] #3368 — Numa container should not restart on UDF failures/crashes
- [ ] #3369 — Move monitor container's error endpoint to numa container
- [ ] #3370 — Daemon server should be able to return critical errors for the UI server
- [ ] #3371 — e2e tests to cover UDF failure scenarios

## Acceptance Criteria

- Monitor container is removed from all vertex pod specs
- Numa container exposes the error endpoint previously owned by the monitor container
- UDF failures do not cause the numa container to restart
- Daemon server API returns critical errors consumable by the UI
- e2e tests pass for UDF failure scenarios

adarsh0728:
Alternative approach: Rebuild the forwarder instead of preserving in flight requests for reconnecting cc @yhl25

PR #3488, opened 2026-06-30
feat: reconnect on map(udf) failures

adarsh0728:

- Adds reconnect/redrive handling for map UDF failures so `numa` does not exit
- Record map UDF failures through `critical_error_total{reason="map_runtime_error"}` 
- Persists runtime errors with `udf` as the fallback container name when the SDK error does not include one
- Removed unit tests that attempted to model map panic recovery with in process SDK servers, since actual panic recovery requires Kubernetes to restart the UDF sidecar. Will have e2e tests as part of https://github.com/numaproj/numaflow/issues/3371

Main changes:
- Adds reconnect config support for map UDF clients.
- Handles unary map send/stream failures through reconnect and redrive.
- Handles batch map send failures, EOT send failures, and partial EOT as redrivable errors.
- Handles stream map request/response stream failures through reconnect and message redrive.
- Adds common map redrive helpers for metrics, runtime error persistence, and reconnect client creation.

### Related issues

Part of #3367 #3368

### Testing

- Kubernetes smoke tests for map reconnect behavior across Go, Java, Python, and Rust SDK.
- Verified unary, batch, and stream map failure scenarios including panic, stream/transport close, batch partial EOT, and stream zero-output EOT.
- Verified `numa` restart count stayed `0` during recoverable UDF failures.
- Verified `critical_error_total{reason="map_runtime_error"}` increments for failure scenarios.
- Verified stable and stream zero output EOT scenarios complete without `map_runtime_error`.
- **Gap**: stream map missing EOT with an otherwise open stream is documented [here](https://github.com/numaproj/numaflow/compare/main...stream-map-missing-eot-fix) and was not included in the reconnect smoke pass.


rust/numaflow-core/src/mapper/map/stream.rs
```diff
@@ -76,92 +79,110 @@ impl MapStreamTask {
...
+                                };
+
+                            self.shared_ctx
+                                .output_tx
+                                .send(msg_handle)
+                                .await
+                                .expect("failed to send response");
+                        }
+                    }
+                    Some(Err(Error::UdfRedrive(status))) => {
```
vigith, line 159: what will happen if some data has been forwarded and we redrive? are we ok with duplicates?

yhl25:
Can you please run a long running validation and endurance test with this change and share the results?

rust/numaflow-core/src/mapper/map/stream.rs
adarsh0728: i think yes, because earlier as well if udf fails before stream closes(and some output already sent downstream)--> we would nack --> numa restarts and redelivers the message, hence duplicate downstream outputs

adarsh0728:
> Can you please run a long running validation and endurance test with this change and share the results?

yes i was planning to run once the feature completes (will have one more PR for [built in sources](https://github.com/numaproj/numaflow/issues/3526) like kafka)

adarsh0728:
> > Can you please run a long running validation and endurance test with this change and share the results?
> 
> yes i was planning to run once the feature completes (will have one more PR for [built in sources](https://github.com/numaproj/numaflow/issues/3526) like kafka)



> Can you please run a long running validation and endurance test with this change and share the results?

Ran a long running validation with chaos. It passed around ~47 iterations and failed only in `full_chaos` during drain, not payload validation.

The failure was `source_acked=1,000,000` but sink stalled at `1,553,865 / 1,583,342`. Missing rows were proportional across unary/batch/stream map, so it didn’t point to one map reconnect path.

Logs showed ISB recovery issues after ISB force delete. No correlation to this PR that map reconnect ACKed fanout early or caused data loss.

The behaviour looks like the pipeline stopped making progress because ISB was unavailable under the hardest chaos mode, and the harness failed after 5 minutes of no sink progress (this can be increased?). I think full chaos ISB recovery can be tracked separately.

vigith (approved):
approving as discussed this ISB stalling is not related to this PR.

rust/numaflow-core/src/mapper/map/batch.rs
```diff
@@ -319,9 +359,31 @@ impl UserDefinedBatchMap {
...
+    ) -> Vec<Result<BatchMapResponse>> {
+        let results = self.batch_once(requests.clone(), cln_token.clone()).await;
+        if results
+            .iter()
+            .any(|result| matches!(result, Err(Error::UdfRedrive(_))))
+        {
+            if let Err(e) = self.reconnect().await {
+                return vec![Err(e)];
+            }
+            return self.batch_once(requests, cln_token).await;
```
vaibhavtiwari33, line 371: Just for my understanding, we retry twice before escalating the redrive request back to the callee. 

nit: Could we document this as a doc comment, as to why twice specifically?

vaibhavtiwari33 (approved):

PR #3502, opened 2026-07-03
fix: return detailed vertex status message when unhealthy

### LINKED ISSUE #3496: Rollup unhealthy vertex reason to pipeline
# Summary

In case of an unhealthy vertex, I see the actual message/reason for that vertex being unhealthy in its status:
```yaml
- lastTransitionTime: "2026-07-01T22:13:35Z"
    message: 'Pod simple-pipeline-cat-0-ko4q9: container "numa" Error'
    reason: PodError
    status: "False"
    type: PodsHealthy
```

But I don't see the same/similar detail in pipeline status:
```yaml
- lastTransitionTime: "2026-07-01T22:13:48Z"
    message: 'Some Vertices are unhealthy: Vertex "cat" is not healthy'
    reason: Unavailable
    status: "False"
    type: VerticesHealthy
```

# Use Cases

Provides visibility into exact issue at pipeline level

---

AyushSrivastava1818:

When a vertex is unhealthy, `CheckVertexStatus` currently returns a generic message:

> `Vertex "<name>" is not healthy`

This hides the more detailed diagnostic information already available in `vertex.Status.Message`.

This PR updates `CheckVertexStatus` to return `vertex.Status.Message` when it is available, while preserving the existing generic fallback message when no detailed status message exists.

Additionally, a unit test has been added to verify the new behavior without changing the existing fallback behavior.

### Related issues

Fixes #3496

### Testing

- Added a unit test covering the detailed status message behavior.
- Existing fallback behavior remains covered by the current tests.
- Ran:

```bash
gofmt -w pkg/reconciler/util.go pkg/reconciler/util_test.go
go test ./pkg/reconciler/...

### DIFF
--- a/pkg/reconciler/util.go
+++ b/pkg/reconciler/util.go
@@ -172,6 +172,9 @@ func CheckVertexStatus(vertices *dfv1.VertexList) (healthy bool, reason string, 
 			return false, "Progressing", `Vertex "` + vertex.Spec.Name + `" Waiting for reconciliation`
 		}
 		if !vertex.Status.IsHealthy() {
+			if vertex.Status.Message != "" {
+				return false, "Unavailable", `Vertex "` + vertex.Spec.Name + `" error: ` + vertex.Status.Message
+			}
 			return false, "Unavailable", `Vertex "` + vertex.Spec.Name + `" is not healthy`
 		}
 	}
--- a/pkg/reconciler/util_test.go
+++ b/pkg/reconciler/util_test.go
@@ -457,6 +457,40 @@ func TestGetVertexStatus(t *testing.T) {
 		assert.Equal(t, "Unavailable", reason)
 		assert.Equal(t, `Vertex "test-vertex" is not healthy`, message)
 	})
+	t.Run("Test Vertex status returns detailed message", func(t *testing.T) {
+		vertices := dfv1.VertexList{
+			Items: []dfv1.Vertex{
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Generation: 2,
+					},
+					Spec: dfv1.VertexSpec{
+						AbstractVertex: dfv1.AbstractVertex{
+							Name: "test-vertex",
+						},
+					},
+					Status: dfv1.VertexStatus{
+						Phase:              "Pending",
+						ObservedGeneration: 2,
+						Message:            "failed to connect to source",
+					},
+				},
+			},
+		}
+
+		vertices.Items[0].Status.Conditions = []metav1.Condition{
+			{
+				Type:   string(dfv1.VertexConditionPodsHealthy),
+				Status: metav1.ConditionTrue,
+			},
+		}
+
+		status, reason, message := CheckVertexStatus(&vertices)
+
+		assert.False(t, status)
+		assert.Equal(t, "Unavailable", reason)
+		assert.Equal(t, `Vertex "test-vertex" error: failed to connect to source`, message)
+	})
 }
 
 var (

### CODE REVIEW

**Comment by adarsh0728** [comment]
can you fix DCO check? (signed commit)

File: pkg/reconciler/util.go [hunk 1]
```diff
@@ -172,6 +172,9 @@ func CheckVertexStatus(vertices *dfv1.VertexList) (healthy bool, reason string,
 			return false, "Progressing", `Vertex "` + vertex.Spec.Name + `" Waiting for reconciliation`
 		}
 		if !vertex.Status.IsHealthy() {
+			if vertex.Status.Message != "" {
+				return false, "Unavailable", vertex.Status.Message
```
vaibhavtiwari33:
```suggestion
				return false, "Unavailable", `Vertex "` + vertex.Spec.Name + `" error: ` + vertex.Status.Message
```

vaibhavtiwari33:
Can you please also provide example of how this message appears in status from a running pipeline that is failing.

AyushSrivastava1818:
Thanks for the review!

I've updated the returned message to include the vertex name and pushed the changes.

Regarding the example, I don't currently have a failing pipeline available to capture an actual `vertex.Status.Message`. The intent of this change is to surface the existing `vertex.Status.Message` from the vertex status instead of returning only the generic `Vertex "<name>" is not healthy` message.

If needed, I can try to reproduce a failing pipeline and share an actual status output.

vaibhavtiwari33:
> If needed, I can try to reproduce a failing pipeline and share an actual status output.

Yes pls

AyushSrivastava1818:
Thanks! I'm setting up a local Numaflow environment to reproduce a failing pipeline and capture an actual `vertex.Status.Message`. I don't have a Kubernetes cluster configured on this machine yet, so it will take me a bit of time to set it up.

I'll update this PR with a real status output as soon as I have it. Thanks for your patience!

AyushSrivastava1818:
Thanks! I reproduced a failing pipeline locally to verify how the message appears in a real scenario.

To reproduce, I modified the example pipeline to use an invalid UDF image:

```yaml
image: quay.io/numaio/numaflow-go/map-cat:this-tag-does-not-exist
```

This caused the `cat` vertex to become unhealthy with an `ImagePullBackOff`. The vertex status contained the following message:

```text
Status.Message:
Degraded: Pod simple-pipeline-cat-0-weclk: container "udf" ImagePullBackOff: Back-off pulling image "quay.io/numaio/numaflow-go/map-cat:this-tag-does-not-exist": ErrImagePull: rpc error: code = NotFound desc = failed to pull and unpack image "quay.io/numaio/numaflow-go/map-cat:this-tag-does-not-exist": failed to resolve reference "quay.io/numaio/numaflow-go/map-cat:this-tag-does-not-exist": not found
```

Without this change, `CheckVertexStatus` returns only:

```text
Vertex "cat" is not healthy
```

With this PR, it returns:

```text
Vertex "cat" error: Degraded: Pod simple-pipeline-cat-0-weclk: container "udf" ImagePullBackOff: Back-off pulling image "...": ErrImagePull: ...
```

which preserves the detailed diagnostic information already available in `vertex.Status.Message` and makes it much easier to identify the underlying issue without having to inspect the `Vertex` resource separately.

I've also attached a screenshot of the reproduced failure showing the vertex status message.
Screenshot 2026-07-04 144645

vaibhavtiwari33:
Thank you! 

cc: @juliev0

juliev0:
awesome! I'll take a look as well

pkg/reconciler/util.go
```diff
@@ -172,6 +172,9 @@ func CheckVertexStatus(vertices *dfv1.VertexList) (healthy bool, reason string,
 			return false, "Progressing", `Vertex "` + vertex.Spec.Name + `" Waiting for reconciliation`
 		}
 		if !vertex.Status.IsHealthy() {
+			if vertex.Status.Message != "" {
+				return false, "Unavailable", `Vertex "` + vertex.Spec.Name + `" error: ` + vertex.Status.Message
+			}
```
juliev0, line 177: This looks good. I confirmed that in [this](https://github.com/numaproj/numaflow/pull/3481/changes) PR the Vertex `Condition.Message` field is the one which has all the details that we need (i.e. underlying container.Reason, container.Message, container.ExitCode), so surfacing that up here seems to include everything.

juliev0: Oh, but I guess you are using Vertex.Status.Message, actually, not the Condition's `Message` field. I assume the Status.Message comes from the Condition.Message but I'll check that as well just in case....

juliev0: I see - that's [here](https://github.com/numaproj/numaflow/blob/main/pkg/apis/numaflow/v1alpha1/vertex_types.go#L863-L867)

juliev0 (approved):

PR #3503, opened 2026-07-03
feat: show mvtx udf metrics on UI

adarsh0728:

- Updates the UI metrics tab to display the new Monovertex UDF metrics
- Filters metrics shown in the UI based on configured containers (udf metrics only if udf configured)
- Filters invalid Prometheus values like "NaN" before chart rendering, so the graph and tooltip do not show NaN samples
- Fixed udf_error metric name for mvtx 
  - Pipeline vertex/forwarder: `forwarder_udf_error_total`
  - MonoVertex: `monovtx_udf_error_total` 
Screenshot 2026-07-03 at 4 23 56 PM


config/base/numaflow-server/numaflow-server-metrics-proxy-config.yaml
```diff
@@ -340,6 +352,30 @@ data:
               filters:
                 - name: pod
                   required: false
+        - metric_name: monovtx_udf_error_total
```
suryapratap-01, line 355: In Rust file `rust/numaflow-core/src/metrics/mod.rs`, line 89, the metric is named `udf_error` and is registered under the udf prefix, so the emitted metric becomes `monovtx_udf_udf_error_total`. But the config change uses `monovtx_udf_error_total`, so the UI/metrics proxy may query a metric that Rust never emits and fail to discover the UDF error metric.

yhl25 (approved):

vigith (approved):

suryapratap-01 (approved):

config/base/numaflow-server/numaflow-server-metrics-proxy-config.yaml
adarsh0728: let me check

adarsh0728: image

PR #3505, opened 2026-07-03
fix(UI): dynamic monovertex dimension for cleaner UI

### LINKED ISSUE #3468: Use bigger dimensions in monovertex card only in case of bypass
can you check if we use these bigger dimensions only in case of bypass, otherwise not.

_Originally posted by @adarsh0728 in https://github.com/numaproj/numaflow/pull/3438#discussion_r3258180693_

suryapratap-01:

MonoVertex cards in the UI graph were always shown at the same large size, even for simple setups like source → sink. That made the graph feel unnecessarily big.

This PR makes the card size depend on what's actually in the MonoVertex, a simple layout gets a smaller card, and it only grows when needed (e.g. extra stages, optional outputs, or bypass paths).

### Related issues

Fixes #3468

### Testing
```bash
yarn test
```

- Unit tests added and updated for the new sizing logic.
- All related UI tests pass (32 tests).

### Notes for reviewers

- Larger dimensions are only used when the spec needs them — not for every MonoVertex.
- Small visual tweaks to internal icons and bypass edges so things stay readable at the smaller sizes.
Screenshot 2026-07-03 at 10 05 30 PM
Screenshot 2026-07-03 at 10 05 45 PM
Screenshot 2026-07-03 at 10 05 51 PM

### DIFF
--- a/ui/src/components/pages/Pipeline/partials/Graph/index.tsx
+++ b/ui/src/components/pages/Pipeline/partials/Graph/index.tsx
@@ -85,12 +85,17 @@ const nodeWidth = 252;
 const nodeHeight = 72;
 const nodeHeightTall = 112;
 const graphDirection = "LR";
+const NODE_DATA_TYPES = {
+  MONO_VERTEX: "monoVertex",
+  MONO_VERTEX_INTERNAL: "monoVertexInternal",
+} as const;
 
 const getNodeLayoutWidth = (node: Node): number => {
   const d = node?.data as Record<string, any> | undefined;
-  if (d?.type === "monoVertex")
+  if (d?.type === NODE_DATA_TYPES.MONO_VERTEX)
     return d.containerWidth ?? MONO_VERTEX_MAX_CONTAINER_WIDTH;
-  if (d?.type === "monoVertexInternal") return MONO_VERTEX_INTERNAL_NODE_WIDTH;
+  if (d?.type === NODE_DATA_TYPES.MONO_VERTEX_INTERNAL)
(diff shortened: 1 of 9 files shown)


ui/src/components/pages/Pipeline/partials/Graph/partials/CustomNode/index.tsx
```diff
@@ -521,7 +535,7 @@ const CustomNode: FC<NodeProps> = ({
       <Box
         className={nodeInputClass}
         onClick={handleClick}
-        style={nodeStyle}
+        style={monoVertexNodeStyle}
```
adarsh0728: rename this to a more generic variable name

ui/src/components/pages/Pipeline/partials/Graph/index.tsx
```diff
@@ -79,23 +84,23 @@ import "./style.css";
...
 const nodeHeightTall = 112;
-const monoVertexNodeWidth = 420;
-const monoVertexNodeHeight = 170;
 const graphDirection = "LR";
 
 const getNodeLayoutWidth = (node: Node): number => {
   const d = node?.data as Record<string, any> | undefined;
-  if (d?.type === "monoVertex") return monoVertexNodeWidth;
-  if (d?.type === "monoVertexInternal") return 32;
+  if (d?.type === "monoVertex")
```
adarsh0728: make node data type list of constants and then compare

adarsh0728 (approved):

PR #3522, opened 2026-07-09
feat: Numaflow Client CLI

### LINKED ISSUE #3520: Proposal: `nfcli` - a CLI to test UDFs without deploying a pipeline
## Problem

Today, seeing a UDF behave end-to-end means building an image, deploying a pipeline, and feeding it data - or hand-constructing `Datum` objects in unit tests.
Related issues:
- [#833 - Easier way to test a Python UDF locally](https://github.com/numaproj/numaflow/issues/833) 
- [#690 - Doc: Add gRPC contract needed to run custom image](https://github.com/numaproj/numaflow/issues/690) - complementary, can act like an executable contract.

## Idea

`nfcli` is a standalone, non-interactive CLI that impersonates the numa container and talks directly to a UDF gRPC server. It connects over a unix domain socket (as numa does in k8s) or TCP (as the Java SDK's local mode does), drives the full protocol choreography - handshake, batching, EOT, window open/close - prints the UDF's responses, and exits with a meaningful code.

Key design decisions:

- **One subcommand per UDF type, all types covered**: `map` (`--mode unary|batch|stream`), `transform`, `reduce` (fixed & sliding, works unchanged for reduce streamers), `session-reduce`, `accumulator`, `sink` (incl. fallback/on-success - same protocol, point at the right socket), `source`, `side-input`, plus `ready <type>` to poll `IsReady` in scripts.
- **The CLI emulates numa's side of windowing**, since numa (not the SDK) owns window assignment: fixed/sliding assignment by event time with one `ReduceFn` stream per window, session OPEN/EXPAND/MERGE/APPEND/CLOSE over a single multiplexed stream, per-key accumulator choreography. End-of-input is treated as the watermark advancing to +∞, so every still-open window closes and a bounded test run is deterministic and always terminates.
- **Scriptable**: `-o json` emits JSONL for golden-file assertions; exit codes distinguish usage error:
	- connect/ready/handshake failure
	- protocol error
	- "protocol OK but the UDF reported failures", e.g. sink `FAILURE`

## Usage

```bash
# minimum options. `--payload` is required, all other message fields are optional
nfcli map --tcp 50051 --payload '{"temp_c": 21.5}'

# Quick one-shot test against a Java UDF in local mode (TCP, default port)
nfcli map --tcp 50051 --payload '{"temp_c": 21.5}' --key sensor-1 --header source=test

# Many messages over UDS, 20 per batch, 500ms between batches
nfcli map --socket /var/run/numaflow/map.sock -f messages.yaml --batch-size 20 --delay 500ms

# Batch / stream map - --mode replaces server-info's MAP_MODE
nfcli map --mode batch --socket /var/run/numaflow/batchmap.sock -f messages.yaml

# Reduce: fixed 60s windows; sliding works too (--window sliding --slide 10s)
nfcli reduce --socket /var/run/numaflow/reduce.sock -f counts.yaml --window fixed --length 60s

# Session reduce with a 10s inactivity gap
nfcli session-reduce --socket /var/run/numaflow/sessionreduce.sock -f clicks.yaml --gap 10s

# Source: read 5 records twice, 1s apart, ack each round, show pending
nfcli source --socket /var/run/numaflow/source.sock --count 5 --rounds 2 --delay 1s --pending

# Sink - fallback/on-success sinks are the same subcommand at their socket
nfcli sink --socket /var/run/numaflow/fb-sink.sock --payload 'poison-pill' --id order-17

# Wait for a server to come up (CI scripts)
nfcli ready map --tcp 50051
```

Sample output (unary map):

```
✓ ready (uds:/var/run/numaflow/map.sock) · handshake ok
[msg-1] 2 results (3.1ms)
  1: keys=[sensor-1] tags=[metric] payload={"temp_f": 70.7}
  2: keys=[sensor-1] tags=[audit]  payload={"seen": true}
[msg-2] DROPPED (tags=[__DROP__])
──
sent=2 · results=2 · dropped=1 · failed=0 · elapsed=18ms
```

## Message file

Single messages go inline on the command line (`--payload`, `--key`, `--event-time`, …). Multi-message scenarios (reduce, sessions, batch map) use a YAML multi-document stream, because it is human-readable, `#` comments, natural multi-line payloads via block scalars, binary via base64, and a `repeat` field for volume:

```yaml
---
payload: |-
  {"user": "alice", "amount": 30,
   "note": "multi-line payloads are fine"}
keys: [alice]
headers: {source: checkout}
eventTime: "+1s"            # relative to --base-time
---
payload: '{"user": "bob", "amount": 25}'
keys: [bob]
eventTime: "+20s"
repeat: 2                   # sent twice, auto ids
---
payloadBase64: "3q2+7w=="   # binary payload
keys: [alice]
eventTime: "+70s"           # lands in the next 60s window
```

Fields per document: `payload`/`payloadBase64`/`payloadFile` (one required), `keys`, `headers`, `eventTime`, `watermark`, `id`, `userMetadata`, `previousVertex`, `repeat`.
- Unknown fields are a hard error.
- Relative times (`+dur`) anchor to `--base-time`, which for the reduce family defaults to the invocation time truncated to the window boundary - so `+0s .. +59s` predictably land in one 60s window.
- Watermark defaults to event time. 
- `-f -` reads from stdin, so generated input can be piped in.

## Status

I've a working implementation. The whole code is in a new directory `rust/numaflow-cli/`. Added as workspace member in `Cargo.toml` (only change in existing code).

## Possible enhancements
- **Assertions**: `--expect-*` flags or an expected-output YAML would turn `nfcli` runs into self-contained CI tests for UDFs; the JSONL output already enables golden-file testing.

---

BulkBeing:
Resolves #3520 

More changes will be made when I use this CLI for `pynumaflow-lite` development.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2522,29 +2522,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "numaflow"
-version = "0.6.0"
-source = "git+https://github.com/numaproj/numaflow-rs?rev=913cc7c8f14af6b93154719e0f68cd02df2912e8#913cc7c8f14af6b93154719e0f68cd02df2912e8"
-dependencies = [
- "chrono",
- "hyper-util",
- "prost 0.14.4",
- "prost-types 0.14.4",
- "serde",
- "serde_json",
- "thiserror 2.0.18",
- "tokio",
- "tokio-stream",
- "tokio-util",
- "tonic",
- "tonic-prost",
- "tonic-prost-build",
- "tonic-types",
- "tracing",
- "uuid",
-]
-
-[[package]]
 name = "numaflow-cli"
 version = "0.1.0"
 dependencies = [
@@ -2553,10 +2530,12 @@ dependencies = [
  "chrono",
  "clap",
  "clap_complete",
+ "futures",
  "humantime",
  "hyper-util",
- "numaflow 0.6.0",
+ "numaflow 0.5.0 (git+https://github.com/numaproj/numaflow-rs?rev=128dedad7ed3ccafc40ef23f9530a27be3ddfd0f)",
  "numaflow-pb",
+ "numaflow-udf-client",
  "prost 0.14.4",
  "prost-types 0.14.4",
  "serde",
@@ -2605,6 +2584,7 @@ dependencies = [
  "numaflow-sqs",
  "numaflow-testing",
  "numaflow-throttling",
+ "numaflow-udf-client",
  "opentelemetry",
  "opentelemetry_sdk",
  "parking_lot",
@@ -2874,6 +2854,25 @@ dependencies = [
  "tokio-util",
  "tracing",
  "trait-variant",
+]
+
+[[package]]
+name = "numaflow-udf-client"
+version = "0.1.0"
+dependencies = [
+ "bytes",
+ "chrono",
+ "hyper-util",
+ "numaflow 0.5.0 (git+https://github.com/numaproj/numaflow-rs?rev=128dedad7ed3ccafc40ef23f9530a27be3ddfd0f)",
+ "numaflow-pb",
+ "prost-types 0.14.4",
+ "tempfile",
+ "thiserror 2.0.18",
+ "tokio",
+ "tokio-stream",
+ "tokio-util",
+ "tonic",
+ "tower",
 ]
 
 [[package]]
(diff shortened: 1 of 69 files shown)


vigith (changes requested):
let's make sure we use numaflow core to manage message life cycle, we should not be reimplementing those.

vigith:
#3582

PR #3528, opened 2026-07-10
Feat/pulsar nack options

### LINKED ISSUE #3492: Implement supported nack options in pulsar
we need to handle nack options for pulsar, can come as a follow-up PR.

_Originally posted by @yhl25 in https://github.com/numaproj/numaflow/pull/3455#discussion_r3499957219_

AyushSrivastava1818:

Adds handling for unsupported per-message NACK options in the Pulsar source.

Currently, the Pulsar extension only supports `nack_with_id`, and the underlying Pulsar client does not expose per-message NACK options. This change detects when per-message `NackOptions` are provided, logs an error indicating that they are unsupported, and continues processing the NACK by ignoring those options.

### Related issues

Fixes #3492

### Testing

* Reviewed the implementation to ensure the error is logged only when any `NackOptions` are present in the NACK batch.
* Verified that the existing `nack_with_id` flow remains unchanged and that NACKs continue to be processed normally.
* Ran `cargo fmt`.

> Note: I could not complete `cargo check` locally because my environment is missing the MSVC linker (`link.exe`). The change is limited to `rust/numaflow-core/src/source/pulsar.rs`, and CI will validate the build.

* This change intentionally ignores unsupported per-message NACK options while logging an error, following the discussion in #3492.
* Consumer-level redelivery configuration is out of scope for this PR and can be implemented separately if needed.

### DIFF
--- a/pkg/reconciler/util.go
+++ b/pkg/reconciler/util.go
@@ -172,6 +172,9 @@ func CheckVertexStatus(vertices *dfv1.VertexList) (healthy bool, reason string, 
 			return false, "Progressing", `Vertex "` + vertex.Spec.Name + `" Waiting for reconciliation`
 		}
 		if !vertex.Status.IsHealthy() {
+			if vertex.Status.Message != "" {
+				return false, "Unavailable", `Vertex "` + vertex.Spec.Name + `" error: ` + vertex.Status.Message
+			}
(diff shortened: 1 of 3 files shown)


pkg/reconciler/util.go
vaibhavtiwari33, line 177: Can you merge the master changes since I believe you've already addressed these

PR #3529, opened 2026-07-10
feat(pulsar): log unsupported per-message nack options

### LINKED ISSUE #3492: Implement supported nack options in pulsar

AyushSrivastava1818:

This PR adds handling for unsupported per-message NACK options in the Pulsar source.

The Pulsar extension currently performs negative acknowledgements using `nack_with_id`, while the underlying Pulsar client does not support per-message NACK options. This change detects when per-message `NackOptions` are provided, logs an error indicating that they are unsupported, and continues processing the NACK by ignoring those options.

### Related issues

Fixes #3492

### Testing

- Verified that unsupported per-message NACK options are detected and logged.
- Verified that the existing `nack_with_id` flow remains unchanged and continues to process NACKs normally.
- Ran `cargo fmt`.

> Note: I was unable to complete `cargo check` locally because my Windows environment is missing the MSVC linker (`link.exe`). The change is limited to `rust/numaflow-core/src/source/pulsar.rs`, and CI will validate the build.

- This change intentionally logs an error and ignores unsupported per-message NACK options, following the maintainer's guidance in #3492.
- Consumer-level redelivery configuration is not included in this PR and can be addressed separately if needed.

### DIFF
--- a/rust/numaflow-core/src/source/pulsar.rs
+++ b/rust/numaflow-core/src/source/pulsar.rs
@@ -105,15 +105,26 @@ impl source::SourceAcker for PulsarSource {
 
     async fn nack(&mut self, offsets: Vec<NackOffset>) -> crate::error::Result<()> {
         let mut pulsar_offsets = Vec::with_capacity(offsets.len());
+        let mut logged = false;
+
         for offset in offsets {
+            if !logged && offset.option.is_some() {
+                tracing::error!(
+                    "Pulsar does not support per-message nack options; ignoring supplied options."
+                );
+                logged = true;
+            }
+
             let Offset::Int(int_offset) = offset.offset else {
                 return Err(Error::Source(format!(
                     "Expected Offset::Int type for Pulsar. offset={:?}",
                     offset.offset
                 )));
             };
+
             pulsar_offsets.push(int_offset.offset as u64);
         }
+
         self.nack_offsets(pulsar_offsets).await.map_err(Into::into)
     }
 }


vaibhavtiwari33:
It would be great to add support for configuring consumer level max redelivery and delay for pulsar.

vaibhavtiwari33 (approved):

rust/numaflow-core/src/source/pulsar.rs
```diff
@@ -105,15 +105,23 @@ impl source::SourceAcker for PulsarSource {
 
     async fn nack(&mut self, offsets: Vec<NackOffset>) -> crate::error::Result<()> {
         let mut pulsar_offsets = Vec::with_capacity(offsets.len());
+        if offsets.iter().any(|offset| offset.option.is_some()) {
+            tracing::error!(
+                "Pulsar does not support per-message nack options; ignoring supplied options."
+            );
+        }
```
vigith: why not move it below to the for loop? why double iterate?

vaibhavtiwari33 (changes requested):
https://github.com/numaproj/numaflow/pull/3529/changes#r3562052446

AyushSrivastava1818:
I used iter().any() to keep the logging separate from the processing, but yess now I can see that it causes a second iteration unnecessarily. I'll move the check into the existing loop and log only once.

rust/numaflow-core/src/source/pulsar.rs
AyushSrivastava1818: I used iter().any() to keep the logging separate from the processing, but yess now I can see that it causes a second iteration unnecessarily. I'll move the check into the existing loop and log only once.

AyushSrivastava1818:
I updated the implementation to perform the check within the existing loop and log only once, avoiding the extra iteration.

vaibhavtiwari33 (approved):

vigith (approved):

PR #3532, opened 2026-07-14
feat(pulsar): support consumer dead letter policy and redelivery delay

### LINKED ISSUE #3530: Add support for configuring consumer level max redelivery and delay for pulsar
It would be great to add support for configuring consumer level max redelivery and delay for pulsar.

_Originally posted by @vaibhavtiwari33 in https://github.com/numaproj/numaflow/issues/3529#issuecomment-4939218645_

AyushSrivastava1818:

This PR adds support for configuring **Pulsar consumer-level Dead Letter Policy** for Pulsar sources.

Specifically, it introduces:

- `deadLetterPolicy` to configure:
  - the dead letter topic
  - the maximum redelivery count before routing messages to the dead letter topic

The Pulsar consumer is configured using the underlying Rust Pulsar client's `with_dead_letter_policy(...)` support.

## Related issue

Fixes #3530

## Testing

Tested locally using:

```bash
make codegen

cd rust
cargo fmt --all --check
cargo check -p numaflow-pulsar
cargo test -p numaflow-pulsar
```

All commands completed successfully.

## Special notes for reviewers

This PR adds end-to-end support for Pulsar consumer dead letter policy across the project, including:

- Kubernetes API definitions
- Protobuf definitions
- OpenAPI specification
- CRDs
- API documentation
- Rust models
- Configuration conversion in `numaflow-core`
- Pulsar consumer configuration in the Rust extension


rust/extns/numaflow-pulsar/src/source.rs
```diff
@@ -97,13 +108,24 @@ impl ConsumerReaderActor {
...
+            .with_options(ConsumerOptions::default().durable(true));
+
+        if let Some(policy) = &config.dead_letter_policy {
+            builder = builder.with_dead_letter_policy(DeadLetterPolicy {
+                max_redeliver_count: policy.max_redelivery,
+                dead_letter_topic: policy.topic.clone(),
+            });
+        }
+
+        builder = builder.with_unacked_message_resend_delay(config.redelivery_delay);
```
vaibhavtiwari33: I don't think we want to use [this](https://github.com/streamnative/pulsar-rs/blob/95a0d73290d56be5d24eb3ed0d5d7d73f3a44343/src/consumer/builder.rs#L169-L176):
> /// The time after which a message is dropped without being acknowledged or nacked
    /// that the message is resent

Our aim with redelivery delay, as part of this change, is to configure sort of a redelivery backoff delay, i.e., once a message is nacked, it should be redelivered only after the backoff time. 

If exponential backoff is supported by the rust client, then we should incorporate that as well. Example for such an [implementation](https://github.com/apache/pulsar/blob/16476b503d11b1091a4bb3e1adebbee1abd8fe02/pulsar-client/src/main/java/org/apache/pulsar/client/impl/MultiplierRedeliveryBackoff.java#L27) is in the java client

vaibhavtiwari33 (changes requested):

vaibhavtiwari33:
Thank you for contributing! One small comment regarding clarification of our aim with `redelivery_delay` (feel free to rename this to `redelivery_backoff` if confusing)

AyushSrivastava1818:
Thanks for the clarification! I had interpreted redelivery_delay as configuring the unacked message resend delay, but I see the intended behavior is a backoff applied after a negative acknowledgement. I'll look into whether the Rust Pulsar client supports configurable negative-ack redelivery backoff (including exponential backoff similar to the Java client). If it isn't currently supported, I'll update the PR accordingly and we can discuss the best approach.

AyushSrivastava1818:
I looked into the Rust Pulsar client implementation. with_unacked_message_resend_delay only configures redelivery for messages that remain neither acknowledged nor negatively acknowledged (ACK timeout). For nack(), the client immediately invokes send_redeliver_unacknowleged_messages(...) and does not expose a configurable negative-ack redelivery backoff (or exponential backoff) similar to the Java client. Given that limitation, would you prefer that I remove the redelivery_delay configuration from this PR for the Rust implementation, or is there another approach you'd like me to take?

vaibhavtiwari33:
Thanks for looking into this. Since the pulsar rust client today doesn't support the config, we can skip redelivery backoff implementation. 

cc: @yhl25

AyushSrivastava1818:
I removed the redeliveryDelay configuration from the Rust implementation and API since the current Rust Pulsar client doesn't support configurable negative-ack redelivery backoff. The PR now only includes consumer dead letter policy support, with the generated models and API artifacts updated accordingly.

pkg/apis/numaflow/v1alpha1/pulsar_source.go
```diff
@@ -1,28 +1,26 @@
...
 	SubscriptionName string `json:"subscriptionName" protobuf:"bytes,4,name=subscriptionName"`
-	// Maximum number of messages that are in not yet acked state. Once this limit is crossed, futher read requests will return empty list.
+
+	// Maximum number of messages that are in not yet acked state.
 	MaxUnack uint32 `json:"maxUnack,omitempty" protobuf:"bytes,5,opt,name=maxUnack"`
+
+	// Consumer level dead letter policy.
+	// +optional
+	DeadLetterPolicy *PulsarDeadLetterPolicy `json:"deadLetterPolicy,omitempty" protobuf:"bytes,7,opt,name=deadLetterPolicy"`
+
```
vaibhavtiwari33: nit: Since this is the newest config to be added to the pulsar source, can we move it to the bottom? Just to keep the protobuf field numbering in order.

pkg/apis/numaflow/v1alpha1/pulsar_source.go
```diff
@@ -1,28 +1,26 @@
...
 
 type PulsarSource struct {
 	ServerAddr       string `json:"serverAddr" protobuf:"bytes,1,name=server_addr"`
 	Topic            string `json:"topic" protobuf:"bytes,2,name=topic"`
 	ConsumerName     string `json:"consumerName" protobuf:"bytes,3,name=consumerName"`
 	SubscriptionName string `json:"subscriptionName" protobuf:"bytes,4,name=subscriptionName"`
-	// Maximum number of messages that are in not yet acked state. Once this limit is crossed, futher read requests will return empty list.
+
+	// Maximum number of messages that are in not yet acked state.
 	MaxUnack uint32 `json:"maxUnack,omitempty" protobuf:"bytes,5,opt,name=maxUnack"`
```
vaibhavtiwari33, line 18: Why is the doc comment here modified? Has the functionality for this field been changed?

rust/numaflow-models/src/models/pulsar_source.rs
```diff
@@ -22,7 +22,9 @@ pub struct PulsarSource {
     pub auth: Option<Box<crate::models::PulsarAuth>>,
     #[serde(rename = "consumerName")]
     pub consumer_name: String,
-    /// Maximum number of messages that are in not yet acked state. Once this limit is crossed, futher read requests will return empty list.
```
vaibhavtiwari33, line 25: Same review comment, unrelated field's doc comment seems to have been modified.

AyushSrivastava1818:
Addressed both review comments. Restored the original MaxUnack documentation comment and moved DeadLetterPolicy below Auth while preserving the protobuf field numbering. Regenerated the generated artifacts and force-pushed the updated changes.

vaibhavtiwari33 (approved):

vaibhavtiwari33:
Please update the PR description to reflect the changes pushed post our further discussion.

PR #3542, opened 2026-07-16
chore: Rust 1.97, faster local image build

BulkBeing:
- Upgrade tool chain to Rust 1.97
- Use debian trixie (instead of bookworm) for dev image
- Removes the need for cargo-chef using buildkit cache mounts
- Added `make image-dev` that builds image in debug mode and other optimizations
- Workflow for building and pushing image with memory profiler (bytehound)

Image building after  a rust file change reduced from 9minute and 45 seconds to 15 seconds.

### DIFF
--- a/docs/user-guide/reference/memory-profiling.md
+++ b/docs/user-guide/reference/memory-profiling.md
@@ -3,62 +3,63 @@
 ## Numa container memory profiling
 
 This section explores attaching [bytehound](https://github.com/koute/bytehound) (heap profiler via `LD_PRELOAD`)
-mem profiler to the numaflow-core process in the **`numa` main container** and captures a `.dat` for analysis offline
+mem profiler to the numaflow-core process in the **`numa` main container** and captures a `.dat` for analysis offline.
 
-The stock numaflow image **cannot** be profiled as-is since the rust binary is built with `-C target-feature=+crt-static`
+The stock numaflow image **cannot** be profiled as-is since the Rust binary is built with `-C target-feature=+crt-static`
 (fully static -> no dynamic loader) and the release image is `FROM scratch` (no `ld.so`/libc).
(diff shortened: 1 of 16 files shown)


docs/user-guide/reference/memory-profiling.md
vigith, line 1: Can we also automatically memory profiler enabled image?

vigith:
> Removes the need for cargo-chef using buildkit cache mounts

why remove now? `buildkit` never had the cache. mount option earlier?

BulkBeing:
> > Removes the need for cargo-chef using buildkit cache mounts
> 
> why remove now? `buildkit` never had the cache. mount option earlier?

We were not caching Rust `target/` directory earlier. I added the last line:
```Dockerfile
RUN --mount=type=cache,target=/usr/local/cargo/registry \
    --mount=type=cache,target=/usr/local/cargo/git \
    --mount=type=cache,target=/numaflow/target \
```
With this, cargo-chef and its multi-stage build is not providing any benefits.

vigith:
memprofile is slowing down the entire build, perhaps we should enable only on `memprofile-*` branches?

BulkBeing:
> memprofile is slowing down the entire build, perhaps we should enable only on `memprofile-*` branches?

That was a test run https://github.com/numaproj/numaflow/actions/runs/29549416717/job/87792037841
Published `quay.io/numaproj/numaflow:faster-builds-1c0c45e-memprofile`.

The CI for memory profile will run only on manual trigger. Once this PR is merged, an option to trigger providing input branch will be available:
Screenshot 2026-07-17 at 8 37 31 AM
(Github UI shows manual trigger only for workflows in main branch)

docs/user-guide/reference/memory-profiling.md
vigith, line 1: Update this file?

vigith (approved):

vigith (approved):

Issue #3219, opened 2026-02-10
Bring your own different ISB implementation


magelisk:
# Summary

We have customers that use more classical message brokers like kafka, rabbitmq, redis etc. Using NATS jetstream is a bit of a turn-off for introducing them to numaflow as it's yet another tool to maintain, security teams aren't as familiar with it, and it's just "different". Having some ability to use a different messaging broker would help ease the introduction into numaflow for these customers. Additionally, it could open up using managed services in cloud environments. 

Many of these customers don't have the exact data-streaming use case that the NATS Jetstream ISB is optimized around. Simple message brokers ala SQS policies is the extent of the needs, so this may help with a features like this if certain rules or guarantees can be relaxed? 

I have no idea what implementation looks like to make it a 'plug-in' type system. But I think the three services that would probably have the most immediate benefit or interest from customers for us would be
1. Pulsar
2. Kafka. 
3. SQS  (We are well aware this is not a good streaming solution, but is useful for getting data processing from external system for longer running type jobs)

---

**Message from the maintainers**:

If you wish to see this enhancement implemented please add a 👍 reaction to this issue! We often sort issues this way to know what to prioritize.

PR #3551, opened 2026-07-20
feat: Generic ISB

BulkBeing:
Make ISB generic throughout the code base using trait objects.

- Introduces `ISBFactory` to create readers, writers, and ISB-owned KV stores across forwarders, watermarks, callbacks, and serving.
- Adds an in-memory ISB/KV backend for programmatic local/test use; moves the existing SimpleBuffer/KV test utilities into core/shared and adds in-memory adapter/factory plus keyed Reduce E2E coverage.
- Watermark idle detection now reuses forwarder-created writers to publish idle WMBs, avoiding duplicate producers.

### Related issues
- https://github.com/numaproj/numaflow/issues/3219
- https://github.com/numaproj/numaflow/issues/2343
- https://github.com/numaproj/numaflow/pull/3582
- 
### Testing
Ran a benchmark with `source -> map -> sink` pipeline. No change in throughput or resource usage compared to current main.
Also, ran 40 rounds of a fixed window reduce sending around 1million events per iteration, 60 second windows, with some processing logic. Data output is validated with the expected precomputed output.

This PR doesn't change anything functionally. The existing pipeline config will work as it is using jetstream ISB. Supporting additional ISB like redis/pulsar will need pipeline config changes, which can be done in a future PR.

- `rust/numaflow-core/src/pipeline/isb/dyn_adapter.rs` is the key abstraction boundary. `trait_variant` ISB traits are not object-safe because async methods return `impl Future`; the dyn twins box those calls while backends continue implementing the existing static traits via blanket impls. New ISB trait methods must be mirrored in both the dyn trait and forwarding impl.
- Idle WMB publishing now requires an integer offset; string offsets return a watermark error.
- `InMemory` is deliberately programmatic-only: the env/Kubernetes loader still accepts only JetStream. It is single-process, non-durable, and intended for tests/local use.

##### Why trait objects instead of generics
- Static dispatch has no material benefit here: ISB operations are network-bound, `async_write` already returns a boxed future, and reader acknowledgements already use spawned tasks and oneshot channels.
- It removes the `NumaflowTypeConfig` / `F: ISBFactory` generic axis, associated reader/writer types, and generic propagation through `PipelineContext` and `run_*`.
- Non-generic watermark/idle components can directly hold `Arc<dyn ISBWriter>`; tests can inject mocks without creating a dedicated type config.


vigith (approved):

vigith:
is there anything left to implement @BulkBeing to make it completely integrated?

BulkBeing:
> is there anything left to implement @BulkBeing to make it completely integrated?

The integration with the traits throughout the code base is complete. May be some ISB trait methods or KVStore trait may change when implement an alternate version like pulsar or redis.
Also, adding another ISB will need pipeline spec change.

PR #3552, opened 2026-07-20
fix: own pipeline cleanup job by its ISB Service to avoid deleting a recreated pipeline's buffers

CameronMatthew:
### What this PR does

Makes the pipeline buffer-cleanup Job (`isbsvc-delete`, job type `cln`) owned by the **InterStepBufferService** it cleans up, instead of clearing its OwnerReferences entirely. Two commits: the behavioral fix (with the matching unit-test update), and a small follow-up refactor that has every `buildISBBatchJob` caller pass the job's owner explicitly, removing the build-then-mutate call site.

### The problem

We do an A/B deployment that alternates namespaces. Each deployment cycle brings up a new set of pipelines and if they are healthy we drain and tear down the old set.

If we do two A/B deployments back-to-back, we see a race where these cleanup jobs hang around after the application is marked as deleted, and then when the new deployment is brought back, up they delete resources on the next generation of the application.

The root cause of this is that numaflow pipelines don't clean up after themselves properly. They don't clean up after themselves properly because this cleanup job doesn't have the right owner.

Not sure what you guys think about this particular solution. Happy to take suggestions. 

We will probably work around this by not reusing the same namespace for new generations of deployment.  But we think it is probably a gap worth filling nevertheless.

### DIFF
--- a/pkg/reconciler/pipeline/controller.go
+++ b/pkg/reconciler/pipeline/controller.go
@@ -631,7 +631,14 @@ func (r *pipelineReconciler) cleanUpBuffers(ctx context.Context, pl *dfv1.Pipeli
 		args = append(args, fmt.Sprintf("--side-inputs-store=%s", pl.GetSideInputsStoreName()))
 
 		batchJob := buildISBBatchJob(pl, r.image, isbSvc.Status.Config, "isbsvc-delete", args, "cln")
-		batchJob.OwnerReferences = []metav1.OwnerReference{}
+		batchJob.OwnerReferences = []metav1.OwnerReference{
+			{
+				APIVersion: dfv1.SchemeGroupVersion.String(),
+				Kind:       dfv1.ISBGroupVersionKind.Kind,
+				Name:       isbSvc.Name,
+				UID:        isbSvc.UID,
+			},
+		}
 		if err := r.client.Create(ctx, batchJob); err != nil && !apierrors.IsAlreadyExists(err) {
 			return fmt.Errorf("failed to create buffer clean up job, err: %w", err)
 		}
--- a/pkg/reconciler/pipeline/controller.go
+++ b/pkg/reconciler/pipeline/controller.go
@@ -329,7 +329,7 @@ func (r *pipelineReconciler) reconcileFixedResources(ctx context.Context, pl *df
 		}
 		args := []string{fmt.Sprintf("--buffers=%s", strings.Join(bfs, ",")), fmt.Sprintf("--buckets=%s", strings.Join(bks, ","))}
 		args = append(args, fmt.Sprintf("--side-inputs-store=%s", pl.GetSideInputsStoreName()))
-		batchJob := buildISBBatchJob(pl, r.image, isbSvc.Status.Config, "isbsvc-create", args, "cre")
+		batchJob := buildISBBatchJob(pl, r.image, isbSvc.Status.Config, "isbsvc-create", args, "cre", *metav1.NewControllerRef(pl.GetObjectMeta(), dfv1.PipelineGroupVersionKind))
 		if err := r.client.Create(ctx, batchJob); err != nil && !apierrors.IsAlreadyExists(err) {
 			r.recorder.Eventf(pl, corev1.EventTypeWarning, "CreateJobForISBCreationFailed", "Failed to create a Job: %w", err.Error())
 			return fmt.Errorf("failed to create ISB creating job, err: %w", err)
@@ -348,7 +348,7 @@ func (r *pipelineReconciler) reconcileFixedResources(ctx context.Context, pl *df
 			bks = append(bks, k)
 		}
 		args := []string{fmt.Sprintf("--buffers=%s", strings.Join(bfs, ",")), fmt.Sprintf("--buckets=%s", strings.Join(bks, ","))}
-		batchJob := buildISBBatchJob(pl, r.image, isbSvc.Status.Config, "isbsvc-delete", args, "del")
+		batchJob := buildISBBatchJob(pl, r.image, isbSvc.Status.Config, "isbsvc-delete", args, "del", *metav1.NewControllerRef(pl.GetObjectMeta(), dfv1.PipelineGroupVersionKind))
 		if err := r.client.Create(ctx, batchJob); err != nil && !apierrors.IsAlreadyExists(err) {
 			r.recorder.Eventf(pl, corev1.EventTypeWarning, "CreateJobForISBDeletionFailed", "Failed to create a Job: %w", err.Error())
 			return fmt.Errorf("failed to create ISB deleting job, err: %w", err)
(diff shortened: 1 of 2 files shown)


CameronMatthew:
Hey! Feel free to ignore this for now while I take a look at Claude's attempt.

Would be interested in your thoughts on the general vibe of this solution though. Was this intentionally not assigned this owner for a reason that I don't understand?

CameronMatthew:
@Koalk

whynowy:
We made the cln jobs not owned by anybody intentionally - it's not right for them to be owned by any objects.

I'm trying to understand your problem, are you doing A/B deployment back to back in one namespace, and using same pipeline? or something else?

Koalk:
Something like that. I think is "less" right to leave orphan jobs that are cleaning up for a Pipeline. Why is it not right to have them as childs of the pipeline resources? I think it would be much cleaner if the pipeline object was to wait for the cleaning tasks to be done before getting deleted.

CameronMatthew:
> We made the cln jobs not owned by anybody intentionally - it's not right for them to be owned by any objects.
> 
> I'm trying to understand your problem, are you doing A/B deployment back to back in one namespace, and using same pipeline? or something else?

We are doing A/B deployments. We alternate between namespaces. So if we have two deployments in quick succession we do:

#### --- 1st Deployment ---
- A up
- Wait for A to be healthy
- B down
#### --- 2nd Deployment ---
- B up
- Wait for B to be healthy
- A down

The orphaned cleanup jobs from B down then interfere with bringing B back up. As the gate between the two jobs is the B application being marked as removed. But B can be marked as removed while these cleanup jobs remain.

whynowy:
Having them to be children of the pipeline would probably end up with job being deleted by pipeline cascading deletion, even before the job is completed. Or we have to use finalizer for the pipeline to wait for the job to be completed, which is something not right.

whynowy:
> > We made the cln jobs not owned by anybody intentionally - it's not right for them to be owned by any objects.
> > I'm trying to understand your problem, are you doing A/B deployment back to back in one namespace, and using same pipeline? or something else?
> 
> We are doing A/B deployments. We alternate between namespaces. So if we have two deployments in quick succession we do:
> 
> #### --- 1st Deployment ---
> * A up
> * Wait for A to be healthy
> * B down
> 
> #### --- 2nd Deployment ---
> * B up
> * Wait for B to be healthy
> * A down
> 
> The orphaned cleanup jobs from B down then interfere with bringing B back up. As the gate between the two jobs is the B application being marked as removed. But B can be marked as removed while these cleanup jobs remain.

If you are doing the A/B deployment in the same namespace, the race condition happens when the new pipeline is created even before the cln job is down, and the creating job might finish before the cln job, which will cause the cln job to fail forever.

Some enhancement you can do from your automation:
1. Do the switching after the cln job is completed.
2. Or use a different name for the new pipeline, for example, adding an index to each deployment, `my-pipeline-01`, `my-pipeline-02`, etc.

CameronMatthew:
> > > We made the cln jobs not owned by anybody intentionally - it's not right for them to be owned by any objects.
> > > I'm trying to understand your problem, are you doing A/B deployment back to back in one namespace, and using same pipeline? or something else?
> > 
> > 
> > We are doing A/B deployments. We alternate between namespaces. So if we have two deployments in quick succession we do:
> > #### --- 1st Deployment ---
> > 
> > * A up
> > * Wait for A to be healthy
> > * B down
> > 
> > #### --- 2nd Deployment ---
> > 
> > * B up
> > * Wait for B to be healthy
> > * A down
> > 
> > The orphaned cleanup jobs from B down then interfere with bringing B back up. As the gate between the two jobs is the B application being marked as removed. But B can be marked as removed while these cleanup jobs remain.
> 
> If you are doing the A/B deployment in the same namespace, the race condition happens when the new pipeline is created even before the cln job is down, and the creating job might finish before the cln job, which will cause the cln job to fail forever.
> 
> Some enhancement you can do from your automation:
> 
> 1. Do the switching after the cln job is completed.
> 2. Or use a different name for the new pipeline, for example, adding an index to each deployment, `my-pipeline-01`, `my-pipeline-02`, etc.

Yeah so we are moving to generations of namespace like you suggest as a workaround.

It still seems like unexpected behavior to have a cleanup job hanging around after the application is deleted though. It was certainly surprising to us when we came across it.

Since we have a solution for our application I'll let you guys decide what you want to do with this.

whynowy:
Correction:

If you are doing the A/B deployment in the same namespace, the race condition happens when the new pipeline is created even before the cln job is down, and the creating job might finish before the cln job, which will cause the ~~cln job~~ new pipeline to fail forever.

For your scenario, I guess the cln job fails because when you do A/B departments in different namespaces, you clean up everything including the isbsvc, which might happen before the cln job is done, then it gets stuck forever.

CameronMatthew:
> Correction:
> 
> If you are doing the A/B deployment in the same namespace, the race condition happens when the new pipeline is created even before the cln job is down, and the creating job might finish before the cln job, which will cause the ~cln job~ new pipeline to fail forever.
> 
> For your scenario, I guess the cln job fails because when you do A/B departments in different namespaces, you clean up everything including the isbsvc, which might happen before the cln job is done, then it gets stuck forever.

So it's not the same namespace, it's two different namespaces that we alternate between. Read A and B in my description above as namespace A and namespace B each containing a pipelines deployment.

You are right that we cleaned up everything including the ISBSVC.

Koalk:
Right, that is why we were wondering why the job is not a child of the pipeline resource from the kubernetes point of view. It creates and cleans the contents of the pipeline and until it finishes it cannot guarantee the pipeline is created/deleted correctly. It should be part of the pipeline finaliser then too which I think is also correct. In the current state, everything is disconnected and prone to leave orphan resources or misreport the state of a pipeline.
I still don't understand the reason you say having it as a child of the pipeline is wrong, what is the problem with it?

whynowy:
> Right, that is why we were wondering why the job is not a child of the pipeline resource from the kubernetes point of view. It creates and cleans the contents of the pipeline and until it finishes it cannot guarantee the pipeline is created/deleted correctly. It should be part of the pipeline finaliser then too which I think is also correct. In the current state, everything is disconnected and prone to leave orphan resources or misreport the state of a pipeline. I still don't understand the reason you say having it as a child of the pipeline is wrong, what is the problem with it?

Think about the process of doing finalizer clean up if we had it controlled by the pipeline controller - when there is pipeline deletion, before any subresources (eg. vertex pods) get cleaned up with cascade deletion (not done through finalizer), the cln job is triggered, and the buffers are deleted from ISBSvc, while the pods are still running, and they error out. Even they are to be deleted, that's not the right process.

Koalk:
isn't that just declaring the correct syncwaves on the vertices and the job? or more like the controller triggering things in the right order. The cascade delete only deletes everything if the finaliser lets it, having the controller orchestrate the order of things is already the patter in use for creating the pipeline isn't it?

Koalk:
We just had a close call with this issue again. Left-over pipeline cleaning jobs that raced with the deletion of the ISBsvc's were accumulating orphaned and outside of the view of argocd too and we started to have an issue provisioning new pods because they were still hogging all the node pod counts. 
Can we at least add a timeout on this jobs so they are not dangling forever when this happens? I still think a pipeline shouldn't be allowed to be deleted until its cleaning jobs are done but if we can't agree on that maybe we can at least make sure they are not alive forever?

whynowy:
> We just had a close call with this issue again. Left-over pipeline cleaning jobs that raced with the deletion of the ISBsvc's were accumulating orphaned and outside of the view of argocd too and we started to have an issue provisioning new pods because they were still hogging all the node pod counts. 
> 
> Can we at least add a timeout on this jobs so they are not dangling forever when this happens? I still think a pipeline shouldn't be allowed to be deleted until its cleaning jobs are done but if we can't agree on that maybe we can at least make sure they are not alive forever?

By default, the ttl is 30s, and the back off limit is 20. You can customize them by specifying them in templates.job, that should be able to mitigate your problem.

PR #3557, opened 2026-07-23
feat(sqs): utilize nack options for SQS

### LINKED ISSUE #3493: Utilize nack options for SQS
we need to handle nack options for sqs, can come as a follow-up PR.

_Originally posted by @yhl25 in https://github.com/numaproj/numaflow/pull/3455#discussion_r3499959367_

AyushSrivastava1818:

Implements support for SQS nack options by utilizing the SQS `ChangeMessageVisibility` API for negative acknowledgements.

This change introduces SQS-specific nack handling by propagating `NackOffset` information from `numaflow-core` to the SQS extension, allowing per-message visibility timeout to be applied when supported.

### Related issues

Fixes #3493

### Testing

- Ran `cargo fmt`.
- Verified the changes at the module level.
- Full workspace verification (`cargo check`) is currently blocked by an unrelated `krb5-src` dependency issue, which is not introduced by this change.

- Introduces `SqsNack` to carry SQS-specific nack information.
- Adds `SQSActorMessage::Nack` and implements handling via the SQS `ChangeMessageVisibility` API.
- Wires `SourceAcker::nack()` in `numaflow-core` to convert generic `NackOffset` values into SQS-specific nack requests.
- Existing ack behavior remains unchanged.

### DIFF
--- a/rust/numaflow-core/src/source/sqs.rs
+++ b/rust/numaflow-core/src/source/sqs.rs
@@ -137,9 +137,21 @@ impl source::SourceAcker for SqsSource {
             };
 
             let visibility_timeout = nack
-                .option
-                .and_then(|opts| opts.delay)
-                .map(|delay_ms| (delay_ms / 1000) as i32);
+            .option
+            .and_then(|opts| opts.delay)
+            .map(|delay_ms| {
+                let visibility_timeout = delay_ms.div_ceil(1000) as i32;
+
+                if delay_ms % 1000 != 0 {
+                    tracing::warn!(
+                        requested_delay_ms = delay_ms,
+                        visibility_timeout_secs = visibility_timeout,
+                        "SQS visibility timeout only supports whole seconds; rounding requested delay up"
+                    );
+                }
+
+                visibility_timeout
+            });
 
             sqs_offsets.push(SqsNack {
                 receipt_handle: string_offset.offset,
--- a/rust/numaflow-core/src/source/sqs.rs
+++ b/rust/numaflow-core/src/source/sqs.rs
@@ -265,7 +265,7 @@ pub mod tests {
         let receive_message_output = get_receive_message_output();
 
         let delete_message_output = get_delete_message_output();
-
+        let change_visibility_output = get_change_message_visibility_output();
         let queue_attributes_output = get_queue_attributes_output();
 
         let sqs_operation_mocks = MockResponseInterceptor::new()
@@ -273,6 +273,7 @@ pub mod tests {
             .with_rule(&queue_url_output)
             .with_rule(&receive_message_output)
             .with_rule(&delete_message_output)
+            .with_rule(&change_visibility_output)
             .with_rule(&queue_attributes_output);
 
         let sqs_client =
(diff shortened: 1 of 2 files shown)


rust/numaflow-core/src/source/sqs.rs
```diff
@@ -125,12 +125,31 @@ impl source::SourceAcker for SqsSource {
         self.ack_offsets(sqs_offsets).await.map_err(Into::into)
     }
 
-    async fn nack(&mut self, _offsets: Vec<NackOffset>) -> crate::error::Result<()> {
-        // SQS doesn't support nack - no-op
-        Ok(())
+    async fn nack(&mut self, offsets: Vec<NackOffset>) -> crate::error::Result<()> {
```
vaibhavtiwari33, line 128: Could we add e2e tests for this. That should automatically cover other newly added methods as well.

vaibhavtiwari33:
Thank you for the contribution, one small comment ^

rust/numaflow-core/src/source/sqs.rs
AyushSrivastava1818: I looked through the existing e2e suite and couldn't find any SQS e2e test.....only HTTP, NATS, and JetStream. Should I add a new SQS-specific e2e test (with the required test infrastructure), or did you have another existing test in mind that you'd like me to extend?

vaibhavtiwari33: Could we target for something like [this](https://github.com/AyushSrivastava1818/numaflow/blob/39606b299bfa1cc4d63096fa39d02a7971c54d31/rust/numaflow-core/src/source/sqs.rs#L262-L276) using mocks

AyushSrivastava1818: Done...I am gonna add it there..

rust/numaflow-core/src/source/sqs.rs
```diff
@@ -125,12 +125,31 @@ impl source::SourceAcker for SqsSource {
...
+                return Err(Error::Source(format!(
+                    "Expected Offset::String type for SQS. offset={:?}",
+                    nack.offset
+                )));
+            };
+
+            let visibility_timeout = nack
+                .option
+                .and_then(|opts| opts.delay)
+                .map(|delay_ms| (delay_ms / 1000) as i32);
```
syayi: Small edge case: delay is in milliseconds, but SQS only accepts whole seconds. With delay_ms / 1000, any delay under 1 second rounds down to 0. If a user is asking to "retry after 500ms" will get an immediate retry instead, which defeats the point of setting a delay.

Should be round up instead for sub second delays, so we never retry earlier than the user asked? Please add a warn log for either rounding up or rounding down

syayi (dismissed):

rust/numaflow-core/src/source/sqs.rs
AyushSrivastava1818: I extended the existing mocked SQS e2e test by adding a mock for change_message_visibility, which is used by the new nack implementation, so the test setup now covers the additional SQS API introduced by this change as well. Let me know if you'd prefer a dedicated test that explicitly exercises the nack path.

rust/numaflow-core/src/source/sqs.rs
AyushSrivastava1818: Updated this to round the delay up to the next whole second using div_ceil, and added a warning log whenever a non-whole-second delay is requested since SQS visibility timeouts only support whole seconds.

rust/numaflow-core/src/source/sqs.rs
vaibhavtiwari33: So, this is great, we've defined what to do when we nack, but unfortunately we're not testing the nack path with this addition.
You can simply add a udsink, instead of the currently used log sink, which nacks 1 message (delay is not necessary) and rest of the test remains same.

AyushSrivastava1818: Done...will make the changes soon

AyushSrivastava1818: Addressed the review comments by extending the SQS e2e test to use a user-defined sink that returns a NACK on the first message and verifies the ChangeMessageVisibility path is exercised.

vaibhavtiwari33 (approved):

Issue #3544, opened 2026-07-17
Support cron-schedule-based autoscaling for MonoVertex and Pipeline vertices


yhl25:
## Summary

Numaflow autoscaling today is purely reactive, scaling between `scale.min` and `scale.max` based on pending messages and processing rate. This proposes **cron-schedule-based autoscaling**: users define time windows with their own `min`/`max` bounds (including 0), and the reactive autoscaler operates within the bounds of the active window.

## Use cases

- **DLQ draining via fallback sink**: Numaflow supports a fallback sink, often pointed at a DLQ. Today, reprocessing DLQ messages requires a separate scheduled job. With cron-based autoscaling, the same MonoVertex/Pipeline can sit at `min: 0` and scale up on a schedule to drain the DLQ, no extra infrastructure.
- **Pre-warming**: scale up ahead of predictable load spikes instead of paying cold-start lag after pending builds up.
- **Cost savings**: force lower bounds during known-idle windows (nights/weekends).

## Proposed spec (starting point)

Extend the existing `Scale` struct (applies to both MonoVertex and Pipeline vertices):

```yaml
scale:
  min: 0
  max: 50
  cron:
    timezone: "America/Los_Angeles"
    schedules:
      - start: "0 2 * * *"   # nightly DLQ drain
        end: "0 3 * * *"
        min: 1
        max: 5
```

Outside any window, base `min`/`max` apply; inside a window, its bounds override the base.

AyushSrivastava1818:
Hi! I'd like to work on this feature if it's available.

I'll first go through the current autoscaling implementation and the existing `Scale` API to understand the flow, then propose the necessary changes for the cron-based scheduling while keeping the current reactive autoscaling behavior intact.

Please assign this issue to me if that works. Thanks!

syayi:
One thing to consider, instead of defining cron schedules per vertex, should the schedule be defined once at the Pipeline level, the whole pipeline comes up/down on one clock, with each vertex still adhering to its own min/max? Would avoid vertices being on mismatched schedules within the same pipeline.

AyushSrivastava1818:
That's a good point. Defining the schedule once at the Pipeline level would definitely simplify the configuration and avoid the possibility of different vertices ending up on conflicting schedules.

[...]

Issue #3054, opened 2025-10-29
Sideinput not working in v1.6: `ERROR numaflow: Schedule("*/1 * * * *\n           ^\n")`


th0ger:
**Describe the bug**

When upgrading from numaflow `v1.5.1` to `v1.6.0`, our sideinputs stops working.

The `numa` sideinput container keeps crashing with `ERROR numaflow: Schedule("*/1 * * * *\n           ^\n")`. It's not clear why the `Schedule` is mentioned in this error (we did not change the pipeline spec), or whether it's the underlying process that the scheduler calls that's failing.

This is blocking our `v1.6.0` migration.

**To Reproduce**

Our sideinput basically connects to kafka and read all events on a topic.

The pipeline spec contains:

```yaml
  sideInputs:
  - name: mysideinput
    container:
      image: sideinputimage
      args:
      - '--config'
      - |
        kafka:
          topic: mytopic
          consumerGroup: mygroup
          brokers:
          - mybroker:9092
    trigger:
      schedule: '*/1 * * * *'
      timezone: Europe/Copenhagen
```

We don't have the logged `\n           ^\n` part in our spec.

Our sideinputimage roughly contains:
```python
class KafkaBulkClient:
    def __init__(...):
        ...
        print(f"KafkaBulkClient initializing with settings: {self}")
        print("Establishing Consumer connection to kafka brokers")
        self.consumer = Consumer(self.consumer_config)
        print(f"Subscribing to kafka topic {self.topic}")
        self.consumer.subscribe([self.topic])
        print("KafkaBulkClient initialized ")

class KafkaSideInput(SideInput):
    def __init__(
        self,
        kafkaclient: KafkaBulkClient,
        ...
    ):

    def retrieve_handler(self, dry: bool = False) -> Response:
        print(f"Sideinput retrieve_handler starting")
        ...
        messages = self.kafkaclient.consume_all_messages()
        ...

kafkaclient = KafkaBulkClient(
    topic=topic,
    consumer_group=consumer_group,
    brokers=brokers,
    tls_enabled=tls_enabled,
    secret_volume_paths=secret_volume_paths,
    ssl_cert_verify=ssl_cert_verify,
)

# Initialize sideinput instance
kafkasideinput = KafkaSideInput(kafkaclient=kafkaclient)

# Start GRPC server
grpc_server = SideInputServer(kafkasideinput)
grpc_server.start()
```

The `numa` container keeps crashing with:
```text
[...]

PR #3558, opened 2026-07-23
feat: cron-based MonoVertex autoscaling spec

### LINKED ISSUE #3546: Cron-based autoscaling: MonoVertex controller changes

devsarvesh92:

Adds cron-window-based autoscaling bounds for MonoVertex.

Users can configure recurring windows with minimum and maximum replica
counts. While a window is active:

- Replicas outside the configured bounds move gradually toward the nearest
  bound.
- Existing scale-up and scale-down cooldowns remain respected.
- Reactive metric-based autoscaling continues within the active bounds.
- Cron bounds are applied before daemon metrics are queried, allowing
  scheduled scale-up from zero.

When cron is omitted, existing autoscaling behavior is unchanged.

Example:

```yaml
scale:
  min: 0
  max: 50
  cron:
    timezone: America/Los_Angeles
    schedules:
      - start: "0 2 * * * *"
        end: "0 3 * * * *"
        min: 1
        max: 5
```

### Related issues

Fixes #3546

Parent issue: #3544

### Testing

- `make test`
- `make pre-push -B`
- `go vet ./pkg/apis/numaflow/v1alpha1 ./pkg/reconciler/monovertex/scaling ./pkg/reconciler/validator`
- `git diff --check`
- Local kind integration:
  - Verified active cron scale-up patches `spec.replicas` before metrics.
  - Verified gradual replica changes honor `replicasPerScaleUp`.
  - Verified the controller reconciles patched replicas and creates worker pods.
  - Verified required timezone is rejected by the CRD schema.
  - Verified incompatible start/end calendar fields are rejected by the webhook.

- Cron configuration is optional.
- When cron is configured, timezone, schedules, min, and max are required.
- Schedule bounds must remain within the parent scale bounds.
- Overlapping schedule detection is deferred; the first active schedule wins.
- `Scale` is shared, so generated Pipeline and Vertex CRDs also contain the
  cron schema. Pipeline runtime support is outside this sub-issue.
- CRD, OpenAPI, protobuf, install manifest, and API documentation changes are
  generated.

### Contributors

- @devsarvesh92
- @learncoder4848


pkg/apis/numaflow/v1alpha1/scale.go
```diff
@@ -16,6 +16,46 @@ limitations under the License.
...
+	"time"
+
+	cron "github.com/robfig/cron/v3"
+)
+
+// CronScheduling defines cron-based autoscaling overrides.
+type CronScheduling struct {
+	// Timezone for interpreting cron expressions. IANA Time Zone Database format.
+	Timezone  string         `json:"timezone" protobuf:"bytes,1,opt,name=timezone"`
+	Schedules []CronSchedule `json:"schedules" protobuf:"bytes,2,rep,name=schedules"`
```
vaibhavtiwari33: ~nit: Could we document the cron format being used. Standard vs extended~

I see it documented below

pkg/apis/numaflow/v1alpha1/scale.go
```diff
@@ -16,6 +16,46 @@ limitations under the License.
...
+}
+
+type CronSchedule struct {
+	// Start of the cron window. Linux cron format (Minute Hour Dom Month Dow).
+	Start string `json:"start" protobuf:"bytes,1,opt,name=start"`
```
vaibhavtiwari33: nit: For sideinputs we're using extended format (6 fields). This might create confusion between users about which one to use.
ref: https://github.com/numaproj/numaflow/issues/3054

pkg/apis/numaflow/v1alpha1/scale.go
```diff
@@ -16,6 +16,46 @@ limitations under the License.
...
+	End string `json:"end" protobuf:"bytes,2,opt,name=end"`
+	// Minimum replicas during this window. Overrides scale.min.
+	Min *int32 `json:"min" protobuf:"varint,3,opt,name=min"`
+	// Maximum replicas during this window. Overrides scale.max.
+	Max *int32 `json:"max" protobuf:"varint,4,opt,name=max"`
+}
+
+// IsActiveAt reports whether time t falls within this cron window.
+func (cs CronSchedule) IsActiveAt(t time.Time) bool {
+	parser := cron.NewParser(cron.Minute | cron.Hour | cron.Dom | cron.Month | cron.Dow)
```
vaibhavtiwari33:
```suggestion
	parser := cron.NewParser(cron.Second | cron.Minute | cron.Hour | cron.Dom | cron.Month | cron.Dow)
```

pkg/apis/numaflow/v1alpha1/scale.go
```diff
@@ -16,6 +16,46 @@ limitations under the License.
...
+import (
+	"time"
+
+	cron "github.com/robfig/cron/v3"
+)
+
+// CronScheduling defines cron-based autoscaling overrides.
+type CronScheduling struct {
+	// Timezone for interpreting cron expressions. IANA Time Zone Database format.
+	Timezone  string         `json:"timezone" protobuf:"bytes,1,opt,name=timezone"`
```
vaibhavtiwari33: nit: Can this be made optional with default being UTC?

pkg/apis/numaflow/v1alpha1/scale.go
```diff
@@ -65,6 +105,45 @@ type Scale struct {
...
+	}
+	loc, err := time.LoadLocation(s.Cron.Timezone)
+	if err != nil {
+		return nil, false
+	}
+	at = at.In(loc)
+	for i := range s.Cron.Schedules {
+		if s.Cron.Schedules[i].IsActiveAt(at) {
+			return &s.Cron.Schedules[i], true
+		}
```
vaibhavtiwari33: We should document that we honor the first matching schedule in case when there exist multiple overlapping schedules.

vaibhavtiwari33:
Could we also please update the [autoscaling doc](https://numaflow.numaproj.io/user-guide/reference/autoscaling/#numaflow-autoscaling) (specifically the mvtx section)

pkg/apis/numaflow/v1alpha1/scale.go
vigith: parsed schedules should be pre-computed, it is computationally heavy and performance will be impacted since it is in hot code path of reconciliation.

pkg/apis/numaflow/v1alpha1/scale.go
```diff
@@ -65,6 +105,45 @@ type Scale struct {
...
+	}
+	return nil, false
+}
+
+// GetEffectiveScaleBounds returns the min/max replicas and whether a cron schedule is active.
+func (s Scale) GetEffectiveScaleBounds() (int32, int32, bool) {
+	minR := s.GetMinReplicas()
+	maxR := s.GetMaxReplicas()
+	sched, active := s.GetActiveCronSchedule(time.Now())
```
vigith: this can make it tough to unit-test, can we pass time.Now() as an arg.

pkg/reconciler/validator/mvtx_validate.go
vigith, line 1:
```
 227 +  if v.Scale.Cron != nil {                                                                                                                                                              
 228 +    return fmt.Errorf("vertex %q: cron autoscaling is not supported for pipeline vertices", v.Name)                                                                                     
 229 +  } 
```

we need to add this pipeline else it will be silently ignored.

pkg/reconciler/validator/mvtx_validate.go
```diff
@@ -78,5 +80,89 @@ func ValidateMonoVertex(mvtx *dfv1.MonoVertex) error {
...
+	}
+	if scale.Disabled {
+		return fmt.Errorf("cron must not be configured when autoscaling is disabled")
+	}
+	if scale.Min != nil && *scale.Min < 0 {
+		return fmt.Errorf("scale.min must not be negative")
+	}
+	if scale.Max != nil && *scale.Max < 0 {
+		return fmt.Errorf("scale.max must not be negative")
+	}
```
vigith, line 103: The `cronMin >= parentMin` and `cronMax <= parentMax` is concerning. This blocks the common pattern where `parentMin=1` (always keep at least one replica running) but during a quiet window you want the cron to scale down to 0. If the intent of cron overrides is to override the parent bounds, this constraint should be relaxed.

pkg/reconciler/monovertex/scaling/scaling.go
```diff
@@ -180,7 +180,17 @@ func (s *Scaler) scaleOneMonoVertex(ctx context.Context, key string, worker int)
...
-	if monoVtx.Spec.Scale.GetMaxReplicas() == monoVtx.Spec.Scale.GetMinReplicas() {
+	minReplicas, maxReplicas, cronActive := monoVtx.Spec.Scale.GetEffectiveScaleBounds()
+	current := int32(monoVtx.Status.Replicas)
+	if cronActive {
+		if current < minReplicas {
+			return s.scaleUp(ctx, monoVtx, current, minReplicas, secondsSinceLastScale, scaleUpCooldown)
+		}
+		if current > maxReplicas {
+			return s.scaleDown(ctx, monoVtx, current, maxReplicas, secondsSinceLastScale, scaleDownCooldown)
+		}
```
vigith, line 280: log.Infof("Cron window active [min=%d, max=%d], current=%d is outside bounds; scaling.", minReplicas, maxReplicas, current)

learncoder4848:
Thanks for the thorough reviews, @vaibhavtiwari33 and @vigith! The latest commit addresses all the comments. Summary below.

## @vigith

- **Pre-compute parsed cron schedules (hot path perf):** Schedules are now parsed once via `CronScheduling.Compile()` and cached in the scaler (`cronScheduleCache`, keyed by `UID/Generation`), so no cron parsing happens on the reconciliation hot path. The cache invalidates automatically on spec changes (Generation) and object recreation (UID).
- **Pass `time.Now()` for testability:** Bound evaluation now flows through `effectiveScaleBoundsAt(scale, parsed, at)` with the time passed in explicitly, so it's unit-testable without wall-clock dependence.
- **Reject cron on pipeline vertices (silent-ignore):** Added validation in `validateVertex` — pipeline vertices with `scale.cron` are now rejected with `cron autoscaling is not supported for pipeline vertices` instead of being silently ignored.
- **Relax `cronMin >= parentMin` / `cronMax <= parentMax`:** Removed these containment checks so cron windows can override the parent bounds (e.g. `parentMin=1` scaling down to `0` during a quiet window). 
- **Log when cron forces scaling outside bounds:** Added the info log when an active cron window scales because current replicas are outside `[min, max]`.

## @vaibhavtiwari33

- **Cron format documentation / consistency with side inputs:** Switched to the six-field extended format (`second minute hour dom month dow`), matching the side-input convention, and documented it. (ref #3054)
- **Timezone optional, default UTC:** `timezone` is now optional and defaults to `UTC` via `CronScheduling.GetTimezone()`; validation and CRDs updated accordingly.
- **Document first-match on overlapping schedules:** Documented that schedules are evaluated in order and the first active schedule takes precedence, both in the API godoc and the autoscaling reference doc.
- **Update autoscaling docs (mvtx section):** Updated `docs/user-guide/reference/autoscaling.md` with the `cron` field, examples, format, timezone default, and precedence behavior.

docs/user-guide/reference/autoscaling.md
```diff
@@ -63,6 +63,13 @@ spec:
     targetProcessingSeconds: 20 # Optional, defaults to 20.
     replicasPerScaleUp: 2 # Optional, defaults to 2.
     replicasPerScaleDown: 2 # Optional, defaults to 2.
+    cron: # Optional; supported only for MonoVertex.
```
vaibhavtiwari33, line 66: nit: 

```suggestion
    cron: # Optional; currently only supported for MonoVertex.
```

Since, I believe there exists an issue to add this functionality for pipeline as well

vigith (approved):
thank you, LGTM

learncoder4848:
> thank you, LGTM

@vigith  Can you approve the awaiting workflows so that we can merge the changes ?

vaibhavtiwari33 (approved):

learncoder4848:
@vigith  can you re-trigger/approve  this [test / E2E Tests (jetstream, e2e) (pull_request)] , don't know what is this

vigith:
> @vigith can you re-trigger/approve this [test / E2E Tests (jetstream, e2e) (pull_request)] , don't know what is this

it is successful now, Github Runners are not very stable

PR #3569, opened 2026-07-30
feat: support cron-based autoscaling for source vertices

### LINKED ISSUE #3547: Cron-based autoscaling: Pipeline controller changes

AyushSrivastava1818:

Adds support for cron-based autoscaling for source vertices.

This change allows source vertices to honor cron schedules defined under `spec.scale.cron`. During an active cron window, the autoscaler applies the configured `min` and `max` replica bounds before performing the normal autoscaling calculations. Outside the scheduled window, the default scale bounds are used.

The implementation includes:
- Parsing and validation of cron schedules.
- Caching parsed cron schedules to avoid repeated parsing.
- Determining effective scale bounds based on the current time.
- Applying cron-specific min/max replica limits for source vertices.
- Unit tests covering cron schedule parsing and effective bound calculation.

### Related issues

Fixes #3547

### Testing

Unit tests:
```bash
go test ./pkg/reconciler/vertex/scaling/...

### DIFF
--- a/docs/user-guide/reference/autoscaling.md
+++ b/docs/user-guide/reference/autoscaling.md
@@ -45,6 +45,13 @@ spec:
         targetBufferAvailability: 50 # Optional, defaults to 50.
         replicasPerScaleUp: 2 # Optional, defaults to 2.
         replicasPerScaleDown: 2 # Optional, defaults to 2.
+        cron: # Optional.
+          timezone: America/Los_Angeles # Optional, defaults to UTC.
+          schedules:
+            - start: "0 0 22 * * *" # Second Minute Hour Day-of-month Month Day-of-week
+              end: "0 0 6 * * *"
+              min: 10
+              max: 20
 ---
 # A MonoVertex example.
 apiVersion: numaflow.numaproj.io/v1alpha1
@@ -63,7 +70,7 @@ spec:
     targetProcessingSeconds: 20 # Optional, defaults to 20.
     replicasPerScaleUp: 2 # Optional, defaults to 2.
     replicasPerScaleDown: 2 # Optional, defaults to 2.
-    cron: # Optional; supported only for MonoVertex.
+    cron: # Optional.
       timezone: America/Los_Angeles # Optional, defaults to UTC.
       schedules:
         - start: "0 0 22 * * *" # Second Minute Hour Day-of-month Month Day-of-week
@@ -122,7 +129,7 @@ spec:
 - `replicasPerScale` - (Deprecated: Use `replicasPerScaleUp` and
   `replicasPerScaleDown` instead, will be removed in v1.5) Maximum number of
   replica change happens in one scale up or down operation, defaults to `2`.
-- `cron` - [MonoVertex only] Defines time-based overrides for `min` and `max`.
+- `cron` - Defines time-based overrides for `min` and `max`.
   Cron expressions use the six-field extended format: `second minute hour
   day-of-month month day-of-week`. `timezone` accepts an IANA time zone and
   defaults to `UTC` when omitted. Each schedule requires `start`, `end`, `min`,
--- a/docs/user-guide/reference/autoscaling.md
+++ b/docs/user-guide/reference/autoscaling.md
@@ -45,13 +45,6 @@ spec:
         targetBufferAvailability: 50 # Optional, defaults to 50.
         replicasPerScaleUp: 2 # Optional, defaults to 2.
         replicasPerScaleDown: 2 # Optional, defaults to 2.
-        cron: # Optional; supported only for source vertices.
-          timezone: America/Los_Angeles # Optional, defaults to UTC.
-          schedules:
-            - start: "0 0 22 * * *" # Second Minute Hour Day-of-month Month Day-of-week
-              end: "0 0 6 * * *"
-              min: 10
-              max: 20
 ---
 # A MonoVertex example.
 apiVersion: numaflow.numaproj.io/v1alpha1
@@ -138,6 +131,34 @@ spec:
   and `max`; its bounds may be lower or higher than the base `scale.min` and
   `scale.max`. When multiple windows overlap, schedules are evaluated in the
   configured order and the first active schedule takes precedence.
+
+#### Cron Autoscaling for Source Vertices
+
+Cron-based autoscaling allows time-based overrides of `min` and `max` replicas.
+It is **only valid for Pipeline source vertices** (and MonoVertices — see the
+MonoVertex example above). Configuring `cron` on a non-source Pipeline vertex
+(UDF or Sink) will be **rejected at validation time** with an error.
+
+```yaml
+apiVersion: numaflow.numaproj.io/v1alpha1
+kind: Pipeline
+metadata:
+  name: my-pipeline
+spec:
+  vertices:
+    - name: input
+      source:
+        generator:
+          rpu: 5
+      scale:
+        cron: # Optional; only valid on source vertices.
+          timezone: America/Los_Angeles # Optional, defaults to UTC.
+          schedules:
+            - start: "0 0 22 * * *" # Second Minute Hour Day-of-month Month Day-of-week
+              end: "0 0 6 * * *"
+              min: 10
+              max: 20
+```
 
 To disable Numaflow autoscaling, set `disabled: true` as following.
 
--- a/docs/user-guide/reference/autoscaling.md
+++ b/docs/user-guide/reference/autoscaling.md
@@ -45,7 +45,7 @@ spec:
         targetBufferAvailability: 50 # Optional, defaults to 50.
         replicasPerScaleUp: 2 # Optional, defaults to 2.
         replicasPerScaleDown: 2 # Optional, defaults to 2.
-        cron: # Optional.
+        cron: # Optional; supported only for source vertices.
           timezone: America/Los_Angeles # Optional, defaults to UTC.
           schedules:
             - start: "0 0 22 * * *" # Second Minute Hour Day-of-month Month Day-of-week
@@ -70,7 +70,7 @@ spec:
     targetProcessingSeconds: 20 # Optional, defaults to 20.
     replicasPerScaleUp: 2 # Optional, defaults to 2.
     replicasPerScaleDown: 2 # Optional, defaults to 2.
-    cron: # Optional.
+    cron: # Optional; supported only for MonoVertex.
       timezone: America/Los_Angeles # Optional, defaults to UTC.
       schedules:
         - start: "0 0 22 * * *" # Second Minute Hour Day-of-month Month Day-of-week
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by AyushSrivastava1818** [comment]
The failing docs-linkcheck job appears unrelated to this PR. The reported 404s are in docs/core-concepts/inter-step-buffer-service.md, docs/core-concepts/inter-step-buffer.md, and docs/user-guide/sources/jetstream.md, which are not modified by this change. The external NATS documentation links currently return 404.

File: pkg/reconciler/vertex/scaling/scaling.go [hunk 1]
```diff
@@ -36,9 +37,62 @@ import (
...
+func parseCronSchedules(cronScheduling *dfv1.CronScheduling) ([]parsedCronSchedule, error) {
+	parser := cron.NewParser(cron.Second | cron.Minute | cron.Hour | cron.Dom | cron.Month | cron.Dow)
+	result := make([]parsedCronSchedule, 0, len(cronScheduling.Schedules))
+	for _, schedule := range cronScheduling.Schedules {
+		start, err := parser.Parse(schedule.Start)
+		if err != nil {
+			return nil, fmt.Errorf("invalid cron start expression %q: %w", schedule.Start, err)
+		}
+		end, err := parser.Parse(schedule.End)
+		if err != nil {
```
vaibhavtiwari33: All of this and a bit more was already added as part of [this PR](https://github.com/numaproj/numaflow/pull/3558/changes#diff-a09812a67dbf3323210dc47ce4689c52f18de51cf0459f536cdf7481dbfe4fae). Maybe your fork doesn't have the latest changes? Could you rebase and try again

adarsh0728:
> The failing docs-linkcheck job appears unrelated to this PR. The reported 404s are in docs/core-concepts/inter-step-buffer-service.md, docs/core-concepts/inter-step-buffer.md, and docs/user-guide/sources/jetstream.md, which are not modified by this change. The external NATS documentation links currently return 404.

`make docs` has been fixed. Please rebase.

pkg/reconciler/vertex/scaling/scaling.go
AyushSrivastava1818: Rebased it again you can check now and let me know if anything else I need to do in this

vaibhavtiwari33: Since monovertex scaling already defines cron parsing logic, let's move it to a common location (utils.go maybe) so that we're not redefining the same thing in both places.

AyushSrivastava1818: Refactored the cron parsing and effective scale bound logic into a shared package (pkg/reconciler/scaling) and updated both Vertex and MonoVertex scaling to use it. The shared helper tests were moved into the new package, and the duplicate tests in vertex/scaling were removed. I've rebased onto the latest main as well. Please take another look when you have a chance.

AyushSrivastava1818:
one more doubt I have that is there a public Slack channel or other community chat for Numaflow contributors?

vaibhavtiwari33:
[From](https://numaflow.numaproj.io/user-guide/FAQ/#3-where-can-i-learn-about-the-latest-releases-of-numaflow): [invite link](https://join.slack.com/t/numaproj/shared_invite/zt-19svuv47m-YKHhsQ~~KK9mBv1E7pNzfg)

vaibhavtiwari33 (approved):

vaibhavtiwari33 (changes requested):
I believe there was a check added in the monovertex PR to prevent marking cron config invalid for pipelines:

https://github.com/numaproj/numaflow/blob/55c3519fd82595d8248d26b4a1cf8c955a63d3af/pkg/reconciler/validator/pipeline_validate.go#L214-L216

Pls remove this check, as well as update the documentation for auto scaling. 

https://github.com/numaproj/numaflow/blob/55c3519fd82595d8248d26b4a1cf8c955a63d3af/docs/user-guide/reference/autoscaling.md?plain=1#L125

https://github.com/numaproj/numaflow/blob/55c3519fd82595d8248d26b4a1cf8c955a63d3af/docs/user-guide/reference/autoscaling.md?plain=1#L66

pkg/reconciler/vertex/scaling/scaling.go
```diff
@@ -209,12 +240,38 @@ func (s *Scaler) scaleOneVertex(ctx context.Context, key string, worker int) err
...
+
+		current := int32(vertex.Status.Replicas)
+		if cronActive {
+			if current < minReplicas {
+				log.Infof("Cron window active [min=%d, max=%d], current=%d is outside bounds; scaling up.", minReplicas, maxReplicas, current)
+				return s.patchVertexReplicas(ctx, vertex, minReplicas)
+			}
+			if current > maxReplicas {
+				log.Infof("Cron window active [min=%d, max=%d], current=%d is outside bounds; scaling down.", minReplicas, maxReplicas, current)
+				return s.patchVertexReplicas(ctx, vertex, maxReplicas)
```
vaibhavtiwari33: Instead of patching replicas immediately, could we add checks similar to what we have for mvtx scaling:

https://github.com/numaproj/numaflow/blob/55c3519fd82595d8248d26b4a1cf8c955a63d3af/pkg/reconciler/monovertex/scaling/scaling.go#L380-L406

usage:

https://github.com/numaproj/numaflow/blob/55c3519fd82595d8248d26b4a1cf8c955a63d3af/pkg/reconciler/monovertex/scaling/scaling.go#L275-L279

vigith (commented):
we cannot attach the `cron` spec to any vertex, it should be only on the source. the current implementation let you attach to any vertex and will be silently ignored on non-source vertices.

vaibhavtiwari33 (approved):

docs/user-guide/reference/autoscaling.md
```diff
@@ -45,6 +45,13 @@ spec:
         targetBufferAvailability: 50 # Optional, defaults to 50.
         replicasPerScaleUp: 2 # Optional, defaults to 2.
         replicasPerScaleDown: 2 # Optional, defaults to 2.
+        cron: # Optional; supported only for source vertices.
+          timezone: America/Los_Angeles # Optional, defaults to UTC.
+          schedules:
+            - start: "0 0 22 * * *" # Second Minute Hour Day-of-month Month Day-of-week
+              end: "0 0 6 * * *"
+              min: 10
+              max: 20
```
vigith: `cron` cannot be added to any vertex. Please add a new section called `sources` and add `cron` there.

vigith (changes requested):

vaibhavtiwari33:
Would it be possible to showcase the validation and cron based scale up for vertices on your local setup?

docs/user-guide/reference/autoscaling.md
AyushSrivastava1818: Moved the cron config out of the generic vertex example and into a new "Cron Autoscaling for Source Vertices" subsection, nested correctly under source:. Also double-checked: validation already rejects cron on non-source vertices (pipeline_validate.go:214), so the docs were the only thing out of sync with the code.

AyushSrivastava1818:
@vaibhavtiwari33 I didn't get a chance to record a local demo yet since I don't have a cluster with the right generator/source setup readily available right now [or: "since this needs a live cron window to actually observe the scale-up, which takes a few extra minutes to set up properly"]. What I can confirm is that all validation paths are covered by unit tests and passing locally:

cron on non-source vertex → correctly rejected
cron on source vertex → correctly accepted
invalid cron config on source vertex → correctly rejected

vigith:
> @vaibhavtiwari33 I didn't get a chance to record a local demo yet since I don't have a cluster with the right generator/source setup readily available right now [or: "since this needs a live cron window to actually observe the scale-up, which takes a few extra minutes to set up properly"]. 

Please setup and try, that is how you learn 🙂

vaibhavtiwari33:
Please test out this functionality by setting up a local cluster. [Quick start guide](https://numaflow.numaproj.io/quick-start/)

AyushSrivastava1818:
@vigith @vaibhavtiwari33 Okay will test this functionality by myself then report it here!

vaibhavtiwari33:
@AyushSrivastava1818 Were you able to test out the current PR?

AyushSrivastava1818:
@vaibhavtiwari33 Not really as I am shifting to my clg as it is reopened now after vacations... apologies from my side will do it as soon as I get time!

vaibhavtiwari33:
Do you have a tentative timeline? Asking since this change is to be part of the next patch release.

AyushSrivastava1818:
@vaibhavtiwari33 i will do it by tomorrow night for sure

AyushSrivastava1818:
Hi @vaibhavtiwari33 @vigith,

Quick update on the local testing.

I was able to set up a local cluster and reproduce the cron autoscaling scenario. During testing I found that the behavior is a bit more involved than I initially expected.

From what I've verified so far:

* validation behaves as expected (cron is accepted only for source vertices and rejected for non-source vertices),
* the cron autoscaler detects the active window and updates the Vertex,
* however, the expected scale-up of the source vertex pods is not happening in my local setup.

I'm currently tracing the reconciliation/runtime path to determine whether this is an issue in my implementation or if I've uncovered an interaction in the existing controller logic. I'd rather understand the exact cause before submitting another change instead of making a speculative fix.

I'll continue investigating and share my findings (and any required fix, if there is one) as soon as I have a confirmed root cause.

vaibhavtiwari33:
What is the setup that you're using? Wht is the source

AyushSrivastava1818:
I'm using a local kind cluster with the latest main plus my PR changes.
For testing I used a simple pipeline with a source, cat and sink connected through the default JetStream ISB service. The cron configuration is applied only on the source vertex with a short time window so I can observe the scaling behaviour locally.
What I'm seeing is that the cron window becomes active and the autoscaler updates the source vertex, but the additional source pods are not getting created. I'm tracing through the reconciliation flow to understand exactly where things are going wrong before making any changes.
If you're referring to a different source setup, please let me know which one you had in mind and I'll test with that as well.

vaibhavtiwari33:
## Test 1

### Spec
```yaml
      scale:
        min: 2
        max: 5
        scaleUpCooldownSeconds: 0 # Optional, defaults to 90.
        scaleDownCooldownSeconds: 0 # Optional, defaults to 90.
        replicasPerScaleUp: 10 # Optional, defaults to 2.
        replicasPerScaleDown: 10 # Optional, defaults to 2.
        cron:
          timezone: America/Los_Angeles # Optional, defaults to UTC.
          schedules:
            - start: "0 54 * * * *" # Second Minute Hour Day-of-month Month Day-of-week
              end: "0 55 * * * *"
              min: 10
              max: 20
```

### Verdict
* Top level min and max are honoured
* Scale up and scale down happens at the specified schedule

### Recording
https://github.com/user-attachments/assets/a7967c75-c115-448f-ad2e-bbf622432ebc

## Test 2

### Spec
```yaml
      scale:
        min: 0
        max: 15
        scaleUpCooldownSeconds: 0 # Optional, defaults to 90.
        scaleDownCooldownSeconds: 0 # Optional, defaults to 90.
        replicasPerScaleUp: 10 # Optional, defaults to 2.
        replicasPerScaleDown: 10 # Optional, defaults to 2.
        cron:
          timezone: America/Los_Angeles # Optional, defaults to UTC.
          schedules:
            - start: "0 54 * * * *" # Second Minute Hour Day-of-month Month Day-of-week
              end: "0 55 * * * *"
              min: 10
              max: 20
```

### Verdict
* Scaling honours cron's min value when given headroom by the top level value
* Scale down didn't happen all the way to zero, but I believe that could be because pending kept fluctuating in simple source

### Recording
https://github.com/user-attachments/assets/827e4029-4ebb-44b9-b864-668df92b8d35

vigith (approved):

yhl25 (approved):

PR #3570, opened 2026-07-30
feat: introduce retry config for transformer and map

### LINKED ISSUE #3220: Built-in Retry on UDFs
# Summary

Have settings to have a message that fails auto retry. We've done this in our own SDKs by keeping track of retries in message envelope and requeuing message via loop so we can keep track of number of retries and how long (TTL) since we tried before we simply expire (drop) the message. But this depends on our own message formats supporting that, I think having something in datum for first-order support would be valuable. Additionally, this requires some logic on the edges to conditionally forward (or not) on retries and this adds complexity to pipeline definition

Losing the order of messages after one fails is generally okay for us, so having messages go back into queue - even to back of queue - is often fine. 

Sink basically supports this today so hopefully a relatively easy or similar behavior on UDFs?

# Use Cases

Any step where we might have intermediate failure would allow retry (external service connection failure, inaccessible file due to some transfer race condition, etc)

---

### LINKED ISSUE #3413: Built-in Retry for Transformer
# Summary

Similar to User Defined Sink retry strategy, can we get fixed and exponential retry support in transformer.

# Use Cases

When would you use this?

---

vaibhavtiwari33:
Introduce a retry config for transformer and map. 
Users can essentially configure a retry strategy, similar to sink. 
Based on the strategy retries are triggered when the user explicitly fails a message via the reserved `\__FAIL__` tag (`Message::failed()`), analogous to the Sink's per-message failure handling.

### Related issues
Fixes #3413 
Fixes #3220 

### Testing
Adds unit tests across transformer, unary, batch, and stream map covering: retry-then-succeed, infinite retry with no config, drop-on-exhaustion (incl. drop metric), and retry-exhaustion (NAK / error propagation).

- Apart from the spec change, changes are mostly in `transformer.rs`, `batch.rs`, `unary.rs`, and `stream.rs`
- Apart from tests, and `make codegen` generated code, other changes are written by this user (reviewed by an agent), please review carefully xD


vigith (commented):
if a single message in a batch is in FAIL will we retry the entire batch? if yes, there is an inherent duplication problem here, right?

vaibhavtiwari33:
> if a single message in a batch is in FAIL will we retry the entire batch?

That is correct. 

> if yes, there is an inherent duplication problem here, right?

Ya, that is currently a caveat with this and per message nack.

vigith (approved):

vaibhavtiwari33:
> if a single message in a batch is in FAIL will we retry the entire batch?

@vigith I think I misunderstood the question here. Just wanted to clarify.  

* In the case of map-batch, if one message in the batch is marked for failure, only that message's input from the batch is retried in the next retry loop. 
* In the case of flatmap, if one output message is marked for failure, the input will be retried again

rust/numaflow-core/src/transformer.rs
```diff
@@ -223,6 +262,8 @@ impl Transformer {
                         transform_handle.clone(),
                         read_msg.clone(),
                         hard_shutdown_token.clone(),
+                        retry_controller.clone(),
```
adarsh0728: We are not preserving the FAIL attempts which are counted inside transform() ? Thinking, if should preserve the FAIL attempts for that message so that on reconnect flow we do not see fresh count at 0

adarsh0728: users may say that they configured 2 retries but seeing more than 2 retries (if reconnect happens)

config/base/crds/full/numaflow.numaproj.io_monovertices.yaml
```diff
@@ -6724,6 +6724,27 @@ spec:
...
+                      retryStrategy:
+                        properties:
+                          backoff:
+                            properties:
+                              cap:
+                                type: string
+                              factor:
+                                type: number
+                              interval:
+                                default: 1ms
```
adarsh0728, line 6736: not specific to this PR, but we should think of better default interval, 1ms is too aggressive.

rust/numaflow-core/src/transformer.rs
vaibhavtiwari33: That's a good catch. Let me try to incorporate this.

vaibhavtiwari33: Updated

yhl25 (commented):
Overall LGTM! Please run performance and validation tests on this.

rust/numaflow-core/src/transformer.rs
```diff
@@ -202,6 +208,9 @@ impl Transformer {
         }
 
         let message_count = msg_handles.len();
+        let dropped_message_count = Arc::new(AtomicUsize::new(0));
+        // Template controller; cloned per message (and per redrive) so each gets a fresh backoff.
```
adarsh0728, line 212:
```suggestion
        // Template cloned once per message; that clone is reused across redrive so FAIL attempts are not reset.
```

rust/numaflow-core/src/mapper/map/batch.rs
```diff
@@ -69,128 +73,174 @@ impl MapBatchTask {
...
+        //
+        // When no OTel layer is registered, `inject_stage_span` returns non-recording spans
+        // and the sys_metadata copy-on-write is skipped — no need to gate this call.
+        let _stage_spans = otel::inject_stage_spans!(
+            self.msg_handles.iter_mut().map(MessageHandle::message_mut),
+            otel::TraceTopology::current(),
+            otel::TraceStage::Map,
+        );
+
+        let mut retry = RetryController::new(&self.retry_config);
```
adarsh0728, line 93: same as transformer,  let's build retry controller outside the UDF call

adarsh0728 (approved):
approving with some comments

rust/numaflow-core/src/mapper/map/batch.rs
vaibhavtiwari33: Let me take this up in a separate PR.

vaibhavtiwari33:
Screenshot 2026-08-19 at 9 48 20 AM

PR #3572, opened 2026-07-31
fix: Ensure batch consumes it's own EOT

### LINKED ISSUE #3571: Batch map: one batch's EOT can fail the next batch with UDF_PARTIAL_RESPONSE
**Describe the bug**

The batch map client can attribute one batch's End-Of-Transmission (EOT) to the *next* batch, failing a healthy batch with `UDF_PARTIAL_RESPONSE(batch_map)` and blaming the user's UDF.

We ran into this on one of our BatchMap nodes which had a `readBatchSize` less than the average throughput per second and thus the batches were consistently full the moment the previous batch completes, so the insert fired immediately and raced the trailing EOT

**To Reproduce**

The window is timing-dependent, it needs the EOT to be read later than the batch's last response, which happens when the EOT lands in a separate flush or the receiver task is descheduled under load.

See [test](https://github.com/callumc34/numaflow/commit/e8b67064122e88fbb9aed7ee717d6e82f212b1d1#diff-00fcf04385a872ae978899f333cf75e64d16d6ca2f3a9450646f368999578590R779)

**Expected behavior**

A batch's EOT should only ever complete or fail that batch.

**Screenshots**

- Kubernetes: [e.g. v1.18.6]
- Numaflow: [e.g. v0.5.1]

**Additional context**
I have drafted a PR which provides a reproducable unit test for this, and implements a fix: https://github.com/numaproj/numaflow/pull/3572

---

callumc34:

The batch map client can attribute one batch's End-Of-Transmission (EOT) to the next batch, failing a healthy batch with UDF_PARTIAL_RESPONSE(batch_map) and blaming the user's UDF.

This PR makes `batch_once` consume the EOT that delimits its own batch before returning.

### Related issues
Fixes #3571

### Testing
Unit test for reproduction which passes with these changes

### DIFF
--- a/rust/numaflow-core/src/mapper/map/batch.rs
+++ b/rust/numaflow-core/src/mapper/map/batch.rs
@@ -776,98 +776,58 @@ mod tests {
 
     /// `batch_once` must consume the EOT that delimits its own batch before returning.
     #[tokio::test]
-    async fn batch_once_consumes_its_eot_before_returning() -> Result<(), Box<dyn Error>> {
-        struct TrailingEotBatchMap;
-        #[tonic::async_trait]
-        impl batchmap::BatchMapper for TrailingEotBatchMap {
-            async fn batchmap(
-                &self,
-                mut input: tokio::sync::mpsc::Receiver<batchmap::Datum>,
-            ) -> Vec<batchmap::BatchResponse> {
-                let mut responses = Vec::new();
-                while let Some(datum) = input.recv().await {
-                    let mut response = batchmap::BatchResponse::from_id(datum.id);
-                    response.append(batchmap::Message {
-                        keys: Some(datum.keys),
-                        value: datum.value,
-                        tags: None,
-                        nack_options: None,
-                    });
-                    responses.push(response);
-                }
-                // Unmatched ids are logged and skipped by the receiver task, which pushes this
-                // batch's EOT well behind its real responses.
-                for i in 0..2000 {
-                    responses.push(batchmap::BatchResponse::from_id(format!("unmatched-{i}")));
-                }
-                responses
-            }
-        }
+    async fn batch_once_consumes_its_eot_before_returning() {
+        // Capacity covers the request and the EOT, so nothing blocks on the request stream.
+        let (read_tx, _read_rx) = mpsc::channel::<MapRequest>(10);
+        let dummy_handle = tokio::spawn(async {});
+        let abort_handle = Arc::new(AbortOnDropHandle::new(dummy_handle));
 
-        let (shutdown_tx, shutdown_rx) = tokio::sync::oneshot::channel();
-        let tmp_dir = TempDir::new()?;
-        let sock_file = tmp_dir.path().join("late_eot.sock");
-        let server_info_file = tmp_dir.path().join("late_eot-server-info");
-        let (ss, si) = (sock_file.clone(), server_info_file.clone());
-        let handle = tokio::spawn(async move {
-            Server::new(TrailingEotBatchMap)
-                .with_socket_file(ss)
-                .with_server_info_file(si)
-                .start_with_shutdown(shutdown_rx)
-                .await
-                .expect("server failed");
-        });
-        tokio::time::sleep(Duration::from_millis(100)).await;
-
-        let mut client = MapClient::new(create_rpc_channel(sock_file).await?);
-        let connection = UserDefinedBatchMap::create_connection(500, &mut client).await?;
-
-        // Held so the state can be inspected synchronously the moment `batch_once` returns.
-        let senders = Arc::clone(&connection.senders);
+        let senders = Arc::new(Mutex::new(BatchSenderMapState::default()));
         let mapper = UserDefinedBatchMap {
-            batch_size: 500,
-            connection: Arc::new(tokio::sync::Mutex::new(connection)),
+            batch_size: 10,
+            connection: Arc::new(tokio::sync::Mutex::new(super::BatchMapConnection {
+                read_tx,
+                senders: Arc::clone(&senders),
+                _handle: abort_handle,
+            })),
             reconnect_config: None,
         };
 
-        for batch in 0..2 {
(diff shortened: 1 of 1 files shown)


rust/numaflow-core/src/mapper/map/batch.rs
```diff
@@ -743,6 +774,102 @@ mod tests {
...
+                        nack_options: None,
+                    });
+                    responses.push(response);
+                }
+                // Unmatched ids are logged and skipped by the receiver task, which pushes this
+                // batch's EOT well behind its real responses.
+                for i in 0..2000 {
+                    responses.push(batchmap::BatchResponse::from_id(format!("unmatched-{i}")));
+                }
+                responses
```
vaibhavtiwari33: We ideally expect messages returned by the UDF to contain IDs that were originally sent to the UDF as part of the datum, and not random IDs.

vaibhavtiwari33: Could you provide a little more detail regarding how you're using the batch map UDF?

callumc34: Have modified the test so it isn't dependent on a UDF

rust/numaflow-core/src/mapper/map/batch.rs
```diff
@@ -743,6 +774,62 @@ mod tests {
...
+        );
+
+        // Consuming the EOT releases the batch.
+        let eot = senders
+            .lock()
+            .unwrap()
+            .eot
+            .take()
+            .expect("batch should have registered an EOT notifier");
+        eot.send(()).expect("batch should be waiting on the EOT");
```
vaibhavtiwari33, line 835: We're testing the functionality that was added as part of this PR. Could we add a test that fails without the changes in the PR? i.e., a test that I can add in the current main branch as is, and it should compile and fail.

vaibhavtiwari33: IMO the batches in map batch are sent serially unlike map stream, so if the UDF returns responses (Vec\<BatchResponse\>) that follow the following pattern:

* ID of each response (id, Vec\<Message\>) is unique
* ID of each response corresponds to what was sent in Datum

We should not see the batch overlapping issue.

vigith: @vaibhavtiwari33 i think we really had a bug. This is because `batch_once` and `receive_batch_responses` are two different tasks, and the check in `EOT` that reads a map that is not scoped to the batch whose EOT it just received.

vaibhavtiwari33: I do see it as a timing bug, as soon as `batch_once` receives `n` IDs that it was waiting for, it unblocks the next batch to start sending data to server. I'm just trying to think of a scenario where EOT from server would be slower than the last response it sent. Anyways, I do agree explicit EOT would be better

rust/numaflow-core/src/mapper/map/batch.rs
```diff
@@ -470,6 +486,20 @@ impl UserDefinedBatchMap {
...
+                if eot.is_err() {
+                    return vec![Err(map_redrive_error(Status::unavailable(
+                        "batch map stream ended before EOT",
+                    )))];
+                }
+            }
+            _ = cln_token.cancelled() => {
+                return vec![Err(Error::Mapper("batch map operation cancelled".to_string()))];
+            }
+        }
```
vigith, line 513:
```suggestion
          const EOT_WAIT_WARN_INTERVAL: std::time::Duration = std::time::Duration::from_secs(30);
          // Wait until this batch's EOT is ready.
          let mut eot_rx = eot_rx;
          loop {
              tokio::select! {
                  eot = &mut eot_rx => {
                      if eot.is_err() {
                          return vec![Err(map_redrive_error(Status::unavailable(
                              "batch map stream ended before EOT",
                          )))];
                      }
                      break;
                  }
                  _ = cln_token.cancelled() => {
                      return vec![Err(Error::Mapper("batch map operation cancelled".to_string()))];                                                                      
                  }
                  _ = tokio::time::sleep(EOT_WAIT_WARN_INTERVAL) => {
                      warn!(
                          ?EOT_WAIT_WARN_INTERVAL,
                          "all batch map responses received but no End-Of-Transmission yet; \
                           the batch is blocked waiting for the UDF's EOT"
                      );
                  }
              }
          }
```

this `warn!` will help if there is a bug and EOT is never reached

rust/numaflow-core/src/mapper/map/batch.rs
callumc34: > We're testing the functionality that was added as part of this PR. Could we add a test that fails without the changes in the PR? i.e., a test that I can add in the current main branch as is, and it should compile and fail.

The test added in this PR would be failing on main currently.

From what we experienced it didn't happen too frequently, but given the function runs 24/7 it was enough to cause crashes commonly.

rust/numaflow-core/src/mapper/map/batch.rs
callumc34: > Could you provide a little more detail regarding how you're using the batch map UDF?

There's a little more info in the issue: https://github.com/numaproj/numaflow/issues/3571

> We ran into this on one of our BatchMap nodes which had a readBatchSize less than the average throughput per second and thus the batches were consistently full the moment the previous batch completes, so the insert fired immediately and raced the trailing EOT

We weren't really doing anything out of the ordinary was just a common Go batchmap UDF, reading the channel -> make a Redis call -> do some processing -> occasionally, store back into Redis. The function was a little slow taking over a second sometimes and the `readBatchSize` and `readBufferTimeout` were probably not updated accordingly, so we ended up encountering this.

vigith:
@callumc34 could you please run the linter and push again?

callumc34:
> @callumc34 could you please run the linter and push again?

Done

vaibhavtiwari33 (approved):

vigith (approved):

PR #3585, opened 2026-08-05
docs: add ArgoCD ignoreDifferences guidance for autoscaled MonoVertex/Vertex

### LINKED ISSUE #3576: Autoscaled monovertex appears out-of-sync in argocd
**Describe the bug**

Our monovertex is allowed to scale as:

```yaml
spec:
  scale:
    max: 4
    min: 1
```

The current pipeline state is scaled to 4:

```yaml
spec:
  replicas: 4
  scale:
    max: 4
    min: 1
   ...
status:
  desiredReplicas: 4
  lastScaledAt: 2026-08-03T10:24:36Z
  lastUpdated: 2026-08-03T10:42:11Z
  phase: Running
  readyReplicas: 4
  replicas: 4
  updatedReadyReplicas: 4
  updatedReplicas: 4
  ...
```

But ArgoCD says desired state is `replicas: 1`, showing out-of-sync.

Image

**To Reproduce**

1. Deploy monovertex with argo, scale 1-4.
2. Apply load so it scales to 4.
3. Observe argocd sync status.

**Expected behavior**
Argo in sync when scaled.

- Numaflow: 1.8.1

---

nirajkumarbarot:
## Problem
   Autoscaled MonoVertex/Vertex resources appear as "OutOfSync" in ArgoCD
   because the autoscaler updates spec.replicas via the scale subresource,
   which diverges from the value committed in Git.

   ## Fix
   Adds a new docs page explaining how to configure `ignoreDifferences`
   in the ArgoCD Application spec so this expected drift isn't flagged
   as a sync error.

   Fixes #3576

### DIFF
--- a/docs/operations/argocd.md
+++ b/docs/operations/argocd.md
@@ -14,7 +14,7 @@ healthy and scaling correctly.
 Add an `ignoreDifferences` entry to your ArgoCD `Application` so it stops
 comparing `spec.replicas` for autoscaled resources:
 
-\`\`\`yaml
+```yaml
 apiVersion: argoproj.io/v1alpha1
 kind: Application
 metadata:
--- a/docs/operations/argocd.md
+++ b/docs/operations/argocd.md
@@ -1,0 +1,39 @@
+# Using Numaflow with ArgoCD
+
+Numaflow's autoscaler updates `spec.replicas` on `MonoVertex` and `Vertex`
+objects directly (via the Kubernetes scale subresource), the same way a
+Horizontal Pod Autoscaler updates `spec.replicas` on a `Deployment`.
+
+If you manage Numaflow resources with ArgoCD, this can cause ArgoCD to
+report the resource as `OutOfSync`, because the live `replicas` value no
+longer matches what's committed in Git — even though the pipeline is
+healthy and scaling correctly.
+
+## Recommended fix
+
+Add an `ignoreDifferences` entry to your ArgoCD `Application` so it stops
+comparing `spec.replicas` for autoscaled resources:
+
+\`\`\`yaml
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: my-numaflow-app
+spec:
+  ignoreDifferences:
+    - group: numaflow.numaproj.io
+      kind: MonoVertex
+      jsonPointers:
+        - /spec/replicas
+    - group: numaflow.numaproj.io
+      kind: Vertex
+      jsonPointers:
+        - /spec/replicas
+  syncPolicy:
+    syncOptions:
+      - RespectIgnoreDifferences=true
+\`\`\`
+
+This tells ArgoCD to treat autoscaler-driven replica changes as expected
+drift instead of a sync error, while still tracking real spec changes
+made in Git.
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -157,6 +157,7 @@ nav:
   - Operator Manual:
       - Releases ⧉: "operations/releases.md"
       - operations/installation.md
+      - Using ArgoCD: operations/argocd.md
       - Validating Webhook: operations/validating-webhook.md
       - Configuration:
           - Controller Configuration: "operations/controller-configmap.md"


vigith:
@nirajkumarbarot please sign your commmit

docs/operations/argocd.md
```diff
@@ -0,0 +1,39 @@
...
+report the resource as `OutOfSync`, because the live `replicas` value no
+longer matches what's committed in Git — even though the pipeline is
+healthy and scaling correctly.
+
+## Recommended fix
+
+Add an `ignoreDifferences` entry to your ArgoCD `Application` so it stops
+comparing `spec.replicas` for autoscaled resources:
+
+\`\`\`yaml
```
vigith:
```suggestion
```yaml
```

docs/operations/argocd.md
```diff
@@ -0,0 +1,39 @@
...
+      jsonPointers:
+        - /spec/replicas
+    - group: numaflow.numaproj.io
+      kind: Vertex
+      jsonPointers:
+        - /spec/replicas
+  syncPolicy:
+    syncOptions:
+      - RespectIgnoreDifferences=true
+\`\`\`
```
vigith, line 35: same here

adarsh0728:
hi @nirajkumarbarot , thank you for taking a [look at this](https://github.com/numaproj/numaflow/issues/3576), we are still figuring out if it requires a fix from numaflow side (or just doc update as you tried out)

docs/operations/argocd.md
nirajkumarbarot: Thank you so much for your review @vigith

nirajkumarbarot:
> hi @nirajkumarbarot , thank you for taking a [look at this](https://github.com/numaproj/numaflow/issues/3576), we are still figuring out if it requires a fix from numaflow side (or just doc update as you tried out)

Thanks for taking a look! Happy to go either way if you all decide this needs a code-side fix (e.g. Numaflow setting ignoreDifferences-friendly defaults, or excluding spec.replicas from what ArgoCD tracks), let me know and I can take a shot at that too. For now I'll get the doc PR cleaned up (signing commits + applying the suggestions) in case it's still useful either way.

nirajkumarbarot:
@adarsh0728 - totally fine either way. This PR documents the workaround on the ArgoCD config side, since spec.replicas on MonoVertex/Vertex is driven by the scale subresource (same pattern as HPA + Deployment). If you all land on wanting a Numaflow-side fix instead (or in addition), happy to help with that too - just let me know the direction and I'll adjust.

Will update this PR shortly with the signed commits.

nirajkumarbarot:
> @nirajkumarbarot please sign your commmit

@vigith - I'll re-push with signed commits (recreating the branch since I can't amend commit messages from the web UI) and apply both formatting suggestions in the process.

docs/operations/argocd.md
```diff
@@ -0,0 +1,39 @@
+# Using Numaflow with ArgoCD
+
+Numaflow's autoscaler updates `spec.replicas` on `MonoVertex` and `Vertex`
+objects directly (via the Kubernetes scale subresource), the same way a
+Horizontal Pod Autoscaler updates `spec.replicas` on a `Deployment`.
+
+If you manage Numaflow resources with ArgoCD, this can cause ArgoCD to
+report the resource as `OutOfSync`, because the live `replicas` value no
+longer matches what's committed in Git — even though the pipeline is
```
whynowy, line 9: This is the most interesting part to me... if `replicas` is not committed to Github, why Argo CD reports out of sync? What kind of logic is Argo CD following to deal with those fields?

nirajkumarbarot: Hi @whynowy 
Good question! This is actually a documented ArgoCD gotcha - Argo's own diffing docs call out "a controller or mutating webhook altering the object after it was submitted to Kubernetes in a manner which contradicts Git" as a known cause of false OutOfSync: https://argo-cd.readthedocs.io/en/stable/user-guide/diffing/

The mechanics here: spec.replicas on MonoVertex/Vertex is exposed via the Kubernetes scale subresource (+kubebuilder:subresource:scale:specpath=.spec.replicas,...), which means it has a schema default. Even when you omit replicas from your Git manifest, ArgoCD still has to materialize a full "desired" object to diff against and for that, it falls back to the CRD's schema default (e.g. 1), not "no opinion on this field." So the computed desired state ends up as replicas: 1 even though you never wrote it in Git.

Then, after the object is created, Numaflow's own autoscaler writes directly to spec.replicas via the scale subresource ,the same mechanism HPA uses on Deployments - completely outside any Git/kubectl apply flow. That live value now diverges from ArgoCD's computed desired value, so it gets flagged OutOfSync, even though nothing is actually wrong.

This is the same category of issue people hit with things like Deployment.spec.strategy.rollingUpdate.maxUnavailable being a server-side default, except here the divergence comes from our own controller rather than the Kubernetes API server.

Happy to add a sentence or two to the doc spelling this out, if that'd help future readers.

whynowy: > Hi @whynowy Good question! This is actually a documented ArgoCD gotcha - Argo's own diffing docs call out "a controller or mutating webhook altering the object after it was submitted to Kubernetes in a manner which contradicts Git" as a known cause of false OutOfSync: https://argo-cd.readthedocs.io/en/stable/user-guide/diffing/
> 
> The mechanics here: spec.replicas on MonoVertex/Vertex is exposed via the Kubernetes scale subresource (+kubebuilder:subresource:scale:specpath=.spec.replicas,...), which means it has a schema default. Even when you omit replicas from your Git manifest, ArgoCD still has to materialize a full "desired" object to diff against and for that, it falls back to the CRD's schema default (e.g. 1), not "no opinion on this field." So the computed desired state ends up as replicas: 1 even though you never wrote it in Git.
> 
> Then, after the object is created, Numaflow's own autoscaler writes directly to spec.replicas via the scale subresource ,the same mechanism HPA uses on Deployments - completely outside any Git/kubectl apply flow. That live value now diverges from ArgoCD's computed desired value, so it gets flagged OutOfSync, even though nothing is actually wrong.
> 
> This is the same category of issue people hit with things like Deployment.spec.strategy.rollingUpdate.maxUnavailable being a server-side default, except here the divergence comes from our own controller rather than the Kubernetes API server.
> 
> Happy to add a sentence or two to the doc spelling this out, if that'd help future readers.

Thanks for all the information! An immediate question from me is, why did we set the `default` for the [replicas](https://github.com/numaproj/numaflow/blob/main/pkg/apis/numaflow/v1alpha1/mono_vertex_types.go#L497) to 1 (This is a question to me... )? I think that is not needed, the controller will compute the value anyways. If we remove the default `1`, will it solve the problem?

whynowy: I believe that will do - we use minimal CRDs internally, which has no default value, I never saw the Argo CD complains.

nirajkumarbarot: Makes sense to me! if there's no schema default, ArgoCD has nothing to materialize for the omitted field, so there's no drift to flag once the autoscaler updates it. I'll dig into where replicas gets its +kubebuilder:default=1 (both MonoVertex and Vertex, since they're both scale-subresource-enabled) and push a commit removing it, so we can see what the diff looks like. Will report back here once it's up.

nirajkumarbarot: > I believe that will do - we use minimal CRDs internally, which has no default value, I never saw the Argo CD complains.

Opened a separate PR for the CRD default removal, since it's a code change independent of the docs fix: https://github.com/numaproj/numaflow/pull/3633

PR #3600, opened 2026-08-17
fix: stricter health check for pipeline/mvtx in UI

### LINKED ISSUE #3294: New pipeline shows healthy despite pods not started
**Describe the bug**

We experienced when recreating pipelines that a cluster had reached max cpu requests (`0/N nodes are available: N Insufficient cpu`). This caused `FailedScheduling` of vertex pods. Vertex was marked `not ready` and pod was `Pending` (all container in pod had ready=false).

But *pipeline showed healty in Argocd, despite not being able to start.*

Only after a long delay, the bad vertex state propagated to make the pipeline show unhealthy.

**To Reproduce**
Create new pipeline on cluster with cpu requests reached.

**Expected behavior**
On new pipeline creation, pipeline should not get into healthy state, before all vertices have become ready/healty.

- Numaflow: 1.7.2

---

suryapratap-01:

Pipelines and MonoVertices could appear healthy when their pods were still Pending because Pending pods often have no container statuses.

Health now requires existing pod checks to pass **and** enough pods to be Ready. A pod is Ready only when it is Running, not terminating, and all application and init containers are Ready.

Pending scheduling errors such as `Insufficient cpu` are shown in the controller status and UI.

### Related issues

Fixes #3294

### Testing

- `go test ./pkg/reconciler/... ./server/apis/v1`
- Covers Pending/Failed/Unknown pods, missing readiness, scheduling messages, restarts, and UI health.

### Notes for reviewers

N.A

Screenshot 2026-08-17 at 5 41 52 PM
Screenshot 2026-08-17 at 6 00 26 PM


adarsh0728:
Currently I think we treat `Healthy` as default state. Let's make it stricter by making `Unhealthy` as the default (the states/checks that we don't cover). This will be easier to implement.

pkg/reconciler/util.go
```diff
@@ -69,6 +89,28 @@ func isPodHealthy(pod *corev1.Pod) (healthy bool, reason string, message string,
...
+			return c.Reason, msg
+		}
+	}
+	msg := pod.Status.Message
+	if msg == "" {
+		msg = "Pod is in Pending phase"
+	}
+	return "Pending", msg
+}
+
```
adarsh0728: let's check for healthy state

pkg/reconciler/util.go
```diff
@@ -43,6 +43,35 @@ func CheckPodsStatus(pods *corev1.PodList) (healthy bool, reason string, message
...
+// CheckPodsStatusWithReadiness requires both the existing pod diagnostics and
+// the desired number of Ready pods to pass before reporting health.
+func CheckPodsStatusWithReadiness(pods *corev1.PodList, desiredReplicas int) (healthy bool, reason string, message string, transientUnhealthy bool) {
+	healthy, reason, message, transientUnhealthy = CheckPodsStatus(pods)
+	if !healthy {
+		return healthy, reason, message, transientUnhealthy
+	}
+
+	readyPods := NumOfReadyPods(*pods)
+	if readyPods < desiredReplicas {
```
adarsh0728: when its autoscaling, number of ready replicas may not be >= desired replicas, but it doesn't mean that its unhealthly. We need to think how to handle this use case.

suryapratap-01: I agree, that's why, rather than introducing a new state like `Progressing`, only Pending pods with `PodScheduled=False` are marked unhealthy. Scale-up Pending pods still count as healthy.

I think this fixes the problem surgically, rather than going through the entire process of adding a new state.

suryapratap-01: Have addressed this in the new code

server/apis/v1/health.go
```diff
@@ -184,6 +184,19 @@ func isVertexHealthy(h *handler, ns string, pipeline string, vertex *dfv1.Vertex
...
+	if cond := vertex.Status.GetCondition(dfv1.VertexConditionPodsHealthy); cond != nil && cond.Status == metav1.ConditionFalse {
+		msg := cond.Message
+		if msg == "" {
+			msg = fmt.Sprintf("Vertex %q pods are not healthy", vertex.Name)
+		}
+		return false, &resourceHealthResponse{
+			Message: msg,
+			Code:    "V3",
+		}, nil
+	}
```
adarsh0728, line 193: update [isVertexHealthy](https://github.com/suryapratap-01/numaflow/blob/deb52566e6c994bc2c22344ed9fe65c71a507462/server/apis/v1/health.go#L167) function comments

adarsh0728 (approved):

adarsh0728:
Approved. Let's plan to test different scenarios once before the release. cc @suryapratap-01

PR #3605, opened 2026-08-21
feat: introduce endpoint field for http source

vaibhavtiwari33:
Provide an endpoint field for http source. 
Currently http source listens on one of the following fixed endpoints (depending on the context):
- `/vertices/<vertex-name>` for pipelines
- `/vertices/<mvtx-name>` for monovertices

This change aims to provide a configuration to allow customizing endpoint the http source listens to:
- `/vertices/<endpoint-value>`

Useful to allow keeping final endpoint consistent when monovertex deployment is being managed by a monovertexrollout for example (each subsequent mvtx deployment has an incrementing number appended to its name, changing the exposed endpoint on each progressive rollout)

### Testing

Screenshot 2026-08-21 at 10 01 44 AM


pkg/reconciler/validator/pipeline_validate.go
```diff
@@ -621,6 +621,39 @@ func validateSource(source dfv1.Source) error {
...
+	return nil
+}
+
+// httpEndpointRegex matches a valid HTTP source endpoint: one or more path
+// segments of URL-safe characters separated by single slashes, with no leading
+// or trailing slash. This rejects empty values, leading slashes (which would
+// produce an unreachable "/vertices//..." route), and path-parameter syntax
+// such as "{id}" or ":id" that would panic the Axum router at pod startup.
+var httpEndpointRegex = regexp.MustCompile(`^[A-Za-z0-9\-._~]+(/[A-Za-z0-9\-._~]+)*$`)
+
```
yhl25, line 640: is there any existing library method that we can resuse?

vaibhavtiwari33: There doesn't seem to be a library that validates against the requirements here:
- no empty string
- no leading/trailing '/'
- no subsequent '/' (eg: `asdf//asdf` )
- no special / query / whitespace characters
- no '{', '}', ':' characters

Simplest one to use would be `url.ParseRequestURI` but that supports leading '/', and '{', '}', ':' characters. 

I can probably simplify the requirement to allow leading '/', but I don't think we should allow other special characters.

yhl25 (commented):
Should we consider not appending the vertex name in the url? Because when we create the service vertex name is already considered.

yhl25 (approved):

vaibhavtiwari33:
> Should we consider not appending the vertex name in the url? Because when we create the service vertex name is already considered.

We can inherently make the endpoint a constant value, that is a fair ask, the only concern is that we'll break the endpoint for anyone already using an http source (apart from our e2e tests). With an optional endpoint value we maintain the backward compatibility.

cc: @vigith @adarsh0728 pls let me know if you guys have any inputs regarding this as well.

adarsh0728:
> > Should we consider not appending the vertex name in the url? Because when we create the service vertex name is already considered.
> 
> We can inherently make the endpoint a constant value, that is a fair ask, the only concern is that we'll break the endpoint for anyone already using an http source (apart from our e2e tests). With an optional endpoint value we maintain the backward compatibility.
> 
> cc: @vigith @adarsh0728 pls let me know if you guys have any inputs regarding this as well.

yeah, optional endpoint option looks simpler and backward compatible.

pkg/apis/numaflow/v1alpha1/http_source.go
```diff
@@ -28,6 +28,10 @@ type HTTPSource struct {
 	// To enable plain HTTP, set ports.http explicitly.
 	// +optional
 	Ports *Ports `json:"ports,omitempty" protobuf:"bytes,3,opt,name=ports"`
+	// Endpoint at which the http source will be available.
```
adarsh0728: nit: Endpoint is actually the path suffix appended to `/vertices/` , we can update this doc(can be misleading)

adarsh0728 (approved):

vigith (approved):

pkg/apis/numaflow/v1alpha1/http_source.go
vaibhavtiwari33: Updated

Issue #2759, opened 2025-06-27
if InterstepBufferService and Pipeline are both deleted, there is no guarantee that the Pipeline Cleanup Job will work


juliev0:
**Describe the bug**

This may not be very high priority but I'm writing an issue for it anyway.

If an isbsvc and its Pipelines are deleted at the same time, the isbsvc will delete the [client auth Secret](https://github.com/numaproj/numaflow/blob/v1.5.0/pkg/reconciler/isbsvc/installer/jetstream.go#L381-L395) which it owns. Meanwhile, the Pipeline Cleanup Job pod can be run simultaneously, attempt to use that same Secret, and fail to find it. Thus, the Job Pod is errored.

In the end, it doesn't matter too much to fail to delete buffers and buckets if the isbsvc itself no longer exists.  So, the main issue is the pod getting left around.  

Pod Container status:
```
  containerStatuses:
  - image: docker.intuit.com/quay-rmt/numaproj/numaflow:v1.4.3
    imageID: ""
    lastState: {}
    name: numa
    ready: false
    restartCount: 0
    started: false
    state:
      waiting:
        message: secret "isbsvc-fmagent-login-async-isbsvc-0-js-client-auth" not found
        reason: CreateContainerConfigError
    volumeMounts:
    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
      name: kube-api-access-m6fpk
      readOnly: true
      recursiveReadOnly: Disabled
    - mountPath: /var/run/secrets/eks.amazonaws.com/serviceaccount
      name: aws-iam-token
      readOnly: true
      recursiveReadOnly: Disabled
```

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

>For quick help and support, join our [slack](https://numaproj.slack.com/join/shared_invite/zt-19svuv47m-YKHhsQ~~KK9mBv1E7pNzfg#/shared-invite/email) channel.

PR #3611, opened 2026-08-25
fix: bound pipeline buffer clean up job lifetime

will-flemmer:
### What this PR does

Sets `activeDeadlineSeconds` on the pipeline buffer clean up job (`isbsvc-delete`, job type `cln`) so it always reaches a terminal state.

### The problem

Refs #2759. When an ISB Service and its Pipelines are deleted together, the ISB Service can delete the `js-client-auth` Secret it owns before the clean up job runs. The job Pod then can't start:

```
state:
  waiting:
    message: secret "isbsvc-<name>-js-client-auth" not found
    reason: CreateContainerConfigError
```

Because the container never *starts*, the Pod never reaches a terminal state and `restartCount` stays at 0. The job stays `active`, so `ttlSecondsAfterFinished: 30` never applies and the job is never garbage collected. Meanwhile the kubelet retries container creation indefinitely.

`backoffLimit: 20` doesn't rescue this. `.status.failed` only advances when something external evicts the Pending Pod, which we measured at a median of ~86 hours per increment — so these jobs would take months to terminate on their own.

### Why activeDeadlineSeconds

It's evaluated by the job controller on wall-clock time and doesn't depend on Pod state, so it terminates the job regardless of how the deletion ordering races. Per the Kubernetes docs it also takes precedence over `backoffLimit`. The clean up job only has work to do while the ISB Service is alive, so one that can't finish within 10 minutes never will.

Set at the `cln` call site rather than in `buildISBBatchJob`, so the `cre` and `del` jobs are unaffected.

### Testing

`Test_cleanupBuffers` asserts the deadline is set. Verified the assertion fails without the change.

### DIFF
--- a/pkg/reconciler/pipeline/controller.go
+++ b/pkg/reconciler/pipeline/controller.go
@@ -57,6 +57,8 @@ const (
 	deprecatedFinalizerName = dfv1.ControllerPipeline
 
 	pauseTimestampPath = `/metadata/annotations/numaflow.numaproj.io~1pause-timestamp`
+
+	cleanUpJobActiveDeadlineSeconds = int64(600)
 )
 
 // pipelineReconciler reconciles a pipeline object.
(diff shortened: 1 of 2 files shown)


whynowy (approved):
LGTM

PR #3614, opened 2026-08-31
fix: harden daemon error cache retention (mvtx) for failed trial rollouts

adarsh0728:

- Aim is to retain MonoVertex runtime errors in the daemon’s in memory cache after pods scale to zero, so trial failures remain visible in the UI (Errors tab) after rollout([using progressive rollout strategy](https://github.com/numaproj/numaplane/blob/f6b9b54ecf6e6629fb39304ed9249a946dda5869/README.md?plain=1#L17)) failure (pods scale down to 0, daemon pod keeps on running for trial MonoVertex) 

- Speed up error capture with 3s worker discovery (while idle) and a 10s scrape interval

- Add unit tests for cache retention and an e2e test that pauses a MonoVertex and verifies errors still flow through the daemon and UX server.

### Related issues
Fixes #

### Testing

Internal testing yet to be done.

This is best effort pull solution (from daemon to numa)


adarsh0728:
This will need changes on Numaplane side.

PR #3615, opened 2026-09-01
feat: SQS source can consume multiple queues

### LINKED ISSUE #3594: feat: support reading from multiple queues in the built-in SQS source
## Problem Statement

Numaflow's built-in SQS source can only consume from a **single** queue. The `SqsSource` spec exposes exactly one `queueName`/`queueOwnerAWSAccountID`/`awsRegion`, and the runtime is pinned to that one queue.

Users who want to feed several SQS queues into one pipeline are forced to run a separate source vertex (or an entire separate pipeline) per queue, or add an upstream aggregator to merge queues — duplicating config, autoscaling, and operational overhead.

We should extend the built-in SQS source to read from **multiple queues** in a single source.

suryapratap-01:

Adds multi-queue SQS source support via `queueNames` (comma-separated) so one source can consume many queues that share region, account, credentials/`assumeRole`, `endpointUrl`, and poll settings. `queueName` and `queueNames` are mutually exclusive.

Messages from configured queues are merged with no cross-queue order. Ack/nack route by origin: offsets are `<queue-name>:<receipt-handle>`, so routing survives `queueNames` reordering. Offsets are only decoded in-process by the source vertex.

A read returns as soon as any queue has data; idle siblings do not hold the batch. Messages that arrive after the read returned are buffered and served first on the next read.

Receive failures are counted per queue; ten consecutive failures take the vertex down (same idea as Kafka). A failing queue does not discard messages already received from healthy queues.

Every message carries origin as **system metadata** — group `sqs`, key `queue_name` — for UDFs/UDSinks. Origin is not written to headers or user metadata. A builtin SQS sink still writes to the destination `queueName` in its spec.

| Field | Where | Notes |
|---|---|---|
| `queueName` | source spec | Single-queue selector |
| `queueNames` | source spec | Comma-separated multi-queue selector; XOR with `queueName` |
| `queue_name` | system metadata, group `sqs` | Origin on every message |

```yaml
source:
  sqs:
    queueNames: "orders-queue,refunds-queue,replay-queue"
    awsRegion: "us-east-1"
    queueOwnerAWSAccountID: "111111111111"
    visibilityTimeout: 30
    maxNumberOfMessages: 10
    waitTimeSeconds: 20
```

Route by origin (map UDF → conditional edges):

```python
def handler(keys: list[str], datum: Datum) -> Messages:
    raw = datum.system_metadata.value("sqs", "queue_name")
    queue = raw.decode("utf-8") if raw else "unknown"
    return Messages(Message(value=datum.value, keys=keys, tags=[queue]))
```

Docs: `docs/user-guide/sources/sqs.md`. Builtin `sink.log` does not print system metadata — use a UDF or UDSink for origin.

### Related issues

Fixes #3594

### Testing

- `cargo test -p numaflow-sqs --lib`
- `cargo test -p numaflow-core --features sqs-tests --lib source::sqs`
- Local [ElasticMQ](https://github.com/softwaremill/elasticmq) (`endpointUrl` e.g. `http://localhost:9324`): same YAML as prod, dummy account/region, no STS. URLs resolve, depths drop, UDSink/UDF sees `queue_name`; bad specs (both fields, empty, duplicates) fail without real SQS.
- AWS (~10K msgs / 3 queues) with `assumeRole`: URLs resolve, payloads appear, origin in system metadata only.

#### Live cluster E2E (2026-09-01)

Controller image + `NUMAFLOW_IMAGE` rolled to this feat; SQS CRs recreated. Three queues, `assumeRole`, IRSA. UDSink logged `sqs`/`queue_name`. Map UDF (above) tagged from that stamp for conditional edges.

| # | Scenario | Result |
|---|---|---|
| 1 | Bad specs (both fields / empty / duplicate / neither) | **PASS** — InvalidSpec, 0 pods |
| 2 | One MonoVertex, `queueNames` ×3 + origin stamp | **PASS** — 2000/queue; last 80 lines q1=27 q2=3 q3=50; depths 2000→1931/1991/1863 in ~70s @ rateLimit 20 |
| 3 | Idle queues must not stall a busy one | **PASS** — q2/q3 empty, q1=2423, `waitTimeSeconds: 20`; first msg ~2s (not 20s); 150 msgs/49s; max gap 1.173s; gaps ≥15s: 0 |
| 4 | Fair share at `readBatchSize: 1` | **PASS** — 90s counts 100/100/100 |
| 5 | Single-queue `queueName` | **PASS** — 200 msgs from that queue only, 0 stamp mismatches; sibling depths unchanged |
| 6 | Origin survives pipeline (UDF tags → conditional sinks) | **PASS** — after 2000×3: 413/0/0 in Q1_out, 0/250/0 in Q2_out, 0/0/413 in Q3_out |
| 7 | Unknown queue fails before consume | **PASS** — CrashLoop / `NonExistentQueue` at `GetQueueUrl`; real queues not drained |
| 8 | Ack routes to origin queue (`<queue-name>:<receipt>`) | **PASS** — after delete + 35s (`VT=30`) depths stayed down; unique body did not reappear after 40s |
| 9 | One dead queue → 10 consecutive receive fails → crash | **SKIPPED** — needs temporary IAM deny of `ReceiveMessage` on one queue; covered by unit tests |

- Queues listed in `queueNames` must share account, region, credentials/`assumeRole`, and poll settings. Different accounts/regions → separate vertices.
- Per-queue receive failure budget is 10 consecutive failures, then the vertex exits.

Screenshot 2026-09-01 at 2 18 17 PM
Screenshot 2026-09-01 at 2 02 52 PM
Screenshot 2026-09-01 at 11 47 28 AM
Screenshot 2026-09-01 at 11 46 37 AM


cosmic-chichu (approved):

yhl25 (approved):

rust/extns/numaflow-sqs/src/source.rs
```diff
@@ -315,44 +389,77 @@ impl SqsActor {
...
+                    })
+                    .collect::<Vec<_>>()
+                    .join(", ");
+                error!(
+                    queue_url = self.queue_url,
+                    failures, "SQS failed to delete one or more messages"
+                );
+                return Err(SqsSourceError::from(Error::Other(format!(
+                    "SQS failed to delete one or more messages: {failures}"
+                ))));
```
vigith, line 459: if we return error, won't the whole batch be retried? if yes, then will the previous successful acks fail on retry?

vigith (changes requested):

rust/extns/numaflow-sqs/src/source.rs
suryapratap-01: hey @vigith, i already had tested these scenarios before with both FIFO and standard queues in us-east-2.

yes, the entire batch is retried. `ack_with_retry` resends the same offset list every 100ms, and this error is mapped to a retryable `Error::Source`.

the messages that were successfully deleted don’t fail on retry either. `DeleteMessageBatch` behaves idempotently here, so re-deleting the same receipt handles returns success instead of `ReceiptHandleIsInvalid`.

results:

1. delete 9 out of 10, then retry all `10 → ok=10 fail=[]`
2. delete the same receipt handle twice → both requests return `HTTP 200`
3. with 15 messages, delete the first 10, then retry all 15, which covers the `chunks(10) `split → `ok=15 fail=[]`

suryapratap-01: hey @vigith , i already had tested these scenarios with both FIFO and standard queues in `us-east-2`.

yes, the entire batch is retried. `ack_with_retry` resends the same offset list every 100ms, and this error is mapped to a retryable `Error::Source`.

the messages that were successfully deleted don’t fail on retry either. `DeleteMessageBatch` behaves idempotently here, so re-deleting the same receipt handles returns success instead of `ReceiptHandleIsInvalid`.

results:

* delete 9 out of 10, then retry all 10 → `ok=10 fail=[]`
* delete the same receipt handle twice → both requests return HTTP 200
* with 15 messages, delete the first 10, then retry all 15, which covers the `chunks(10)` split → `ok=15 fail=[]`

vigith (approved):

PR #3630, opened 2026-09-03
ci: add weekly grouped dependabot config for go, npm, cargo, and github-actions

vigith:
Too many dependabot PRs

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,7 @@ updates:
       interval: "weekly"
     groups:
       go-dependencies:
+        applies-to: security-updates
         patterns:
           - "*"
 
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,41 @@
...
+updates:
+  # Go modules (root)
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      go-dependencies:
+        patterns:
+          - "*"
```
vaibhavtiwari33, line 13:
```suggestion
      go-dependencies:
        applies-to: security-updates
        patterns:
          - "*"
```

Based on the documentation [here](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#groups--), under `groups`, the `applies-to` when unspecified defaults to version updates. We should limit it to security updates

vaibhavtiwari33 (approved):

yhl25 (approved):

PR #3633, opened 2026-09-04
fix: remove default replicas value from MonoVertex/Vertex CRD

nirajkumarbarot:
## Problem
`spec.replicas` on MonoVertex/Vertex has a `+kubebuilder:default=1` (and the same default baked into the generated CRD YAMLs, including the `minimal` variant). Since replicas is managed via the scale subresource by the autoscaler, this schema default causes ArgoCD to materialize a "desired" value of `replicas: 1` even when the field is omitted from Git - leading to a false OutOfSync once the autoscaler updates the live value.

## Fix
Removes the default from:
- Go struct markers in `vertex_types.go` / `monovertex_types.go`
- Generated CRD manifests in `config/base/crds/full/`
- Generated CRD manifests in `config/base/crds/minimal/`

The controller already treats a `nil`/unset `Replicas` as `1` in Go (`getReplicas()` on both `Vertex` and `MonoVertex`), so removing the schema-level default doesn't change runtime behavior.

Pipeline's CRD does not define its own `spec.replicas` (each vertex's replica count is tracked on its own MonoVertex/Vertex object), so no changes were needed there.

Discussed in #3585.
Related: #3576

### DIFF
--- a/config/base/crds/full/numaflow.numaproj.io_monovertices.yaml
+++ b/config/base/crds/full/numaflow.numaproj.io_monovertices.yaml
@@ -2871,7 +2871,6 @@ spec:
               priorityClassName:
                 type: string
               replicas:
-                default: 1
                 format: int32
                 type: integer
               resourceClaims:
--- a/config/base/crds/full/numaflow.numaproj.io_vertices.yaml
+++ b/config/base/crds/full/numaflow.numaproj.io_vertices.yaml
@@ -2533,7 +2533,6 @@ spec:
               priorityClassName:
                 type: string
               replicas:
-                default: 1
                 format: int32
                 type: integer
               resourceClaims:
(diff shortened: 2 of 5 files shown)


vigith:
please run `make codegen` and also make the DCO pass

vaibhavtiwari33:
@nirajkumarbarot please don't make CRD changes individually for this kubebuilder annotation marker change. You can simply run `make codegen` to get the desired effect.

nirajkumarbarot:
> @nirajkumarbarot please don't make CRD changes individually for this kubebuilder annotation marker change. You can simply run `make codegen` to get the desired effect.

Thanks @vaibhavtiwari33. That makes sense, that's a much cleaner way to do it. I am currently working entirely through the GitHub web UI without a local dev environment set up, so I can't run make codegen directly. 
If it is easier, happy to have a maintainer/you run make codegen and push the regenerated files on top of my branch; I would be happy and thankful with edits from maintainers.

PR #3638, opened 2026-09-10
fix: daemon should not be recreated on scale changes

adarsh0728:

Daemon was getting recreated when scale.min / scale.max changed on progressive failure, hence persistence of errors in the UI was not happening. This occurred because reconciler hashes the full daemon deployment spec and deletes/recreates the deployment when the hash changes. Hash changed because daemon deployment object embeds the full spec in NUMAFLOW_MONO_VERTEX_OBJECT.

### Related issues
Fixes #

### Testing
event consumer test 5 is the trial mvtx - both application/worker pod and daemon pod running

Screenshot 2026-09-10 at 3 24 37 PM

post analysis failure, worker pod scaled down as scale.min / scale.max is set to 0 - and same daemon pod is still running, no recreation

Screenshot 2026-09-10 at 3 27 49 PM

errors visible even when there are no pods

Screenshot 2026-09-10 at 3 28 44 PM
Screenshot 2026-09-10 at 3 28 52 PM

Any change in lookbackSeconds will recreate daemon, which is okay as it is used in metrics calculation

### DIFF
--- a/pkg/apis/numaflow/v1alpha1/const.go
+++ b/pkg/apis/numaflow/v1alpha1/const.go
@@ -190,7 +190,6 @@ const (
 	DefaultCooldownSeconds          = 90  // Default cooldown seconds after a scaling operation
 	DefaultZeroReplicaSleepSeconds  = 120 // Default sleep time in seconds after scaling down to 0, before peeking
 	DefaultMaxReplicas              = 50  // Default max replicas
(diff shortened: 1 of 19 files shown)


whynowy:
The way to detect the number of replicas from daemon can be switched to using DNS resolver, this way it does not need to pass the min/max to the daemon.

whynowy (approved):

PR #3640, opened 2026-09-10
feat: Introduce a KVStoreFactory trait

vaibhavtiwari33:
Side-input crate currently hardcodes kv store creation using jetstream context. 
This change aims to not only generalize that but also introduce a KVStoreFactory that allows extensibility into the factory creation for different types of KVStores. The respective ISBFactories can then hold the corresponding KVStore factories instead of a direct context/kv stores. 
So, in-theory, we can use a different kv store (eg: jetstream) with a different buffer (eg: pulsar)

Related: https://github.com/numaproj/numaflow/issues/2343

### Testing

Working sideinput example that updates sideinput file with current time

Screenshot 2026-09-11 at 5 24 17 PM


rust/numaflow-core/src/pipeline/isb/jetstream/factory.rs
```diff
@@ -127,6 +105,6 @@ mod tests {
         let factory = JetStreamFactory::new(context.clone());
 
         // Verify the context is accessible
-        assert!(std::ptr::eq(factory.context(), &factory.context));
+        assert!(std::ptr::eq(factory.context(), factory.context()));
```
vigith: what are we testing here? just curious :)

vigith (approved):

rust/numaflow-core/src/pipeline/isb/jetstream/factory.rs
vaibhavtiwari33: Just that the js context was created and it doesn't panic. Let me remove this.

Issue #3651, opened 2026-09-16
enforce that the PVC is `ReadWriteOncePod` for reduce pods


vigith:
> Let's enforce that the PVC is `ReadWriteOncePod` for reduce pods so we can remove fencing. `ReadWriteOncePod` was introduced in K8s 1.22, so should be well adopted. 

 _Originally posted by @vigith in [#3192](https://github.com/numaproj/numaflow/issues/3192#issuecomment-3849865195)_

We introduced `ReadWriteOncePod` in https://github.com/numaproj/numaflow/pull/2865 and  then reverted it in https://github.com/numaproj/numaflow/pull/3070.

I think we should fix it.

vigith:
also relates to https://github.com/numaproj/numaflow/issues/3648

PR #3656, opened 2026-09-18
feat: allow a startupProbe on the numa container. Fixes #3648

### LINKED ISSUE #3648: Allow a startupProbe on the numa and udf containers
# Summary

`ContainerTemplate` and the UDF `Container` type expose `livenessProbe` and `readinessProbe` but not `startupProbe`, so a slow-booting vertex has to buy that time out of its steady-state liveness settings.

On reduce vertices the two timeouts actively conflict. A reduce vertex waits up to 5 minutes for a stale fence file before taking the WAL lock by force ([reduce_forwarder.rs#L513](https://github.com/numaproj/numaflow/blob/v1.8.4/rust/numaflow-core/src/pipeline/forwarder/reduce_forwarder.rs#L513)), and serves no probe endpoint while it waits. The liveness budget is `initialDelaySeconds + (failureThreshold - 1) * periodSeconds`, which on the [defaults](https://github.com/numaproj/numaflow/blob/v1.8.4/pkg/apis/numaflow/v1alpha1/const.go#L254-L261) is:

| container | probe | budget |
|---|---|---|
| `numa` | `/livez` | 4m20s |
| `udf` | `/sidecar-livez` | 4m30s |

Both expire before the fence wait completes, so the pod is killed at the moment recovery would have succeeded — then does the same again on restart. We saw a reduce vertex crash-loop for 16 minutes this way on v1.8.4. No data was lost, but it processed nothing throughout.

## Proposal

Add `StartupProbe *Probe` to `ContainerTemplate` and the UDF `Container` type, constructed **only when set** — no default, so nothing changes for existing users. Both liveness probes are already built from the template this way ([vertex_types.go#L338](https://github.com/numaproj/numaflow/blob/v1.8.4/pkg/apis/numaflow/v1alpha1/vertex_types.go#L338), [udf.go#L80](https://github.com/numaproj/numaflow/blob/v1.8.4/pkg/apis/numaflow/v1alpha1/udf.go#L80)). Happy to raise a PR.

Related to #3192 but distinct: that treats the 5 minutes as latency, whereas this is about the wait not completing at all under the default probes. A `startupProbe` helps regardless of how the fencing question is settled.

*Checked on `v1.8.4` (current release) and `main` as of 2026-09-16 — unchanged in both.*

# Use Cases

- Reduce vertices recovering a stale fence, as above.
- Any vertex with a slow first start — large WAL replay, cold cache, a udf that loads a model. Today the only lever is a wider `failureThreshold`, which permanently blunts steady-state detection to buy a one-off boot window.

Workaround until then: `initialDelaySeconds: 120` on both probes, giving a 360s budget against the 300s fence timeout. `60` is not enough — it yields exactly 300s.

---

chriscarter90:

`ContainerTemplate` exposes `livenessProbe` and `readinessProbe` but not `startupProbe`, so a vertex that is slow to start has to buy that time out of its steady-state liveness settings. The only lever available today is a wider `failureThreshold`, which permanently blunts failure detection in order to pay for a one-off boot window.

On reduce vertices this actively conflicts. A reduce vertex waits up to 5 minutes for a stale fence file before taking the WAL lock by force, and serves no probe endpoint while it waits. The liveness budget is `initialDelaySeconds + (failureThreshold - 1) * periodSeconds`, which on the defaults is 4m20s for the `numa` container. That expires before the fence wait completes, so the pod is killed at the moment recovery would have succeeded, then does the same again on restart. We saw a reduce vertex crash-loop for 16 minutes this way on v1.8.4; no data was lost, but it processed nothing throughout.

This adds `StartupProbe *Probe` to `ContainerTemplate`, applied to the `numa` container. It is constructed only when set, so there is no default and nothing changes for existing users. The probe runs the same handler as the container's liveness probe, and any timing field left unset falls back to that liveness probe's own value, so setting a single field is enough to get a usable probe. `successThreshold` is deliberately left unset for Kubernetes to default it to 1, as startup probes require.

### Scope: why not the user-defined containers

An earlier revision of this PR also added `startupProbe` to the user-defined `Container` type, covering `udf`, `udsink`, `fb-udsink`, `on-success-udsink`, `transformer`, `udsource` and the serving sidecar. Local testing showed that this cannot work, and it has been removed.

Those containers run as native sidecars, and their probe endpoint `/sidecar-livez` is served by the `numa` container. Kubernetes does not start the main containers until every sidecar reports started, and a sidecar that carries a `startupProbe` is only considered started once that probe succeeds. The probe therefore gates itself on an endpoint that cannot exist yet, and the pod never starts. On a reduce vertex the sidecar logged 25 consecutive `connection refused` probe failures over four minutes and the pod sat at `Init:1/2` indefinitely; the same pipeline reaches `2/2 Running` without it.

This is worth noting against the original issue, which asked for `udf` as well: the `udf` container's liveness budget does expire during the fence wait, but a `startupProbe` cannot address it, since the only endpoint it could probe is the one `numa` is not yet serving. That likely needs a separate change to serve `/sidecar-livez` earlier.

### Related issues

Fixes #3648

### Testing

Unit tests covering the new helper in isolation, `ContainerTemplate.ApplyToContainer`, and the resulting `Vertex` and `MonoVertex` pod specs — including that no startup probe is produced unless one is configured, and that the liveness probe keeps its own settings when one is.

Verified on a local cluster (minikube) with a reduce pipeline, against a controller built from this branch:

- `numa` container renders `/livez` with the configured `failureThreshold: 60` and `periodSeconds: 10`, plus `timeoutSeconds: 30` inherited from the liveness probe and `successThreshold: 1` defaulted by Kubernetes; the vertex reaches `2/2 Running`
- the `udf` sidecar renders no startup probe
- a vertex with no `startupProbe` configured renders none at all

Also `make pre-push -B` (clean and idempotent), the full Go unit suite, and `cargo test --workspace`.

Most of the diff is `make codegen` output: protobuf, deepcopy, OpenAPI, swagger, the CRDs under `config/`, `docs/APIs.md` and the Rust models.

`startupProbeFrom` returns nil when a container has no liveness probe. In practice that is only the init containers, which have none and for which Kubernetes does not support a startup probe anyway.


vigith:
I have resolved a merge conflict

vigith:
looks like github UI based merge conflict introduced a typo.. @chriscarter90 could you please resolve?

chriscarter90:
@vigith Hopefully that should sort itself out now.

vigith (approved):

vaibhavtiwari33:
>Not tested against a live cluster, and the service-backed Rust feature tests were compiled but not run.

Could we please test this once in a live/local cluster?

chriscarter90:
@vigith could you possibly re-run the docs check failure please? doesn't look related to my changes.

chriscarter90:
@vaibhavtiwari33 Tested on a local cluster now, and it turned up a problem that changed the scope of this PR.

Running a reduce pipeline on minikube against a controller built from this branch:

- **`numa` container** — works as intended. The probe renders as `/livez` with the configured `failureThreshold: 60` and `periodSeconds: 10`, plus `timeoutSeconds: 30` inherited from the liveness probe and `successThreshold: 1` defaulted by Kubernetes. The vertex reaches `2/2 Running`, and a vertex with no `startupProbe` configured gets none at all.

- **`udf` sidecar** — deadlocks the pod, and I've removed support for it. User-defined containers run as native sidecars, and Kubernetes won't start the main containers until every sidecar reports started; a sidecar carrying a `startupProbe` is only started once that probe succeeds. But `/sidecar-livez` is served by the `numa` container, which can't start yet. The sidecar logged 25 consecutive `connection refused` probe failures over four minutes and the pod sat at `Init:1/2` indefinitely. The same pipeline reaches `2/2 Running` without it.

So this PR now offers `startupProbe` on `ContainerTemplate` only, and the title and description are updated to match.

One consequence worth flagging against #3648: the `udf` container's liveness budget does also expire during the fence wait, but a `startupProbe` can't address it for the reason above. That likely needs a separate change to serve `/sidecar-livez` earlier — happy to raise a follow-up issue if you think it's worth pursuing.

vigith:
I think we should resolve #3651 sooner than later.

whynowy (approved):

PR #3657, opened 2026-09-18
Add TLS caCertSecret support to Pulsar source and sink

sbbuilds:
Add an optional `tls` field (PulsarSource.tls / PulsarSink.tls) that lets a
MonoVertex trust a custom or self-signed CA when connecting to a Pulsar
broker over TLS, mirroring the existing `auth` field's secret-reference
pattern.

Signed-off-by: Stephanie Borja borja_stephanie@bah.com

Numaflow's PulsarSource/PulsarSink could not connect to a Pulsar broker over
TLS if the broker's certificate was signed by a custom or self-signed CA
(e.g. an internal cert-manager issuer) — only the system default trust
store was used. This adds an optional `tls.caCertSecret` field (mirroring
the existing `auth` secret-reference pattern) that lets the Rust Pulsar
client trust a CA supplied via a Kubernetes Secret, so MonoVertex can
connect over Pulsar's TLS port instead of requiring plaintext or a
publicly-trusted CA.

### Related issues
N/A — internal fork, no linked GitHub issue for this change.

### Testing
Manual end-to-end validation on a dedicated no-auth, TLS-enabled Pulsar
pool (auth intentionally off to isolate TLS/CA behavior from auth
concerns):

- Deployed a MonoVertex with PulsarSource/PulsarSink using
  `tls.caCertSecret` pointed at a cert-manager-issued self-signed broker CA.
- Confirmed the connection actually used the TLS port (6651, not plaintext
  6650) via socket-level inspection (`/proc/net/tcp`).
- Verified end-to-end produce/consume: broker subscription stats showed
  `msgOutCounter` incrementing and `msgBacklog` at 0.
- Cross-checked trust independently with the Java `pulsar-client` CLI using
  the same CA cert, to confirm the CA/cert chain itself (not just Numaflow)
  was valid.
- `go build ./...` and `cargo check --workspace` both pass clean.
- Regenerated CRDs/protobuf/OpenAPI/swagger/install manifests via
  `make codegen`; diff against the committed generated files was empty,
  confirming they were already in sync with the API change.

- Only server-authentication (one-way TLS) is supported: `caCertSecret` is
  honored, but mutual TLS (presenting a client cert/key) is not, since the
  underlying `pulsar` Rust crate's `PulsarBuilder` doesn't currently expose
  a way to set a client certificate for the broker connection.
- `tls.insecureSkipVerify` exists for test/debug use only — prefer
  `caCertSecret`, since it adds the CA to the trust root instead of
  disabling verification (and rustls has historically ignored the insecure
  flag against Pulsar brokers anyway).
- No changes to authentication behavior; TLS and auth are independent knobs
  on the same `PulsarSource`/`PulsarSink` spec.


vigith:
@sbbuilds could you please fix the DCO?

vaibhavtiwari33:
Please run `make codegen` to fix codegen errors. The unit tests also seem to be failing (due to non-lint related errors)

vigith (approved):

syayi (approved):

Issue #3645, opened 2026-09-14
EOF Accumulator Feature not progressing watermark


th0ger:
**Describe the bug**

https://github.com/numaproj/numaflow-python/issues/357#issuecomment-5524451483

The EOF implementation described in https://github.com/numaproj/numaflow-go/issues/228 did not progress our watermarks. They were just stuck, and the pipeline_processing_lag kept growing indefinitely (hours and days).

**To Reproduce**

Setup Accumulator with pynumaflow SDK, that returns without any Messages to output.

**Expected behavior**

Watermark progressing.

**Environment (please complete the following information):**

- Numaflow: 1.8.3
- pynumaflow: 0.14.0

**Additional info**

```python
            await output.put(
                Message(
                    value=b"",
                    keys=[],
                    tags=[DROP],
                    event_time=self.latest_wm,
                    watermark=self.latest_wm,
                )
            )
```

After adding above DROP code, the watermark progresses.
This proves that DROP is still needed, hence EOF implementation must not be fully functional.

---

**Message from the maintainers**:

Impacted by this bug? Give it a 👍. We often sort issues this way to know what to prioritize.

>For quick help and support, join our [slack](https://numaproj.slack.com/join/shared_invite/zt-19svuv47m-YKHhsQ~~KK9mBv1E7pNzfg#/shared-invite/email) channel.

akarazhev:
I think the report and the "skip, the timeout closes the window" advice in numaflow-go#228 are both right, for different keys.

In core (372bd73, same as v1.8.3) the vertex watermark follows the oldest tracked event time ([isb.rs#L182-L189](https://github.com/numaproj/numaflow/blob/372bd73a2ca32e09480f12c51f9ffeed54783645/rust/numaflow-core/src/watermark/isb.rs#L182-L189)). Those times are only cleared by windows returned in UDF responses ([reducer.rs#L166-L168](https://github.com/numaproj/numaflow/blob/372bd73a2ca32e09480f12c51f9ffeed54783645/rust/numaflow-core/src/reduce/reducer/unaligned/reducer.rs#L166-L168)), and a window times out only after its key goes idle ([accumulator.rs#L183](https://github.com/numaproj/numaflow/blob/372bd73a2ca32e09480f12c51f9ffeed54783645/rust/numaflow-core/src/reduce/reducer/unaligned/windower/accumulator.rs#L183)).

[...]

PR #3660, opened 2026-09-22
docs: correct when an accumulator UDF must respond to keep the watermark moving

akarazhev:

The "Dropping or skipping messages" section currently states that an explicit drop is never needed and that window
cleanup no longer depends on the UDF emitting output. That holds for a key that falls idle, but not for a key that keeps
receiving data: the outbound watermark is bounded by the oldest tracked event time, a tracked event time is released
only by a response from the UDF, and the window timeout is measured against the watermark. A key that keeps receiving
data while the UDF emits nothing for it therefore stalls the watermark, never times out, and grows state, with no error
reported anywhere.

This is the conclusion reached in #3645, where the request was to document it: "This will confuse the user, we need to
document this properly."

### Related issues

Refs #3645
Refs numaproj/numaflow-python#357, numaproj/numaflow-java#217

### Testing

Documentation only, no code changes.

The v1.8.1 note is kept and scoped rather than removed — it is accurate for keys that fall idle. If the drop helper is
restored in the SDKs, this section should link to it.


docs/user-guide/user-defined-functions/reduce/windowing/accumulator.md
```diff
@@ -219,21 +219,28 @@ persisted state (WAL), release the tracked messages, and let the watermark advan
 ### Dropping or skipping messages
 
 Because the Accumulator is a reduce-family operation — a stream of `Datum` in, a stream of `Datum` out — there is **no
-requirement for a one-to-one mapping between input and output**. If you do not want to forward a particular `Datum`,
-simply do not emit it; you do **not** need to emit an explicit "drop" message to account for it.
+requirement for a one-to-one mapping between input and output**. Individual `Datum`s can be skipped by simply not
+emitting them.
+
+There is one exception. The outbound watermark of the vertex cannot move past the oldest event time the vertex is still
```
vaibhavtiwari33:
```suggestion
There is one caveat. The outbound watermark of the vertex cannot move past the oldest event time the vertex is still
```

vaibhavtiwari33 (approved):

docs/user-guide/user-defined-functions/reduce/windowing/accumulator.md
```diff
@@ -219,21 +219,28 @@ persisted state (WAL), release the tracked messages, and let the watermark advan
...
+
+There is one caveat. The outbound watermark of the vertex cannot move past the oldest event time the vertex is still
+tracking, and a tracked event time is released only by a response from the UDF for that key — a `Datum`, a dropped
+message, or the EOF that follows a window close. A key that keeps receiving data while the UDF emits nothing for it
+therefore holds the watermark in place: the window timeout is measured against the watermark (see
+[`timeout`](#timeout)), so the window never closes, no EOF is produced, and the state for that key keeps growing.
+Nothing reports an error; the only symptom is a processing lag that keeps growing.
+
+**If your UDF filters out every message for a key that keeps receiving data, emit something back for that key** — for
+example a message tagged to be dropped — so that the watermark can progress.
```
vigith:
```suggestion
example a message tagged to be dropped (e.g., `MessageToDrop` / `to_drop()`) — so that the watermark can progress.
```

vigith (approved):

